PR #25, opened 2017-02-04
Create auto-prop field attributes proposal specification

HaloFour:
Proposal specification for field-targeting attributes for auto-implemented properties.

### DIFF
--- a/proposals/auto-prop-field-attrs.md
+++ b/proposals/auto-prop-field-attrs.md
@@ -56,7 +56,10 @@ public class Foo {
 ## Drawbacks
 [drawbacks]: #drawbacks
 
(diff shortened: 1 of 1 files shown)


gafter:
@dotnet/ldm Absent objections, I would champion this small change.

proposals/auto-prop-field-attrs.md
```diff
@@ -0,0 +1,69 @@
...
+public class Foo {
+    [field: NonSerialized]
+    public event EventHandler MyEvent;
+}
+```
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+Currently attempting to apply an attribute to the field of an auto-implemented property produces a compiler warning that the attributes in that block will be ignored.  There is a chance that there is existing code containing these ignored attribute blocks which will result in a potential change in behavior if the feature is properly supported.
```
gafter: @HaloFour Could you please note that possible breaking change in the drawbacks?

HaloFour:
Updated.  I did not realize that the compiler didn't validate the attribute usage when explicitly attempting to field-target those attributes.

PR #31, opened 2017-02-09
Add some championed proposals.

gafter:


proposals/records.md
```diff
@@ -0,0 +1,504 @@
...
+    ;
+
+with_initializer
+    : identifier '=' expression
+    ;
+```
+
+The token `with` is a new context-sensitive keyword.
+
+Each *identifier* on the left of a *with_initilaizer* must bind to an accessible instance field or property of the type of the *primary_expression* of the *with_expression*. There may be no duplicated name among these identifiers of a given *with_expression*.
```
333fred: Spelling: *with_initilaizer*->*with_initializer*

PR #50, opened 2017-02-09
Updating the `intptr-operators` proposal to include the list of operators.

tannergooding:
FYI. @MadsTorgersen, @gafter

### DIFF
--- a/proposals/intptr-operators.md
+++ b/proposals/intptr-operators.md
@@ -30,7 +30,245 @@ As such, the language should begin to support these operators to help advance th
 ## Detailed design
 [design]: #detailed-design
 
-The full set of operators supported are defined in `III.1.5` of the Common Language Infrastructure specification (`ECMA-335`). The specification is available here: [https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf)
+The full set of operators supported are defined in `III.1.5` of the Common Language Infrastructure specification (`ECMA-335`). The specification is available here: [https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf).
+* A summary of the operators is provided below for convenience.
+* The unverifiable operators defined by the CLI spec are not listed and are not currently part of this proposal (although it may be worth considering these as well).
+* Providing a keyword (such as `nint` and `nuint`) nor providing a way to for literals to be declared for `System.IntPtr` and `System.UIntPtr` (such as 0n) is part of this proposal (although it may be worth considering these as well).
+
+#### Unary Plus Operator
+
+```C#
+System.IntPtr operator +(System.IntPtr)
+```
+
+```C#
+System.UIntPtr operator +(System.UIntPtr)
+```
+
+#### Unary Minus Operator
+
+```C#
+System.IntPtr operator -(System.IntPtr)
+```
+
+#### Bitwise Complement Operator
+
+```C#
+System.IntPtr operator ~(System.IntPtr)
+```
+
+```C#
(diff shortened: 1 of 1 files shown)


tannergooding:
Also FYI: @migueldeicaza, since you showed interest in this proposal on the mailing list (in relation to Xamarin)

proposals/intptr-operators.md
```diff
@@ -30,7 +30,245 @@ As such, the language should begin to support these operators to help advance th
...
+```
+
+```C#
+System.UIntPtr operator ~(System.UIntPtr)
+```
+
+#### Cast Operators
+
+```C#
+operator sbyte(System.IntPtr)               // Truncate
```
gafter: Explicit conversions are not specified as methods (if it were, these would need to be `explicit` oeprators), but are rather specified as changes to the spec section on explicit conversions. But this is fine for now.

proposals/intptr-operators.md
```diff
@@ -30,7 +30,245 @@ As such, the language should begin to support these operators to help advance th
...
+operator uint(System.IntPtr)                // Truncate
+operator ulong(System.IntPtr)               // Zero Extend
+
+operator System.IntPtr(int)                 // Sign Extend
+operator System.IntPtr(long)                // Truncate
+
+operator System.IntPtr(uint)                // Sign Extend
+operator System.IntPtr(ulong)               // Truncate
+
+operator System.IntPtr(System.IntPtr)
```
gafter: Should this one return UIntPtr?

proposals/intptr-operators.md
```diff
@@ -30,7 +30,245 @@ As such, the language should begin to support these operators to help advance th
...
+#### Division Operator
+
+```C#
+System.IntPtr operator /(int, System.IntPtr)
+System.IntPtr operator /(System.IntPtr, int)
+System.IntPtr operator /(System.IntPtr, System.IntPtr)
+```
+
+```C#
+System.UIntPtr operator /(uint, USystem.IntPtr)
```
gafter: I think `USystem.IntPtr` should be `System.UIntPtr`

proposals/intptr-operators.md
tannergooding: No this was meant to be casting to same type (so `System.IntPtr x = (System.IntPtr)System.IntPtr.Zero`) for example, it (like the unary plus operators) is effectively a no-op.

The ones that return `System.UIntPtr` are just below these.

I did

proposals/intptr-operators.md
tannergooding: Yes, will fix.

proposals/intptr-operators.md
tannergooding: Yes, I wasn't sure how best to express these, but they were meant to all be `explicit operator`

tannergooding: I have updated to explicitly list these as `explicit operator`.

tannergooding:
I believe all feedback has been addressed here.

gafter:
@MadsTorgersen If you're good with the proposal in this state, please let me know and I'll merge this PR and create a tracking issue.

MadsTorgersen:
Looks good!

dgrunwald:
What's the semantics of these cast operators? The same as if `int`/`long` (depending on platform) was used in place of `IntPtr`? That would mean you get overflow checks based on the current `checked`/`unchecked` context.
While that would make sense, it's a breaking change from the present [crazy behavior](https://github.com/icsharpcode/ILSpy/blob/master/doc/IntPtr.txt#L27) that these casts exhibit. Worse, there might be programs out there that rely on the current behavior for security (overflow checks), so changing the language could introduce security vulnerabilities into existing programs.

tannergooding:
@dgrunwald, there are a few proposals around about `intptr` in general.

The two most likely (at least last I checked) are either:
* A new type
* A new keyword that uses partial erasure under the hood (I am in favor of this approach, personally)

In either case, this would not be a breaking change.

PR #70, opened 2017-02-12
Add ISSUE_TEMPLATE.md for pre-proposal discussions.

alrz:


alrz:
Note sure if it's desired to fill the whole template for pre-proposal discussion, otherwise please close this. :)

scottdorman:
How is this really any different than the proposal template? If I'm submitting an issue to request a new language feature, that's a proposal and should just follow the proposal template. If I'm submitting an issue because of a spec bug, asking a question about a proposal, or some other reason that's not a language feature, then this template doesn't provide any benefit.

alrz:
This *is* the proposal template. Think of it as some kind of pre-requisite for a feature request, because not everyone knows that there is  actually a template to follow. Eventually, we would have more detailed feature requests for LDT to consider. If you are not requesting a feature, you can simply clear the text. :)

scottdorman:
In that case, I'd suggest we call it something other than "issue template" or otherwise indicate that if you're entering an issue to request a language feature, that issue should follow the proposal template.

A concern is that people are going to see that as the issue template, decide it's too complex to follow for just a simple spec defect and not bother, or simply leave all of the sections in place with the default text and enter their issue somewhere else in the body of the issue, or try to figure out how to force-fit their issue in to the template.

alrz:
That file name is required by GitHub to use it as the issue template.

I'll defer to the team to decide on exact detail required (if any at all) at this stage (pre-proposal).

scottdorman:
It's only required to be so named if you want it to act as the default template that gets used when someone opens a new issue.

gafter:
I don't think we want a template for issues on this repo.

PR #71, opened 2017-02-12
Add null coalescing expression to the list of affected language constructs

alrz:


gafter:
This change is not correct. The proposed specification change would have no effect on the null coalescing expression unless some separate change were made to the specification for the null coalescing expression.

alrz:
@gafter I thought `var x = obj as Cat ?? obj as Dog;` would work with this change, no?

gafter:
I don't see anything in the specification for `??` that references the section of the spec that this proposal changes. So no.

PR #81, opened 2017-02-13
Update the spec for handling constant patterns (implementation in progress)

### LINKED ISSUE #69: Constant patterns in 'is' expression question: why different from 'switch'?
Can somebody explain why this `is` constant pattern:
```c#
bool Is42(byte b) => b is 42;
```
Works not the same way as similar `switch` statement:
```c#
bool Is42(byte b) {
    switch (b) {
        case 42: return true;
        default: return false;
    }
}
```

`switch` statement performs implicit conversion of literal expressions under `case` labels to *"govering-type"* (`byte` in the example above), while `is` expression just uses "nominal"/"default" type of literal expression and in fact first method never returns `true`.

Is there a deep reason why `is` constant pattern do not performs such implicit conversion of literal based on LHS type?

gafter:
Fixes #69

### DIFF
--- a/proposals/patterns.md
+++ b/proposals/patterns.md
@@ -137,11 +137,7 @@ The condition of the `if` statement is `true` at runtime and the variable `v` ho
 
 #### Constant Pattern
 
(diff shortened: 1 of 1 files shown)


gafter:
@dotnet/csharplangdesign Comments are welcome. I'm planning to implement this imminently to address https://github.com/dotnet/roslyn/issues/17103

Issue #80, opened 2017-02-13
Proposal: C# should provide better interop with unmanaged function pointers.


tannergooding:
Ported from https://github.com/dotnet/roslyn/issues/13240

**Issue**
C# has no mechanism for declaring unmanaged function pointers.

While there are a couple of mechanisms designed to work with and call unmanaged function pointers, these mechanisms may:
- Incur a higher level of overhead
- Require caching of the managed delegate
- Emit sub-optimal instructions for invoking the unmanaged function pointer
- Have issues when working with other unsafe/unmanaged code

**Workaround**
Declare the function pointer using `IntPtr` and then use the `Marshal.GetDelegateForFunctionPointer` method to convert to a managed `Delegate`. Such as:

``` C#
[UnmanagedFunctionPointer(CallingConvention.StdCall)]
public delegate void _D();

public struct S
{
    private IntPtr pD;

    public void D()
    {
        var funcptr = Marshal.GetDelegateForFunctionPointer<_D>(pD);
        funcptr();
    }
}

public static class C
{
    [DllImport("SomeBinary.dll")]
    public static extern void CreateS(out S pS);
}
```

-or-

Declare a managed delegate and rely on the marshalling behavior to convert as appropriate. Such As:

``` C#
[UnmanagedFunctionPointer(CallingConvention.StdCall)]
public delegate void _D();

public struct S
{
    [MarshalAs(UnmanagedType.FunctionPtr)]
    public _D D;
}

public static class C
{
    [DllImport("SomeBinary.dll")]
    public static extern void CreateS(out S pS);
}
```

**Proposal**
The C# language should be extended to support function pointers in an unsafe context. Such as:

``` C#
unsafe public struct S
{
    public void (*D)();
}
```

The compilation should fail if `unsafe` is missing.

**Advantages**
The compiler could take advantage of the fact this is an unsafe function pointer and emit the `calli` instruction when invoking.

The compiler could provide some level of type safety when invoking the function pointer (ensuring the arguments passed in match the types required).

This would allow better interop with unmanaged languages and types that use function pointers.

This would allow a user to more readily implement VTBLs for unmanaged interop when required.

tannergooding:
FYI. @jaredpar, who had some thoughts on how this could be implemented in a safe-manner.

jaredpar:
The idea I had was for a more general purpose C# feature: static delegates.  

``` C#
static delegate int Func();
```

A `static delegate` differs from a normal `delegate` in the following ways:

[...]

PR #126, opened 2017-02-16
Adding the static-delegate proposal.

tannergooding:
FYI. @MadsTorgersen, @gafter 

This is the official proposal for https://github.com/dotnet/csharplang/issues/80, which @jaredpar has agreed to champion.

### DIFF
--- a/proposals/static-delegates.md
+++ b/proposals/static-delegates.md
@@ -53,7 +53,9 @@ Invocation of the callback would be implemented by the `calli` instruction.
 ## Drawbacks
 [drawbacks]: #drawbacks
 
-TBD
+Static Delegates would not work with existing APIs that use regular delegates (one would need to wrap said static delegate in a regular delegate of the same signature).
+
+Additional work would be needed to make Static Delegate readily usable in the core framework.
 
 ## Alternatives
 [alternatives]: #alternatives
--- a/proposals/static-delegates.md
+++ b/proposals/static-delegates.md
@@ -33,15 +33,20 @@ One could additionally attribute the declaration with something similar to `Syst
 The declaration would get translated into an internal representation by the compiler that is similar to the following
 
 ```C#
-struct <name>
+struct <Func>e__StaticDelegate
 {
     IntPtr pFunction;
 
-    static int Func();
+    static int WellKnownCompilerName();
 }
 ```
 
-That is to say, it is internally represented by a struct that has a single member of type `IntPtr` (such a struct is blittable and does not incur any heap allocations). The member contains the address of the function that is to be the callback. Additionally, the type declares a method matching the method signature of the callback.
+That is to say, it is internally represented by a struct that has a single member of type `IntPtr` (such a struct is blittable and does not incur any heap allocations):
+* The member contains the address of the function that is to be the callback.
+* The type declares a method matching the method signature of the callback.
+* The name of the struct should not be user-constructable (as we do with other internally generated backing structures).
+ * For example: fixed size buffers generate a struct with a name in the format of `<name>e__FixedBuffer` (`<` and `>` are part of the identifier and make the identifier not constructable in C#, but still useable in IL).
+* The name of the method declaration should be a well known name used across all static delegate types (this allows the compiler to know the name to look for when determining the signature).
 
 The value of the static delegate can only be bound to a static method that matches the signature of the callback.
 
@@ -49,11 +54,11 @@ Chaining callbacks together is not supported.
 
 Invocation of the callback would be implemented by the `calli` instruction.
 
-
 ## Drawbacks
 [drawbacks]: #drawbacks
 
 Static Delegates would not work with existing APIs that use regular delegates (one would need to wrap said static delegate in a regular delegate of the same signature).
+* Given that `System.Delegate` is represented internally as a set of `object` and `IntPtr` fields (http://source.dot.net/#System.Private.CoreLib/src/System/Delegate.cs), it would be possible to allow implicit conversion of a static delegate to a `System.Delegate` that has a matching method signature. It would also be possible to provide an explicit conversion in the opposite direction, provided the `System.Delegate` conformed to all the requirements of being a static delegate.
 
 Additional work would be needed to make Static Delegate readily usable in the core framework.
 
--- a/proposals/static-delegates.md
+++ b/proposals/static-delegates.md
@@ -1,0 +1,72 @@
+# Static Delegates
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Provide a general-purpose, lightweight callback capability to the C# language.
+
+## Motivation
+[motivation]: #motivation
+
+Today, users have the ability to create callbacks using the `System.Delegate` type. However, these are fairly heavyweight (such as requiring a heap allocation and always having handling for chaining callbacks together).
+
+Additionally, `System.Delegate` does not provide the best interop with unmanaged function pointers, namely due being non-blittable and requiring marshalling anytime it crosses the managed/unmanaged boundary.
+
+With a few minor tweaks, we could provide a new type of delegate that is lightweight, general-purpose, and interops well with native code.
+
+## Detailed design
+[design]: #detailed-design
+
+One would declare a static delegate via the following:
+
+```C#
+static delegate int Func()
+```
+
+One could additionally attribute the declaration with something similar to `System.Runtime.InteropServices.UnmanagedFunctionPointer` so that the calling convention, string marshalling, and set last error behavior can be controlled. NOTE: Using `System.Runtime.InteropServices.UnmanagedFunctionPointer` itself will not work, as it is only usable on Delegates.
+
+The declaration would get translated into an internal representation by the compiler that is similar to the following
+
+```C#
+struct <name>
+{
+    IntPtr pFunction;
+
+    static int Func();
+}
+```
+
+That is to say, it is internally represented by a struct that has a single member of type `IntPtr` (such a struct is blittable and does not incur any heap allocations). The member contains the address of the function that is to be the callback. Additionally, the type declares a method matching the method signature of the callback.
+
+The value of the static delegate can only be bound to a static method that matches the signature of the callback.
+
+Chaining callbacks together is not supported.
+
+Invocation of the callback would be implemented by the `calli` instruction.
+
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+TBD
+
+## Alternatives
+[alternatives]: #alternatives
+
+TBD
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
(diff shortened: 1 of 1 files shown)


proposals/static-delegates.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+TBD
+
+## Alternatives
+[alternatives]: #alternatives
+
+TBD
```
tannergooding: C# does not currently provide any mechanism to expose the `calli` instruction. So, I believe the only alternative would be to implement the required functionality in pure IL.

proposals/static-delegates.md
```diff
@@ -0,0 +1,72 @@
...
+
+Chaining callbacks together is not supported.
+
+Invocation of the callback would be implemented by the `calli` instruction.
+
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+TBD
```
tannergooding: I can't really think of any drawbacks here. The design described above should be completely verifiable and relatively simple to implement.

proposals/static-delegates.md
```diff
@@ -0,0 +1,72 @@
...
+{
+    IntPtr pFunction;
+
+    static int Func();
+}
+```
+
+That is to say, it is internally represented by a struct that has a single member of type `IntPtr` (such a struct is blittable and does not incur any heap allocations). The member contains the address of the function that is to be the callback. Additionally, the type declares a method matching the method signature of the callback.
+
+The value of the static delegate can only be bound to a static method that matches the signature of the callback.
```
tannergooding: For better interop with unmanaged code, it might be desirable to allow assigning the value of a static delegate to that of an `IntPtr` (in an unsafe context only).

However, given that the static delegate type is itself blittable, the only scenario this would improve is if you had an unmanaged function that returned a `void*` instead of a typed function pointer (I can't think of any APIs off the top of my head that do this and I wouldn't think this would be common practice anyways).

As such, I left that off the proposal.

alrz:
So we will have the whole new delegate types`StaticAction<..>` etc?

proposals/static-delegates.md
tannergooding: I thought of some drawbacks last night, mostly due to the comment by @alrz 😄

tannergooding:
@alrz, I believe so. Although that would be part of a coordinated effort with the framework team (for betterness), rather than part of the proposed language spec/

proposals/static-delegates.md
```diff
@@ -0,0 +1,74 @@
...
+
+Chaining callbacks together is not supported.
+
+Invocation of the callback would be implemented by the `calli` instruction.
+
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+Static Delegates would not work with existing APIs that use regular delegates (one would need to wrap said static delegate in a regular delegate of the same signature).
```
tannergooding: It is probably worth noting that `System.Delegate` is represented (internally) as a set of `object` and `IntPtr` fields (http://source.dot.net/#System.Private.CoreLib/src/System/Delegate.cs). It may be possible to provide a one way conversion from `static delegate` to `System.Delegate` (this could possibly be bidirectional if we can validate the `Delegate` conforms to the requirements of a static delegate).

proposals/static-delegates.md
```diff
@@ -0,0 +1,74 @@
+# Static Delegates
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
```
jaredpar: What is this mark down trick doing?

proposals/static-delegates.md
```diff
@@ -0,0 +1,74 @@
...
+```C#
+static delegate int Func()
+```
+
+One could additionally attribute the declaration with something similar to `System.Runtime.InteropServices.UnmanagedFunctionPointer` so that the calling convention, string marshalling, and set last error behavior can be controlled. NOTE: Using `System.Runtime.InteropServices.UnmanagedFunctionPointer` itself will not work, as it is only usable on Delegates.
+
+The declaration would get translated into an internal representation by the compiler that is similar to the following
+
+```C#
+struct <name>
```
jaredpar: The `<name>` here should be `Func`.

proposals/static-delegates.md
```diff
@@ -0,0 +1,74 @@
...
+{
+    IntPtr pFunction;
+
+    static int Func();
```
jaredpar: The name here should be a standard name used across all static function types.  That way the compiler knows what name to look for when determining the signature.  

Not sure if that's going too far or not for the language spec though as it's a bit of implementation detail of the compiler.

proposals/static-delegates.md
```diff
@@ -0,0 +1,74 @@
...
+
+That is to say, it is internally represented by a struct that has a single member of type `IntPtr` (such a struct is blittable and does not incur any heap allocations). The member contains the address of the function that is to be the callback. Additionally, the type declares a method matching the method signature of the callback.
+
+The value of the static delegate can only be bound to a static method that matches the signature of the callback.
+
+Chaining callbacks together is not supported.
+
+Invocation of the callback would be implemented by the `calli` instruction.
+
+
```
jaredpar: double blank line.

proposals/static-delegates.md
jaredpar: This incompatibility could be addressed though.  The language could allow an implicit conversion from any `static delegate` to a normal `delegate` with compatible signatures.

proposals/static-delegates.md
tannergooding: It allows you to link to the sub-section of the document (this is based on the [proposal-template.md](https://github.com/dotnet/csharplang/blob/master/proposals/proposal-template.md)).

proposals/static-delegates.md
tannergooding: Yes, which is why I called it out here. I will add a note to the proposal indicating that we could workaround this.

proposals/static-delegates.md
tannergooding: I put a placeholder here because I think just `Func` isn't enough, it would need to be a non-user constructable identifier as we do with other backing structs. For example, with fixed size buffers we do `<name>e__FixedBuffer` (the `<` and `>` are not placeholders but part of the actual identifier), so in this case we would do something like `<Func>e__StaticDelegate` (or something similar, I would imagine).

proposals/static-delegates.md
tannergooding: I'm not sure either. @MadsTorgersen or @gafter, how should we represent placeholder data like this?

proposals/static-delegates.md
```diff
@@ -0,0 +1,74 @@
...
+With a few minor tweaks, we could provide a new type of delegate that is lightweight, general-purpose, and interops well with native code.
+
+## Detailed design
+[design]: #detailed-design
+
+One would declare a static delegate via the following:
+
+```C#
+static delegate int Func()
```
alrz: Bikeshed: `static delegate` could be confusing such that it can only refer to static members (in fact I'm already confused and not sure if that's the case), I'd suggest `struct delegate` though that's probably your last concern. :)

tannergooding: @alrz, static delegates can only refer to static methods. This is because delegates which refer to instance method also require you to carry a reference to the `this` object and that makes them inherently non-blittable.

proposals/static-delegates.md
tannergooding: Note added.

proposals/static-delegates.md
tannergooding: Changed the placeholder and provided a text explanation/example.

proposals/static-delegates.md
tannergooding: I updated this to be WellKnownCompilerName for the time being and added a text description indicating as such.

proposals/static-delegates.md
tannergooding: Removed.

proposals/static-delegates.md
tannergooding: Also, maybe `value delegate` to better match the naming convention for other types we have done similar things (`ValueTuple`, `ValueTask`, etc...), although I don't much care what the name is, provided I can use the feature.

jaredpar: I prefer `static delegate` over `value delegate`.  

The `Value` pattern is used today for types which are being made a `struct` type.  The `Value` prefix is given to distinguish them from the well know `class` types.  If there were a well known base type for this feature then I agree that calling it `ValueDelegate` would make sense.  That's not the case though here, there is in fact no common base type to refer to.  

In this case it's a delegate that can only refer to `static` members.  There are existing examples in the language of using `static` to modify existing types such that it can only contain `static` content: `static class`.  The `static delegate` notation is just feeding into this existing usage.

proposals/static-delegates.md
jaredpar: Yep.

proposals/static-delegates.md
alrz: @tannergooding What about open delegates? Delegate.CreateDelegate would certainly not work with this. Will there an API to create an open static delegate that directly points to a member of the input parameter?

gafter: I'm confident that this document correctly reflects the current (very preliminary) state of the proposal. We can discuss possible changes to it in the linked issue.

tannergooding:
@jaredpar, I believe (since you are championing this) the proposal needs your explicit approval before it can be merged. Let me know if you think any further modifications or clarifications are necessary.

jaredpar:
CC @dotnet/csharplangdesign

jaredpar:
@tannergooding one other aspect I think we need to make note of somewhere: this will require some run time changes in the following areas:

- Verification: the verifier has special knowledge of the manner in which normal delegates are created.  Essentially to ensure the method token being passed to the constructor is indeed a valid signature.  This same process needs to be applied to static delegates. 
- Code unloading: for environments where code unloading is supported (AppDomains for example) there likely needs to be special knowledge of static delegates.  They represent a new type which can essentially prevent DLLs from being unloaded. 

I'm unsure if the proposal needs to go that deep here.  Will defer to @gafter if that is appropriate or not here.

tannergooding:
@jaredpar is that actually the case though?

For any `calli` instruction, the following correctness & verification exists:
image


In order to assign the value of a static delegate, one would additionally need to either have the declaration come from native code (unverifiable) or one would need to use the `ldftn` instruction, which has the following correctness & verification:
image

My understanding of that is that, for verifiable code, the verifier should already be able to handle combinations of `ldftn` and `calli`, or is that not the case?

tannergooding:
@jaredpar, @gafter. Is there anything else that needs to be figured out here to get the proposal merged (it can certainly be updated further, post-merge, if need be)?

gafter:
I think this is fine to merge (@jaredpar). We can refine it later as needed.

tannergooding:
https://github.com/dotnet/csharplang/issues/302

Issue #187, opened 2017-02-26
Champion "blittable"


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

eyalsk:
Is this a proposal to improve existing blittable types or something? what's this about? :)

vbcodec:
@eyalsk  It si [Redacted] ::)

ufcpp:
Is this related to https://github.com/dotnet/roslyn/issues/2209 ?

bbarry:
and/or [dotnet/roslyn#7592](https://github.com/dotnet/roslyn/issues/7592)? Or something to do with compile time evaluation for structs?

jaredpar:
Sorry for the confusion.  Will reopen when we have the proposal ready.

jaredpar:
Sorry there isn't more details here.  This is being used as a reminder for me to add more details.

vbcodec:
verfy funny :)
Rename to 'Reservation place for any incoming proposal'

jaredpar:
Proposal added here https://github.com/dotnet/csharplang/pull/206

ufcpp:
This is good idea! but I have some questions

- How is this related to [Blittable Types in Interop Marshaling](https://msdn.microsoft.com/en-us/library/75dwhxf7)? The same? or similar but bit different?
- Can `IntPtr` and `UIntPtr` be blittable? They are blittable in the Interop Marshaling and unmanaged in the current C# spec.

eyalsk:
@jaredpar Just a thought.

Why can't it be an attribute?

```
[Blittable]
struct Point 
{
    public int X;
    public int Y;
}
```

And instead of adding a new `blittable` constraint such in the following case:

```
void Hash<T>(T value) where T : blittable struct
{
    using (T* p = &value) { 
        ...
    }
}
```

Allow attributes to be used as a generic constraint? so something like this.

```
void Hash<T>(T value) where T : struct, Blittable
{
    using (T* p = &value) { 
        ...
    }
}
```

Just seems like a more general feature that can be used here and elsewhere in the framework. :)

eyalsk:
[...]

Issue #78, opened 2017-02-13
Proposal: Fixed Sized Buffers should support formatted value types that contain only the already supported types.


tannergooding:
Ported from https://github.com/dotnet/roslyn/issues/11018

At present, fixed size buffers only support `bool`, `byte`, `char`, `short`, `int`, `long`, `sbyte`, `ushort`, `uint`, `ulong`, `float`, and `double`.

This should be expanded to also support formatted value types (structs) that contain only the above supported types.

This would allow interop code, such as the following, to be written:

```C#
public struct DXGI_RGB
{
    public float Red;
    public float Green;
    public float Blue;
}

unsafe public struct DXGI_GAMMA_CONTROL
{
    public DXGI_RGB Scale;
    public DXGI_RGB Offset;
    public fixed DXGI_RGB GammaCurve[1025];
}
```

**Other Thoughts:**
The existing `backing structure` that is emitted by the compiler to support fixed sized buffers would likely need to be modified to support this.

With the current 'backing structure' a modification to the private fields in a struct would cause a break in the generated code. For example:

Assembly A:

``` C#
public struct S1
{
    private float _x;
}
```

Assembly B:

``` C#
unsafe public struct S2
{
    public fixed S1 Buffer[10];
}
```

The above code will generate something akin to the following:

Assembly A:

``` C#
public struct S1
{
    private float _x;
}
```

Assembly B:

``` C#
unsafe public struct S2
{
    [CompilerGenerated, StructLayout(LayoutKind.Sequential, Pack = 0, Size = 40), UnsafeValueType]
    public struct <Buffer>e__FixedBuffer
    {
        public S1 FixedElementField;
    }

    [FixedBufferAttribute(typeof(S1), 10)]
    public <Buffer>e__FixedBuffer Buffer;
}
```

Then, modifying Assembly A in the following manner will cause a break (the fixed buffer has a size of 40, when it should have a size of 80, due to the additional field added).

``` C#
public struct S1
{
    private float _x;
    private float _y;
}
```

This could be worked around, by modifying the generated structure in the following manner:

``` C#
unsafe public struct S2
{
    [CompilerGenerated, StructLayout(LayoutKind.Sequential, Pack = 1), UnsafeValueType]
    public struct <Buffer>e__FixedBuffer
    {
        public S1 _0;
        public S1 _1;
        public S1 _2;
        public S1 _3;
        public S1 _4;
        public S1 _5;
        public S1 _6;
        public S1 _7;
[...]

PR #206, opened 2017-02-28
Initial blittable proposal

jaredpar:
This is the initial proposal for the blittable feature as tracked by #187.

### DIFF
--- a/proposals/blittable.md
+++ b/proposals/blittable.md
@@ -1,0 +1,161 @@
+# Blittable Types
+
+* [x] Proposed
+* [ ] Prototype
+* [ ] Implementation
+* [ ] Specification
+
+## Summary
+[summary]: #summary
+
+The blittable feature will give language enforcement to the class of types known as "unmanaged types" in the C# language spec.  This is defined in section 18.2 as a type which is not a reference type and doesn't contain reference type fields at any level of nesting.  
+
+## Motivation
+[motivation]: #motivation
+
+The primary motivation is to make it easier to author low level interop code in C#.  Blittable types are one of the core building blocks for interop code, yet abstracting around them is difficult today due to a lack of declarative language support.  
+
+This is the case even though blittable types are well defined in the language spec and many features, like pointers, can operate only on them.  The language chooses to let structs implicitly opt into being blittable by virtue of their construction with no avenue for opting out of such a classification.
+
+While attractive in small applications this is more difficult to manage across a large set of libraries authored by different developers.  It means small field additions to structs can cause compile and runtime breaks in downstream consumers with no warning to the developers that made the change.  This spooky action at a distance is one of the core problems with having an implicit opt in model here. 
+
+A declarative, explicit opt in model makes it easier to code in this area.  Developers can be very explicit about the types they intend for interop.  This gives them compiler help when mistakes are made in their application and in any libraries they consume.  
+
+The lack of compile time support also makes it difficult to abstract over blittable types.  It's not possible for instance to author common helper routines using generic code:
+
+``` c#
+void Hash<T>(T value) where T : blittable struct
+{
+    using (T* p = &value) { 
+        ...
+    }
+}
+```
+
+Instead developers are forced to rewrite virtually the same code for all of their blittable types:
+
+``` c#
+void Hash(Point p) { 
+    ...
+}
+
+void Hash(Time t) { 
+    ...
+}
+```
+
+The lack of constraints here also make it impossible to have efficient conversions between streams of data and more structured types.  This is an operation that is common in networking stacks and serialization layers:
+
+``` c#
+Span<byte> Convert<T>(ref T value) where T : blittable {
+    ...
+}
+```
+
+Such routines are advantageous because they are provably safe at compile time and allocation free.  Interop authors today can not due this (even though it's at a layer where perf is critical).  Instead they need to rely on allocating routines that have expensive runtime checks to verify values are correctly blittable.
+
+## Detailed design
+[design]: #detailed-design
+
+The language will introduce a new declaration modifier named `blittable` that can be applied to `struct` definitions.  
+
+``` c#
+blittable struct Point 
+{
+    public int X;
+    public int Y;
+}
+```
+
+The compiler will enforce that the fields of such a struct definition fit one of the following categories:
+
+- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
+- Any `enum` type
+- Any pointer type which points to a `blittable` type
+- Any user defined struct explicitly declared as `blittable`
+
+Any type fitting the definition above, or any element in the list, is considered a valid `blittable` type.
+
+``` c#
+struct User
+{
+    string FirstName;
+    string LastName;
+}
+
+blittable struct ItemData
+{
+    // Error: blittable struct cannot contain field of type User which is not blittable
+    User User;
+    int Id;
+}
+```
+
+Note that a user defined struct must be explicitly declared as `blittable` in order to meet the requirements above.  This is required even if the struct otherwise meets the requirements of `blittable`. 
+
+``` c#
+struct SimplePoint
+{
+    public int X;
+    public int Y;
+}
+
+blittable struct Data
+{
+    // Error: blittable struct cannot contain field of type SimplePoint which is not blittable
+    SimplePoint Point;
+}
+```
+
+The language will also support the ability to constrain generic type parameters to be `blittable` types.  
+
+``` C#
+void M<T>(T p) where T : blittable struct
+{
+    ...
+}
+
+M<Point>(); // Ok
+M<User>();  // Error: Type User does not satisfy the blittable constraint.
+```
+
+One of the primary motivations for `blittable` structs is ease of interop.  As such it's imperative that such a type have it's field ordered in a sequential, or explicit, layout.  An auto layout of fields makes it impossible to reliably interop the data.  
+
+The default today is for a sequential layout so this doesn't represent a substantial change.  However the compiler will make it illegal to mark such types as having an auto layout.  
+
+``` c#
+// Error: A blittable struct may not be marked with an auto layout
+[StructLayout(LayoutKind.Auto)]
+blittable struct LayoutExample 
+{
+    ...
+}
+```
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+The primary drawback of this feature is that it serves a small number of developers: typically low level library authors or frameworks.  Hence it's spending precious language time for a small number of developers. 
+
+Yet these frameworks are are often the basis for the majority of .NET applications out there.  Hence performance / correctness wins at this level can have a riple effect on the .NET ecosystem.  This makes the feature worth considering even with the limited audience. 
+
+There will also likely be a small transition period after this is released where core libraries move to adopt it.  Types like [System.Runtime.InteropServices.ComTypes.FILETIME](https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.comtypes.filetime(v=vs.110).aspx) are common in interop code.  Until it is marked as `blittable` in source though, developers won't be able to depend on it in their libraries.  
+
+## Alternatives
+[alternatives]: #alternatives
+
+There are a couple of alternatives to consider:
+
+- The status quo:  The feature is not justified on its own merits and developers continue to use the implicit opt in behavior.
+- Generic constraint only: The blittable keyword is used on generic constraints only.  This allows for developers to author efficient helper libraries.  But the types involved lack any declarative support and hence are fragile across distributed development. 
+
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- blittable vs. unmanaged.  The F# language has a very [similar feature](https://docs.microsoft.com/en-us/dotnet/articles/fsharp/language-reference/generics/constraints) which uses the keyword unmanaged. The blittable name comes from the use in Midori.  May want to look to precedence here and use unmanaged instead. 
+- Verifier: does the verifier / runtime need to be updated to understad the use of pointers to generic type parameters?  Or can it simply work as is without changes. 
+
+## Design meetings
+
+n/a


jaredpar:
CC @dotnet/csharplangdesign

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+void M<T>(T p) where T : blittable struct
+{
+    ...
+}
+
+M<Point>(); // Ok
+M<User>();  // Error: Type User does not satisfy the blittable constraint.
+```
+
+One of the primary motivations for `blittable` structs is ease of interop.  As such it's imperative that such a type have it's field ordered in a sequential, or explicit, layout.  An auto layout of fields makes it impossible to reliably interop the data.
```
333fred: Nit: second "it's" -> "its"

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+
+void Hash(Time t) { 
+    ...
+}
+```
+
+The lack of constraints here also make it impossible to have efficient conversions between streams of data and more structured types.  This is an operation that is common in networking stacks and serialization layers:
+
+``` c#
+Span<byte> Convert<T>(ref T value) where T : blittable {
```
jcouv: Should the constraint be `blittable struct`, or is this intentional?

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+
+While attractive in small applications this is more difficult to manage across a large set of libraries authored by different developers.  It means small field additions to structs can cause compile and runtime breaks in downstream consumers with no warning to the developers that made the change.  This spooky action at a distance is one of the core problems with having an implicit opt in model here. 
+
+A declarative, explicit opt in model makes it easier to code in this area.  Developers can be very explicit about the types they intend for interop.  This gives them compiler help when mistakes are made in their application and in any libraries they consume.  
+
+The lack of compile time support also makes it difficult to abstract over blittable types.  It's not possible for instance to author common helper routines using generic code:
+
+``` c#
+void Hash<T>(T value) where T : blittable struct
```
tannergooding: nit: You use `blittable struct` here and `blittable` below. We should probably be consistent with one or the other.

I personally prefer just `blittable`, since you can't have a `blittable` class.

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+blittable struct Point 
+{
+    public int X;
+    public int Y;
+}
+```
+
+The compiler will enforce that the fields of such a struct definition fit one of the following categories:
+
+- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
```
tannergooding: Please, Please, Please do not leave off `IntPtr` and `UIntPtr`. These types are always blittable and are one of the core interop types.

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+- Any `enum` type
+- Any pointer type which points to a `blittable` type
+- Any user defined struct explicitly declared as `blittable`
+
+Any type fitting the definition above, or any element in the list, is considered a valid `blittable` type.
```
tannergooding: I thought you wanted to add a section for fixed arrays as well (https://github.com/dotnet/csharplang/issues/78)?

tannergooding:
@jaredpar, what type of 'contract' are we attempting to declare with `blittable`, other than this type only declares types that can be used in unmanaged code without marshaling?

Are we also saying that:
* the type is expected to never change in the future, or
* the initial field layout of the struct won't change, or
* is it ok to add new fields to the end of a blittable struct, or
* maybe that the size of a struct will never shrink

I think these are important questions, partially because some things are still safe, provided the consumer is aware they can happen. For example, exposing new members on the end of a struct is technically OK, provided the user isn't assuming a fixed size.

Also:
* are fields expected to be exposed directly (public), or
* is marking fields private and exposing them via properties still ok

This is probably important because in the first means the user can view the direct field layout, where-as the latter means someone can abstract it away if the need/want.

Some of this may also become important when tied in with Static Delegates (https://github.com/dotnet/csharplang/pull/126) as users can manually implement VTBLs for interop code (and may want to mark those as blittable).

KodrAus:
@tannergooding I would expect the contract purely to assert the data layout is the same between managed and unmanaged code, and maintaining those other assumptions would still be the responsibility of the programmer, if that's what they need.

I haven't looked at #126 yet though, so maybe that context would shift my thinking.

proposals/blittable.md
jaredpar: This was mentioned in the issue as well.  It was an ommission, will add before merging.

proposals/blittable.md
jaredpar: I considered that but as we were discussing I'm not sure why we should limit this to blittable types, seems we can trivially expand it to any type out there.  Working to validate if that's legal or not.  If it's not then I'll go back and expand this proposal to do that.

I should explicitly mention in the proposal though that a fixed array of blittable types is considered a valid field type inside a blittable struct.

proposals/blittable.md
jaredpar: Will make it consistent.  My preference is still for `blittable struct` though because I want to allow for other future features.  A `class` may not be `blittable` but what about a `static delegate`?

proposals/blittable.md
jaredpar: Good catch.  Should've been blittable struct.

proposals/blittable.md
jaredpar: Its correct I think :trollface:

jaredpar:
@tannergooding 

>  For example, exposing new members on the end of a struct is technically OK, provided the user isn't assuming a fixed size.

There are no restrictions on future changes to the `blittable struct`.  One of the goals of blittable though is to make it much more apparent the ramifications of making changes like adding fields.  

For instance most developers wouldn't think twice about adding a field to a struct which already has at least one public constructor.  Not really going to break anyone with that.  Adding it to a blittable struct though is more likely to raise the correct questions from reviewers and the developer in question. 

> are fields expected to be exposed directly (public), or is marking fields private and exposing them via properties still ok

Visibility of fields does not factor into whether a struct can / can't be marked as `blittable`.  Both because it shouldn't matter IMHO and it's necessary for it to be a faithful implementation of `unmanaged type`.

tannergooding:
👍 That is about what I expected

proposals/blittable.md
gafter: > Rule 1: When you mean it is or it has, use an apostrophe. Rule 2: When you are using its as a possessive, don't use the apostrophe.

proposals/blittable.md
Tornhoof: Your list contains quite a few non blittable types, e.g. decimal, char, bool. Are those mistakes or so you plan to change this?
If you plan to change this, what about guid, datetime, datetimeoffset, timespan?

tannergooding: `Decimal`, `Guid`, `DateTime`, `DateTimeOffset`, and `TimeSpan` are implicitly blittable as they are structs composed of blittable types.

`char` and `bool` are in the category of "sometimes blittable", so we should be careful about including them.

jaredpar: > Decimal, Guid, DateTime, DateTimeOffset, and TimeSpan are implicitly blittable as they are structs composed of blittable types.

I've waffled a bit on whether these should be special cased as implicitly blittable, or waiting for source annotations to happen.  Leaning towards implicit.  

> char and bool are in the category of "sometimes blittable", so we should be careful about including them.

Been way down this rabbit hole in Midori.  The basic problem comes down to whether you believe bilttable means ... 

1. A type which has no field, at any level of nesting, whose type is a reference type 
1. 1 + the constraint that all possible bit patterns represent a valid instance of the type 

In Midori we actually separated this out into two features: primitive and blittable.  All blittable are primitive but not the other way around.  The distinction was interesting but IMHO wasn't worth the extra complexity. 

In any case though, this feature definitely targets 1 above.  It's the language manifestation of the existing spec concept of "unmanaged type".

tannergooding: I think if we are explicitly stating that this is for 'unmanaged types' (that is types for which the unmanaged and managed representation is bit for bit identical) then we should not be including 'char' or 'bool', since the statement is not always true for these types.

String is technically blittable in some cases (and there is logic for pinning a string rather than copying in those instances).

Tornhoof: @tannergooding  `DateTime` and `DateTimeOffset` are currently not implicitly blittable as they are structs marked with layout Auto, even though they are composed of blittable types.
I guess that would need fixing for this proposal, otherwise this might exclude quite a few user defined blittable types (especially for serialization).
Still a pitty that nothing really can be done for `bool` besides using a classic BlittableBool struct wrapper.

Thank you both for your detailed explanations.

proposals/blittable.md
whoisj: @jaredpar can you say "blittable type" to cover all possible blittable types?

jaredpar: @whoisj not quite sure what you are asking there.

whoisj: @jaredpar seemed like confusion of using `blittable struct` vs `blittable`, I was just suggesting use of the word "type" in place of "struct" to keep it vague but accurate. 'Tis all. 😉

tannergooding: @jaredpar, a `static delegate`, as currently proposed is a `blittable struct` 😉

mattwar:
Maybe use keyword 'fixed' instead of adding 'blittable' which is not a keyword and is really jargon-y.

proposals/blittable.md
tannergooding: Ah yes, I missed the Auto layout attribute. It might be worth logging bugs to remove that. At least in the case of `DateTime`, it currently only contains a single field (a `UInt64`) and will have the same layout regardless. For `DateTimeOffset`, it contains a `DateTime` and a `Int16`, so the layout should be the same as well.

KodrAus:
Personally, I think `blittable` is better than `fixed` which already means something else as a keyword. I don't have any problem with `blittable`, since it communicates and binds some behaviour of the data type it's added to. It makes the most sense for `struct`s, which could change, but to communicate intent on a delegate I think is also useful.

mattwar:
Fixed already sort of implies/means blittable in C#.  The keyword _fixed_ is used on fixed sized array fields, making them inline to the struct and can only have blittable element types.  A _fixed_ field is blittable.  An entire _fixed_ struct would be the same.

tannergooding:
@mattwar, the issue with that logic is that `fixed` currently only covers a small portion of what `blittable` covers: https://github.com/dotnet/csharplang/issues/78

tannergooding:
That is, `fixed` would first need to be extended to support all blittable types (including things like `IntPtr`, which is blittable, but has a variant size, depending on the platform) before it could be interchangeable with 'blittable'.

mattwar:
@tannergooding I'm assuming any proposal to add blittable structs would include changing fixed arrays to include these types as well.

whoisj:
I'm honestly fine with `blittable`, `fixed`, `unmanaged`, or `itjustworksuseit` so long as I get the functionality. 😁 

Seriously `fixed` is good because it give an indication that the memory won't be moved, so a pointer is meaningful. `blittable` is good because, like `unmanaged` informs that the structure can be copied and/or modified directly without impeding / impacting the GC.

That said, can we please get an `IFixable`, `IBlittable`, or (and I really like this one) `IUnmanagable` interface that enabled the `fixed` statement to be used on any type that can have a pointer, um, pointed at it?

Please please please, with a cherry on top? 🍨 🍒

proposals/blittable.md
Tornhoof: Filed bugs https://github.com/dotnet/coreclr/issues/10448 and https://github.com/dotnet/coreclr/issues/10449 for the layoutkind issue

omariom:
`flat`, no?

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+    public int X;
+    public int Y;
+}
+```
+
+The compiler will enforce that the fields of such a struct definition fit one of the following categories:
+
+- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
+- Any `enum` type
+- Any pointer type which points to a `blittable` type
```
jkotas: Why can't this be any pointer? It should not matter what the pointer points to.

jkotas: Also `void*` should be blittable

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+The primary drawback of this feature is that it serves a small number of developers: typically low level library authors or frameworks.  Hence it's spending precious language time for a small number of developers. 
+
+Yet these frameworks are are often the basis for the majority of .NET applications out there.  Hence performance / correctness wins at this level can have a riple effect on the .NET ecosystem.  This makes the feature worth considering even with the limited audience. 
+
+There will also likely be a small transition period after this is released where core libraries move to adopt it.  Types like [System.Runtime.InteropServices.ComTypes.FILETIME](https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.comtypes.filetime(v=vs.110).aspx) are common in interop code.  Until it is marked as `blittable` in source though, developers won't be able to depend on it in their libraries.  
+
```
joeante: This is a very important feature for Unity, that goes far beyond low level library authors. This is something many of our customers would use. In games generally we very much like the ability to pack our memory in a very controlled way.

proposals/blittable.md
omariom: Then I withdraw my  `flat` proposal :)

proposals/blittable.md
whoisj: Once we have `blittable` we can start discussing things like an `IFixable<T>` interface `where T : blittable`, which would enable `fixed (void* p = myIFixable)` and other useful syntax candy.

xoofx:
While I like the idea to be able to explicitly mark a struct with a blittable attributes/keyword so that compile time checking is not only done on generic bindings but also on struct declaration itself, but the fact that the implicit behavior is not possible is quite annoying, as it will require to go over all existing structs that are blittable today and add this attribute.

By introducing the keyword, it will not change the fact that the JIT/AOT will always check that the struct is actually blittable, even if the method has a generic "blittable" constraint, it will have to check the type.

Why couldn't we allow both, implicit (for legacy, that would still allow generic constraint on blittable) and explicit (for code that can be ugraded today easily)?  On a struct that doesn't have a blittable attribute, the generic constraint blittable would check at least that the struct is transitively blittable compatible at least when it was compiled...
The implicit scenario could allow a smoother transition...

whoisj:
> Why couldn't we allow both, implicit ...

What is the cost delta of requiring the keyword and pure run-time detection?

xoofx:
> What is the cost delta of requiring the keyword and pure run-time detection?

As I see it, the keyword would only be used at compile time (both for struct and generic constraint) but I don't expect it to be used at runtime, as the runtime has to check anyway invalid usages of fixed (e.g via IL patching). The check at runtime is relatively small, blittable is computed once for a type, and checked when compiling a method that requires to "fixed" a T

whoisj:
Assuming the cost is as minimal as @xoof suggests, then I support the idea to back-port the feature via detection.

whoisj:
> The policy there is to remove all private struct fields (a terrible idea for compilers, but a reality that must be considered).

... and that was my concern (sorry for not articulating as nicely as you did, but yes the "reference assemblies" worried me). Thanks @jaredpar. Finally - with .Net Core and the "death of reflection" (kill it, kill it with fire! :fire:), is it even possible to know what a private field is reliably?

xoofx:
> The problem with implicit detection is that it basically renders this feature useless. 

Not entirely, it would make it half-useless 😅  (we would still benefit from having the generic constraint working at compile time). But I agree that it is not great in terms of language coherency...

> The policy there is to remove all private struct fields 

Just to understand, what is the deal currently with them, could you elaborate?

tannergooding:
@xoofx. Imagine you have the following:
```C#
class C
{
    private int _value;

    public int Value
    {
        get
        {
            return _value;
        }
    }
}

struct S
{
    private C _c;

    public C c
    {
        get
        {
            return _c;
        }
    }
}
```

In the implementation assembly, `S` will never satisfy the blittable requirement.

However, it is fairly common to have 'reference' assemblies. For reference assemblies, all private members are stripped (generally), so we end up with:

```C#
class C
{
    public int Value
    {
        get
        {
            throw new NotImplementedException();
        }
    }
}

struct S
{
    public C c
    {
        get
        {
            throw new NotImplementedException();
        }
    }
}
```

Now, `S` will pass the blittable requirement (as far as the compiler is concernced, since it has no fields) and we could potentially end up with unverifiable, unjitabble, or unsafe code.

xoofx:
@tannergooding han, thanks for the precision, never looked exactly how much the ref assemblies kept around the original internal meta... so... "fixing" the reference assembly generation tools could actually make the implicit proposal "half-possible"? 😁 

kidding, let's keep it explicit 👍

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+    ...
+}
+```
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+The primary drawback of this feature is that it serves a small number of developers: typically low level library authors or frameworks.  Hence it's spending precious language time for a small number of developers. 
+
+Yet these frameworks are are often the basis for the majority of .NET applications out there.  Hence performance / correctness wins at this level can have a riple effect on the .NET ecosystem.  This makes the feature worth considering even with the limited audience.
```
Flamanis: `frameworks are are`
Just a small nitpick from someone who knows basically nothing about what is said in the proposal.

jnm2: While we're at it, `riple effect` should be `ripple effect`.

proposals/blittable.md
jnm2: Yeah, rule of thumb I say is if it has a apostrophe, read it in your head as "it is" and see it if makes sense.

Ultrahead:
As I mentioned [here](https://github.com/dotnet/csharplang/issues/492#issuecomment-298903559), please consider adding an IsBlittable getter to both, the Type and ValueType classes.

proposals/blittable.md
Ultrahead: Will this allow the use of generic pointers? Like: `void Hash<T>(T* pointer) where T : blittable struct`

proposals/blittable.md
jamesqo: > Please, Please, Please do not leave off IntPtr and UIntPtr. These types are always blittable and are one of the core interop types.

Wouldn't it be more helpful to special-case actual pointer types instead, e.g. `void*`?

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+
+The lack of constraints here also make it impossible to have efficient conversions between streams of data and more structured types.  This is an operation that is common in networking stacks and serialization layers:
+
+``` c#
+Span<byte> Convert<T>(ref T value) where T : blittable {
+    ...
+}
+```
+
+Such routines are advantageous because they are provably safe at compile time and allocation free.  Interop authors today can not due this (even though it's at a layer where perf is critical).  Instead they need to rely on allocating routines that have expensive runtime checks to verify values are correctly blittable.
```
Tragetaschen: "due" -> "do"

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+    ...
+}
+```
+
+The lack of constraints here also make it impossible to have efficient conversions between streams of data and more structured types.  This is an operation that is common in networking stacks and serialization layers:
+
+``` c#
+Span<byte> Convert<T>(ref T value) where T : blittable {
+    ...
+}
```
jcouv: From recent chat in LDM, the question came up: is `ValueTuple<int, int>` blittable? (I assume it should) How would we achieve that?

jkotas: I do not think `ValueTuple<int, int>` should be blittable. This feature is for low-level interop. Limiting use of abstractions in low-level interop is fine. For example, generic PInvokes are not allowed and I do not think that it was ever raised as problem.

> ValueTuple types are implemented with [StructLayout(LayoutKind.Auto)]

That makes them non-blittable by definition. This optimization would need to be undone for `ValueTuple<int, int>` to be blittable.

tannergooding: For a large number of generic types, such as `ValueTuple`, they are designed to be used with more than just blittable types, and there is no real way to make it 'blittable' without breaking back-compat (under the current proposal).

With the current proposal, structs will need to be explicitly marked as blittable (`blittable struct`). If they are missing this decoration then they will not be considered blittable.

There are already limitation around using generics with any kind of P/Invoke. However, if we had a `blittable` constraint, there is really no reason why this support could not be added (just like there is no reason to block taking the address of a generic type with blittable constraints on each typeparam).

jkotas: > limitation around using generics with any kind of P/Invoke
> no reason to block taking the address of a generic type with blittable constraints

These two are pretty different:

Taking address of generic type is simple, well-defined operation. You can do it today using `System.Runtime.CompilerServices.Unsafe`. It is just a C# language limitation that you cannot write it directly.

Generics in P/Invokes is not simple and well defined. It is limitation of the PInvoke marshalling baked into the runtime. PInvoke marshalling rules are pretty complex. Throwing generics into the mix would make them even more complex.

4creators:
#749

proposals/blittable.md
weitzhandler: @jnm2 *and see ~it~ if
😛

GaTechThomas: Simple mneumonic for its/it's: Think of the apostrophe as the dot on an 'i'.

ghord:
Is this proposal even under active work? It's scheduled for 7.2 yet no progress is being made. Do you think it will be delayed to 7.X or abandoned altogether?

jaredpar:
@ghord i just forgot to merge the PR 😦 

>  It's scheduled for 7.2 yet no progress is being made

It is not scheduled for 7.2. The notations on this repo around milestones are more aspirations / suggestions to the compiler team. Essentially "it would be fine if this made it for 7.2". 

It is currently being considered for 7.3 by the compiler team.

tomrus88:
> It is currently being considered for 7.3 by the compiler team
It was removed from Language Feature Status completely recently, plans changed?

jcouv:
@tomrus88 Work has not started on that feature, but we're still planning to work on it in 7.3.

4creators:
@jcouv 
Would it help if I would volunteer to work on the implementation with C# team? This feature and static delegates are of high importance to me.

jaredpar:
> It was removed from Language Feature Status completely recently, plans changed?

The language feature status page is meant to be status on actual changes happening in the compiler at the time. Essentially work that is scheduled / in progress that has resources allocated to it. It's not a commitment, can always decided to delay / cancel work. But it does try to add transparency to what the compiler team is working on.

In this case we had scheduled the blittable work and then as we got closer to the end of C# 7.2 we decided it simply didn't fit and we backed off. This was due to a combination of design issues and time constraints. 

> Would it help if I would volunteer to work on the implementation with C# team? T

For blittable we have devs assigned to it: @OmarTawfik. It's planned for C# 7.3 but as with all features that's subject to change for any number of reasons 😄 

> This feature and static delegates are of high importance to me.

Static delegates aer trickier because they have runtime implications that we need to work out. Still trying to figure that out.

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+There are a couple of alternatives to consider:
+
+- The status quo:  The feature is not justified on its own merits and developers continue to use the implicit opt in behavior.
+- Generic constraint only: The blittable keyword is used on generic constraints only.  This allows for developers to author efficient helper libraries.  But the types involved lack any declarative support and hence are fragile across distributed development. 
+
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- blittable vs. unmanaged.  The F# language has a very [similar feature](https://docs.microsoft.com/en-us/dotnet/articles/fsharp/language-reference/generics/constraints) which uses the keyword unmanaged. The blittable name comes from the use in Midori.  May want to look to precedence here and use unmanaged instead.
```
jkotas: "blittable type" has an existing definition in CLR interop already: https://docs.microsoft.com/en-us/dotnet/framework/interop/blittable-and-non-blittable-types . The existing definition is different from your definition, e.g. `char` and `bool` are not blittable in the existing definition. It may be a good idea to call it something else to avoid confusion.

tannergooding: It's probably worthwhile to keep this as blittable and remove `char` and `bool` from the definition.

That being said, `bool` and `char` are special in that they are considered "sometimes" blittable and, last I checked, the (CoreCLR) marshaller will still blit `bool` and `char` when the appropriate conditions are met (just like it will pin and directly marshal a `string` when the right conditions are met).

proposals/blittable.md
jnm2: @GaTechThomas mnemonic
(I'm well aware that I'm in a glass house wiyh typos 😂)

proposals/blittable.md
jkotas: The existing definition also have "The following complex types are also blittable types: One-dimensional arrays of blittable types". What would you do about that?

proposals/blittable.md
```diff
@@ -0,0 +1,161 @@
...
+
+This is the case even though blittable types are well defined in the language spec and many features, like pointers, can operate only on them.  The language chooses to let structs implicitly opt into being blittable by virtue of their construction with no avenue for opting out of such a classification.
+
+While attractive in small applications this is more difficult to manage across a large set of libraries authored by different developers.  It means small field additions to structs can cause compile and runtime breaks in downstream consumers with no warning to the developers that made the change.  This spooky action at a distance is one of the core problems with having an implicit opt in model here. 
+
+A declarative, explicit opt in model makes it easier to code in this area.  Developers can be very explicit about the types they intend for interop.  This gives them compiler help when mistakes are made in their application and in any libraries they consume.  
+
+The lack of compile time support also makes it difficult to abstract over blittable types.  It's not possible for instance to author common helper routines using generic code:
```
jkotas: This is possible using `System.Runtime.CompilerServices.Unsafe` package.

proposals/blittable.md
jaredpar: I think we are going to go with the term "unmanaged type". It is already in use in the C# spec and F#.

PR #207, opened 2017-02-28
Redundant brace removal v2

DavidArno:


alrz:
I think you don't need to open a PR for a proposal right away as it's mentioned in the README.md.

>For new feature proposals, however, please raise them for discussion, and only submit a proposal as a pull request if invited to do so by a member of the Language Design Team (a "champion").

DavidArno:
I'm pushing back against that advice in the readme, as I think it poor advice. Having the PR means the team can merge it, even before it's championed and accepted/rejected in the LDM. That way, the proposal is there for all to see, rather than only existing as a discussion thread in an issue. Even if later rejected, it can then be moved to `proposals/rejected` as a better permanent record than just a closed issue.

DavidArno:
Oh and @alrz, I believe you previously suggested I had some of the terms wrong in the proposal. I'd appreciate it if you could highlight those for me and I'll fix them.

PR #264, opened 2017-03-13
Safety rules for ref-like types like `Span<T>`

VSadov:
Please do not merge yet.

### DIFF
--- a/proposals/span-safety.md
+++ b/proposals/span-safety.md
@@ -20,7 +20,7 @@ C# compiler already has a concept of a restricted types that covers special plat
 
 In this document the "ref like" types include only `Span<T>` and related types. And the ref-like variables mean variables of "ref-like" types.  
 
-In order to force ref-like variables to be stack only,we need the following restrictions: 
+In order to force ref-like variables to be stack only, we need the following restrictions: 
 
 - ref-like type cannot be a type of an array element
 - ref-like type cannot be used as a generic type argument
@@ -45,7 +45,7 @@ Example:
 
 Span<T> Caller()
 {
-	Span<T> safeToReturn Span<T> safeToReturn;
+	Span<T> safeToReturn;
 
 	if (TodaIsNotFriday())
 	{
--- a/proposals/span-safety.md
+++ b/proposals/span-safety.md
@@ -4,7 +4,7 @@ The main reason for the additional safety rules when dealing with types like `Sp
  
 There are three reasons why `Span<T>` must be a stack-only type.
 
-1. `Span<T>` is basically a struct containing `{reference to data, offset limit}`. Concurrent "tearing" of such struct would lead to the possibility of out of range accesses and type-safety violations, which basically means "GC heap corruption".
+1. `Span<T>` is semantically a struct containing a reference and a range - `(ref T data, int length)`. Regardless of actual implementation, writes to such struct would not be atomic. Concurrent "tearing" of such struct would lead to the possibility of `length` not matching the `data`, causing out-of-range accesses and type-safety violations, which ultimately could result in GC heap corruption in seemingly "safe" code.
 2. Some implementations of `Span<T>` literally contain a managed pointer in one of its fields. Managed pointers are not supported as fields of heap objects and code that manages to put a managed pointer on the GC heap typically crashes at JIT time.
 3. It is permitted for a `Span<T>` to refer to data in the local stack frame - individual local variables or `stackalloc`-ed arrays. A scenario when an instance of a `Span<T>` outlives the referred data to would lead to undefined behavior, including type-safety violations and heap corruptions.
 
@@ -18,7 +18,7 @@ At the minimum the restrictions described below would need to apply to `Span<T>`
 
 C# compiler already has a concept of a restricted types that covers special platform types such as TypedReference. In some cases those types are referred as ref-like types as well. We should probably use some different term to refer to `TypedReference` and similar types - like `restricted types` to avoid confusion.
 
-In this document the "ref like" types include only `Span<T>` and related types. And a ref-like variables mean a variable of a "ref-like" type.  
+In this document the "ref like" types include only `Span<T>` and related types. And the ref-like variables mean variables of "ref-like" types.  
 
 In order to force ref-like variables to be stack only,we need the following restrictions: 
 
@@ -66,7 +66,9 @@ void Callee(ref Span<int> arg1)
 
 ```
 
-Note: returning by a ordinary writeable reference appears to be ok.
+Note: returning by a ordinary writeable reference appears to be ok by itself, but with values on heap being impossible, values in local frame being not returnable and variables passed from the caller being readonly, it may not matter.
+
+Another way to look at this is that writeable references to a ref-like variable would never be exposed outside of the containing method, except, of course, if that method is a constructor which has to take `this` by a writeable ref.
 
 ## `ref-like` types must be readonly structs. ##
 
--- a/proposals/span-safety.md
+++ b/proposals/span-safety.md
@@ -6,7 +6,7 @@ There are three reasons why `Span<T>` must be a stack-only type.
 
 1. `Span<T>` is semantically a struct containing a reference and a range - `(ref T data, int length)`. Regardless of actual implementation, writes to such struct would not be atomic. Concurrent "tearing" of such struct would lead to the possibility of `length` not matching the `data`, causing out-of-range accesses and type-safety violations, which ultimately could result in GC heap corruption in seemingly "safe" code.
 2. Some implementations of `Span<T>` literally contain a managed pointer in one of its fields. Managed pointers are not supported as fields of heap objects and code that manages to put a managed pointer on the GC heap typically crashes at JIT time.
-3. It is permitted for a `Span<T>` to refer to data in the local stack frame - individual local variables or `stackalloc`-ed arrays. A scenario when an instance of a `Span<T>` outlives the referred data to would lead to undefined behavior, including type-safety violations and heap corruptions.
+3. It is permitted for a `Span<T>` to refer to data in the local stack frame - individual local variables or `stackalloc`-ed arrays. A scenario when an instance of a `Span<T>` outlives the referred data would lead to undefined behavior, including type-safety violations and heap corruptions.
 
 All the above problems would be alleviated if instances of `Span<T>` would be allowed as stack data types only. In fact for the `#3` we need a stronger guarantee. Details on that to follow below.
 
@@ -16,7 +16,7 @@ At the minimum the restrictions described below would need to apply to `Span<T>`
 
 ## `ref-like` types must be stack-only. ##
 
-C# compiler already has a concept of a restricted types that covers special platform types such as TypedReference. In some cases those types are referred as ref-like types as well. We should probably use some different term to refer to `TypedReference` and similar types - like `restricted types` to avoid confusion.
+C# compiler already has a concept of a restricted types that covers special platform types such as `TypedReference`. In some cases those types are referred as ref-like types as well. We should probably use some different term to refer to `TypedReference` and similar types - like `restricted types` to avoid confusion.
 
 In this document the "ref like" types include only `Span<T>` and related types. And the ref-like variables mean variables of "ref-like" types.  
 
@@ -25,7 +25,7 @@ In order to force ref-like variables to be stack only,we need the following rest
 - ref-like type cannot be a type of an array element
 - ref-like type cannot be used as a generic type argument
 - ref-like variable cannot be boxed
-- ref-like type cannot be a field type
+- ref-like type cannot be a field of ordinary not ref-like type
 - indirect restrictions, such as disallowed use of ref-like types in async methods, which are really a result of disallowing ref-like typed fields.
 
 Note:
--- a/proposals/span-safety.md
+++ b/proposals/span-safety.md
@@ -45,11 +45,11 @@ Example:
 
 Span<T> Caller()
 {
-	Span<T> safeToReturn;
+	Span<T> safeToReturn = GetSafeToReturnSpan(array);
 
 	if (TodaIsNotFriday())
 	{
-		// callee can "poison" any assumption that we can made about safeToReturn
+		// callee can "poison" any assumption that we have made about safeToReturn
 		Callee(ref safeToReturn);
 	}
 
@@ -59,7 +59,7 @@ Span<T> Caller()
 
 void Callee(ref Span<int> arg1)
 {
-	// local refernces are escaping. 
+	// local references are escaping. 
 	// what rule would disallow this?
 	arg1 = new Span<T>(localloc(...));
 }
@@ -74,7 +74,7 @@ Another way to look at this is that writeable references to a ref-like variable 
 
 - ref-like types must be readonly. 
 
-This is basically follows from the previous "only pass by read-only references" rule when applied to "this" inside members of ref-like types.
+This basically follows from the previous "only pass by read-only references" rule when applied to "this" inside members of ref-like types.
 
 ## not all `ref-like` variables are safe to return. ##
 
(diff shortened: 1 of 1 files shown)


proposals/span-safety.md
```diff
@@ -0,0 +1,102 @@
+# Compile time enforcement of safety for ref-like types #
+
+The main reason for the additional safety rules when dealing with types like `Span<T>` is that such types must be confined to the execution stack.
+ 
+There are three reasons why `Span<T>` must be a stack-only type.
+
+1. `Span<T>` is basically a struct containing `{reference to data, offset limit}`. Concurrent "tearing" of such struct would lead to the possibility of out of range accesses and type-safety violations, which basically means "GC heap corruption".
```
jaredpar: I would give the tuple example here.  Essentially: `Span<T>` is logically a `(ref T data, int length)` structure.

proposals/span-safety.md
```diff
@@ -0,0 +1,102 @@
...
+
+At the minimum the restrictions described below would need to apply to `Span<T>`.
+
+## `ref-like` types must be stack-only. ##
+
+C# compiler already has a concept of a restricted types that covers special platform types such as TypedReference. In some cases those types are referred as ref-like types as well. We should probably use some different term to refer to `TypedReference` and similar types - like `restricted types` to avoid confusion.
+
+In this document the "ref like" types include only `Span<T>` and related types. And the ref-like variables mean variables of "ref-like" types.  
+
+In order to force ref-like variables to be stack only,we need the following restrictions:
```
jaredpar: Need a space after comma

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+- it is ok to pass a ref-like variable by reference to a call as long as the reference is **_readonly_** . 
+The reason is that exposing ref-like variables via writeable references make it nearly impossible to reason about their safety.
+
+Example:
+
+```cs
+
+Span<T> Caller()
+{
+	Span<T> safeToReturn Span<T> safeToReturn;
```
pakrym: Duplication.

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+		Callee(ref safeToReturn);
+	}
+
+	// is this safe?
+	return safeToReturn;
+}
+
+void Callee(ref Span<int> arg1)
+{
+	// local refernces are escaping.
```
jkotas: Typo: refernces

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+
+All the above problems would be alleviated if instances of `Span<T>` would be allowed as stack data types only. In fact for the `#3` we need a stronger guarantee. Details on that to follow below.
+
+An additional problem arises due to composition. It would be generally desirable to build more complex data types that would embed `Span<T>` instances. Such composite types would have to be structs and would share all the hazards and requirements of `Span<T>`. As a result the safety rules described here should be viewed as applicable to the whole range of **_ref-like types_**
+
+At the minimum the restrictions described below would need to apply to `Span<T>`.
+
+## `ref-like` types must be stack-only. ##
+
+C# compiler already has a concept of a restricted types that covers special platform types such as `TypedReference`. In some cases those types are referred as ref-like types as well. We should probably use some different term to refer to `TypedReference` and similar types - like `restricted types` to avoid confusion.
```
ahsonkhan: > a restricted types

Should be `a restricted type`

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+void Callee(ref Span<int> arg1)
+{
+	// local references are escaping. 
+	// what rule would disallow this?
+	arg1 = new Span<T>(localloc(...));
+}
+
+```
+
+Note: returning by a ordinary writeable reference appears to be ok by itself, but with values on heap being impossible, values in local frame being not returnable and variables passed from the caller being readonly, it may not matter.
```
ahsonkhan: > a ordinary

`an ordinary`

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. `Span<T>` is semantically a struct containing a reference and a range - `(ref T data, int length)`. Regardless of actual implementation, writes to such struct would not be atomic. Concurrent "tearing" of such struct would lead to the possibility of `length` not matching the `data`, causing out-of-range accesses and type-safety violations, which ultimately could result in GC heap corruption in seemingly "safe" code.
+2. Some implementations of `Span<T>` literally contain a managed pointer in one of its fields. Managed pointers are not supported as fields of heap objects and code that manages to put a managed pointer on the GC heap typically crashes at JIT time.
+3. It is permitted for a `Span<T>` to refer to data in the local stack frame - individual local variables or `stackalloc`-ed arrays. A scenario when an instance of a `Span<T>` outlives the referred data would lead to undefined behavior, including type-safety violations and heap corruptions.
+
+All the above problems would be alleviated if instances of `Span<T>` would be allowed as stack data types only. In fact for the `#3` we need a stronger guarantee. Details on that to follow below.
```
ahsonkhan: > Details on that to follow below. 

Can you add a link/reference to the "details"?

jkotas:
BTW: Any thoughts on how the ref-like types are going to be identified?

proposals/span-safety.md
jkotas: There is also `byref-like type` defined in ECMA spec. I think that one is fine - we are just making it to cover more types and relaxing the rules.

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+C# compiler already has a concept of a restricted types that covers special platform types such as `TypedReference`. In some cases those types are referred as ref-like types as well. We should probably use some different term to refer to `TypedReference` and similar types - like `restricted types` to avoid confusion.
+
+In this document the "ref like" types include only `Span<T>` and related types. And the ref-like variables mean variables of "ref-like" types.  
+
+In order to force ref-like variables to be stack only, we need the following restrictions: 
+
+- ref-like type cannot be a type of an array element
+- ref-like type cannot be used as a generic type argument
+- ref-like variable cannot be boxed
+- ref-like type cannot be a field of ordinary not ref-like type
```
MichalStrehovsky: Should "*ref-like type cannot be a static field of any type*" be added to this list?

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+Note:
+
+- it is ok to allow returning ref-like values from methods (additional rules apply when referring to local data)
+- it is ok for a ref-like type be a field of another struct as long as the struct itself is ref-like.
+
+
+## ref-like variables can be passed as `in` parameters ##
+
+- it is ok to pass a ref-like variable by reference to a call as long as the reference is **_readonly_** . 
+The reason is that exposing ref-like variables via writeable references make it nearly impossible to reason about their safety.
```
MichalStrehovsky: Do we need any special rules for unmanaged pointers? When I type `unsafe`, I'll still be able to e.g. `stackalloc` an array of spans (useful), or shoot myself in the foot (not so useful), like I can with `TypedReference`, right?

```csharp
using System;

unsafe class Program
{
    static void Leak(TypedReference* tr)
    {
        int i = 456;

        // Whoops
        *tr = __makeref(i);
    }

    static void Main()
    {
        TypedReference tr;
        Leak(&tr);
        Console.WriteLine(__refvalue(tr, int));
    }
}
```

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+
+Reassignment of ref-like variables to new values can easily circumvent scoping/life-time invariants. A particular concern is when a ref-like variable with a bigger span is assigned a ref-like variable that refers to a variable from an inner scope.
+It would be even worse if the LHS is a returnable variable, but the RHS is not a returnable value.  
+
+There are ways to prevent such problems:
+
+The most permissive approach would be alias/value tracking via fixed point data-flow analysis. It is very expensive and complex approach to be practical, but will be called out here for completeness.
+
+Simpler solutions are:
+- make ref-like variables "single-assignment" similarly to the solution we have for ordinary ref variables.
```
KrzysztofCwalina: I don't think we can make them single assignment. It's super common to do span = span.Slice(count);

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+
+```
+
+Note: returning by a ordinary writeable reference appears to be ok by itself, but with values on heap being impossible, values in local frame being not returnable and variables passed from the caller being readonly, it may not matter.
+
+Another way to look at this is that writeable references to a ref-like variable would never be exposed outside of the containing method, except, of course, if that method is a constructor which has to take `this` by a writeable ref.
+
+## `ref-like` types must be readonly structs. ##
+
+- ref-like types must be readonly.
```
ektrah: Is it safe to have a ref-like type that is not readonly if all ref-like fields are readonly?

Use case: I'm passing around mutable structs by reference for incrementally reading from a ReadOnlySpan ([struct](https://github.com/ektrah/nsec/blob/master/src/Cryptography/Formatting/Asn1Reader.cs), [usage](https://github.com/ektrah/nsec/blob/master/src/Cryptography/ChaCha20Poly1305.cs#L191-L203)) / incrementally writing to a Span ([struct](https://github.com/ektrah/nsec/blob/master/src/Cryptography/Formatting/Asn1Writer.cs), [usage](https://github.com/ektrah/nsec/blob/master/src/Cryptography/ChaCha20Poly1305.cs#L205-L213)).

VSadov: @ektrah - I need to think about it, but at first glance it seems unsafe. Any method on a non-readonly struct would get "this" by a read-write reference and as such can completely replace it with anything:

```c#

SomeMethod()
{
    // we can check if r1 is safe to return based on `arg1` and `arg2`
    Asn1Reader r1 =  GetReader(arg1, arg2);

    r1.Foo();

    // can we return or even use r1 after the call above?
    return r1;
}

struct Asn1Reader
{
      public void Foo()
      {
            // this will invalidate the assumptions that caller could make about `s1`
            this = new Asn1Reader(new ReadonlySpan<byte>(stackalloc byte[16]));
      }
}
```

ektrah: @VSadov - `stackalloc` requires `Foo` to be `unsafe`, which probably means that `Foo` can do unsafe things with `this` even if the type is readonly. This might be a better example:

```csharp
struct MyStruct
{
    readonly ref int _x;

    public MyStruct(ref int x)
    {
        ref _x = ref x;
    }

    public void Foo()
    {
        int y = 123;
        this = new MyStruct(ref y);
    }

    public MyStruct Bar()
    {
        int y = 123;
        return new MyStruct(ref y);
    }
}
```

To me it seems that assigning to `this` should follow the same rules as returning a `ref`, i.e., the assignment in `Bar` should be rejected for the same reason as the return statement in `Baz`.

proposals/span-safety.md
```diff
@@ -0,0 +1,104 @@
...
+
+In this document the "ref like" types include only `Span<T>` and related types. And the ref-like variables mean variables of "ref-like" types.  
+
+In order to force ref-like variables to be stack only, we need the following restrictions: 
+
+- ref-like type cannot be a type of an array element
+- ref-like type cannot be used as a generic type argument
+- ref-like variable cannot be boxed
+- ref-like type cannot be a field of ordinary not ref-like type
+- indirect restrictions, such as disallowed use of ref-like types in async methods, which are really a result of disallowing ref-like typed fields.
```
svick: What about edge cases, where ref-like types could be allowed? For example:

1.  Values in async methods that are immediately passed to another method. E.g. even inside an async method, `F(GetRefLike())` or `F(await task, GetRefLike())` could be allowed (though `F(GetRefLike(), await task)` couldn't).
2. Locals in async methods with no awaits.
3. Locals that are part of a closure of a "simple" local function (i.e. one whose closure is a `struct`, not a `class`).

I'm not sure what would be the value in allowing such cases, especially since the rules could become byzantine and closely tied to implementation details of the compiler. But I think it's worth considering.

proposals/span-safety.md
VSadov: Yes, ref-like type cannot be a type of a static field. It can only be an instance field of another ref-like type.

proposals/span-safety.md
VSadov: Unsafe code is a tough subject. Some unsafe code has well-defined behavior even if not safe.  Your example may actually run just fine depending on implementation. 
C# as a language generally does not have a lot of guarantees about unsafe code.

I'd say the sample is legal, but its behavior is implementation specific and thus undefined.

proposals/span-safety.md
VSadov: @KrzysztofCwalina - would it be ok to allow this only for spans that are free from referring to local data?

With local data you would have to do  'var anotherSpan = span.Slice(count);` 

The reason is that inside a method could be many nested scopes and it would be very hard to enforce that variables from outer scopes do not end up referring indirectly to variables from inner scopes if they can be reassigned.  
On the other hand, everything outside of the method lives as long as you have a reference to it, so it is safe to assign to anything.

proposals/span-safety.md
VSadov: @ektrah  - your examples are correct.

The idea with stacalloc is to have some syntax that would allow wrapping stack allocated data into a span (without leaking the ptr of course). Such construct could be used without "unsafe", since it would be type-safe, but we would need to make the resulting span not-returnable.

`  new ReadonlySpan<byte>(stackalloc byte[16]) ` is one possible example of such syntax. It is TBD though.

proposals/span-safety.md
VSadov: I think we can look at preexisting behavior of types like TypedReference here. To make rules simple we may outright disallow ref-like types in async methods. Disallowing just the cases when ref-like is captured could lead to cases where something works in optimized code and does not in debug.

proposals/span-safety.md
VSadov: @jkotas - any suggestions? 
Ideally it should be something that makes Span unconsumable by old code - like a `modreq`, but it is not possible to put `modreq` on a type.

VSadov: We can use an ordinary attribute, but it is easily ignorable.

proposals/span-safety.md
ektrah: @VSadov - A safe `stackalloc` would be nice. I'm already using unsafe `stackalloc` with `Span<T>` quite a few times: [1](https://github.com/ektrah/nsec/blob/a29d00598a58f5963e7e62940284f48306d929be/src/Cryptography/Formatting/KeyFormatter.cs#L75-L79), [2](https://github.com/ektrah/nsec/blob/a29d00598a58f5963e7e62940284f48306d929be/src/Cryptography/Sha512.cs#L72-L76), [3](https://github.com/ektrah/nsec/blob/a29d00598a58f5963e7e62940284f48306d929be/src/Cryptography/HkdfSha256.cs#L121-L125), [4](https://github.com/ektrah/nsec/blob/a29d00598a58f5963e7e62940284f48306d929be/src/Cryptography/HmacSha256.cs#L138-L142), [5](https://github.com/ektrah/nsec/blob/a29d00598a58f5963e7e62940284f48306d929be/src/Cryptography/Ed25519.cs#L109-L113).

And it would really, really nice if it was possible to encapsulate a ref-like type together with mutable state, as I currently do in the structs linked above.

proposals/span-safety.md
jkotas: IIRC, the solution we have discussed was to mark Span with ObsoleteAttribute with special message pattern, and make newer compilers to not emit warnings for ObsoleteAttribute with this special pattern.

VSadov:
Based on feedback, the rules around passing span like types byref have been refined.
Because of that some of the feedback here has become inapplicable. The rest have been applied to the updated draft. 

Since the discussion here is fairly long and concerns are captured in the updated document, I am closing this to start new discussion here:
https://github.com/dotnet/csharplang/pull/472

PR #270, opened 2017-03-14
Change reachability

agocke:
I think local functions bodies should only be considered reachable if they are used by a reachable expression.

### DIFF
--- a/proposals/csharp-7.0/local-functions.md
+++ b/proposals/csharp-7.0/local-functions.md
@@ -67,4 +67,5 @@ using local functions with capturing.
 
 We add to the spec
 
-> The body of a statement-bodied lambda expression or local function is considered reachable.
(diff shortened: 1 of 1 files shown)


agocke:
cc @gafter

gafter:
@agocke There is no concept of "reachable expression" in the spec.

gafter:
@agocke What do you want to do with this PR?

agocke:
@gafter Sorry, didn't see your comment before. How about, "if there is a call or delegate conversion to the local function in a reachable statement?

gafter:
@agocke Sounds reasonable, but I suspect that isn't something we implement today. How would we handle `M(false && LocalFunction())`?

agocke:
@gafter Yup, you're right, I bet that's considered unreachable. I'm gonna close this and think about it some more.

PR #286, opened 2017-03-18
Fix invalid character in unsafe-code.md

niklaskarl:
A invisible control character prevented the title from being displayed correctly.

### DIFF
--- a/spec/unsafe-code.md
+++ b/spec/unsafe-code.md
@@ -1,4 +1,4 @@
-﻿# Unsafe code
+# Unsafe code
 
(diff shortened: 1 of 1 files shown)


svick:
It's not just unsafe-code, it seems all the other documents contain the invisible BOM character too.

gafter:
The title looks OK to me: https://github.com/dotnet/csharplang/blob/master/spec/unsafe-code.md

svick:
@gafter This PR was made shortly after [GitHub switched to new markdown parser](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown), so maybe it was a bug in it that was since fixed?

gafter:
Since there no longer appears to be a problem here, closing.

Issue #287, opened 2017-03-19
Champion "CallerArgumentExpression"


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

See https://github.com/dotnet/csharplang/issues/205

HaloFour:
If expansion of the supported caller info attributes is going to be discussed in the LDM is there a chance that some quantity of #87 be mentioned as well?

CyrusNajmabadi:
@HaloFour I think that would be reasonable.  If we're going to do more in this space, we probably want to do it all at once.

jamesqo:
Related corefx proposal that would take advantage of this feature: dotnet/corefx#17068

jamesqo:
@CyrusNajmabadi 

> I think that would be reasonable. If we're going to do more in this space, we probably want to do it all at once.

Do I need to modify the proposal I'm working on to include his suggestions, or will they be reviewed just by looking at the issue?

gafter:
My intent is to bring up this one request to the design group. It sounds like @CyrusNajmabadi will bring up the others in that discussion.

alrz:
Related: https://github.com/dotnet/csharplang/issues/443 

```cs
void M(object param, [CallerArgumentExpression(nameof(param))] string arg = null) {}
```

jamesqo:
@alrz I agree with the proposal, but it isn't strictly necessary for this to move forward. The end user will not have to touch this attribute. Only a handful of methods in the framework will use it, and for those it's fine to simply hard-code the string.

jamesqo:
Great to see that this has been discussed in the LDM, I wish I'd commented just 2 days later so I would have seen that :smile: I see this is the text for the `CallerArgumentExpression` section:

> Push to 7.X and start process for getting the attribute in place.

Does that mean I should open a proposal in the corefx repo?

jamesqo:
I opened a corefx proposal for this, see https://github.com/dotnet/corefx/issues/21809.

paulomorgado:
Call me strange, but I still don't feel comfortable with optional arguments in public APIs.

To the proposal in question, the callee shouldn't be required to know anything about the caller. And the caller should know what it was doing just by the call stack of the exception. Otherwise, you have worst problems in your hands and this is not the solution.

alrz:
Caller attributes are not a new concept so I think we should be past that argument by now.

[...]

Issue #205, opened 2017-02-28
Proposal: Allow expressions of method parameters to be stringified


jamesqo:
## Motivation

When an exception is thrown or an assert/unit test fails, it's not always obvious what went wrong just from looking at the stack trace. Part of the reason for this is that you can't *see* the condition that failed. If you write `Debug.Assert(array != null)` and it fails, the stack trace will only contain the filename and the line number. You will have to open up a text editor and navigate to that file to see what went wrong. It would be ideal if `Debug.Assert` actually showed you the string `"array != null"` in the failure message.

## Proposal

Add a `CallerArgumentExpression` attribute allowing callees to 'stringify' the expressions passed in at a callsite. The constructor of the attribute will take an `string` argument specifying the name of the argument to stringify.

```cs
public static class BetterDebug
{
    public static void Assert(bool condition, [CallerArgumentExpression("condition")] string expression = null)
    {
        Debug.Assert(condition, expression);
    }
}

public void MyArrayMethod(T[] array)
{
    BetterDebug.Assert(array != null); // Upon failure, the message will be "array != null"
}
```

- The attribute will work the same as other `Caller*` attributes like `CallerMemberName`, `CallerFilePath`, etc. The compiler will replace the optional parameter with the string of the expression of the argument having the name specified in the attribute.

- The compiler will verify that the name is valid at compile-time, e.g. `[CallerArgumentExpression(null)]` or `[CallerArgumentExpression("cnodition")]` in this example will raise an error.

The use cases for this are not limited to `Debug.Assert`. Consider the possibility of creating a helper class for argument validation without needing to use `nameof`:

```cs
public static class Requires
{
    public static void NotNull<T>(T argument, [CallerArgumentExpression("argument")] string expression = null)
    {
        if (argument == null) throw new ArgumentNullException(expression);
    }

    public static void Range(bool condition, [CallerArgumentExpression("condition")] string expression = null)
    {
        if (!condition) throw new ArgumentOutOfRangeException(message: $"{expression} failed.");
    }

    public static void CanRead(Stream stream, [CallerArgumentExpression("stream")] string expression = null)
    {
        if (!stream.CanRead) throw new InvalidOperationException(message: $"{expression} can't read.");
    }
}
```

Now we can validate parameters as follows

```cs
public static void MyHighPerfByteArrayStreamMethod(
    byte[] buffer, int index, int count, Stream stream)
{
    // No nameof! Each parameter name only has to be typed once.
[...]

PR #347, opened 2017-03-25
Add proposal for CallerArgumentExpression attribute

jamesqo:
This is the official proposal for https://github.com/dotnet/csharplang/issues/205, which was championed in https://github.com/dotnet/csharplang/issues/287.

/cc @gafter, @tannergooding, @HaloFour

### DIFF
--- a/proposals/caller-argument-expression.md
+++ b/proposals/caller-argument-expression.md
@@ -1,0 +1,241 @@
+# CallerArgumentExpression
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Allow developers to capture the expressions passed to a method, to enable better error messages in diagnostic/testing APIs and reduce keystrokes.
+
+## Motivation
+[motivation]: #motivation
+
+When an assertion or argument validation fails, the developer wants to know as much as possible about where and why it failed. However, today's diagnostic APIs do not fully facilitate this. Consider the following method:
+
+```cs
+T Single<T>(this T[] array)
+{
+    Debug.Assert(array != null);
+    Debug.Assert(array.Length == 1);
+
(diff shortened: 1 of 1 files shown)


proposals/caller-argument-expression.md
```diff
@@ -0,0 +1,245 @@
...
+        var callerInfo = new CallerInfo();
+        callerInfo.FilePath = @"C:\Bar.cs";
+        callerInfo.Method = MethodBase.GetCurrentMethod();
+        callerInfo.ArgumentExpressions = new string[] { "false" };
+        Debug.Assert(false, callerInfo);
+    }
+}
+```
+
+This was originally proposed at https://github.com/dotnet/csharplang/issues/87.
```
jamesqo: @HaloFour, I liked your idea of having a `CallerInfo` struct to enscapulate all the necessary information about the caller. I took your idea and expanded upon it a bit in the alternatives section of this proposal.

proposals/caller-argument-expression.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+A proposal to add such a helper class to the framework is underway at https://github.com/dotnet/corefx/issues/17068. If this language feature was implemented, the proposal could be updated to take advantage of this feature.
+
+### Extra details
+
+- Like the other `Caller*` attributes, such as `CallerMemberName`, this attribute may only be used on parameters with default values.
+- Multiple parameters marked with `CallerArgumentExpression` are permitted, as shown above.
+- The attribute's namespace will be `System.Runtime.CompilerServices`.
+- In the case `null` or a string that does not correspond to a parameter name (e.g. `"cnodition"`) is provided, an error will be raised during compilation.
```
jnm2: In the case **of** null and `"cnodition"` -> `"condition"`

jnm2: Missed that this was merged, so https://github.com/dotnet/csharplang/pull/537

PR #375, opened 2017-03-28
Add proposal for attributes on lambdas

stephentoub:


stephentoub:
> If attributes will be allowed on lambdas, should they be also allowed on local functions (including their parameters and type parameters)?

I think they should (in particular I really want to be able to write local DllImports), but I also think it should be treated separately from attributes on lambdas.

> Also, I assume adding an attribute on the return value of a lambda will work using the normal [return: SomeAttribute] syntax?

I would hope so.  I've added a comment to that effect.

proposals/lambda-attributes.md
```diff
@@ -0,0 +1,106 @@
...
+- Multiple attributes may be applied to the parameters of an anonymous method or lambda, using either the comma-delimited or full-attribute syntax, e.g.
+
+```C#
+([FirstAttribute, SecondAttribute] i) => { ... } // Ok
+([FirstAttribute] [SecondAttribute] i) => { ... } // Ok
+```
+
+- `return`-targeted attributes may also be used on lambdas, e.g.
+```C#
+([return: SomeAttribute] i => { ... }) // Ok
```
alrz: Per line 69, this should be `([return: SomeAttribute] (i) => { ... }) // Ok`

stephentoub: Yes, thanks, typo.

alrz:
- Should we extend `AttributeTargets` to include lambda, local function, etc?
- The use cases mentioned here doesn't need attributes to be emitted to the assembly but perhaps "source-only attributes " aka "attribute everywhere" is left for a separate proposal. (?)

HaloFour:
@alrz 

> * Should we extend `AttributeTargets` to include lambda, local function, etc?

Considering that neither lambdas nor local functions exist as far as the CLR concerns I don't think it makes a lot of sense to change that shared enum.

I'd rather a well defined set of rules as to which and how attributes can be applied to those constructs in the language and then what the language does with them, if anything.  The language could introduce new targeting keywords to eliminate ambiguity, but I don't think that would be necessary.

alrz:
@HaloFour 
>Considering that neither lambdas nor local functions exist as far as the CLR concerns I don't think it makes a lot of sense to change that shared enum.

Right. It could be even a breaking change because of `AttributeTargets.All`. 

I think that would make more sense for source-only attributes, as for `CaptureNone` example (1) it doesn't need to be emitted in the assembly, and (2) it doesn't make sense on anything other than lambdas and local functions so ideally, it should be used as a source-only attribute and also limited to only lambdas and local functions (which makes the former a requirement).

alrz:
Apparently there isn't a "champion" issue for this proposal.

PR #381, opened 2017-03-29
Add proposal for readonly locals/parameters

stephentoub:
https://github.com/dotnet/csharplang/issues/188
https://github.com/dotnet/roslyn/issues/115

### DIFF
--- a/proposals/readonly_locals.md
+++ b/proposals/readonly_locals.md
@@ -1,0 +1,92 @@
+# readonly locals and parameters
+
+* [x] Proposed
+* [ ] Prototype
+* [ ] Implementation
+* [ ] Specification
+
+## Summary
+[summary]: #summary
+
+Allow locals and parameters to be annotated as readonly in order to prevent shallow mutation of those locals and parameters.
+
+## Motivation
+[motivation]: #motivation
+
+Today, the `readonly` keyword can be applied to fields; this has the effect of ensuring that a field can only
+be written to during construction (static construction in the case of a static field, or instance construction in the case of an instance field),
+which helps developers avoid mistakes by accidentally overwriting state which should not be modified. But fields aren’t the only places developers
+want to ensure that values aren’t mutated.  In particular, it’s common to create a local variable to store temporary state, and accidentally updating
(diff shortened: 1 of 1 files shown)


alrz:
Would all this be applied for the `let` shorthand? (from https://github.com/dotnet/roslyn/issues/16182)
```cs
// ref local
ref var x = e;
ref let x = e;
// deconstruction
var (x, y) = e;
let (x, y) = e;
(var x, var y) = e;
(let x, let y) = e;
// out var
M(out var x);
M(out let x);
// deconstructing out var
M(out var (x, y));
M(out let (x, y));
// just for clarity
foreach (let x in arr)
using (let x = e)
// ref iterator variable
foreach(ref var x in arr)
foreach(ref let x in arr)
// variable patterns
e is var x
e is let x
case var x:
case let x:
// lambda parameters
(x, y) => {};
(let x, let y) => {};
// deconstructing lambda parameters
((x, y)) => {};
(let (x, y)) => {};
````

stephentoub:
> // ref local
> // ref iterator variable

I've explicitly left `ref` out of this proposal, as noted in the end section.  That's not a comment on whether I believe it should be supported, but that I think ref readonly is its own standalone follow-on feature to this.

> // deconstruction
> // out var
> // deconstructing out var
> // variable patterns
> // lambda parameters
> // deconstructing lambda parameters

I would expect all of these to be supported unless there's a specific reason found they can't be.  Basically anywhere you can currently define `T local` or `var local`, I would expect you'd be able to use `readonly T local` or `let local`.

> // just for clarity

I'm not sure what the value to supporting this is.

alrz:
>I'm not sure what the value to supporting this is. 

Just to make it more visible and consistent over all `let` and `var` locals. I think an analyzer could also suggest to use `let` in those places.

svick:
> This would have no effect on what the caller of the method is able to pass to the parameter

I take it the intention is that `readonly` for parameters will have no effect whatsoever on code outside the method? It is not represented in metadata in any way, a method with `readonly` can override method without `readonly` and vice versa, etc.

Is is worth pointing this out in the proposal?

PR #405, opened 2017-04-04
Update default interface methods proposal

gafter:


proposals/default-interface-methods.md
```diff
@@ -1,43 +1,227 @@
...
+    override abstract void M();
+}
+class C : IB { } // error: class 'C' does not implement 'IA.M'.
+```
+
+The `abstract` modifier is not required in the declaration of `IB.M` (that is the default in interfaces), but it is probably good practice to be explicit in an override declaration.
+
+This is useful in derived interfaces where the default implementation of a method is inappropriate and a more appropriate implementation should be provided by implementing classes.
+
+> ***Open Issue:*** Should reabstraction be permitted?
```
gafter: We discussed this and decided "NO", however I've reopened this as the decision contradicts the goals of the feature, as described in https://github.com/dotnet/csharplang/issues/406

PR #408, opened 2017-04-05
Added alternative to the proposal.

Grauenwolf:


gafter:
This proposed alternative only addresses one of the motivations for the proposal.

PR #440, opened 2017-04-15
Add proposal for property backing field feature.

anders-liu:


gafter:
Proposals from the community are welcome as new issues. The LDM creates proposal documents for features that have at least one LDM member to "champion" it.

PR #462, opened 2017-04-19
Modify default interface methods proposal based on 2017-04-18 LDM

gafter:


gafter:
@AlekseyTs @MadsTorgersen FYI

proposals/default-interface-methods.md
```diff
@@ -138,6 +138,11 @@ interface IB : IA
 
 Override declarations in interfaces may not be declared `sealed`.
 
+Public `virtual` function members in an interface may be overriden in a derived interface either implicitly (by using the `override` modifier in a public declaration in the derived interface) or explicitly (by qualifying the name in the override declaration with the interface type that originally declared the method, and omitting an access modifier).
```
AlekseyTs: > by using the `override` modifier in a public declaration in the derived interface

Should the accessibility instead match the member being overridden? We need to think more about implicit overrides, we might not want them to appear as a new member in the derived interface.

gafter: This paragraph is about public members. Matching the accessibility of a public member means the overriding member is public. Are you suggesting there should be an open question whether an implicit override in an interface introduces a new member?

The next paragraph is about other access levels. According to this specification and yesterday's LDM, there are no implicit overrides except for public members.

AlekseyTs: > Are you suggesting there should be an open question whether an implicit override in an interface introduces a new member?

Yes, thanks.

AlekseyTs:
LGTM

PR #472, opened 2017-04-21
Updated draft for the safety of ref-like types such as `Span<T>`

VSadov:


proposals/span-safety.md
```diff
@@ -0,0 +1,188 @@
...
+- ref-like type cannot be a static field
+- ref-like type cannot be an instance field of ordinary not ref-like type
+- indirect restrictions, such as disallowed use of ref-like typed parameters and locals in async methods, which are really a result of disallowing ref-like typed fields.
+
+Note:
+
+- it is ok to return ref-like values from methods/lambdas, including by reference.
+- it is ok to pass ref-like values to methods/lambdas, including by reference.
+- it is ok for a ref-like type be a field of another struct as long as the struct itself is ref-like.
+- it is ok for an intermediate span values to be used in lambdas/async methods.
```
davkean: This needs explaining - you've not explained "intermediate span values" previously.

proposals/span-safety.md
```diff
@@ -0,0 +1,188 @@
...
+
+- "What comes in is what comes out" rule for span-returning expressions:
+  If any operand of an expression is a stack-referring value, the result is a stack-referring value.
+  This applies equally to ternary expressions, indexers, method calls, and so on. 
+  Receiver of an invocation/property/indexer is considered an operand here.
+ 
+  In particular: _Slice of a stack-referring value is a stack-referring value._  
+
+- "No Mixing with refs to ordinary" rule: 
+  Stack-referring spans can be passed as an argument as long as no other argument is an ordinary span variable (not a stack-referring local), and:
```
davkean: What's the rationale for this? It's not clear to me (not being involved in this, why this rule exists)

proposals/span-safety.md
VSadov: Basicaly   `M1(M2())` is ok in an async method, even if `M2()` returns a span. Will add an example.

proposals/span-safety.md
VSadov: If we classified one variable as safe to return and another variable as stack-referring, we will need to prevent assignments from second to the first, since that would invalidate the classification (and we do not want to do global data flow analysis - it is hardly tractable).

The goal here is to prevent such assignments while not causing too much inconvenience. The call boundary is an issue though.

If I make a call like `M1(p1: ref safeToReturn)`, I am risking that `M1` will do `p1 = stackalloc int[10];`

In a more complex case I am making a call like `M2(p1: ref safeToReturn, p2: unsafeToReturn)` and `M2` does `p1 = p2;`.  

We need to make both scenarios illegal.  
The original proposal simply disallowed passing spans by reference. It was found too restricting so we need something else.

In this proposal assigning stack-referring values to parameters (or to anything other than stack-returning locals) is disallowed. That handles the `p1 = stackalloc int[10];`case.

The other case is harder. Inside `M1` we cannot tell when assigning second parameter to the first is unsafe, so we disallow at the call side passing an ordinary span variable by reference and any kind of stack-referring spans in the same call, so that `M1` could assign one parameter to another without worrying.

Thanks for commenting. I will add these details to the document.

jkotas:
We are going to need attribute to mark ref-like types - to avoid spooky action at a distance when Span is embedded in other structs.

VSadov:
@jkotas  - yes, we will use [Obsolete] with a well-known message in metadata. That will also allow us to error on compilers that are not aware of span requirements.

And in the source we will use `ref struct S1 `

We had some follow up discussions on that. I will add that to the document.

jkotas:
`[Obsolete]` with a well-known message may be fine to poison the type for old compilers; but I do not think that it should be the ultimate way to identify these types. Note that these types need to be identified accross the board - in the runtime, tools, etc.

VSadov:
It is possible to add another attribute, but it would always have to be combined with the Obsolete.
It is possible, but seems a bit redundant.

VSadov:
Note that the Obsolete message must be precise or we cannot distinguish it from an ordinary obsolete attribute.

It does feel that having two attributes having distinct meanings is more direct way of doing this, but there is also the part that they will have to be always emitted in pairs makes it seem a bit redundant.

We should think about this. 
There is nothing inherently wrong with having "IsRefStructAttribute"

proposals/span-safety.md
```diff
@@ -37,6 +37,57 @@ Note:
...
+	public Span<T> first;
+	public Span<T> second;
+} 
+
+// error: arrays of ref-like types are not allowed. 
+TwoSpans<T>[] arr = null;
+
+``` 
+
+Designating a struct as ref-like will allow the struct to have ref-like instance fields and will also make all the requirements of ref-like types applicable to the struct.
```
svick: Will it somehow be possible to create a ref-like struct that contains `ref` fields?

proposals/span-safety.md
```diff
@@ -0,0 +1,238 @@
...
+## Stack-referring spans in safe code.
+
+There are two main scenarios where stack-referring spans could come into being.  
+- Stackalloc spans.
+  To enable safe wrapping of stack allocated memory in a span, we will need a special syntax that would not allow intermediate pointer to escape.
+  A preliminary solution is to use target-typing.
+
+```cs
+  //This usage of stackalloc does not require unsafe 
+  Span<int> sp = stackalloc int[100];
```
mgravell: Probably simply "no" (and that's fine), but: is there any utility to exposing something akin to array initializer syntax here?

    Span<int> sp = stackalloc { 1, 42, 13 };

(Note preserving the `stackalloc` keyword to retain the intent in the code)

mikedn:
Can perhaps `modreq` be used instead of this `Obsolete` based hack? It seems that the main problem with an old compiler using these types is the possibility of exposing references to stack locations that have been "freed" and this can only happen via method calls.

Other improper uses of ref-like types will be blocked by the runtime. Sure, it would be it would be nice to be blocked at compile time but using `Obsolete` for that is quite ugly.

proposals/span-safety.md
mikedn: This is interesting. It doesn't require unsafe but the resulting code is unverifiable. It would seem that we're moving towards a system where the managed compiler(s) are responsible for ensuring type safety and not the runtime.

davkean:
@mikedn It's an interesting idea, and definitely falls in the spirit of what they were designed for (think of some the C++ modifiers; IsUdtReturn, IsByValue, etc).

omariom:
@VSadov 

Will such scenario  be allowed?
```C#

private void M()
{
    var myStruct = new MyStruct(stackalloc int[10]);
    Foo(myStruct);
    Bar(myStruct.ReadOnlyView);
}

private void Foo(MyStruct myStruct)
{
}

private void Bar(ReadOnlySpan<int> roSpan)
{
}

reflike struct MyStruct
{
    private Span<int> span;

    MyStruct(Span<int> span) => this.span = span;
    
    public ReadOnlySpan<int> ReadOnlyView => span.AsReadOnly(); 
}


```

VSadov:
@omariom - not exactly like that, stackalloc can be used only when initializing a local, so you would need one more span-typed variable, but yes, conceptually the above example will work. At least according to the rules in the proposal.

mgravell:
to me, though, it does also prompt the question of whether there should
also be a syntax for initializing a stack-allocated span into a local
*without* requiring "unsafe" (for "stackalloc"). Probably ties more into
#535 than this though - is specific to span.

On 4 Jun 2017 3:40 a.m., "Vladimir Sadov" <notifications@github.com> wrote:

@omariom <https://github.com/omariom> - not exactly like that, stackalloc
can be used only when initializing a local, so you would need one more
span-typed variable, but yes, conceptually the above example will work. At
least according to the rules in the proposal.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/dotnet/csharplang/pull/472#issuecomment-306013922>, or mute
the thread

omariom:
@VSadov 
> stackalloc can be used only when initializing a local, so you would need one more span-typed variable

It should be relaxed for safe (Spans) case.

PR #588, opened 2017-05-16
Correct URL for "Reference assemblies"

mshahrouri:
current Url for "Reference assemblies" point to a not fond URL, I've just added to correct Url


gafter:
Thanks!

PR #610, opened 2017-05-20
Corrected patterns.md to use discard pattern ('_') in examples

lachbaer:
Corrected patterns.md to use discard pattern ('_') in examples instead of star (`*`).


gafter:
Thanks!

PR #620, opened 2017-05-25
Fixed broken link for reference assemblies link in the language version history

DavidArno:
The link to the reference assemblies spec for v7.1 is broken. Fixed.


DavidArno:
This is a duplicate PR, so closing.

PR #635, opened 2017-05-30
[Small Fix] Typos In Documentation

BraedonWooding:
In the README.md file under the spec folder location there are two typos, they are just URL link typos, that didn't actually link to the right section, which annoyed me a little so this is just a quick fix for those typos :D.


gafter:
Thank you!

Issue #570, opened 2017-05-11
Champion "Non-trailing named parameters"


jcouv:
- [ ] Proposal added 
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

Relates to discussion https://github.com/dotnet/csharplang/issues/518

jcouv:
Prototype started:
* https://github.com/jcouv/roslyn/tree/named-parameters (C#)
* https://github.com/jcouv/roslyn/tree/named-parameters-vb (VB)

jcouv:
The feature is merged into master branch for 7.2 (Visual Studio 2017 version 15.5).
https://github.com/dotnet/roslyn/pull/20805

PR #638, opened 2017-05-30
Proposal for "non-trailing named arguments"

jcouv:
This is a proposal for championed issue https://github.com/dotnet/csharplang/issues/570

@dotnet/roslyn-compiler @MadsTorgersen for review.


proposals/non-trailing-named-arguments.md
```diff
@@ -0,0 +1,82 @@
...
+
+- The status quo
+- Providing IDE assistance to fill-in all the names of trailing arguments when you type specific a name in the middle.
+
+Both of those suffer from more verbosity, as they introduce multiple named arguments even if you just need one name of a literal at the beginning of the argument list.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- implications for F#
```
jcouv: @dsyme @cartermp This feature may encourage parameters in APIs to be ordered more "badly", so it is good to think about the implications for F#.
I'm trying to line up this feature with C# 7.2, so we have some time to discuss (we're just about to escrow C# 7.1).

proposals/non-trailing-named-arguments.md
```diff
@@ -0,0 +1,82 @@
...
+
+A function member is said to be an applicable function member with respect to an argument list A when all of the following are true:
+* [... existing rules ...]
+* For each named argument in A, if its corresponding parameter (as described in §7.5.1.1) does not have the same position as that argument, then all the following arguments must be named.
+
+In particular, this prevents invoking `void M(bool a = true, bool b = true, bool c = true, );` with `M(c: false, valueB);`. The first argument is used non-positionally (the argument is used in first position, but the parameter named "c" is in third position), so the following arguments should be named.
+
+In other words, non-trailing named arguments are only allowed when the name and the position result in finding the same corresponding parameter.
+
+## Detailed design (VB)
```
jcouv: @AnthonyDGreen I'll add VB as well.

CyrusNajmabadi:
Note: as part of this, we would want to fix up completion.  Right now, once you type a single named-argument, then we suppress completion for everything but named-arguments for subsequent positions.  We'd now only want to do the suppression if your named argument did not match the right position.

Also, we would definitely want the IDE feature/code-style to say that literals should have named arguments.  Note the options would likely be: 

Prefer named arguments for literals:

1. Yes
2. When there is more than one argument
3. No

The reason for '2' is that you may often have apis like "frob.SetIsEnabled(true)".  In these cases, you don't really need the named argument as the meaning is clear in context of the method name.

proposals/non-trailing-named-arguments.md
cartermp: I believe that this would make such code uncallable in F# without changes to the F# compiler to allow non-trailing named parameters.  No idea how much work is involved in the F# side of things.

jcouv: I didn't understand the concern. Would you have an example?
The feature does not affect API declarations (you still need to name all *parameters*). 

The only impact to APIs is indirect and a bit speculative (how people choose to order their parameters).

jcouv:
@CyrusNajmabadi I made a list of work items (including the two you mentioned above) at https://github.com/dotnet/roslyn/issues/19890
Let me know if I should file a specific issue for the code-style feature.

proposals/non-trailing-named-arguments.md
cartermp: Had some folks talk about it yesterday, and I see now that there isn't any concern here w.r.t breaking.

PR #661, opened 2017-06-06
Clarify the specified semantics of a throw expression.

### LINKED ISSUE #656: Effect of throw expressions on reachability is unspecified
As described in the commentary on dotnet/roslyn#19984, the impact of throw expressions on reachability analysis is not currently clear. Currently the specification for [throw expressions](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.0/throw-expression.md) only says this:

> Note: the rest of the semantics of the throw expression are identical to the semantics of the *throw_statement* in the current language specification.

I believe that the impact of this new construct on reachability should be explicit rather than implicit, and propose the following changes as part of the Throw expression feature.

### Syntax

The following is a more complete representation of the intended syntax of this feature.

```diff
+throw_expression
+    : 'throw' expression
+    ;
 
 null_coalescing_expression
     : conditional_or_expression
     | conditional_or_expression '??' null_coalescing_expression
+    | conditional_or_expression '??' throw_expression
     ;
 
 conditional_expression
     : null_coalescing_expression
-    | null_coalescing_expression '?' expression ':' expression
+    | null_coalescing_expression '?' expression_or_throw ':' expression_or_throw
     ;
 
+expression_or_throw
+    : expression
+    | throw_expression
+    ;
 
 anonymous_function_body
-    : expression
+    : expression_or_throw
     | block
     ;

 arrow_expression_body // NOTE this doesn't actually appear to exist anywhere yet
-    : '=>' expression ';'
+    : '=>' expression_or_throw ';'
     ;
```

#### Intent

The intent of the semantics for a throw_expression include the expansion of the concept of end point reachability to expressions. The end point of an expression is, in general, reachable unless evaluation of the complete expression necessarily includes the evaluation of a *throw_expression*. For the purposes of this determination, the value of constant expressions are considered for the purposes of determining reachability for conditionally-executed expressions, including:

* The ternary operator, where the condition may be a constant
* Null coalescing expressions, where the left hand side may be a constant
* Conditional logical operators (short-circuiting), where the left hand side may be a constant
* Null conditional operators (`?.` and `?[`)

:memo: While expressions have start and end point reachability as a result of this change, a warning for unreachable subexpressions may be omitted for compatibility purposes. For example, the following code was legal prior to C# 7, and did not report a warning related to reachability:

```csharp
var ignored = true ? Reachable() : Unreachable();
```

#### Details

⚠️ The following is really just notes for now.

* *declaration_statement*
  * The end point of a *declaration_statement* is reachable if the end point of its last *local_variable_initializer* is reachable. If the *declaration_statement* does not include any initializers, then the end point is reachable if the start point is reachable.
* *expression_statement*
  * TODO
* *local_variable_initializer*
  * TODO
* *expression*
  * Unless otherwise specified, the end point of an *expression* is reachable if the start point is reachable.
* *array_initializer*
  * TODO
* *local_variable_initializer_unsafe*
  * TODO
* *throw_expression*
  * The end point of a *throw_expression* is never reachable
* *invocation_expression*
  * The end point of an *invocation_expression* is reachable if it has no *argument_list*, or if the end point of its *argument_list* is reachable.
* *element_access*
  * The end point of an *element_access* is reachable if the end point of its *expression_list* is reachable
*

gafter:
Fixes #656 

@sharwell FYI

### DIFF
--- a/proposals/csharp-7.0/throw-expression.md
+++ b/proposals/csharp-7.0/throw-expression.md
@@ -17,6 +17,10 @@ The type rules are as follows:
 - A *throw_expression* has no type.
 - A *throw_expression* is convertible to every type by an implicit conversion.
 
+A *throw expression* throws the value produced by evaluating the *null_coalescing_expression*, which must denote a value of the class type `System.Exception`, of a class type that derives from `System.Exception` or of a type parameter type that has `System.Exception` (or a subclass thereof) as its effective base class. If evaluation of the expression produces `null`, a `System.NullReferenceException` is thrown instead.
+
+The behavior at runtime of the evaluation of a *throw expression* is the same [as specified for a *throw statement*](https://github.com/dotnet/csharplang/blob/master/spec/statements.md#the-throw-statement).
+
 The flow-analysis rules are as follows:
 
 - For every variable *v*, *v* is definitely assigned before the *null_coalescing_expression* of a *throw_expression* iff it is definitely assigned before the *throw_expression*.
@@ -26,5 +30,3 @@ A *throw expression* is permitted in only the following syntactic contexts:
 - As the second or third operand of a ternary conditional operator `?:`
 - As the second operand of a null coalescing operator `??`
 - As the body of an expression-bodied lambda or method.
-
-> Note: the rest of the semantics of the throw expression are identical to the semantics of the *throw_statement* in the current language specification.


sharwell:
:bulb: It would be good to modify the grammar as well. I'm having a hard time relating it back to an actual change in the spec.

PR #685, opened 2017-06-16
Adding the 'fixed-sized-buffers' proposal.

tannergooding:
FYI. @jaredpar

### DIFF
--- a/proposals/fixed-sized-buffers.md
+++ b/proposals/fixed-sized-buffers.md
@@ -1,0 +1,71 @@
+# Fixed Sized Buffers
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Provide a general-purpose and safe mechanism for declaring fixed sized buffers to the C# language.
+
+## Motivation
+[motivation]: #motivation
+
+Today, users have the ability to create fixed-sized buffers in an unsafe-context. However, this requires the user to deal with pointers, manually perform bounds checks, and only supports a limited set of types (`bool`, `byte`, `char`, `short`, `int`, `long`, `sbyte`, `ushort`, `uint`, `ulong`, `float`, and `double`).
+
+With a few minor tweaks, we could provide general-purpose fixed-sized buffers which support any type, can be used in a safe context, and have automatic bounds checking performed.
+
+## Detailed design
+[design]: #detailed-design
+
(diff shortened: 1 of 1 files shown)


proposals/fixed-sized-buffers.md
```diff
@@ -0,0 +1,72 @@
...
+
+```C#
+public fixed DXGI_RGB GammaCurve[1025];
+```
+
+The declaration would get translated into an internal representation by the compiler that is similar to the following
+
+```C#
+[CompilerGenerated, StructLayout(LayoutKind.Sequential, Pack = 1)]
+struct <GammaCurve>e__StaticDelegate
```
svick: Nit: since this has nothing to do with static delegates, the type name probably shouldn't contain `StaticDelegate`.

proposals/fixed-sized-buffers.md
```diff
@@ -0,0 +1,72 @@
...
+    private DXGI_RGB _e0;
+    private DXGI_RGB _e1;
+    // _e2 ... _e1023
+    private DXGI_RGB _e1024;
+
+    public ref DXGI_RGB this[int index]
+    {
+        get;
+        set;
+    }
```
svick: Since `ref`-returning indexers can't have setters, was this supposed to be a `ref`-returning `get`-only indexer?

tannergooding: Yes

proposals/fixed-sized-buffers.md
tannergooding: Will Fix. Cloned one of my other proposals as a template and apparently missed this part :)

tannergooding:
There are several instances where fixed-sized-buffers are multi-dimensional arrays (ex: `float m[4][4]`, as declared in C/C++).

It is probably worth considering whether multi-dimensional and/or jagged arrays should be supported for fixed-sized buffers as well.

tannergooding:
I wonder if using `ref` in the declaration would be worthwhile; such as: `public ref DXGI_RGB GammaCurve[1025];`

This would allow us to distinguish between `legacy` fixed-buffers (those that are unsafe and return a pointer) and `new` fixed-buffers (those that are safe and return a `ref`). It would also clearly expose that the array returns a `ref`.

@jaredpar, thoughts?

jaredpar:
> I wonder if using ref in the declaration would be worthwhile; such as: public ref DXGI_RGB GammaCurve[1025];

Can't do this. This is the exact syntax we would use if we allowed ref fields in the future.

tannergooding:
@jaredpar, am I good to merge this?

PR #718, opened 2017-07-03
Fix changes from #537

jamesqo:
See https://github.com/dotnet/csharplang/commit/7d6fe57574f8dce80b62c82d0f75acb3e98f4556#commitcomment-22897524. @jnm2 unintentionally corrected an intentional typo.

/cc @jnm2, @gafter

### DIFF
--- a/proposals/caller-argument-expression.md
+++ b/proposals/caller-argument-expression.md
@@ -138,7 +138,7 @@ A proposal to add such a helper class to the framework is underway at https://gi
 - Like the other `Caller*` attributes, such as `CallerMemberName`, this attribute may only be used on parameters with default values.
 - Multiple parameters marked with `CallerArgumentExpression` are permitted, as shown above.
 - The attribute's namespace will be `System.Runtime.CompilerServices`.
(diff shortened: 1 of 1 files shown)


jnm2:
@jamesqo Embarrassing that I misunderstood it. Now that I do, the sentence would read more clearly if you say <code>In the **event that</code>.

jnm2:
The proposed wording is not confusing at all now. So sorry about that!

ElizaKons:
test

gafter:
The bullet that is being corrected is being corrected as intended. However, that behavior is one of the kinds of breaking changes we avoid. Specifically, a C# 6 compiler will allow you to write the erroneous code, which would then be an error in C# N. I'd prefer the behavior is to just yield an empty string.

jamesqo:
@gafter 

> I'd prefer the behavior is to just yield an empty string.

Fixed the spec! Thanks for your feedback.

jcouv:
@gafter Ok to merge?

jcouv:
@gafter Ok to merge?

PR #719, opened 2017-07-03
Update CAE spec for extension methods

jamesqo:
I'd completely overlooked extension methods when I first thought of the CAE proposal, but luckily it didn't ruin everything. I've updated the spec to explain how the attribute will behave when it references a `this` parameter in response to https://github.com/dotnet/csharplang/issues/287#issuecomment-312611587.

/cc @yaakov-h, @gafter


jcouv:
I'll go ahead and merge. Thanks for the updated proposal.
When we get to prototyping, I suspect we'll find out this is harder to support than the main case. But we'll have to see.

PR #743, opened 2017-07-14
Correct rendering of Expressions table

Iiridayn:
In a table `x | y` doesn't render correctly - one must write `<code>x &#124; y</code>` - the entity to avoid pipe conflict, the `<code>` tags to render the entity with the same result. See https://stackoverflow.com/a/17320389/118153 for further details.


jnm2:
This came up in the past and we fixed it using `\|` instead of `&#124;` IIRC. People jump to HTML too fast without trying escaping the markdown.

```markdown
| |
|-|
| \|  |
```

Yields
| |
|-|
| \|  |

Iiridayn:
@jnm2 thank you - looks much cleaner this way

jcouv:
Good catch. Thanks

PR #759, opened 2017-07-21
Small fixes related to async

svick:
* Added `async` modifier to grammar
* Fixed link to *Async functions*
* Added *Async functions* to TOC

### DIFF
--- a/spec/classes.md
+++ b/spec/classes.md
@@ -1723,6 +1723,7 @@ method_modifier
     | 'override'
     | 'abstract'
     | 'extern'
+    | 'async'
     | method_modifier_unsafe
     ;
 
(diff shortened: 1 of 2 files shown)


spec/classes.md
```diff
@@ -1723,6 +1723,7 @@ method_modifier
     | 'override'
     | 'abstract'
     | 'extern'
+    | 'async'
```
jcouv: @gafter I see that the code treats `async` as a modifier. Can you confirm that's the case from the language perspective as well? Thanks

gafter: Yes, the language spec treats `async` as a modifier that can be placed anywhere among the other modifiers.

PR #819, opened 2017-08-13
Fixed the example in the motivation section

Shiney:
Previously the method was defined with the name 'Method' but called with the name 'M'


jcouv:
Thanks for the fix. The change looks good to me, but the baseline doc has changed.

jcouv:
@Shiney This PR couldn't be merged because the underlying document has changed in the meanwhile. Sorry we took too long to review.
Since I didn't hear back from you, I'll close the PR. Feel free to reopen with the conflict resolved. Thanks

Issue #43, opened 2017-02-09
Champion "Async Streams"


gafter:
- [X] Proposal added: [proposals/async-streams.md](https://github.com/dotnet/csharplang/blob/master/proposals/async-streams.md)
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

gafter:
See also https://github.com/dotnet/roslyn/issues/261

gafter:
This would be something like

``` c#
interface IAsyncEnumerable<T>
{
        IAsyncEnumerator<T> GetEnumerator();
}
interface IAsyncEnumerator<T>: IAsyncDisposable
{
        Task<bool> MoveNextAsync();
        T Current { get; }
}
```

With the following elements yet to be designed
- `IAsyncDisposable`
- async foreach loops
- Declaring async iterator methods
- Some kind of query expression support

onovotny:
@gafter I thought we settled on 

``` csharp
interface IAsyncEnumerable<T>
{
        IAsyncEnumerator<T> GetEnumerator(CancellationToken cancellation);
}
```

And potentially

```
static class AsyncEnumerableExtensions
{
    static IAsyncEnumerator<T> GetEnumerator(this IAsyncEnumerable<T> source) => source.GetEnumerator(CancellationToken.None);
}
```

juepiezhongren:
After the appearance of "IAsyncEnumerable" in Azure Service Fabric, we want more.........
We want the entire async implementation of .net collection BCL part!!!!!!
AsyncLinq: SelectAsync... AsyncParalellLinq: AsAsyncParalell... await yield!!!!!

As ecmascript is proposing with generator async, there is to be the fulfilment in .net!
And with what i have mentioned being implemented, .net(C# and F#) is going to be the most developer-friendly language for big-data, and service-fabric is gonna boosting a completely new ecosystem with a firm asynchronous programming foundation.

wish these come true earlier!!!!

oliverjanik:
Any news on this? Seems this interface has been defined in several places already (IX, EF7)

benaadams:
As brought up in roslyn thread
```csharp
interface IAsyncEnumerator<T>: IAsyncDisposable
{
    Task<bool> MoveNextAsync();
    T Current { get; }
}
```
Has issues with parallelazation/concurrency/atomicity which is a desirable property of an async system (e.g. double gets on Current or missed items on Current)

[...]

PR #864, opened 2017-08-30
Add strawman proposal for async iterators

stephentoub:
This is by no means a finished design; rather, it's a starting point for subsequent discussions, design, and prototyping.

Contributes to #43
cc: @MadsTorgersen

### DIFF
--- a/proposals/async-streams.md
+++ b/proposals/async-streams.md
@@ -1,0 +1,793 @@
+# Async Streams
+
+* [x] Proposed
+* [ ] Prototype
+* [ ] Implementation
+* [ ] Specification
+
+## Summary
+[summary]: #summary
+
+C# has support for iterator methods and async methods, but no support for a method that is both an iterator and an async method.  We should rectify this by allowing for await to be used in a new form of iterator, one that returns an `IAsyncEnumerable<T>` or `IAsyncEnumerator<T>` rather than an `IEnumerable<T>` or `IEnumerator<T>`.  An optional `IAsyncDisposable` interface is also used to enable for asynchronous cleanup.
+
+## Related discussion
+- https://github.com/dotnet/roslyn/issues/261
+- https://github.com/dotnet/roslyn/issues/114
+
+## Detailed design
+[design]: #detailed-design
+
+## Interfaces
+
+### IAsyncDisposable
+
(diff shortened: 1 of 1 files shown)


proposals/async-streams.md
```diff
@@ -0,0 +1,789 @@
...
+    while (await enumerator.MoveNextAsync())
+    {
+        Use(enumerator.Current);
+    }
+}
+finally { await enumerator.DisposeAsync(); }
+```
+
+Discarded options considered:
+- _`ValueTask<bool> MoveNextAsync(); T current { get; }`_: There's no benefit to using `ValueTask<bool>` instead of `Task<bool>`, as all possible `Task<bool>` values for synchronous completion can be cached, for asynchronous completion a `Task<bool>` is allocated anyway to back the `ValueTask<bool>`, and `ValueTask<bool>` adds additional overheads not present with `Task<bool>` (e.g. it has both a `T` and `Task<T>` field, which means its builder does to, which means it increases the size of state machine types that await it).
```
jnm2: Typo, `does to` should be `does too`

stephentoub: Fixed. Thanks.

proposals/async-streams.md
```diff
@@ -0,0 +1,789 @@
...
+{
+    while (await enumerator.MoveNextAsync())
+    {
+       T item = enumerator.Current;
+       ...
+    }
+}
+finally
+{
+    await enumerator.DisposeAsync(); // omitted, along with the try/catch, if the enumerator doesn't expose DisposeAsync
```
jnm2: `try/catch` should be `try/finally`

proposals/async-streams.md
```diff
@@ -0,0 +1,789 @@
...
```
jnm2: `try/catch` should be `try/finally`

stephentoub: Fixed

proposals/async-streams.md
stephentoub: Fixed.

jnm2:
Everything in this proposal is profoundly satisfying.
I can see it either way, but my preference is for WaitForNextAsync/TryGetNext.

CyrusNajmabadi:
Q1: Should DisposeAsync take a cancellation token?

PR #866, opened 2017-08-30
Update async stream proposal based on design meeting

stephentoub:
cc: @MadsTorgersen

### DIFF
--- a/proposals/async-streams.md
+++ b/proposals/async-streams.md
@@ -8,7 +8,7 @@
 ## Summary
 [summary]: #summary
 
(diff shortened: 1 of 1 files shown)


proposals/async-streams.md
```diff
@@ -365,70 +365,62 @@ Discarded options considered:
...
 {
-    public static class AsyncEnumerableExtensions
+    ct.ThrowIfCancellationRequested();
+    await UseAsync(i, ct);
+    ...
+}
+```
+If you want to pass a `CancellationToken` into an iterator, you simply do so as an argument, just as with other `async` methods:
+```C#
+static async IAsyncEnumerable<T> GetData(CancellationToken cancellationToken = default)
```
sharwell: :bulb: Would be good to remove the `= default` here

Issue #628, opened 2017-05-27
Move Proposals to Respective Language Version Directory


shaggygi:
Based on [Language Feature Status](https://github.com/dotnet/roslyn/blob/master/docs/Language%20Feature%20Status.md#language-feature-status) in Roslyn repo, should the following proposal files be moved?

### [csharp-7.1](https://github.com/dotnet/csharplang/tree/master/proposals/csharp-7.1)
- [Async Main](https://github.com/dotnet/csharplang/blob/master/proposals/async-main.md)
- [Default Expressions](https://github.com/dotnet/csharplang/blob/master/proposals/target-typed-default.md)
- [Ref Assemblies](https://github.com/dotnet/roslyn/blob/master/docs/features/refout.md)
- [Pattern-matching with generics](https://github.com/dotnet/csharplang/blob/master/proposals/generics-pattern-match.md)

If so, related [Roslyn page](https://github.com/dotnet/roslyn/blob/master/docs/Language%20Feature%20Status.md#language-feature-status) would need to be updated to new paths, as well.

shaggygi:
cc @jcouv 
Looks like you recently added links to proposals.

jcouv:
Here's a PR to do that: https://github.com/dotnet/csharplang/pull/870

PR #870, opened 2017-09-01
Move C# 7.1 proposals to proper folder

jcouv:
@gafter @MadsTorgersen @jaredpar 
The proposals folder is going to grow fast if we don't organize the files. So I'd like to move the C# 7.1 proposals to the C# 7.1 folder.
The trouble is that people may have pointed to the old locations :-(
For instance, we've referenced the old locations from github issues, discussions, stackoverflow threads and blog posts.

Let me know what you think. Maybe it is better to keep this as a flat folder?

Fixes https://github.com/dotnet/csharplang/issues/628


jcouv:
From discussion with Neal, the path change is ok. People who were pointing to old path will figure it out.

PR #963, opened 2017-10-03
Spec for ref local reassignment.

gafter:
Here is a proposed spec for ref local reassignment

/cc @dotnet/ldm @dotnet/roslyn-compiler @agocke

### DIFF
--- a/proposals/csharp-7.2/ref-local-reassignment.md
+++ b/proposals/csharp-7.2/ref-local-reassignment.md
@@ -1,0 +1,27 @@
+# Ref Local Reassignment
+
+In C# 7.2, we add support for rebinding the referent of a ref local variable or a ref parameter.
+
+We add the following to the set of `assignment_operator`s.
+
+``` antlr
+assignment_operator
+    : '=' 'ref'
+    ;
+```
+
+The `=ref` operator is called the ***ref assignment operator***. It is not a *compound assignment operator*. The left operand must be an expression that binds to a ref local variable, a ref parameter, or an out parameter. The right operand must be an expression that yields an lvalue designating a value of the same type as the left operand.
+
+The right operand must be definitely assigned at the point of the ref assignment.
+
+When the left operand binds to an `out` parameter, it is an error if that `out` parameter has not been definitely assigned at at the beginning of the ref assignment operator.
+
+If the left operand is a writeable ref (i.e. it designates anything other than a `ref readonly` local or parameter), then the right operand may not be a `ref readonly` local or parameter.
+
+The ref assignment operator yields an lvalue of the assigned type. It is writeable if the left operand is writeable (i.e. not `ref readonly`).
+
+The safety rules for this operator are:
+
+- For a ref reassignment `e1 = ref e2`, the *ref-safe-to-escape* of `e2` must be at least as wide a scope as the *ref-safe-to-escape* of `e1`.
+
+Where *ref-safe-to-escape* is defined in https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md
--- a/proposals/csharp-7.2/ref-local-reassignment.md
+++ b/proposals/csharp-7.2/ref-local-reassignment.md
@@ -16,9 +16,9 @@ The right operand must be definitely assigned at the point of the ref assignment
 
 When the left operand binds to an `out` parameter, it is an error if that `out` parameter has not been definitely assigned at at the beginning of the ref assignment operator.
 
-If the left operand is a writeable ref (i.e. it designates anything other than a `ref readonly` local or parameter), then the right operand may not be a `ref readonly` local or parameter.
+If the left operand is a writeable ref (i.e. it designates anything other than a `ref readonly` local or  `in` parameter), then the right operand may not be a `ref readonly` local or parameter.
 
-The ref assignment operator yields an lvalue of the assigned type. It is writeable if the left operand is writeable (i.e. not `ref readonly`).
+The ref assignment operator yields an lvalue of the assigned type. It is writeable if the left operand is writeable (i.e. not `ref readonly` or `in`).
 
 The safety rules for this operator are:
 
--- a/proposals/csharp-7.2/ref-local-reassignment.md
+++ b/proposals/csharp-7.2/ref-local-reassignment.md
@@ -10,13 +10,13 @@ assignment_operator
     ;
 ```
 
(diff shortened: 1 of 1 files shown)


benaadams:
So you could reassign `ref readonly`s? (the left operand/holder variable rather than what it refers to)

Hadn't thought about that; but makes sense as is the chain `ref->value` that is readonly not the whole chain `variable->ref->value`

For writable `ref`s I assume syntax is:
```csharp
ref int currentRef = ref oldRef;
ref currentRef = ref newRef; // reassign
```
For `readonly ref`s does the syntax include `readonly`? 

As per declaration+init?
```csharp
ref readonly int currentRef = ref readonly oldRef;
ref readonly currentRef = ref readonly newRef; // reassign
```

Or is `readonly` part of the type/declaration e.g. `readonly int` so its only `ref`
```csharp
ref readonly int currentRef = ref readonly oldRef;
ref currentRef = ref readonly newRef; // reassign
ref currentRef = ref newRef; // auto-convert writable ref to readonly ref in reassign?
```

jcouv:
@benaadams Per OP the reassignment the syntax is `currentRef = ref newRef; // reassign` (no `ref` on the left). This will allow using it as an expression without redundant `ref`'s.

benaadams:
The left side being `ref` is inferred from the right? e.g.

```
currentRef = newRef; // assign/copy value to pointed at ref
currentRef = ref newRef; // reassingn ref
```

That works

jcouv:
@benaadams It's not obvious, but I think the `ref` is properly seen as belonging with the `=`, not to the right-hand-side.

proposals/csharp-7.2/ref-local-reassignment.md
```diff
@@ -0,0 +1,27 @@
...
+    ;
+```
+
+The `=ref` operator is called the ***ref assignment operator***. It is not a *compound assignment operator*. The left operand must be an expression that binds to a ref local variable, a ref parameter, or an out parameter. The right operand must be an expression that yields an lvalue designating a value of the same type as the left operand.
+
+The right operand must be definitely assigned at the point of the ref assignment.
+
+When the left operand binds to an `out` parameter, it is an error if that `out` parameter has not been definitely assigned at at the beginning of the ref assignment operator.
+
+If the left operand is a writeable ref (i.e. it designates anything other than a `ref readonly` local or  `in` parameter), then the right operand may not be a `ref readonly` local or parameter.
```
VSadov: I’d just say “must be a writeable lvalue”. That would cover fields (except readonly), ref/val locals (except readonly), ref ternaries (when both branches are writeable), etc... 
We state same requirements when passing by ‘ref’.

proposals/csharp-7.2/ref-local-reassignment.md
```diff
@@ -0,0 +1,27 @@
+# Ref Local Reassignment
+
+In C# 7.2, we add support for rebinding the referent of a ref local variable or a ref parameter.
```
VSadov: There is nothing technically wrong with rebinding ref parameters. I guess - why not. Similarly as we allow assigning to byval params.

benaadams: Could it cause confusion? I assume it would be detaching from the passed ref; but it may be assumed that its changing the passed ref (in the callee) rather than just the local parameter ref? 

Unless it actually is ofc... :)

proposals/csharp-7.2/ref-local-reassignment.md
```diff
@@ -0,0 +1,27 @@
...
+
+When the left operand binds to an `out` parameter, it is an error if that `out` parameter has not been definitely assigned at at the beginning of the ref assignment operator.
+
+If the left operand is a writeable ref (i.e. it designates anything other than a `ref readonly` local or  `in` parameter), then the right operand may not be a `ref readonly` local or parameter.
+
+The ref assignment operator yields an lvalue of the assigned type. It is writeable if the left operand is writeable (i.e. not `ref readonly` or `in`).
+
+The safety rules for this operator are:
+
+- For a ref reassignment `e1 = ref e2`, the *ref-safe-to-escape* of `e2` must be at least as wide a scope as the *ref-safe-to-escape* of `e1`.
```
VSadov: When ref local declared with no initializer it has the “returnable” level, right?

gafter: If I recall correctly, we decided to defer supporting "ref local declared with no initializer" to some future revision.

VSadov: That would work too. 
Although I think the need will arise quickly. Perhaps as soon as we start testing this :)

proposals/csharp-7.2/ref-local-reassignment.md
VSadov: Caller will see no effects from/after param rebinding by callee.  But that could be the reason why callee does this...
 It will not be the most common feature to use - like assignments to ordinary parameters, but nothing is wrong with it technically.

benaadams: Maybe I'm just over thinking it :) 

People don't think changing a non-ref param changes it in callee and its not a `ref ref` or `&ref` type.

Also is probably handy so you have a starter `ref` that you then change rather than having to introduce a local copy that you then modify.

e.g.
```csharp
ref T FindFrom(ref T current)
{
    ref var local = ref current;
    while (!local.IsCorrect)
    {
        local = ref _array[local.Next];
    }
    
    return ref local;
}
```
would be better as
```csharp
ref T FindFrom(ref T current)
{
    while (!current.IsCorrect)
    {
        current = ref _array[current.Next];
    }
    
    return ref current;
}
```

benaadams: Though in practice it would be more like
```csharp
static T NotFound;

T[] _array;

ref T FindFrom(ref T current, out bool success)
{
    var maxSteps = _array.Length;
    success = true;
    do
    {
        if (current.IsCorrect)
        {
            return ref current;
        }
        else if (current.Next < 0)
        {
            break;
        }
        current = ref _array[current.Next];
        maxSteps--;
    } while (maxSteps > 0);

    success = false;
    return ref NotFound;
}
```

VSadov: Also need to clarify position on rebinding of ‘this’ in struct methods. In theory all the arguments for allowing rebinding of ref parameters apply to ‘this’ in a struct.  
So is the ‘this = ref something’ allowed in a struct method?

gafter: Just as "this" in a class acts like a parameter, it cannot be assigned, "this" in a struct acts like a ref parameter, but cannot be rebound. I'll add that to the spec.

proposals/csharp-7.2/ref-local-reassignment.md
gafter: Will do.

PR #979, opened 2017-10-06
Disallow ref-reassignment of struct `this`

agocke:
I propose we add an additional rule to disallow ref-reassignment of the struct `this` parameter. I can think of no good reason to allow this.

### DIFF
--- a/proposals/csharp-7.2/ref-local-reassignment.md
+++ b/proposals/csharp-7.2/ref-local-reassignment.md
@@ -14,6 +14,8 @@ The `=ref` operator is called the ***ref assignment operator***. It is not a *co
 
 The right operand must be definitely assigned at the point of the ref assignment.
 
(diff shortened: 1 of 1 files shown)


agocke:
and @jaredpar @jcouv @OmarTawfik @gafter

benaadams:
Is already there? https://github.com/dotnet/csharplang/pull/963/commits/cd1a6a90dc7b21e970922c63583a17251b6e4dae

> The left operand must be an expression that binds to a ref local variable, a ref parameter **(other than `this`)**, or an out parameter.

benaadams:
Was brought up in https://github.com/dotnet/csharplang/pull/963#discussion_r142947651

agocke:
@benaadams You're right! This is covered in the previous clause. Great!

PR #1036, opened 2017-10-23
Update nullable-reference-types.md

v-andrew:
It should be prohibited to convert nullable to non-null reference without check


svick:
If you disagree with some aspect of a proposal, you can voice your opinion on the related issue (in this case, https://github.com/dotnet/csharplang/issues/36). I don't think submitting a Pull Request is the right way to do that.

alrz:
It already produces a warning.

v-andrew:
@svick Thanks

PR #1100, opened 2017-11-12
Proposal for the Pattern based fixed statement

VSadov:
Defining a pattern that would allow types to participate in fixed statements.

LDM https://github.com/dotnet/csharplang/blob/master/meetings/2017/LDM-2017-12-04.md discussed this

### DIFF
--- a/proposals/pattern-based-fixed.md
+++ b/proposals/pattern-based-fixed.md
@@ -39,7 +39,6 @@ A viable pattern-based “fixed” need to:
 -	Prescribe the behavior in "empty" case when there is nothing to refer to. 
 -	Should not push API authors toward design decisions that hurt the use of the type outside of `fixed`.
 
-
 I think the above could be satisfied by recognizing a specially named ref-returning member:
  `ref [readonly] T DangerousGetPinnableReference()`.
 
@@ -47,18 +46,16 @@ In order to be used by the `fixed` statement the following conditions must be me
 
 1)	There is only one such member provided for a type.
 1)	Returns by `ref` or `ref readonly`. 
-(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
+(`readonly` is permitted so that authors of immutable/readonly types could implement the pattern without adding writeable API that could be used in safe code)
 1)  T is an unmanaged type.
-(since `T*` becomes the pointer type)
+(since `T*` becomes the pointer type. The restriction will naturally expand if/when the notion of "unmanaged" is expanded)
 1)	Returns managed `nullptr` when there is no data to pin – probably the cheapest way to convey emptiness.
-(note that “” string returns a ref to “\0” since strings are zero-terminated)
-
+(note that “” string returns a ref to '\0' since strings are null-terminated)
 
 Alternatively for the `#3` we can allow the result in empty cases be undefined or implementation-specific. 
 That, however, may make the API more dangerous and prone to abuse and unintended compatibility burdens. 
 
 ## *Translation* ##
-
 
 ```C#
     fixed(byte* ptr = thing)
@@ -121,7 +118,6 @@ Users can introduce DangerousGetPinnableReference or similar member and use it a
 ```
 
 There is no solution for `System.String` if alternative solution is desired.
-
 
 ## Unresolved questions
 [unresolved]: #unresolved-questions
--- a/proposals/pattern-based-fixed.md
+++ b/proposals/pattern-based-fixed.md
@@ -24,7 +24,7 @@ fixed(byte* ptr = byteArray)
 
 ```
 
-The set of types that can participate in `foreach` is hardcoded and limited to arrays and `System.String`. Hardcoding "special" types does not scale when new primitives such as `ImmutableArray<T>`, `Span<T>`, `Utf8String` are introduced. 
+The set of types that can participate in `fixed` is hardcoded and limited to arrays and `System.String`. Hardcoding "special" types does not scale when new primitives such as `ImmutableArray<T>`, `Span<T>`, `Utf8String` are introduced. 
 
 In addition, the current solution for `System.String` relies on a fairly rigid API. The shape of the API implies that `System.String` is a contiguous object that embeds UTF16 encoded data at a fixed offset from the object header. Such approach has been found problematic in several proposals that could require changes to the underlying layout. 
 It would be desirable to be able to switch to something more flexible that decouples `System.String` object from its internal representation for the purpose of unmanaged interop. 
@@ -43,7 +43,7 @@ A viable pattern-based “fixed” need to:
 I think the above could be satisfied by recognizing a specially named ref-returning member:
  `ref [readonly] T DangerousGetPinnableReference()`.
 
-In order to be used by the `foreach` statement the following conditions must be met:
+In order to be used by the `fixed` statement the following conditions must be met:
 
 1)	There is only one such member provided for a type.
 1)	Returns by `ref` or `ref readonly`. 
@@ -127,6 +127,8 @@ There is no solution for `System.String` if alternative solution is desired.
 [unresolved]: #unresolved-questions
 
 - [ ] Behavior in "empty" state. - `nullptr` or `undefined` ? 
+- [ ] Should the extension methods be considered ? 
+- [ ] If a pattern is detected on `System.String`, should it win over ? 
 
 ## Design meetings
 
--- a/proposals/pattern-based-fixed.md
+++ b/proposals/pattern-based-fixed.md
@@ -1,0 +1,133 @@
+# Pattern-based `fixed` statement
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Introduce a pattern that would allow types to participate in `fixed` statements. 
+
+## Motivation
+[motivation]: #motivation
+
+The language provides a mechanism for pinning managed data and obtain a native pointer to the underlying buffer.
+
+```C#
+fixed(byte* ptr = byteArray)
+{
+   // ptr is a native pointer to the first element of the array
+   // byteArray is protected from being moved/collected by the GC for the duration of this block 
+}
+
+```
+
+The set of types that can participate in `foreach` is hardcoded and limited to arrays and `System.String`. Hardcoding "special" types does not scale when new primitives such as `ImmutableArray<T>`, `Span<T>`, `Utf8String` are introduced. 
+
+In addition, the current solution for `System.String` relies on a fairly rigid API. The shape of the API implies that `System.String` is a contiguous object that embeds UTF16 encoded data at a fixed offset from the object header. Such approach has been found problematic in several proposals that could require changes to the underlying layout. 
+It would be desirable to be able to switch to something more flexible that decouples `System.String` object from its internal representation for the purpose of unmanaged interop. 
+
+## Detailed design
+[design]: #detailed-design
+
+## *Pattern* ##
+A viable pattern-based “fixed” need to:
+-	Provide the managed references to pin the instance and to initialize the pointer (preferably this is the same reference)
+-	Convey unambiguously the type of the unmanaged element   (i.e. “char” for “string”)
+-	Prescribe the behavior in "empty" case when there is nothing to refer to. 
+-	Should not push API authors toward design decisions that hurt the use of the type outside of `fixed`.
+
+
+I think the above could be satisfied by recognizing a specially named ref-returning member:
+ `ref [readonly] T DangerousGetPinnableReference()`.
+
+In order to be used by the `foreach` statement the following conditions must be met:
+
+1)	There is only one such member provided for a type.
+1)	Returns by `ref` or `ref readonly`. 
+(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
+1)  T is an unmanaged type.
+(since `T*` becomes the pointer type)
+1)	Returns managed `nullptr` when there is no data to pin – probably the cheapest way to convey emptiness.
+(note that “” string returns a ref to “\0” since strings are zero-terminated)
+
+
+Alternatively for the `#3` we can allow the result in empty cases be undefined or implementation-specific. 
+That, however, may make the API more dangerous and prone to abuse and unintended compatibility burdens. 
+
+## *Translation* ##
+
+
+```C#
+    fixed(byte* ptr = thing)
+    { 
+		// <BODY>
+    }
+```
+
+becomes the following pseudocode (not all expressible in C#)
+
+```C#
+
+    byte* ptr;
+    // specially decorated "pinned" IL local slot, not visible to user code.
+    pinned ref byte _pinned;
+
+    try
+    {
+        // NOTE: null check is omitted for value types 
+        // NOTE: `thing` is evaluated only once (temporary is introduced if necessary) 
+        if (thing != null)
+        {
+            // obtain and "pin" the reference
+            _pinned = ref thing.DangerousGetPinnableReference();
+
+            // unsafe cast in IL
+            ptr = (byte*)_pinned;
+        }
+        else
+        {
+            ptr = default(byte*);
+        }
+
+        // <BODY> 
+    }
+    finally   // finally can be omitted when not observable
+    {
+        // "unpin" the object
+        _pinned = nullptr;
+    }
+
+```
+
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+- DangerousGetPinnableReference is intended to be used only in `fixed`, but nothing prevents its use in safe code, so implementor must keep that in mind.
+
+## Alternatives
+[alternatives]: #alternatives
+
+Users can introduce DangerousGetPinnableReference or similar member and use it as
+ 
+```C#
+    fixed(byte* ptr = thing.DangerousGetPinnableReference())
+    { 
+		// <BODY>
+    }
+```
+
+There is no solution for `System.String` if alternative solution is desired.
+
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- [ ] Behavior in "empty" state. - `nullptr` or `undefined` ? 
+
+## Design meetings
+
+None yet. 


proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,133 @@
...
+```C#
+fixed(byte* ptr = byteArray)
+{
+   // ptr is a native pointer to the first element of the array
+   // byteArray is protected from being moved/collected by the GC for the duration of this block 
+}
+
+```
+
+The set of types that can participate in `foreach` is hardcoded and limited to arrays and `System.String`. Hardcoding "special" types does not scale when new primitives such as `ImmutableArray<T>`, `Span<T>`, `Utf8String` are introduced.
```
CyrusNajmabadi: it's confusing to me that you're starting out talking about 'fixed', but then talk about 'foreach' here.  Did you mean 'fixed' in this sentence?

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,133 @@
...
+-	Prescribe the behavior in "empty" case when there is nothing to refer to. 
+-	Should not push API authors toward design decisions that hurt the use of the type outside of `fixed`.
+
+
+I think the above could be satisfied by recognizing a specially named ref-returning member:
+ `ref [readonly] T DangerousGetPinnableReference()`.
+
+In order to be used by the `foreach` statement the following conditions must be met:
+
+1)	There is only one such member provided for a type.
```
stephentoub: Extension methods would be supported, presumably?

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,133 @@
...
+1)	Returns by `ref` or `ref readonly`. 
+(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
+1)  T is an unmanaged type.
+(since `T*` becomes the pointer type)
+1)	Returns managed `nullptr` when there is no data to pin � probably the cheapest way to convey emptiness.
```
stephentoub: How do I implement that in C#?
```C#
class MyType
{
    public ref readonly byte DangerousGetPinnableReference()
    {
        return ...; // what do I write here to return a managed nullptr?
    }
}
```

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,133 @@
...
+(note that �� string returns a ref to �\0� since strings are zero-terminated)
+
+
+Alternatively for the `#3` we can allow the result in empty cases be undefined or implementation-specific. 
+That, however, may make the API more dangerous and prone to abuse and unintended compatibility burdens.
```
stephentoub: I think it needs to be defined.  We left it as undefined for a while for arrays, and that just led to writing less efficient code, since the call sites would then end up duplicating checks that were emitted by the compiler anyway.

proposals/pattern-based-fixed.md
VSadov: Of course “fixed” :)
Was tslking a lot about foreach lately too..,

proposals/pattern-based-fixed.md
VSadov: I was thinking about this and was not sure. The method would typically be very coupled with internal implementation of the type. 

On the other hand if type cannot be changed, can extension be provided? (likely encapsulating unsavory hacks against implementation, but at least in one place)

Extensions can be allowed, but not everything that can be allowed should be allowed. I will add this to the list of things to discuss.

stephentoub: > The method would typically be very coupled with internal implementation of the type.

Yes, but that doesn't prevent it from being an extension method, which might be done purely for usability purposes.  For example, today these DangerousGetPinnableReference methods on span/readonlyspan are marked as non-browsable, so that they don't show up in normal IntelliSense browsing.  But that's a bad experience when you actually do need them, as it's a long name for which you don't get completion support.  We're talking instead about making them extension methods that live on the System.Runtime.InteropServices.MemoryMarshal class, also in Corelib, without the attribute.  So such DangerousGetPinnableReference methods could be extensions but still live in the same assembly.

proposals/pattern-based-fixed.md
VSadov: Yes.
Or “default(Span<T>).FangerousGetPinnable..” for generics.

We do not need to make this easy or common/popular. 
‘nullptr’ references should not generally exist. Except maybe in unsafe/interop cases.

proposals/pattern-based-fixed.md
VSadov: fixed :)

proposals/pattern-based-fixed.md
VSadov: Yes, "undefined" often leads to "defined in tribal knowledge" situation. 
Implementer still ends up bound by compat, while users need to program defensively.

I have added this as a point to discuss since we did "undefined" in this area before.  
The spec for current cases of "fixed" still says "undefined" although the behavior is well known and hardly can be changed.

stephentoub: > The spec for current cases of "fixed" still says "undefined"

We'd agreed to update the spec here:
https://github.com/dotnet/roslyn/issues/6707
and did so here:
https://github.com/dotnet/csharplang/pull/16
though I don't know if those changes have propagated anywhere "real".

proposals/pattern-based-fixed.md
jkotas: > Or for generic cases
> `return ref  default(Span<T>).DangerousGetPinnable..`

This will not work for the Span itself - it would call itself recursively.

Also, it needs to produce fast code (e.g. constant zero). Getting JIT to produce that out of `default(Span<T>).DangerousGetPinnable...` would be a ton of work for the JIT.

I think `Unsafe.AsRef<T>(null)` should work.

proposals/pattern-based-fixed.md
jkotas: > We're talking instead about making them non-extension methods that live on the System.Runtime.InteropServices.MemoryMarshal class, also in Corelib, without the attribute.

We should have both the extension method for direct use, and (instance?) method for the pinning pattern.

- The (instance?) method for pinning should have the safe-guard for empty spans to return null.
- The extension method on MemoryMarshal should not have the safe-guard for empty spans.

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+-	Prescribe the behavior in "empty" case when there is nothing to refer to. 
+-	Should not push API authors toward design decisions that hurt the use of the type outside of `fixed`.
+
+
+I think the above could be satisfied by recognizing a specially named ref-returning member:
+ `ref [readonly] T DangerousGetPinnableReference()`.
+
+In order to be used by the `fixed` statement the following conditions must be met:
+
+1)	There is only one such member provided for a type.
```
jkotas: Nit: Should this be `1)`, `2)`, `3)`, ... .

proposals/pattern-based-fixed.md
stephentoub: > and (instance?) method for the pinning pattern

Right, we just planned to defer the pinning pattern one until it was actually a language feature.

> We should have both the extension method for direct use, 

But not an extension method, right?  Just a static method on a marshaling class.

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+1)	Returns by `ref` or `ref readonly`. 
+(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
```
jkotas: ... authors of immutable/readonly types could implement the pattern *without unsafe constructs*.

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+
+There is no solution for `System.String` if alternative solution is desired.
+
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- [ ] Behavior in "empty" state. - `nullptr` or `undefined` ? 
+- [ ] Should the extension methods be considered ? 
+- [ ] If a pattern is detected on `System.String`, should it win over ?
```
jkotas: Yes!

proposals/pattern-based-fixed.md
jkotas: > But not an extension method, right

Right.

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+1)	There is only one such member provided for a type.
+1)	Returns by `ref` or `ref readonly`. 
+(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
+1)  T is an unmanaged type.
+(since `T*` becomes the pointer type)
+1)	Returns managed `nullptr` when there is no data to pin � probably the cheapest way to convey emptiness.
+(note that �� string returns a ref to �\0� since strings are zero-terminated)
+
+
+Alternatively for the `#3` we can allow the result in empty cases be undefined or implementation-specific.
```
jkotas: There is difference in the strength of these conditions: 1) and 2) are enforced by the compiler; 3) is just a recommendation for the implementers of the `DangerousGetPinnableReference` API.

proposals/pattern-based-fixed.md
VSadov: that is just to let markdown to number them at render time. convenient if need to add/remove/reorder an item

proposals/pattern-based-fixed.md
VSadov: Also think so. Otherwise - what is the point?

proposals/pattern-based-fixed.md
VSadov: Yes, it is just a recommendation, but I believe we have to recommend what to do, so that end users could have uniform expectations. 
(and the bugs about unexpected behavior would go to the type owners).

VSadov:
@dotnet/roslyn-compiler @MadsTorgersen  - FYI

proposals/pattern-based-fixed.md
VSadov: that is just to let markdown to number them at render time. convenient if need to add/remove/reorder an item

ufcpp:
How would you like TryGet pattern for getting pinnable reference?
For instance, [the compact string](https://github.com/dotnet/coreclr/issues/7083) needs ASCII (`byte*`)/UTF-16 (`char*`) switching.

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+
+In order to be used by the `fixed` statement the following conditions must be met:
+
+1)	There is only one such member provided for a type.
+1)	Returns by `ref` or `ref readonly`. 
+(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
+1)  T is an unmanaged type.
+(since `T*` becomes the pointer type)
+1)	Returns managed `nullptr` when there is no data to pin � probably the cheapest way to convey emptiness.
+(note that �� string returns a ref to �\0� since strings are zero-terminated)
```
4creators: nit: zero terminated in ASCII is "x30" (hex) since strings are 'null-terminated' (x00)

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+   // ptr is a native pointer to the first element of the array
+   // byteArray is protected from being moved/collected by the GC for the duration of this block 
+}
+
+```
+
+The set of types that can participate in `fixed` is hardcoded and limited to arrays and `System.String`. Hardcoding "special" types does not scale when new primitives such as `ImmutableArray<T>`, `Span<T>`, `Utf8String` are introduced. 
+
+In addition, the current solution for `System.String` relies on a fairly rigid API. The shape of the API implies that `System.String` is a contiguous object that embeds UTF16 encoded data at a fixed offset from the object header. Such approach has been found problematic in several proposals that could require changes to the underlying layout. 
+It would be desirable to be able to switch to something more flexible that decouples `System.String` object from its internal representation for the purpose of unmanaged interop.
```
jaredpar: Did you mean for there to be a paragraph break here?

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+[design]: #detailed-design
+
+## *Pattern* ##
+A viable pattern-based �fixed� need to:
+-	Provide the managed references to pin the instance and to initialize the pointer (preferably this is the same reference)
+-	Convey unambiguously the type of the unmanaged element   (i.e. �char� for �string�)
+-	Prescribe the behavior in "empty" case when there is nothing to refer to. 
+-	Should not push API authors toward design decisions that hurt the use of the type outside of `fixed`.
+
+
```
jaredpar: double blank line

proposals/pattern-based-fixed.md
jaredpar: did you mean fro a link break here?

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+1)	There is only one such member provided for a type.
+1)	Returns by `ref` or `ref readonly`. 
+(readonly is permitted so that authors of immutable/readonly types could implement the pattern)
+1)  T is an unmanaged type.
+(since `T*` becomes the pointer type)
+1)	Returns managed `nullptr` when there is no data to pin � probably the cheapest way to convey emptiness.
+(note that �� string returns a ref to �\0� since strings are zero-terminated)
+
+
```
jaredpar: double blank line

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,135 @@
...
+## *Pattern* ##
+A viable pattern-based �fixed� need to:
+-	Provide the managed references to pin the instance and to initialize the pointer (preferably this is the same reference)
+-	Convey unambiguously the type of the unmanaged element   (i.e. �char� for �string�)
+-	Prescribe the behavior in "empty" case when there is nothing to refer to. 
+-	Should not push API authors toward design decisions that hurt the use of the type outside of `fixed`.
+
+
+I think the above could be satisfied by recognizing a specially named ref-returning member:
+ `ref [readonly] T DangerousGetPinnableReference()`.
```
KrzysztofCwalina: Can this be an extension method?

VSadov: @KrzysztofCwalina - it is one of the items to discuss. 
Allowing extension methods is not strictly required for the feature to function and support for that can be added later.

stephentoub: https://github.com/dotnet/csharplang/pull/1100#discussion_r150416143

VSadov:
@ufcpp - the compact or compressed string can easily implement `DangerousGetPinnableReference` by expanding itself into a UTF16 zero-terminated `char[]` buffer and returning a reference to that. 

It may also retain the buffer for the future reuse (perhaps weakly).

proposals/pattern-based-fixed.md
VSadov: Yes, I wanted "## Detailed design " to start a new paragraph

proposals/pattern-based-fixed.md
VSadov: fixed

---
In reply to: [150590172](https://github.com/dotnet/csharplang/pull/1100#discussion_r150590172) [](ancestors = 150590172)

proposals/pattern-based-fixed.md
VSadov: Unsafe.AsRef<T>((void*)0);    could also work...

These are just possible solutions for "nullptr'.  They are not pretty, but perhaps should not be.

---
In reply to: [150422216](https://github.com/dotnet/csharplang/pull/1100#discussion_r150422216) [](ancestors = 150422216)

VSadov:
DangerousGetPinnableReference was discussed today at LDM. 

The major takeaways: 
- DangerousGetPinnableReference as a method name is **ok**.  
Having no better options, why change.?
It should probably be hidden from intellisense and documented as "for use by compilers" - that is up to the API designer though.

- Position on allowing DangerousGetPinnableReference to be an extension method - is **ok**
There is no reason to disallow that.

Ideally the method would be usable in `unsafe` code only, but we have no formal way of forcing that. The feature did not seem worth introducing whole new "fixed operator" as some syntax that could be translated into otherwise uncallable method.

GrabYourPitchforks:
Per our internal email thread, this proposal would assist a scenario I've been working on where I'm trying to extract a `readonly ref char` from a `string` instance. The scenario is possible today through pinning, but this proposal makes it easier and much more optimized.

proposals/pattern-based-fixed.md
```diff
@@ -0,0 +1,131 @@
...
+
+In order to be used by the `fixed` statement the following conditions must be met:
+
+1)	There is only one such member provided for a type.
+1)	Returns by `ref` or `ref readonly`. 
+(`readonly` is permitted so that authors of immutable/readonly types could implement the pattern without adding writeable API that could be used in safe code)
+1)  T is an unmanaged type.
```
tannergooding: Is `unmanaged` interchangeable with `blittable`, or does it have some other meaning here?

jkotas:
Now that `DangerousGetPinnableReference` is gone from the public surface (https://github.com/dotnet/corefx/issues/23881), we should change the name of the method to just `GetPinnableReference`. There is nothing dangerous about it if it returns readonly span and null for empty Span (https://github.com/dotnet/corefx/issues/23881#issuecomment-352191055).

@VSadov Do you think this will make it for .NET Core 2.1?

VSadov:
As long as the fault mode for the method is to return nulptr byref, it seems fairly dangerous. 
Ordinary safe methods should not return or pass or be expected to handle nulptr byrefs. There is no safe way to even check if a byref is not bound to concrete storage - FWIW it was an impossible case since V1 - virtually no existing code would be prepared to handle that now.

I’d prefer this not be callable in source, but it is fairly hard to arrange. Besides in some advanced cases direct use could be justified. 
Calling this method directly should be generally discouraged and having “Dangerous” in the name would let the API owner be able to say “told you so” when some one gets in trouble.

jkotas:
> As long as the fault mode for the method is to return nulptr byref, it seems fairly dangerous.

Depends on how you define dangerous. We have used dangerous to mean unsafe code.

There is nothing unsafe with derefencing nulptr byref. You will get a clean null reference exception.

VSadov:
Merging this, since this has been opened for a long time, and even discussed in LDM.

PR #1122, opened 2017-11-20
Stackalloc array initializers.

VSadov:
In short: Ordinary arrays can be created through the following syntax:

```C#
new int[3]
new int[3] { 1, 2, 3 }
new int[] { 1, 2, 3 }
new[] { 1, 2, 3 }
```

We should allow stack allocated arrays be created through:

```C#
stackalloc int[3]				// currently allowed
stackalloc int[3] { 1, 2, 3 }
stackalloc int[] { 1, 2, 3 }
stackalloc[] { 1, 2, 3 }
```

### DIFF
--- a/proposals/stackalloc-array-initializers.md
+++ b/proposals/stackalloc-array-initializers.md
@@ -1,0 +1,65 @@
+# Stackalloc array initializers.
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Allow array initializer syntax to be used with `stackalloc` 
+
+## Motivation
+[motivation]: #motivation
+
+Ordinary arrays can have their elements initialized at creation time. It seems reasonable to allow that in `stackalloc` case.
+
+The question of why such syntax is not allowed with `stackalloc` arises fairly frequently.  
+See, for example, https://github.com/dotnet/csharplang/issues/1112  
+
+## Detailed design
+
+Ordinary arrays can be created through the following syntax:
+
+```C#
+new int[3]
+new int[3] { 1, 2, 3 }
+new int[] { 1, 2, 3 }
+new[] { 1, 2, 3 }
+```
+
+We should allow stack allocated arrays be created through:  
+
+```C#
+stackalloc int[3]				// currently allowed
+stackalloc int[3] { 1, 2, 3 }
+stackalloc int[] { 1, 2, 3 }
+stackalloc[] { 1, 2, 3 }
+```
+
+The semantics of all cases is roughly the same as with arrays.  
+For example: in the last case the element type is inferred from the initializer and must be an "unmanaged" type.
+
+NOTE: the feature is not dependent on the target being a `Span<T>`. It is just as applicable in `T*` case, so it does not seem reasonable to predicate it on `Span<T>` case.  
+
+## Translation ##
+
+The naive implementation could just initialize the array right after creation through a series of element-wise assignments.  
+
+Similarly to the case with arrays, it might be possible and desirable to detect cases where all or most of the elements are blittable types and use more efficient techniques by copying over the pre-created state of all the constant elements. 
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+## Alternatives
+[alternatives]: #alternatives
+
+This is a convenience feature. It is possible to just do nothing.
+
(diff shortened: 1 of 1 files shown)


VSadov:
FYI: @jaredpar @stephentoub @jkotas @KrzysztofCwalina @MadsTorgersen

benaadams:
Skip zero init? Or not be considered for triggering for `.locals init` to avoid double assign

proposals/stackalloc-array-initializers.md
```diff
@@ -0,0 +1,65 @@
...
+stackalloc int[3]				// currently allowed
+stackalloc int[3] { 1, 2, 3 }
+stackalloc int[] { 1, 2, 3 }
+stackalloc[] { 1, 2, 3 }
+```
+
+The semantics of all cases is roughly the same as with arrays.  
+For example: in the last case the element type is inferred from the initializer and must be an "unmanaged" type.
+
+NOTE: the feature is not dependent on the target being a `Span<T>`. It is just as applicable in `T*` case, so it does not seem reasonable to predicate it on `Span<T>` case.
```
jkotas: What are the requirements on the target? Does it work for any type with `void*, int` length constructor?

(I assume that this is answered in the spec for the currently allowed case without initializer. The link to that spec may be helpful.)

proposals/stackalloc-array-initializers.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Translation ##
+
+The naive implementation could just initialize the array right after creation through a series of element-wise assignments.  
+
+Similarly to the case with arrays, it might be possible and desirable to detect cases where all or most of the elements are blittable types and use more efficient techniques by copying over the pre-created state of all the constant elements. 
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
```
KrzysztofCwalina: I like how there are no drawbacks :-)

proposals/stackalloc-array-initializers.md
VSadov: It works with any type that has implicit conversion from `Span<T>`.

proposals/stackalloc-array-initializers.md
VSadov: Well, implementing a feature has cost, but it kind of obvious drawback. Could not find any other.

proposals/stackalloc-array-initializers.md
VSadov: From the language semantics `stackalloc T[size]` expression has `Span<T>` type, unless it matches backwards-compatible pattern where it must have type `T*`. 

The implementation will call well-known constructor `Span<T>(void*, int)`. Essentially there is one "magic" type - `Span<T>` and one magic method `Span<T>(void*, int)`.  

Other types opt-in by providing a conversion operator. It is more intentional than matching `(void*, int)` constructors on random types - who knows what they expect, is that always `{ptr, size}`, is that `size` in bytes on elements?  
With `Span<T>` there could be no doubts on `{ptr, size}` part.

This is also how `stackalloc` always worked. - you could always participate by having a conversion from `T*` .  Opting in via a conversion is not a new feature.

jkotas: > Essentially there is one "magic" type - `Span<T>` and one "magic" method `Span<T>(void*, int)`, that we know about.

Ok, that make sense. I was under impression that you are somehow doing this without knowing about magic `Span<T>` type.

VSadov: We tried avoiding magic types, but arrived to a conclusion that for a robust pattern we would need one type - a generally accepted representation of `{ptr, size}` tuple.

And then realized, that `Span<T>` works just fine for that purpose  :-)

stephentoub: Since it's based on `Span<T>`, I assume I'd be able to write:
```C#
ReadOnlySpan<char> ros = stackalloc char[] { char };
...
```
?  Today to do the same thing I need to write:
```C#
Span<char> s = stackalloc char[1];
s[0] = char;
ReadOnlySpan<char> ros = s;
...
```

VSadov: @stephentoub  Yes. That should work.

alrz:
What about `Span<int> s = { 1, 2, 3 };` similar to `int[] a = { 1, 2, 3 };`

VSadov:
Merging this, since there are Approvals and no known objections. 
Will be discussed in the LDM at the next opportunity.

PR #1134, opened 2017-11-22
Fix ANTLR in Expression

zccz14:
I try to assembly all the ANTLR spec together and my ANTLR Syntax Inspector found there's no definition of  `interpolated_string`, but the `interpolated_string_expression`.


jcouv:
Merged. Thanks

PR #1166, opened 2017-12-03
More details on indexable fixed-sized buffers proposal.

VSadov:


VSadov:
CC: @gafter @tannergooding @jcouv @jaredpar

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
-    }
+    public ref T this[int index] => ref (uint)index <= 1024u ?
+                                         ref RefAdd<T>(ref _e0, index):
+                                         throw new IndexOutOfRange();
 }
 ```
 
+Since such fixed-sized buffers no longer require use of `fixed`, it makes sense to allow any element type.  
+
+> NOTE: `fixed` will still be supported, but only if the element type is `blittable`
```
tannergooding: Today, unsafe `fixed` buffers only support a subset of types that would be considered `blittable`. Is this expected to be expanded to include all `blittable` types?

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
-* This can quickly become 'unwieldly' for large arrays
-* The JIT may not behave well with types that contain a large number of fields
- * We may be able to work with the JIT/Runtime to provide a special flag they can use (basically, it would give the contained type and the element count, then they just need to compute the size of the base struct and multiply, rather than actually checking each field).
+* There could be some challenges with backwards compatibility, but given that the existing fixed-sized buffers only work with a selection of primitive types, it should be possible for the compiler to continue "just-working" if the user treats the fixed-buffer as a pointer. 
+* Incompatible constructs may need to use slightly different `v2` encoding to hide the fields from old compiler.
+* Packing is not weel defined in IL spec for generic types. While the approach should work, we will be bordering on undocumnted behavior. We should make that documented and make sure other JITs like Mono have the same behavior.
```
mikedn: weel - well

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
+* Specifying a separate type for every length (an possibly another for `readonly` fields, if supported) will have impact on metadata. It will be bound by the number of arrays of different sizes in the given app.
```
tannergooding: It might be possible to limit this somewhat by declaring well known sizes and constructing buffers of other sizes from the small ones.  Ex: `struct <Buffer>e__FixedBuffer_1024<T>` could contain four elements of `struct <Buffer>e__FixedBuffer_256<T>`, rather than 1024 elements of `T`.

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
+* There could be some challenges with backwards compatibility, but given that the existing fixed-sized buffers only work with a selection of primitive types, it should be possible for the compiler to continue "just-working" if the user treats the fixed-buffer as a pointer. 
+* Incompatible constructs may need to use slightly different `v2` encoding to hide the fields from old compiler.
+* Packing is not weel defined in IL spec for generic types. While the approach should work, we will be bordering on undocumnted behavior. We should make that documented and make sure other JITs like Mono have the same behavior.
+* Specifying a separate type for every length (an possibly another for `readonly` fields, if supported) will have impact on metadata. It will be bound by the number of arrays of different sizes in the given app. 
+* `ref` math is not formally verifiable (since it is unsafe). We will need to find a way to update verification rules to know that our use is ok.
```
mikedn: That seems problematic given the proposed implementation. Perhaps something like `new Span<T>(ref _e0, 1024)[index]` would be easier to verify - the verifier should simply "see" that a span is being constructed from a `ref` to a field and that there are enough fields after it for the span length to be correct. Though AFAIR this particular span constructor is internal...

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
-Manually declare your structures
+Manually declare your structures and use unsafe code to construct indexers.
 
 ## Unresolved questions
 [unresolved]: #unresolved-questions
 
-Can/should this additionally be extended to stackalloc which provides a very similar mechanism.
+- should we allow `readonly`?  (with readonly indexer)
+- should we allow array initializers?
+- is `fixed` keyword necessary?
```
mikedn: Hmm, `fixed` seems pointless if the construct is safe.

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
+- `foreach`?
```
mikedn: Via span?

proposals/fixed-sized-buffers.md
```diff
@@ -29,42 +31,51 @@ public fixed DXGI_RGB GammaCurve[1025];
...
 
-Manually declare your structures
+Manually declare your structures and use unsafe code to construct indexers.
 
 ## Unresolved questions
 [unresolved]: #unresolved-questions
 
-Can/should this additionally be extended to stackalloc which provides a very similar mechanism.
+- should we allow `readonly`?  (with readonly indexer)
+- should we allow array initializers?
```
mikedn: Yes

proposals/fixed-sized-buffers.md
tannergooding: I think we need some syntax to disambiguate normal arrays from inline arrays. Although I'm not sure if `fixed` is the right way to do that 😄

tannergooding:
Making fixed-sized buffers more commonplace also runs the risk of hitting the max struct size limitation and/or blowing the stack (for structs in particular). Maybe an analyzer would be useful to help catch those cases (since they are runtime implementation details, rather than some hard limit)?

proposals/fixed-sized-buffers.md
mikedn: Well, the syntax is rather different already:
```C#
struct Foo {
    int[] array;
    int buffer[3];
}
```
I'm sure some will be confused at first but I doubt that this confusion will have negative long term consequences. In the worst case they'll think that copying such a struct will result in a new struct that references the old buffer :smile:

> Although I'm not sure if fixed is the right way to do that

`inline` perhaps? Not an existing keyword so perhaps problematic. or `struct`?
```C#
struct Foo {
    inline int buffer[3];
    struct int buffer[3]; // kind of communicates that we're really dealing with a hidden, anonymous struct?
}
```
Both seem a bit silly though.

tannergooding: I would prefer `inline` over `struct`, as it seems a bit more clear. But you're right, the syntax is already different so maybe there won't be any issues

ufcpp:
Does this proposal conflict with the ref-returns spec? IIRC, struct is intentionally disallowed to return its fileds by reference.

```cs
struct StructField
{
    int _v;
    public ref int Value => ref _v; // Not Allowed
}

class ClassField
{
    int _v;
    public ref int Value => ref _v; // OK
}

struct ClassInStruct
{
    int[] _items;
    public ClassInStruct(int[] items) => _items = items;

    public ref int Value => ref _items[0]; // OK!
}

class Program
{
    static ref int F()
    {
        var x = new ClassInStruct(new int[] { 1 });

        // OK!
        // This is safe-to-return because strurct never return its fields by-ref
        return ref x.Value;
    }

    static void Main()
    {
        ref var v = ref F();
        v = 10;
    }
}
```

mikedn:
> Does this proposal conflict with the ref-returns spec? IIRC, struct is intentionally disallowed to return its fileds by reference

Hmm, so it seems. But I'm not sure that the indexer has to actually exist. The buffer fields can be public  and then the compiler can generate the necessary code inline. Especially when the index is constant there's no need to generate anything other than a `ldflda`.

VSadov:
Good point. For the tracking of ref returns indexing should be treated similar to field accesses.

proposals/fixed-sized-buffers.md
VSadov: The ‘slow’ span may not be able to do such wrapping since it needs the outer object when data is on the heap.  
‘foreach’ would need to be supported intrinsically.

proposals/fixed-sized-buffers.md
VSadov: ‘blittable’ is a subset of ‘unmanaged’ types that includes types which (and their arrays) have the same native/com representation as in the managed code.  

It is a bit of a circular definition that ‘fixed’ allows only blittable types. 

If the set of blittable typea is expanded (why not enums, for example), then they will be allowed in ‘fixed’. - if that is what you are asking.

tannergooding: Enums are one of the `sometimes` blittable types, as they also include `bool` and `char` (at the IL level).

`fixed` today doesn't support `IntPtr`, `UIntPtr`, or structs that are composed of only blittable types (`Vector4` or `Complex` for example).

I was wonting to confirm that after this support goes in, will those will be supported in the `unsafe` fixed-sized buffer (the one that exists today) context as well as the `safe` fixed-sized buffer (the one being proposed here) context

jaredpar: I don't think we plan to make any changes to the existing unsafe fixed feature. The reason being that if we were starting from scratch today it's likely a feature we simply wouldn't add. Instead we'd focus on the experience being desigend here: safe, fixed buffers that work for all types.

tannergooding: That is prpbably fine, as long as the new functionality can still be used with unsafe and interop code. And as long as you can have an unsafe struct that also contains a "safe" fixed sized buffer...

VSadov: FWIW. Once there is an indexer, you will likely be able to pin it. 
Any other ref-returning indexer can be pinned, assuming it has an unmanaged type, so it feels like indexable `fixed` should be pinnable as well.

```C#
fixed (S* p = &thing.buffer[0])
{
        // ... use p here
}
```
You can use pinned pointer for anything, including interop. That would be unsafe - with all the issues regarding marshaling, block-copying and blittability (https://en.wikipedia.org/wiki/Blittable_types.)

I think this will allow you to get your scenario working. It would be unsafe and possibly platform specific.

As for 

```C#
fixed (S* p = thing.buffer)
{
        // ... use p here
}
```

We should at least support what worked before because of compat - assuming we use the same or similar syntax. Not sure if we need to go further.

tannergooding: You should also continue to be able to take the address of 'thing' (assuming 'thing' is a struct, 'thing.buffer' is a fixed array for a blittable type and the other members of 'thing' are also blittable).

I think the backing data structure being generic will hinder this scenario, unless the compiler relaxes the restriction here (taking the address of a generic type, provided it is blittable, is part of the blittable proposal).

VSadov:
Merging this, since this has been opened for a long time, and is an adjustment to already existing proposal that is mostly uncontroversial.

Note: it is still a proposal and has not been discussed by LDM.

omariom:
Can fields  declaration be dropped? For the most cases  the size could be calculated and put to StructLayout.Size. Less burden on metadata

VSadov:
@omariom - not if we want to go beyond primitive types.

Reference types will need fields. 
Private fields of structs may not be all known statically (ie: reference assemblies, versioning).

tannergooding:
The field declarations also allow variable sized primitive types (`IntPtr` and `UIntPtr`) and for structs containing those types to work in this scenario as well.

omariom:
Oh.. I thought it was about 'unmanaged types' only and types of different size (like IntPtr) would be an exception. 

Then yes, using it for well-known primitives only  doesn't make much sense

PR #1223, opened 2017-12-28
Allow suppressing emit of `localsinit` flag via an attribute.

VSadov:
It does not seem to be big enough to be a real C# language feature (as in "required by all conforming C# implementations"), but we can have it as a compiler feature where compiler may optionally recognize a specific attribute as a directive/hint to omit the flag.

There are few scenarios where forced zero-initialization may noticeably impact performance. 

The feature could be implemented as a compiler switch - similar to `/optimize`, but it could be more valuable when can be applied more selectively to specific methods or types.

### DIFF
--- a/proposals/skip-localsinit.md
+++ b/proposals/skip-localsinit.md
@@ -8,7 +8,7 @@
 ## Summary
 [summary]: #summary
 
-Allow user to suppress emitting of `localsinit` flag by applying an attribute. 
+Allow suppressing emit of `localsinit` flag via `SkipLocalsInitAttribute` attribute. 
 
 ## Motivation
 [motivation]: #motivation
--- a/proposals/skip-localsinit.md
+++ b/proposals/skip-localsinit.md
@@ -1,0 +1,91 @@
+# Suppress emitting of `localsinit` flag.
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Allow user to suppress emitting of `localsinit` flag by applying an attribute. 
+
+## Motivation
+[motivation]: #motivation
+
+
+### Background
+Per CLR spec local variables that do not contain references are not initialized to a particular value by the VM/JIT. Reading from such variables without initialization is type-safe, but otherwise the behavior is undefined and implementation specific. Typically uninitialized locals contain whatever values were left in the memory that is now occupied by the stack frame. That could lead to nondeterministic behavior and hard to reproduce bugs. 
+
+There are two ways to "assign" a local variable: 
+- by storing a value or 
+- by specifying `localsinit` flag which forces everything that is allocated form the local memory pool to be zero-initialized
+NOTE: this includes both local variables and `stackalloc` data.    
+
+Use of uninitialized data is discouraged and is not allowed in verifiable code. While it might be possible to prove that by the means of flow analysis, it is permitted for the verification algorithm to be conservative and simply require that `localsinit` is set.
+
+Historically C# compiler emits `localsinit` flag on all methods that declare locals.
+
+While C# employs definite-assignment analysis which is more strict than what CLR spec would require (C# also needs to consider scoping of locals), it is not strictly guaranteed that the resulting code would be formally verifiable:
+- CLR and C# rules may not agree on whether passing a local as `out` argument is a `use`.
+- CLR and C# rules may not agree on treatment of conditional branches when conditions are known (constant propagation).
+- CLR could as well simply require `localinits`, since that is permitted.  
+
+### Problem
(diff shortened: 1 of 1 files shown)


VSadov:
CC: @jkotas @jaredpar @KrzysztofCwalina

VSadov:
This is related to the https://github.com/dotnet/roslyn/issues/23951

One change would help another.

proposals/skip-localsinit.md
```diff
@@ -0,0 +1,91 @@
...
+- omit `localinits` flag always.
+Even worse than above.
+
+- omit `localinits` flag unless `stackalloc` is used in the method body.
+Does not address the most requested scenario and may turn code unverifiable with no option to revert that back.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should the attribute be actually emitted to metadata?
```
jkotas: Also, are we going to add the attribute to public surface so that folks can use it without local definition? (I assume that local private per-assembly definition would be possible as well.)

proposals/skip-localsinit.md
```diff
@@ -0,0 +1,91 @@
...
+In high-performance application the cost of forced zero-initialization could be noticeable. It is particularly noticeable when `stackalloc` is used.
+
+In some cases JIT can elide initial zero-initialization of individual locals when such initialization is "killed" by subsequent assignments. Not all JITs do this and such optimization has limits. It does not help with `stackalloc`.
+
+To illustrate that the problem is real - there is a known bug where a method not containing any `IL` locals would not have `localsinit` flag. The bug is already being exploited by users by putting `stackalloc` into such methods - intentionally to avoid initialization costs. That is despite the fact that absence of `IL` locals is an unstable metric and may vary depending on changes in codegen strategy. 
+The bug should be fixed and users should get a more documented and reliable way of suppressing the flag. 
+
+## Detailed design
+
+Allow specifying `System.Runtime.CompilerServices.SkipLocalsInitAttribute` as a way to tell the compiler to not emit `localsinit` flag.
```
jkotas: Nit: Would a better name of the attribute be `NoLocalsInitAttribute`?

I believe that attribute names do not typically start with a verb.

VSadov: I would be ok with any attribute name as long as it tells that it is suppressing `localsinit` flag - because that is all what it does.  

There might be more suggestions for the attribute name from which we can chose.

tannergooding: Any reason we can't add this as a flag to MethodImplOptions?

proposals/skip-localsinit.md
VSadov: Compiler does not care about the identity of attribute types, so it can be defined anywhere including an internal type in source.  
Even though it is not a requirement, adding the attribute to public surface might be useful.

I did not mentioned that intentionally, to not get into matters  of nuget packages, reference assemblies, netstandard v-next and the like. - To defer to people who know better what treatment would be appropriate in this case. 
I should add this to "unresolved questions" though.

proposals/skip-localsinit.md
tannergooding: Ex: `[MethodImpl(MethodImplOptions.NoLocalsInit)]`

It would fit in well with the existing usage of `MethodImplOptions`. I one of the drawbacks would be that `MethodImplAttribute` is only `Constructor | Method` currently.

jkotas: MethodImplOptions are understood by the runtime, not by the compiler. Would the compiler  be responsible for masking this bit out before saving it to the binary?

VSadov: I thought about `MethodImplOptions`. 
It is attractive, since there is no need for a new type, however `MethodImpl` can only target methods and constructors and expanding its applicability would not be trivial since it impacts run/jit time behavior as well.

I will list this in "alternative solutions".  
If the idea of allowing the attribute on method containers falls through for some other reasons, then this will be a viable option.

VSadov: The metadata emit is listed as one of the items that needs to be discussed further.  
Attributes that have compile-time effect at declaration is a new scenario to us.

IMO - Compiler will likely not interfere with how attribute is emitted. At least to not cause versioning/compat problems when switching between compilers.
 
That could be an issue with `MethodImplOptions` since `MethodImpl` is a "pseudoattribute" and options are emitted into bit fields in the MethodImpl metadata table. Not sure how a new bit would fit there.

Joe4evr: > Attributes that have compile-time effect at declaration is a new scenario to us.

What about `[Conditional]`? 🤔

VSadov: @Joe4evr  `[Conditional]` has effect at the place where method is called, but at the point of declaration (like where `Debug.Assert` is declared) it is just an attribute - it gets emitted to metadata.

VSadov:
Merging this, since this has been opened for a long time, and has approvals (as a viable proposal). 

NOTE: The proposal does not have direct impact on the language, but has effect on the compiler.
(that is why it fits in csharplang repo)

PR #1250, opened 2018-01-12
should implement IAsyncDisposable correctly

bbarry:
need to cast to the type in case it is implemented explicitly and the viable alternative should also implement

### DIFF
--- a/proposals/async-streams.md
+++ b/proposals/async-streams.md
@@ -72,7 +72,7 @@ try
         Use(enumerator.Current);
     }
 }
-finally { await ((IAsyncDisposable)enumerator).DisposeAsync(); }
+finally { await enumerator.DisposeAsync(); }
 ```
 
 Discarded options considered:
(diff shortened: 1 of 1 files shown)


proposals/async-streams.md
```diff
@@ -72,7 +72,7 @@ try
         Use(enumerator.Current);
     }
 }
-finally { await enumerator.DisposeAsync(); }
+finally { await ((IAsyncDisposable)enumerator).DisposeAsync(); }
```
yaakov-h: Cast is unnecessary since `IAsyncDisposable<out T>` implements `IAsyncDisposable`.

bbarry: Fair, not true for a pattern based implementation though. I'll remove.

jcouv:
Tagging @stephentoub FYI

proposals/async-streams.md
```diff
@@ -103,16 +103,20 @@ namespace System.Collections.Generic
 ```
 TryGetNext is used in an inner loop to consume items with a single interface call as long as they're available synchronously.  When the next item can't be retrieved synchronously, it returns false, and any time it returns false, a caller must subsequently invoke WaitForNextAsync to either wait for the next item to be available or to determine that there will never be another item. Typical consumption (without additional language features) would look like:
 ```C#
-IAsyncEnumerable<T> enumerator = enumerable.GetAsyncEnumerator();
-while (await enumerator.WaitForNextAsync())
+IAsyncEnumerator<T> enumerator = enumerable.GetAsyncEnumerator();
+try
 {
-    while (true)
+    while (await enumerator.WaitForNextAsync())
```
benaadams: Still think this shouldn't use the work `Wait` as that's normally blocking. 

`await enumerator.WhenNextAsync()`?

E.g. `Task.Wait` vs `Task.When`

Issue #1282, opened 2018-01-26
Converted constants in a switch statement [spec bug and proposed fix]


gafter:
There appears to be a bug in the [specification of the `switch` statement](https://github.com/dotnet/csharplang/blob/master/spec/statements.md#the-switch-statement) (ECMA 13.8.3). This appear in the spec:

> The constant expression of each `case` label must denote a value that is implicitly convertible ([Implicit conversions](https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-conversions)) to the governing type of the `switch` statement. A compile-time error occurs if two or more `case` labels in the same `switch` statement specify the same constant value.

This permits the following code, which the compiler rejects:

``` c#
    {
        const int N1 = 1;
        const short N2 = 1;
        int x = 1;
        switch (x)
        {
            case N1:
            case N2: // error CS0152: The switch statement contains multiple cases with the label value '1'
        		break;
        }
    }
```

This is permitted by the spec because `N1` and `N2` do not "specify the same constant value". They are not even of the same type.

The simple "fix" would be to require that the ***converted expressions*** do not specify the same constant value. However, that would expose a different problem. In the current spec, a null literal is not a constant value after it has been converted to a nullable type such as `int?`. As a result, the modified rule would permit multiple `null` cases (because they do not specify the same constant value after conversion, as they do not in fact specify a constant value at all after conversion).

I believe the way to fix these issues is to
1. Modify the quoted section to say something like "A compile-time error occurs if two or more `case` labels in the same `switch` statement specify the same constant value after the conversion."
2. In the section [Constant expressions](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#constant-expressions) (ECMA 12.20), in the bullet list following the introductory sentence "The following conversions are permitted in constant expressions:", add a new bullet reading "Null literal conversions".

These proposed changes correspond to what the Roslyn compiler actually does.

[This is a duplicate of an [open issue being tracked in the C# ECMA specification](https://github.com/ECMA-TC49-TG2/spec/issues/942). I have created this copy for reference by the Roslyn team. -@gafter]

Unknown6656:
@gafter : I openend a pull request.

AlekseyTs:
I do not believe the "Constant expressions" section should change. Values of nullable types should never be considered as valid constant expressions.

gafter:
@AlekseyTs Do you have an alternate proposal to fix the spec that aligns with your belief?

[...]

PR #1283, opened 2018-01-26
Addressing the issue dotnet/csharplang#1282

Unknown6656:
See the issue https://github.com/dotnet/csharplang/issues/1282 raised by @gafter.

### DIFF
--- a/spec/expressions.md
+++ b/spec/expressions.md
@@ -4821,6 +4821,7 @@ A constant expression must be the `null` literal or a value with one of  the fol
 *  References to `const` parameters or local variables
 *  Parenthesized sub-expressions, which are themselves constant expressions.
 *  Cast expressions, provided the target type is one of the types listed above.
+*  `null` literal conversions
 *  `checked` and `unchecked` expressions
 *  Default value expressions
 *  Nameof expressions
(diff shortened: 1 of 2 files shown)


gafter:
We do not have a process for evolving the normative text of the specification at the moment. We expect the process will involve moving to the ECMA version of the specification as a starting point, with the integration of many Microsoft changes, and the ECMA committee (including Microsoft members) will probably handle the change requests from there. Whatever the process is, it won't be accepting PRs into this particular version of the specification, except perhaps for things like typographical errors introduced in the translation to markdown. This version of the spec is largely a placeholder for a process yet to be established.

I'm going to leave this PR open for now, but I think you should expect it to be closed.

@MadsTorgersen Hopefully you can say more at an appropriate time?

spec/expressions.md
```diff
@@ -4821,6 +4821,7 @@ A constant expression must be the `null` literal or a value with one of  the fol
 *  References to `const` parameters or local variables
 *  Parenthesized sub-expressions, which are themselves constant expressions.
 *  Cast expressions, provided the target type is one of the types listed above.
+*  `null` literal conversions
```
gafter: The name of the null literal conversion does not use a code font. I think this would be

> * Null literal conversions

Unknown6656: Thank you, @gafter . I updated the line.

Unknown6656:
Thank you very much, @gafter for your quick and elaborate response.

AdamSpeight2008:
Changes should be okay if the specifications on here explicitly state it is non-ECMA standard.

gafter:
@AdamSpeight2008 We have no intention of developing a language to diverge from ECMA. It is our intent to develop future versions of the specification jointly with ECMA. So taking changes to the spec of the core language that intentionally diverge from ECMA would be counterproductive.

PR #1306, opened 2018-02-03
Create target-typed-new.md

alrz:
[Rendered](https://github.com/alrz/csharplang/blob/patch-2/proposals/target-typed-new.md)

### DIFF
--- a/proposals/target-typed-new.md
+++ b/proposals/target-typed-new.md
@@ -1,0 +1,100 @@
+
+# Target-typed `new` expressions
+
+* [x] Proposed
+* [x] [Prototype](https://github.com/alrz/roslyn/tree/features/implicit-new)
+* [ ] Implementation
+* [ ] Specification
+
+## Summary
+[summary]: #summary
+
+Do not require type specification for constructors when the type is known. 
+
+## Motivation
+[motivation]: #motivation
+
+Allow field initialization without duplicating the type.
+```cs
+Dictionary<string, List<int>> field = new() {
+    { "item1", new() { 1, 2, 3 } }
+};
+```
+Allow omitting the type when it can be inferred from usage.
+```cs
+XmlReader.Create(reader, new() { IgnoreWhitespace = true });
+```
+
+## Detailed design
+[design]: #detailed-design
+
+The *object_creation_expression* syntax would be modified to make the *type* optional when parentheses are present. This is required to address the ambiguity with *anonymous_object_creation_expression*.
+```antlr
+object_creation_expression
+    : 'new' type? '(' argument_list? ')' object_or_collection_initializer?
+    | 'new' type object_or_collection_initializer
+    ;
+```
+The target-type is resolved through a conversion that only exists if and only if there exists a single best accessible constructor with regard to provided arguments. Target of the initialization is determined after the conversion succeeded.
+
+The type of the expression would be inferred from the target-type which would be required to be one of the following:
+
+- **Any struct type**
+- **Any reference type**
+- **Any type parameter** with a constructor constraint
+
+The following types do not contribute to the conversion:
+
+- **Enum types:** not all enum types contain the constant zero, so it should be desirable to use the explicit enum member.
+- **Delegate types:** if the type is inferrable, an anonymous function can already be used.
+- **Interface types:** this is a niche feature and it should be preferable to explicitly mention the type.
+- **Tuple types:** tuple literals should be used for such expressions.
+- **Array types:** arrays need a special syntax to provide the length.
+
+
+All the other types that are not permitted in the *object_creation_expression* are excluded as well, for instance, pointer types.
+
+>**Open Issue:** What other types should be excluded?
+
+
+
+Note that any restriction on permitted types would raise the success rate of the overload resolution. For example, if we exclude primitive types, the following would successfully compile.
+```cs
+class C {}
+void M(C c) {}
+void M(int i) {}
+
+M(new());
+```
+Otherwise, it would fail with a `CS0121` error.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+Since we're relying on overload resolution of the target-type members, there is a high chance that any change in one of the participating members (target-type's constructors or call-site method overloads, if any) would result in breaking the compilation. For instance,
+```cs
+class Foo { }
+class Bar { }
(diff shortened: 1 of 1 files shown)


proposals/target-typed-new.md
```diff
@@ -85,14 +85,14 @@ The invocation compiles until a overload like `void M(Bar bar) {}` is added alon
 ## Alternatives
 [alternatives]: #alternatives
 
-Most of complains about types being too long to duplicate in field initialization is *type arguments* not the type itself, we could infer only type arguments like `new Dictionary(...)` (or similar) instead of the whole type, in which case, it would be a lot less likely to break compilation with adding a constructor or an overloaded member.
+Most of complains about types being too long to duplicate in field initialization is about *type arguments* not the type itself, we could infer only type arguments like `new Dictionary(...)` (or similar) instead of the whole type, in which case, it would be a lot less likely to break compilation with adding a constructor or an overloaded member.
```
yaakov-h: Perhaps "complaints"?

alrz: Yes, thanks.

jcouv:
Thanks
Tagging @gafter as well

jcouv:
Because of the possible ramifications, this will be discussed in the context of C# 8.0 (as opposed to a point release). I'll keep you updated once we manage to clear some LDM backlog and can get this on the agenda. I'd expect in at least a month. Thanks for your patience.

proposals/target-typed-new.md
```diff
@@ -0,0 +1,100 @@
...
+- **Array types:** arrays need a special syntax to provide the length.
+
+
+All the other types that are not permitted in the *object_creation_expression* are excluded as well, for instance, pointer types.
+
+>**Open Issue:** What other types should be excluded?
+
+
+
+Note that any restriction on permitted types would raise the success rate of the overload resolution. For example, if we exclude primitive types, the following would successfully compile.
```
gafter: Rather than excluding primitive types, I suggest we ought to exclude the default constructor for a value type (unless the constructor appears explicitly in metadata). That would have the effect of excluding the primitive types and the default constructor or most value types as well. If you wanted to use the default value of such types you could write `default`.

proposals/target-typed-new.md
```diff
@@ -0,0 +1,100 @@
...
+XmlReader.Create(reader, new() { IgnoreWhitespace = true });
+```
+
+## Detailed design
+[design]: #detailed-design
+
+The *object_creation_expression* syntax would be modified to make the *type* optional when parentheses are present. This is required to address the ambiguity with *anonymous_object_creation_expression*.
+```antlr
+object_creation_expression
+    : 'new' type? '(' argument_list? ')' object_or_collection_initializer?
```
gafter: It is not clear how or when the `object_or_collection_initializer` is bound. I presume it is bound as part of the conversion (i.e. late), but that errors within it do not affect whether or not the given constructor is inferred. That should be made explicit.

proposals/target-typed-new.md
```diff
@@ -0,0 +1,100 @@
...
+```
+The target-type is resolved through a conversion that only exists if and only if there exists a single best accessible constructor with regard to provided arguments. Target of the initialization is determined after the conversion succeeded.
+
+The type of the expression would be inferred from the target-type which would be required to be one of the following:
+
+- **Any struct type**
+- **Any reference type**
+- **Any type parameter** with a constructor constraint
+
+The following types do not contribute to the conversion:
```
gafter: Perhaps say "with the following exceptions", as it otherwise appears to disagree with the above.

proposals/target-typed-new.md
alrz: > that errors within it do not affect whether or not the given constructor is inferred

we don't bind the initializer when we try to create the conversion, it only happens after the conversion succeeded. This is mentioned in the next line:

>Target of the initialization is determined after the conversion succeeded.

alrz:
@gafter this is ready for merge. thanks.

proposals/target-typed-new.md
```diff
@@ -0,0 +1,99 @@
...
+
+Note that any restriction on permitted types would raise the success rate of the overload resolution. For example:
+```cs
+class C {}
+void M(C c) {}
+void M(int i) {}
+
+M(new());
+```
+This would fail with a `CS0121` error, considering the restriction on the default constructor for value types.
```
jcouv: >CS0121 [](start = 24, length = 6)

Perhaps quote the message for the error

proposals/target-typed-new.md
```diff
@@ -0,0 +1,99 @@
...
+    | 'new' type object_or_collection_initializer
+    ;
+```
+The target-type is resolved through a conversion that only exists if and only if there exists a single best accessible constructor with regard to provided arguments. Target of the initialization is determined after the conversion succeeded.
+
+The type of the expression would be inferred from the target-type which would be required to be one of the following:
+
+- **Any struct type**
+- **Any reference type**
+- **Any type parameter** with a constructor constraint
```
alrz: I think we should add "a constructor or a struct constraint" because in this case default and new wouldn't be identical.

proposals/target-typed-new.md
```diff
@@ -0,0 +1,99 @@
...
+
+- **Enum types:** not all enum types contain the constant zero, so it should be desirable to use the explicit enum member.
+- **Delegate types:** if the type is inferrable, an anonymous function can already be used.
+- **Interface types:** this is a niche feature and it should be preferable to explicitly mention the type.
+- **Tuple types:** tuple literals should be used for such expressions.
+- **Array types:** arrays need a special syntax to provide the length.
+
+All the other types that are not permitted in the *object_creation_expression* are excluded as well, for instance, pointer types.
+
+The default constructor for a value type (unless appeared explicitly in metadata) is ought to be excluded, which would have the effect of excluding the primitive types and the default constructor for most value types. If you wanted to use the default value of such types you could write `default`.
```
gafter: >is ought [](start = 82, length = 8)

"is" or "ought to be"? ... "would have" or "has" the effect? Please take a position one way or the other (though whatever position you take you can leave it as an open issue that we might reconsider).

alrz: I was just quoting your words, I think that's why it came out wired.

PR #1373, opened 2018-03-11
Make example use <see> how it is intended to be used

carlreinke:
`<see>` is meant to be used inline, so the examples should use it inline.


sharwell:
💡 The alternative to the approach in this pull request is replace the top-level `<see>` elements with `<seealso>` elements (the top-level equivalent).

carlreinke:
The examples are supposed to demonstrate how to use `<see>`, so using `<seealso>` would defeat the purpose.

PR #1375, opened 2018-03-12
Update lexical-structure.md

epicbugs:
The first paragraph of Lexical Structure defines what is a source file in way that is not entirely correct (by stating that it is "an ordered sequence"). Every sequence is ordered already by definition as each and every element of any sequence has a distinct and unique position. Also, there is no such thing as "an unordered sequence". Hence the term "an ordered sequence" is something of a semantic pleonasm. If there is a need to stress that the order in which characters appear in the source file is of crucial importance, why not use "strictly ordered collection".

### DIFF
--- a/spec/lexical-structure.md
+++ b/spec/lexical-structure.md
@@ -2,7 +2,7 @@
 
 ## Programs
 
-A C# ***program*** consists of one or more ***source files***, known formally as ***compilation units*** ([Compilation units](namespaces.md#compilation-units)). A source file is an ordered sequence of Unicode characters. Source files typically have a one-to-one correspondence with files in a file system, but this correspondence is not required. For maximal portability, it is recommended that files in a file system be encoded with the UTF-8 encoding.
+A C# ***program*** consists of one or more ***source files***, known formally as ***compilation units*** ([Compilation units](namespaces.md#compilation-units)). A source file is a strictly ordered collection of Unicode characters. Source files typically have a one-to-one correspondence with files in a file system, but this correspondence is not required. For maximal portability, it is recommended that files in a file system be encoded with the UTF-8 encoding.
 
 Conceptually speaking, a program is compiled using three steps:
 
--- a/spec/lexical-structure.md
+++ b/spec/lexical-structure.md
@@ -2,7 +2,7 @@
 
 ## Programs
 
-A C# ***program*** consists of one or more ***source files***, known formally as ***compilation units*** ([Compilation units](namespaces.md#compilation-units)). A source file is a strictly ordered collection of Unicode characters. Source files typically have a one-to-one correspondence with files in a file system, but this correspondence is not required. For maximal portability, it is recommended that files in a file system be encoded with the UTF-8 encoding.
+A C# ***program*** consists of one or more ***source files***, known formally as ***compilation units*** ([Compilation units](namespaces.md#compilation-units)). A source file is a sequence of Unicode characters. Source files typically have a one-to-one correspondence with files in a file system, but this correspondence is not required. For maximal portability, it is recommended that files in a file system be encoded with the UTF-8 encoding.
 
 Conceptually speaking, a program is compiled using three steps:
 


svick:
I don't think "strictly ordered collection" is an improvement, I think using just "sequence" would be fine.

epicbugs:
@svick Replaced.

alrz:
Although I agree "sequence" does imply "ordered" but I think that is just to make it explicit.

epicbugs:
@alrz IMHO saying you can have "an ordered sequence" is the equivalent of talking about triangular triangles and whole number integers. It's just silly. Besides, I don't think this document is meant for those who don't know what a sequence is. But if there really is an absolutely unavoidable need to make the property of ordering explicit, then "an ordered collection" is a more precise term. No?

CyrusNajmabadi:
> Although I agree "sequence" does imply "ordered" but I think that is just to make it explicit.

I agree with this.  I think "ordered sequence" reads fine and acts as just a form of emphasis.

epicbugs:
@CyrusNajmabadi Sorry, I'm not following. To emphasize what? That the content of a source file is not random?

alrz:
I think that emphasizes the order in which we parse the source, that's where the "lexical order" comes from.

epicbugs:
@alrz Still not following. As far as I know, single characters don't have a lexical order except in a sense of the alphabet they appear in. Tokens are a another matter but that's on a level where single characters are no longer relevant. How is this "lexical" order of characters in a source file different from their positional order? To me it seems they are the same while you guys seem to insisting that they aren't.

CyrusNajmabadi:
> To emphasize what?  that the 'sequence' is 'ordered'.

epicbugs:
@CyrusNajmabadi @alrz Any sequence is ordered by definition, and conversly, there is no such thing as an un-ordered sequence. The qualifier in question is not only redundant, but it is also misleading -- to a knowledgeable person it is redundant, and to a less educated person it is misleading as it implies a sequence as such has no default order or can have several orderings at once which is also wrong.

CyrusNajmabadi:
> to a knowledgeable person 

I'm knowledgeable, and it reads just fine to me.  It comes across as emphasis to me and does not feel misleading.   :)

alrz:
>there is no such thing as an un-ordered sequence. The qualifier in question is not only redundant, but it is also misleading

It signifies the fact that the order is important and is respected. you can consume a sequence in whichever order you want, regardless. So I think that is not neither redundant nor misleading.

epicbugs:
@CyrusNajmabadi One reason it reads fine to you can be that you've simply gotten used to the redundancy and give too much importance to the feeling that something would be missing without it. Another explanation is that an ephasis is needed but would be better served by using another word or another phrase.

Either way, my suggestion here would be to instead take the classical (Strunk and White) way: "Omit needless words". But if you really are sure it is needed, then by all means, why not highlight the object or relationship you think it emphasizes (because I don't see it).

epicbugs:
@alrz I understand your reasoning but I still don't see where you're coming from. What makes you say that it's discussing the consumption of that sequence? I don't see anything about consumption in that sentence. What I see being described is the low-level structure of a source file.

CyrusNajmabadi:
> Another explanation is that an ephasis is needed but would be better served by using another word or another phrase.

Enh.  :)

I'm just saying: it seems fine.  Nothing feels unclear or misleading to me.  I don't feel like removing the word makes much of a change.

epicbugs:
@CyrusNajmabadi Omit needless words. The text is long enough even without  them.

jnm2:
https://docs.microsoft.com/style-guide/top-10-tips-style-voice#be-brief 😈

CyrusNajmabadi:
> @CyrusNajmabadi Omit needless words.

I don't want the spec to just be a formal specification of C# (which would be the smallest and most concise possible spec).  I want it to contain *tons* of normative text that help me, as the reader, understand better :)

gafter:
The premise of this fix is that 'The first paragraph of Lexical Structure defines what is a source file in way that is not entirely correct (by stating that it is "an ordered sequence").'

However, that *is* entirely correct, though admittedly redundant. The redundancy was intentional, so I don't see a reason to accept this change.

epicbugs:
@alrz @CyrusNajmabadi @gafter Be as it may, I've made my case. Intention is something that's evident only to those involved, and effects don't follow from intention alone. Thank you for your time and attention, and if there's anything else you'd like to add, feel free to reply.

PR #1389, opened 2018-03-16
Fixed typo

cincuranet:


jnm2:
I'm not sure this is a typo. https://en.wikipedia.org/wiki/If_and_only_if

cincuranet:
Ahh, completely forgot about that.

PR #1393, opened 2018-03-17
spec/lexical-structure.md: fixed definition of delimited comment

pkulikov:
According to the current definition of the delimited comment in the [specification](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#comments) the line 
```csharp
/*/
```
 is a valid comment, which is not true.

### DIFF
--- a/spec/lexical-structure.md
+++ b/spec/lexical-structure.md
@@ -110,7 +110,7 @@ new_line_character
     ;
 
 delimited_comment
-    : '/*' delimited_comment_section* asterisk* '/'
+    : '/*' delimited_comment_section* asterisk* '*/'
     ;
 
 delimited_comment_section
--- a/spec/lexical-structure.md
+++ b/spec/lexical-structure.md
@@ -110,7 +110,7 @@ new_line_character
     ;
 
 delimited_comment
-    : '/*' delimited_comment_section* asterisk* '*/'
+    : '/*' delimited_comment_section* asterisk+ '/'
     ;
 
 delimited_comment_section
@@ -673,7 +673,7 @@ open_brace_escape_sequence
     : '{{'
     ;
 
-    close_brace_escape_sequence
+close_brace_escape_sequence
     : '}}'
     ;
     


spec/lexical-structure.md
```diff
@@ -110,7 +110,7 @@ new_line_character
     ;
 
 delimited_comment
-    : '/*' delimited_comment_section* asterisk* '/'
+    : '/*' delimited_comment_section* asterisk* '*/'
```
gafter: The correct fix is to change `asterisk*` to `asterisk+`.

A similar change needs to be made in `delimited_comment_section` to remove an ambiguity.

pkulikov: @gafter I'll change `asterisk*` to `asterisk+`; just want to check with you that suggested fix is also correct. I've chosen that variant to match the sentence:
>**Delimited comments** start with the characters `/*` and end with the characters `*/`.

I don't understand your remark about `delimited_comment_section`. Similar change means
```antlr
delimited_comment_section
    : '/'
    | asterisk+ not_slash_or_asterisk
    ;
```

gafter: I was wrong about `delimited_comment_section`.

gafter:
Before we pull this change, we should determine whether it is the specification or the compiler that has a bug.

pkulikov:
@gafter the same page, just before the block with formal definition, mentions that delimited comments are surrounded by `/*` and `*/`. I really doubt that the compiler has a bug. How do we determine?

jnm2:
I hope the compiler keeps its current behavior because of this cool pattern I discovered a long time ago:

```cs
/*
DebugCodeA();
/*/
DebugCodeB();
//*/
```

To switch between the two, add an extra character before the first comment:
```cs
//*
DebugCodeA();
/*/
DebugCodeB();
//*/
```

gafter:
> the same page, just before the block with formal definition, mentions that delimited comments are surrounded by `/*` and `*/`. I really doubt that the compiler has a bug. How do we determine?

If the compiler doesn't comply with the formal definition (by accepting `/*/` as a comment) then it has a bug. Either that or there is a bug in the spec. We have to decide which it is.

pkulikov:
@gafter I think it's the bug in the draft in this repo. This PR aligns the draft with the [ECMA-334.pdf](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-334.pdf), see page 17 of the document.

pkulikov:
@gafter 
> If the compiler doesn't comply with the formal definition (by accepting /*/ as a comment) then it has a bug.

The compiler doesn't accept `/*/` as a comment. So, I think this spec draft has a bug.

PR #1444, opened 2018-04-06
Add raw notes

MadsTorgersen:


CyrusNajmabadi:
> If we have a range operator `1..10`, would we similarly have a range pattern? How would it work?
```
+
+``` c#
+    if (ch is in 'a' to 'z')
+    switch (ch) {
+        case in 'a' to 'z':
```

How about ```ch in ['a'..'z']"``` (or whatever syntax we have for a range that is inclusive?.

PR #1512, opened 2018-05-11
Add singleton proposal for C# language

csblo:


HaloFour:
Proposals should be submitted as Issues and discussed before a PR is submitted.  Any proposal needs to be "championed" by a member of the design team in order to be considered.

https://github.com/dotnet/csharplang#design-process

theunrepentantgeek:
The singleton pattern is more of a *runtime policy* than anything else; enforcement within the language eliminates a lot of useful possibilities - such as unit testing, singleton-per-request, singleton-per-transaction, and so on.

DanFTRX:
Is this just a lazy loaded readonly?

csblo:
Thanks for comments, after reading them, I don't think that I'll maintain my proposal.

PR #1593, opened 2018-06-01
Clarify Span<T> rules

jaredpar:
While discussing `Span<T>` safety rules with the F# team I realized we
hadn't fully documented our dependency on the shape of the `Span<T>`
API. Added that plus some of the future considerations we discussed for
`Span<T>`.

https://github.com/Microsoft/visualfsharp/pull/4888

### DIFF
--- a/proposals/csharp-7.2/span-safety.md
+++ b/proposals/csharp-7.2/span-safety.md
@@ -236,6 +236,20 @@ A `new` expression that invokes a constructor obeys the same rules as a method i
 
 In addition *safe-to-escape* is no wider than the smallest of the *safe-to-escape* of all arguments/operands of the object initializer expressions, recursively, if initializer is present. 
 
+## Span constructor
+The language relies on `Span<T>` not having a constructor of the following form:
+
+``` csharp
+void Example(ref int x)
+{
+    // Create a span of length one
+    var span = new Span<int>(ref x); 
+}
+```
+
+Such a constructor makes `Span<T>` which are used as fields indistinguishable from a `ref` field. The safety rules described in this document
+depend on `ref` fields not being a valid construct in C#, or .NET.
+
 ## `default` expressions
 
 A `default` expression is *safe-to-escape* from the entire enclosing method.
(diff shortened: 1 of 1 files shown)


jaredpar:
CC @gafter, @dsyme, @TIHan, @VSadov

jaredpar:
CC @OmarTawfik

proposals/csharp-7.2/span-safety.md
```diff
@@ -280,3 +294,37 @@ We wish to ensure that no `ref` local variable, and no variable of `ref struct`
...
+    // Simpler to just allow length one span
+    var span2 = new Span<int>(ref x);
+    Use(span2);
+}
+```
+
+This feature gets more compelling if we lift the restrictions on (fixed sized buffers)[https://github.com/dotnet/csharplang/blob/master/proposals/fixed-sized-buffers.md] as it would
+allow for `Span<T>` instances of even greater length. 
+
+If there is ever a need to go down this path then the language could accommodate this by ensuring such `Span<T>` instances
```
gafter: Not sure this would work. By the time we would want to make that change, there might already be a user-declared type
``` c#
internal ref struct MySpan<T>
{
    public MySpan(ref T data) ...;
}
```
This API is perfectly safe today because it doesn't (has no safe way to) capture the data.
``` c#
MySpan<int> M(int x)
{
    return new MySpan<int>(ref x);
}
```
Although this is safe today, but would become illegal under the rule changes we would need to permit a `Span` that can capture a reference.

jaredpar: That constructor though couldn't be implemented in a way that captured `ref T data`. If it attempted to do so via `Span<T>` then it would be an error as the value wouldn't be escapable beyond the constructor body and hence couldn't be assigned to a field.

VSadov: Yes, that would work. Because:
- we do not have such constructors on Span/ReadOnly span right now. 
- with suggested restrictions the resulting instances will never escape the scope that contains the constructor invocation, so they will not introduce any new escaping scenarios for the outer scopes.

VSadov: I would also add that in a case when the ref argument is  "definitely referring to a heap object" - like a ref to an array element, for example, the resulting instance could be allowed to escape. It feels like it might be another common case.

proposals/csharp-7.2/span-safety.md
```diff
@@ -280,3 +294,37 @@ We wish to ensure that no `ref` local variable, and no variable of `ref struct`
 > ``` c#
 > Foo(new Span<int>(...), await e2);
 > ```
+
+# Future Considerations
+
+## Length one Span<T> over ref values
+Though legal today there are cases where creating a length one `Span<T>` instance over a value would be beneficial:
```
VSadov: typo?   _not_ legal.

PR #1635, opened 2018-06-15
Fix code format

lindexi:
We use [Allman style](http://en.wikipedia.org/wiki/Indent_style#Allman_style) braces, where each brace begins on a new line. 

We should use `_camelCase` for internal and private fields

See: [coding-style.md](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md )


jcouv:
Thanks

PR #1640, opened 2018-06-15
operator exclamation mark-question mark -dot proposal.md

Anysharp:
short form of (b==null)?null:b.c would become b!?.c

### DIFF
--- a/proposals/proposal-template.md
+++ b/proposals/proposal-template.md
@@ -1,4 +1,4 @@
-# FEATURE_NAME
+# !?. operator
 
 * [x] Proposed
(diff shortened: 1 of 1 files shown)


amis92:
1. Please read README. Especially the last paragraph of the first section: https://github.com/dotnet/csharplang#c-language-design

Citing:
> For new feature proposals, however, please raise them for discussion, and only submit a proposal as a pull request if invited to do so by a member of the Language Design Team (a "champion").

2. What is the difference between your proposal and the current null conditional access operator? (ref: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators)

Anysharp:
Sorry, this feature already exists since c# 6.0, I missed it when I searched for it. It's simply the "?." operator.

Issue #1655, opened 2018-06-22
Typo in `proposals/csharp-7.2/readonly-ref.md`, or bad implementation in Roslyn?


stakx:
@zvirja noticed in https://github.com/castleproject/Core/pull/370#issuecomment-399370350 that there appears to be a typo in [proposals/csharp-7.2/readonly-ref.md](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/readonly-ref.md#metadata-representation-of-in-parameters):

> ## Metadata representation of `in` parameters
>
> [...] In addition, if the method is *abstract* or *virtual*, then the signature of such parameters (and only such parameters) must have `modreq[System.Runtime.CompilerServices.IsReadOnlyAttribute]`.

Roslyn actually appears to encode a `modreq([mscorlib]System.Runtime.InteropServices.InAttribute)`.

So which party has got it right: Roslyn, or the proposal in this repository?

I suppose the answer will be that Roslyn is right, that proposal seems out-of-date. Or has specification really not started yet even though the feature has already shipped?

> * [X] Proposed
> * [X] Prototype
> * [X] Implementation: Started
> * [ ] Specification: Not Started

bbarry:
`InAttribute` has been in the framework since at least 1.1 and shouldn't need to be duplicated and is defined in the mscorlib assembly in .net standard 2.0: https://github.com/dotnet/standard/blob/e2cbc66c56edf5dc43cc00902ead04d20e32795e/netstandard/ref/mscorlib.cs#L10036

The test cases in Roslyn utilize the assembly qualifier, but the actual code in the compiler doesn't seem to depend on anything more than the namespace qualified type.  I think it would be a breaking change to the compiler to require an assembly qualifier and is inconsistent with other well known types. 

Perhaps the business of `modreq` should be said differently. The CIL representation is omitted elsewhere in places where the language spec discusses attribute usage.  Elsewhere in the specification language such as "`...Attribute` is used on ..."  and `...Attribute` is defined ...`.

Maybe this would be better:

> In addition, if the method is *abstract* or *virtual*, then the signature of such parameters (and only such parameters) must have the attribute `System.Runtime.CompilerServices.IsReadOnlyAttribute` defined.

stakx:
Seeing that #1658 was approved and merged, I will take this to mean that the updated documentation reflects how things ought to be. Closing.

PR #1657, opened 2018-06-22
incorrect type specified

bbarry:
InAttribute should be used for `in` parameters

### DIFF
--- a/proposals/csharp-7.2/readonly-ref.md
+++ b/proposals/csharp-7.2/readonly-ref.md
@@ -257,9 +257,9 @@ The life time of the argument temporaries matches the closest encompassing scope
 The formal life time of temporary variables is semantically significant in scenarios involving escape analysis of variables returned by reference.
 
 ## Metadata representation of `in` parameters.
-When `System.Runtime.CompilerServices.IsReadOnlyAttribute` is applied to a byref parameter, it means that the the parameter is an `in` parameter.
+When `System.Runtime.InteropServices.InAttribute` is applied to a byref parameter, it means that the the parameter is an `in` parameter.
 
-In addition, if the method is *abstract* or *virtual*, then the signature of such parameters (and only such parameters) must have `modreq[System.Runtime.CompilerServices.IsReadOnlyAttribute]`.
+In addition, if the method is *abstract* or *virtual*, then the signature of such parameters (and only such parameters) must have `modreq[System.Runtime.InteropServices.InAttribute]`.
 
 **Motivation**: this is done to ensure that in a case of method overriding/implementing the `in` parameters match.
 


bbarry:
fixes #1655

proposals/csharp-7.2/readonly-ref.md
```diff
@@ -257,9 +257,9 @@ The life time of the argument temporaries matches the closest encompassing scope
 The formal life time of temporary variables is semantically significant in scenarios involving escape analysis of variables returned by reference.
 
 ## Metadata representation of `in` parameters.
-When `System.Runtime.CompilerServices.IsReadOnlyAttribute` is applied to a byref parameter, it means that the the parameter is an `in` parameter.
+When `System.Runtime.InteropServices.InAttribute` is applied to a byref parameter, it means that the the parameter is an `in` parameter.
```
stakx: I think this is wrong. The Roslyn compiler places a `System.Runtime.CompilerServices.IsReadOnlyAttribute` custom attribute (and optionally a `System.Runtime.InteropServices.InAttribute`) for the required custom modifier.

Note that `ParameterInfo.GetCustomAttributes()` may report a custom attribute of type `System.Runtime.InteropServices.InAttribute`, but that is actually a fiction produced by Reflection.

proposals/csharp-7.2/readonly-ref.md
```diff
@@ -257,9 +257,9 @@ The life time of the argument temporaries matches the closest encompassing scope
...
 
-In addition, if the method is *abstract* or *virtual*, then the signature of such parameters (and only such parameters) must have `modreq[System.Runtime.CompilerServices.IsReadOnlyAttribute]`.
+In addition, if the method is *abstract* or *virtual*, then the signature of such parameters (and only such parameters) must have `modreq[System.Runtime.InteropServices.InAttribute]`.
```
stakx: This appears to be what Roslyn is doing, but before this gets merged, I'd like to hear from the language design team whether this is a typo or a implementation bug (which, if so, probably cannot be reversed anymore now that the feature has shipped).

proposals/csharp-7.2/readonly-ref.md
stakx: (Furthermore, it might be good to be more precise about what "is applied to" means: Applied as a custom attribute, or as a required custom modifier? But that probably needs to be discussed in the issue first.)

bbarry: agreed, will close this and submit a new request with just the other change

bbarry:
closing to remove the first change from this

proposals/csharp-7.2/readonly-ref.md
stakx: Why not wait until we have an answer from the C# language team?

bbarry: I was reviewing the tests in Roslyn to determine this. In any case, branches and PRs are cheap and easy to dismiss.

PR #1658, opened 2018-06-22
abstract/virtual parameters use InAttribute

bbarry:
An `in` parameter must have the attribute `[IsReadOnly]`. It may have the attribute `[In]` if and only if it is either abstract or virtual.

issue #1655


stakx:
Like I commented in your previous issue, I think it would be good to wait for an answer from the C# language design team before suggesting a change, so that we know for sure whether (and how exactly) it should be changed.

tannergooding:
CC. @jcouv, @jaredpar. This improves the documentation to better match the actual behavior. Can we merge?

PR #1667, opened 2018-06-25
Proposal: "nominal records" for C#

agocke:
This is the first part of a set of proposals about data manipulation and representation in C#, but can be viewed as a standalone language feature.

### DIFF
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -1,0 +1,241 @@
+
+# Working with Data
+
+When we talk about C# and start to talk about data, the conversation often
+moves to talk about advanced data structures and complicated use cases. Here
+I'd like to do the opposite: talk about simple data and the representations
+we use for it.
+
+To start, let's talk about what data is and what it isn't.
+
+Data *is* a collection of values with potentially heterogenous types. Some
+examples include
+
+* Database rows
+* JSON/XML messages
+* Login info
+* Configuration options
+
+Data *is not*
+
+* A process
+* A computation
+* A conversation
+* Interactive
+* An object
+
+What you can do with data
+
+* Name it
+* Read it
+* Modify it (or prevent modification)
+* Compose it
+* Compare it
+* Copy it
+
+What you can't do
+
+* Call it
+* Query it
+
+For an object-oriented language this may seem strange, because isn't
+everything an object? In some sense, you can view data as a degenerate object
+-- fields with pure transparency. But this also misses the point. The point
+of object-oriented architecture is to bundle state and behavior and provide
+composable objects that can interactively respond to the system, like cells
+in an organism. There's value in this structure, but it also creates a
+binding between the data and the behavior. By creating data individually we
+allow the data to shift contexts and allow other components to define their
+own behaviors.
+
+## What does it look like in C#?
+
+C# has a couple different ways to represent simple data, but I contend that
+all have fundamental problems.
+
+**Anonymous types**. Things look pretty good at first: you can name, it
+can read it, and you can compare it easily (equality is automatically
+defined). You can't modify it, though -- anonymous types are always immutable
+and you can't easily create a copy with only one change. The real problem
+is composibility. You can't use it as a real type anywhere outside the
+current method and you can't nest it in other data structures, except as
+an object.
+
+**Tuples**. Tuples are a lot like anonymous types. You can provide names,
+read the elements, modify them, and compare them. You can't make them
+immutable, but the real problem is in composition. Tuples aren't really
+an abstraction -- you describe the data structure in full in every
+place you use it. This makes it hard to expand tuples past a certain
+size and makes it difficult to compose with other data structures because
+you cannot refer to them by name.
+
+**Classes/Structs**. This is by far the most common representation of data
+in C#. A canonical example looks something like this:
+
+```C#
+public class LoginResource
+{
+    public string Username { get; set; }
+    public string Password { get; set; }
+    public bool RememberMe { get; set; } = false;
+}
+```
+
+This feature provides names, for both the members and the data structure, it
+provides easy nominal composition, and is easily composible with all other
+data structures. It also provides a convenient syntax for creation by
+interacting directly with the named data, e.g.
+
+```C#
+var x = new LoginResource {
+    Username = "andy",
+    Password = password
+}
+```
+
+Unfortunately, there are still serious problems. There is no piecewise
+comparer implicitly defined for C# classes, so if you want simple data
+comparison, the real example looks like
+[this](https://gist.github.com/agocke/6ba6e64f77f1212ba7292bfd1f1000e5)
+
+Immutable data is also a problem. The object initializer syntax provides
+a simple name-based mechanism to create a data type. With `readonly`
+fields or properties, a constructor must be used instead. This creates
+another set of problems:
+
+1. A constructor must be manually defined.
+1. The constructor parameters are ordered, while the fields are not.
+   Consumers have now taken a dependency on the parameter ordering.
+1. The constructor must be maintained with any field changes.
+
+There is also no way to create a copy of a data structure with readonly
+fields with only one item changed. A new type must be constructed manually.
+
+
+## Proposal
+
+To resolve many of these issues, I propse a new modifier for classes and structs: `data`.
+`data` classes or structs are meant to satisfy the goals listed above by doing the
+following things:
+
+1. Automatically generating `Equals`, `GetHashCode`, `ToString`, `==`, `!=`, and `IEquatable<T>`
+   based on the member data of the type.
+1. Allow object initializers to also initialize readonly members.
+
+### Equality
+
+First, the generation of equality support. Data members are only public
+fields and auto-properties. This allows data classes to have private
+implementation details for things like caching without giving up simple
+equality semantics. There are a few places this could be problematic. For
+instance, only auto-properties are considered data members by default, but
+it's not uncommon to have some simple validation included in a property
+getter that does not meaningfully change the semantics, e.g.
+
+```C#
+{
+    ...
+    private int _field;
+    public int Field
+    {
+        get
+        {
+            Debug.Assert(_field >= 0);
+            return _field;
+        }
+    }
+}
+```
+
+To support these cases and provide an easy escape hatch, I propose a
+new attribute, `DataMemberAttribute` with a boolean flag argument on the
+constructor. This allows users to override the normal behavior and include
+or exclude extra members in equality. The previous example would now read:
+
+```C#
+{
+    ...
+    private int _field;
+
+    [DataMember(true)]
+    public int Field
+    {
+        get
+        {
+            Debug.Assert(_field >= 0);
+            return _field;
+        }
+    }
+}
+```
+
+Equality itself would be defined in terms of its data members. A `data` type
+is equal to another `data` type when there is an implicit conversion between
+the target type and the source type and each of the corresponding members
+are equal. The members are compared by `==` if it is available. Otherwise,
+the method `Equals` is tried according to overload resolution rules (st. an
+`Equals` method with an identity conversion to the target type is preferred
+over the virtual `Equals(object)` method).
+
+Note that the equality defined in this manner is not symmetric in the case
+of record subtyping. This topic will be addressed in more detail later, but
+it's important to note for equality that `a.Equals(b)` may be true if `B`
+is a subtype of `A` (and therefore has an implicit conversion) and all of
+the shared members between `A` and `B` are equal. In contrast, `A` would
+likely not have an implicit conversion to `B`, so they would not be
+considered equal.
+
+`GetHashCode` would be implemented by calling `GetHashCode` on each of
+the data members and these would be combined using a symmetric operation
+(like `xor`) to prevent ordering from affecting the results.
+
+### Readonly initialization
+
+Support for `readonly` members in object initializers may seem like a small
+feature, but it's important that making a `data` type readonly not come with
+a lot of extra ceremony. The essence of a `data` type is a set of named fields
+and that should stay true regardless of whether or not the fields are 
+`readonly`. It may be tempting for implementation simplicity to try to use
+constructors instead, but this is a design smell that conflates positional
+semantics with `readonly` semantics. Requiring initialization via constructor
+means that field order becomes a public API and requires careful versioning,
+which is not true of mutable `data` types and is a constraint that should
+be irrelevant to `readonly` semantics.
+
+One way to remove dependence on a constructor is simply not make the members
+`readonly` in metadata. The CLR treats `readonly` mostly as guidance -- it
+can easily be overriden using reflection anyway. Most of the safety of
+`readonly` members in C# is not provided by the runtime, but by C# safety
+rules. One way we could enfore compiler rules would be to generate public
+`get`-only properties and make the backing field public and mutable. Object
+initializers would be able to set the properties, but user code wouldn't be
+able to because the backing properties are unspeakable.
+
+One problem with this strategy is `readonly` fields. In that case there is
+no backing field to hide. There are two possible solutions. The first is
+to forbid public `readonly` fields and require properties. The second is
+to make all fields into properties automatically. This is strange because
+we would be generating a property from a field syntax. However, it removes
+what will be a meaningless restriction for the user, only mandated by
+implementation difficulties. Property substition will never be a perfect
+abstraction (reflection will be able to see properties, for example) but the
+solution would probably be able to match user expectations for the vast
+majority of cases. The properties would also be `ref-readonly` returning, so
+even uses of `in` or `ref-readonly` would function as expected.
+
+If data classes contain any `readonly` members that do not have initializers,
+they also do not define a default public constructor like other classes.
+Instead, they define a protected constructor with no arguments, and an
+unspeakable public "initialization" method. This method is called when using
+an object initializer and the compiler verifies that all `readonly` members
+are initialized, or an error is produced.
+
+
+## Extensible data classes (data class subtyping)
+
+Like normal C# classes, data classes are not sealed by default and can be
+inherited from in base classes. If there are any readonly members without
+default initialization in the base class, the subclass is required to
+define a protected constructor. The constructor must assign all readonly
+members of the base class before the constructor ends, or an error is
+produced.
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -308,7 +308,7 @@ the method `Equals` is tried according to overload resolution rules (st. an
 `Equals` method with an identity conversion to the target type is preferred
 over the virtual `Equals(object)` method).
 
-There is also one hidden data member, `EqualityContractOrigin`, that is
+There is also one hidden data member, `protected virtual Type EqualityContractOrigin { get; }`, that is
 always considered in equality. By default this member always returns the
 static type of its containing type, i.e. `typeof(Containing)`. This means
 that sub-classes are not, by default, considered equal to their base classes,
@@ -318,8 +318,7 @@ overridden, but then it is the user's responsibility to ensure that they
 abide by the appropriate contract.
 
 `GetHashCode` would be implemented by calling `GetHashCode` on each of
-the data members and these would be combined using a symmetric operation
-(like `+`) to prevent ordering from affecting the results.
+the data members.
 
 ### Readonly initialization
 
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -57,7 +57,7 @@ all have fundamental problems.
 can read it, and you can compare it easily (equality is automatically
 defined). You can't modify it, though -- anonymous types are always immutable
 and you can't easily create a copy with only one change. The real problem
-is composibility. You can't use it as a real type anywhere outside the
+is composability. You can't use it as a real type anywhere outside the
 current method and you can't nest it in other data structures, except as
 an object.
 
@@ -82,7 +82,7 @@ public class LoginResource
 ```
 
 This feature provides names, for both the members and the data structure, it
-provides easy nominal composition, and is easily composible with all other
+provides easy nominal composition, and is easily composable with all other
 data structures. It also provides a convenient syntax for creation by
 interacting directly with the named data, e.g.
 
@@ -209,7 +209,7 @@ public class LoginResource : IEquatable<LoginResource>
 
     protected LoginResource() { }
 
-    public static LoginResource Init() => new LoginResource();
+    public static LoginResource <>Init() => new LoginResource();
 
     public override bool Equals(object obj)
         => obj is LoginResource resource && Equals(resource);
@@ -250,6 +250,14 @@ public class LoginResource : IEquatable<LoginResource>
         return !(resource1 == resource2);
     }
 }
+```
+
+The above usage would translate into:
+
+```C#
+var x = LoginResource.<>Init();
+x.<>Backing_Username = "andy";
+x.<>Backing_Password = password;
 ```
 
 ### Equality
@@ -335,7 +343,7 @@ be irrelevant to `readonly` semantics.
 
 One way to remove dependence on a constructor is simply not make the members
 `readonly` in metadata. The CLR treats `readonly` mostly as guidance -- it
-can easily be overriden using reflection anyway. Most of the safety of
+can easily be overridden using reflection anyway. Most of the safety of
 `readonly` members in C# is not provided by the runtime, but by C# safety
 rules. One way we could enforce compiler rules would be to generate public
 `get`-only properties and make the backing field public and mutable. Object
@@ -348,7 +356,7 @@ to forbid public `readonly` fields and require properties. The second is
 to make all fields into properties automatically. This is strange because
 we would be generating a property from a field syntax. However, it removes
 what will be a meaningless restriction for the user, only mandated by
-implementation difficulties. Property substition will never be a perfect
+implementation difficulties. Property substitution will never be a perfect
 abstraction (reflection will be able to see properties, for example) but the
 solution would probably be able to match user expectations for the vast
 majority of cases. The properties would also be `ref readonly` returning, so
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -162,15 +162,14 @@ fields with only one item changed. A new type must be constructed manually.
 ## Proposal
 
 To resolve many of these issues, I propose a new modifier for classes and structs: `data`.
-`data` classes or structs are meant to satisfy the goals listed above by doing the
-following things:
+Since this proposal focuses on unorderd, *named* data, I'll refer to this feature as 
+*nominal records*, in contrast to the existing [records proposal](https://github.com/dotnet/csharplang/blob/master/proposals/records.md), which I'll
+refer to as *positional records*. Nominal records are meant to satisfy the
+goals listed above by doing the following things:
 
 1. Automatically generating `Equals`, `GetHashCode`, `ToString`, `==`, `!=`, and `IEquatable<T>`
    based on the member data of the type.
 1. Allow object initializers to also initialize readonly members.
-
-Data classes or structs represent unordered, *named* data, like the simple
-`LoginResource` class that people write today.
 
 The LoginResource class now could be defined as
 
@@ -195,21 +194,66 @@ var x = new LoginResource {
 Note that `RememberMe` must have an initializer to avoid a warning in the
 object initializer about an unset read-only property.
 
-However, the generated class code would look like:
+**Why have nominal records?**
+
+Nominal records provide two major use cases that are not addressed by positional
+records. The first is adoption convenience. The most common use of a
+record-like type in C# today is a simple class with an object initializer.
+Transforming this into a positional record would require replacing all fields
+with parameters on the primary constructor, replacing all uses of object
+initializers with constructor calls, and explicitly providing default values
+for all parameters which are optional (where this may have been implicitly
+supplied before by simply not putting an initializer on a field or property).
+
+The second, more important use case is API evolution. Positional records make
+it difficult to add or re-order members without producing a breaking change.
+Reordering is essentially impossible for positional records without
+introducing a binary breaking change. Adding new members is difficult and
+requires manual intervention to create constructor overloads with the old
+members which must be carefully maintained. For many uses of records, like
+"options containers" (see
+[Roslyn](http://source.roslyn.io/#Microsoft.CodeAnalysis/Compilation/CompilationOptions.cs,29a3818343614151),
+[Benchmark.NET](https://github.com/dotnet/BenchmarkDotNet/blob/master/src/BenchmarkDotNet/ConsoleArguments/CommandLineOptions.cs)
+for examples), this is a headache. Often data should not have positional
+dependence, and retrofitting independence later is quite difficult.
+
+### Code generation
+
+Aside from places where semantics are underspecified, this section is mostly
+for edification. The implementation of the feature is irrelevant to the
+language design itself. The implementation may change significantly as it is
+actually developed.
+
+That being said, the current proposal is that the generated class code would
+look like:
 
 ```C#
 public class LoginResource : IEquatable<LoginResource>
 {
-    public string <>Backing_Username;
-    public string Username => <>Backing_Username;
-    public string <>Backing_Password;
-    public string Password => <>Backing_Password;
-    public string <>Backing_RememberMe = false;
-    public bool RememberMe => <>Backing_RememberMe;
+    public struct <>Initializer
+    {
+        public string Username;
+        public string Password;
+        public bool RememberMe;
 
-    protected LoginResource() { }
+        public static <>Initializer Create()
+        {
+            var init = new <>Initializer();
+            init.RememberMe = false;
+            return init;
+        }
+    }
 
-    public static LoginResource <>Init() => new LoginResource();
+    private readonly <>Initializer _init;
+
+    public string Username => _init.Username;
+    public string Password => _init.Password;
+    public bool RememberMe => _init.RememberMe;
+
+    public LoginResource(<>Initializer init)
+    {
+        _init = init;
+    }
 
     public override bool Equals(object obj)
         => obj is LoginResource resource && Equals(resource);
@@ -252,16 +296,19 @@ public class LoginResource : IEquatable<LoginResource>
 }
 ```
 
-The above usage would translate into:
+If any of the generated members listed here conflict with user-written
+members, the user written members take precedence and those generated members
+are skipped. However, the default constructor allowing object initializer usage
+will always be created.
+
+The above usage of the positional record would translate into:
 
 ```C#
-var x = LoginResource.<>Init();
-x.<>Backing_Username = "andy";
-x.<>Backing_Password = password;
+var init = LoginResource.<>Initializer().Create();
+init.Username = "andy";
+init.Password = password;
+var x = new LoginResource(init);
 ```
-
-If any of the generated members conflict with user-written members, the user written
-members take precedence and those generated members are skipped.
 
 ### Equality
 
@@ -333,16 +380,11 @@ the data members.
 
 ### Readonly initialization
 
-Support for `readonly` members in object initializers may seem like a small
-feature, but it's important that making a `data` type readonly not come with
-a lot of extra ceremony. The essence of a `data` type is a set of named fields
-and that should stay true regardless of whether or not the fields are 
-`readonly`. It may be tempting for implementation simplicity to try to use
-constructors instead, but this is a design smell that conflates positional
-semantics with `readonly` semantics. Requiring initialization via constructor
-means that field order becomes a public API and requires careful versioning,
-which is not true of mutable `data` types and is a constraint that should
-be irrelevant to `readonly` semantics.
+There are two reasonable implementations: making backing fields public,
+mutable, but unspeakable, or using a struct initializer type in a generated
+constructor.
+
+** Mutable unspeakable backing fields **
 
 One way to remove dependence on a constructor is simply not make the members
 `readonly` in metadata. The CLR treats `readonly` mostly as guidance -- it
@@ -372,6 +414,27 @@ unspeakable public "initialization" method. This method is called when using
 an object initializer and the compiler verifies that all `readonly` members
 are initialized, or an error is produced.
 
+** Struct initializer type **
+
+Another design, the one currently preferred, allows `readonly` members to be
+initialized via the unspeakable `<>Initializer` struct. Making the struct
+unspeakable allows compilers to enforce rules about readonly initialization.
+Default values for read-only members would be lowered into the static
+`Create` function on the initializer, since past JIT bugs have sometimes
+caused no-argument struct constructors to be skipped. This also affects all
+writes to the read-only members in user-written constructorsOnly read-only
+members would be present in the <>Initializer struct -- mutable members would
+be compiled directly onto the class, as they are today. Support for
+`readonly` members in object initializers may seem like a small feature, but
+it's important that making a `data` type readonly not come with a lot of
+extra ceremony. The essence of a `data` type is a set of named fields and
+that should stay true regardless of whether or not the fields are `readonly`.
+It may be tempting for implementation simplicity to try to use constructors
+instead, but this is a design smell that conflates positional semantics with
+`readonly` semantics. Requiring initialization via constructor means that
+field order becomes a public API and requires careful versioning, which is
+not true of mutable `data` types and is a constraint that should be
+irrelevant to `readonly` semantics.
 
 ## Extensible data classes (data class subtyping)
 
@@ -380,7 +443,7 @@ inherited from in sub-classes.
 
 In non-`data` sub-classes, if there are any readonly members without
 default initialization in the base class, the subclass is required to
-define a protected constructor. The constructor must assign all readonly
+define at least one constructor. The constructor must assign all readonly
 members of the base class before the constructor ends, or an error is
 produced.
 
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -260,6 +260,9 @@ x.<>Backing_Username = "andy";
 x.<>Backing_Password = password;
 ```
 
+If any of the generated members conflict with user-written members, the user written
+members take precedence and those generated members are skipped.
+
 ### Equality
 
 First, the generation of equality support. Data members are only public
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -77,7 +77,7 @@ public class LoginResource
 {
     public string Username { get; set; }
     public string Password { get; set; }
-    public bool RememberMe { get; set; } = false;
+    public bool RememberMe { get; set; }
 }
 ```
 
@@ -95,8 +95,53 @@ var x = new LoginResource {
 
 Unfortunately, there are still serious problems. There is no piecewise
 comparer implicitly defined for C# classes, so if you want simple data
-comparison, the real example looks like
-[this](https://gist.github.com/agocke/6ba6e64f77f1212ba7292bfd1f1000e5)
+comparison, the real example looks like this:
+
+```C#
+using System;
+
+public class LoginResource : IEquatable<LoginResource>
+{
+    public string Username { get; set; }
+    public string Password { get; set; }
+    public bool RememberMe { get; set; } = false;
+
+    public override bool Equals(object obj)
+        => obj is LoginResource resource && Equals(resource);
+
+    public bool Equals(LoginResource other)
+    {
+        return other != null &&
+               Username == other.Username &&
+               Password == other.Password &&
+               RememberMe == other.RememberMe;
+    }
+
+    public override int GetHashCode()
+    {
+        var hashCode = -736459255;
+        hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Username);
+        hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Password);
+        hashCode = hashCode * -1521134295 + RememberMe.GetHashCode();
+        return hashCode;
+    }
+
+    public override string ToString()
+    {
+        return $"{{{nameof(Username)} = {Username}, {nameof(Password)} = {Password}, {nameof(RememberMe)} = {RememberMe}}}";
+    }
+
+    public static bool operator ==(LoginResource resource1, LoginResource resource2)
+    {
+        return EqualityComparer<LoginResource>.Default.Equals(resource1, resource2);
+    }
+
+    public static bool operator !=(LoginResource resource1, LoginResource resource2)
+    {
+        return !(resource1 == resource2);
+    }
+}
+```
 
 Immutable data is also a problem. The object initializer syntax provides
 a simple name-based mechanism to create a data type. With `readonly`
@@ -107,14 +152,16 @@ another set of problems:
 1. The constructor parameters are ordered, while the fields are not.
    Consumers have now taken a dependency on the parameter ordering.
 1. The constructor must be maintained with any field changes.
+1. Constructor parameter names don't necessarily line up with field/property
+names hence named argument passing doesn't have the same ease of use as
+object initializers.
 
 There is also no way to create a copy of a data structure with readonly
 fields with only one item changed. A new type must be constructed manually.
 
-
 ## Proposal
 
-To resolve many of these issues, I propse a new modifier for classes and structs: `data`.
+To resolve many of these issues, I propose a new modifier for classes and structs: `data`.
 `data` classes or structs are meant to satisfy the goals listed above by doing the
 following things:
 
@@ -122,15 +169,88 @@ following things:
    based on the member data of the type.
 1. Allow object initializers to also initialize readonly members.
 
+Data classes or structs represent unordered, *named* data, like the simple
+`LoginResource` class that people write today.
+
+The LoginResource class would now be defined as
+
+```C#
+public data class LoginResource
+{
+    public string Username { get; set; }
+    public string Password { get; set; }
+    public bool RememberMe { get; set; }
+}
+```
+
+and the use would be identical:
+
+```C#
+var x = new LoginResource {
+    Username = "andy",
+    Password = password
+}
+```
+
+However, the generated class code would look like:
+
+```C#
+public class LoginResource : IEquatable<LoginResource>
+{
+    public string Username { get; set; }
+    public string Password { get; set; }
+    public bool RememberMe { get; set; }
+
+    public override bool Equals(object obj)
+        => obj is LoginResource resource && Equals(resource);
+
+    public bool Equals(LoginResource other)
+    {
+        return other != null &&
+               EqualityContractOrigin == other.EqualityContractOrigin &&
+               Username == other.Username &&
+               Password == other.Password &&
+               RememberMe == other.RememberMe;
+    }
+
+    protected virtual Type EqualityContractOrigin => typeof(LoginResource);
+
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            return EqualityComparer<string>.Default.GetHashCode(Username) +
+                EqualityComparer<string>.Default.GetHashCode(Password) +
+                RememberMe.GetHashCode();
+        }
+    }
+
+    public override string ToString()
+    {
+        return $"{{{nameof(Username)} = {Username}, {nameof(Password)} = {Password}, {nameof(RememberMe)} = {RememberMe}}}";
+    }
+
+    public static bool operator ==(LoginResource resource1, LoginResource resource2)
+    {
+        return EqualityComparer<LoginResource>.Default.Equals(resource1, resource2);
+    }
+
+    public static bool operator !=(LoginResource resource1, LoginResource resource2)
+    {
+        return !(resource1 == resource2);
+    }
+}
+```
+
 ### Equality
 
 First, the generation of equality support. Data members are only public
 fields and auto-properties. This allows data classes to have private
-implementation details for things like caching without giving up simple
-equality semantics. There are a few places this could be problematic. For
-instance, only auto-properties are considered data members by default, but
-it's not uncommon to have some simple validation included in a property
-getter that does not meaningfully change the semantics, e.g.
+implementation details without giving up simple equality semantics. There are
+a few places this could be problematic. For instance, only auto-properties
+are considered data members by default, but it's not uncommon to have some
+simple validation included in a property getter that does not meaningfully
+change the semantics, e.g.
 
 ```C#
 {
@@ -143,6 +263,7 @@ getter that does not meaningfully change the semantics, e.g.
             Debug.Assert(_field >= 0);
             return _field;
         }
+        set { ... }
     }
 }
 ```
@@ -177,17 +298,18 @@ the method `Equals` is tried according to overload resolution rules (st. an
 `Equals` method with an identity conversion to the target type is preferred
 over the virtual `Equals(object)` method).
 
-Note that the equality defined in this manner is not symmetric in the case
-of record subtyping. This topic will be addressed in more detail later, but
-it's important to note for equality that `a.Equals(b)` may be true if `B`
-is a subtype of `A` (and therefore has an implicit conversion) and all of
-the shared members between `A` and `B` are equal. In contrast, `A` would
-likely not have an implicit conversion to `B`, so they would not be
-considered equal.
+There is also one hidden data member, `EqualityContractOrigin`, that is
+always considered in equality. By default this member always returns the
+static type of its containing type, i.e. `typeof(Containing)`. This means
+that sub-classes are not, by default, considered equal to their base classes,
+or vice versa. This also ensures that equality is commutative and `GetHashCode`
+matches the results of `Equals`. These methods are virtual, so they can be
+overridden, but then it is the user's responsibility to ensure that they
+abide by the appropriate contract.
 
 `GetHashCode` would be implemented by calling `GetHashCode` on each of
 the data members and these would be combined using a symmetric operation
-(like `xor`) to prevent ordering from affecting the results.
+(like `+`) to prevent ordering from affecting the results.
 
 ### Readonly initialization
 
@@ -206,7 +328,7 @@ One way to remove dependence on a constructor is simply not make the members
 `readonly` in metadata. The CLR treats `readonly` mostly as guidance -- it
 can easily be overriden using reflection anyway. Most of the safety of
 `readonly` members in C# is not provided by the runtime, but by C# safety
-rules. One way we could enfore compiler rules would be to generate public
+rules. One way we could enforce compiler rules would be to generate public
 `get`-only properties and make the backing field public and mutable. Object
 initializers would be able to set the properties, but user code wouldn't be
 able to because the backing properties are unspeakable.
@@ -234,8 +356,17 @@ are initialized, or an error is produced.
 ## Extensible data classes (data class subtyping)
 
 Like normal C# classes, data classes are not sealed by default and can be
-inherited from in base classes. If there are any readonly members without
+inherited from in sub-classes. 
+
+In non-`data` sub-classes, if there are any readonly members without
 default initialization in the base class, the subclass is required to
 define a protected constructor. The constructor must assign all readonly
 members of the base class before the constructor ends, or an error is
 produced.
+
+In `data` sub-classes, the requirements of the base become requirements
+of the sub-class, such that initialization of the sub-class must also
+initialize all of the required members of the base.
+
+
+## TODO: "With"-ers
--- a/proposals/data-classes.md
+++ b/proposals/data-classes.md
@@ -172,14 +172,14 @@ following things:
 Data classes or structs represent unordered, *named* data, like the simple
 `LoginResource` class that people write today.
 
-The LoginResource class would now be defined as
+The LoginResource class now could be defined as
 
 ```C#
 public data class LoginResource
 {
-    public string Username { get; set; }
-    public string Password { get; set; }
-    public bool RememberMe { get; set; }
+    public string Username { get; }
+    public string Password { get; }
+    public bool RememberMe { get; } = false;
 }
 ```
 
@@ -189,17 +189,27 @@ and the use would be identical:
 var x = new LoginResource {
     Username = "andy",
     Password = password
-}
+};
 ```
+
+Note that `RememberMe` must have an initializer to avoid a warning in the
+object initializer about an unset read-only property.
 
 However, the generated class code would look like:
 
 ```C#
 public class LoginResource : IEquatable<LoginResource>
 {
-    public string Username { get; set; }
-    public string Password { get; set; }
-    public bool RememberMe { get; set; }
+    public string <>Backing_Username;
+    public string Username => <>Backing_Username;
+    public string <>Backing_Password;
+    public string Password => <>Backing_Password;
+    public string <>Backing_RememberMe = false;
+    public bool RememberMe => <>Backing_RememberMe;
+
+    protected LoginResource() { }
+
+    public static LoginResource Init() => new LoginResource();
 
     public override bool Equals(object obj)
         => obj is LoginResource resource && Equals(resource);
@@ -342,8 +352,8 @@ what will be a meaningless restriction for the user, only mandated by
 implementation difficulties. Property substition will never be a perfect
 abstraction (reflection will be able to see properties, for example) but the
 solution would probably be able to match user expectations for the vast
-majority of cases. The properties would also be `ref-readonly` returning, so
-even uses of `in` or `ref-readonly` would function as expected.
+majority of cases. The properties would also be `ref readonly` returning, so
+even uses of `in` or `ref readonly` would function as expected.
 
 If data classes contain any `readonly` members that do not have initializers,
 they also do not define a default public constructor like other classes.


agocke:
cc @jaredpar

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+1. Automatically generating `Equals`, `GetHashCode`, `ToString`, `==`, `!=`, and `IEquatable<T>`
+   based on the member data of the type.
+1. Allow object initializers to also initialize readonly members.
+
+### Equality
+
+First, the generation of equality support. Data members are only public
+fields and auto-properties. This allows data classes to have private
+implementation details for things like caching without giving up simple
+equality semantics. There are a few places this could be problematic. For
```
CyrusNajmabadi: I'm not saying this is wrong.  But this def feels like it needs to be expanded upon, esp. give the above sections on what data is/isn't.  Talking about impl-details(i.e computation) and caching, after somewhat saying that data is not these things isn't wrong, but it's very subtle and muddies things a bit.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+of record subtyping. This topic will be addressed in more detail later, but
+it's important to note for equality that `a.Equals(b)` may be true if `B`
+is a subtype of `A` (and therefore has an implicit conversion) and all of
+the shared members between `A` and `B` are equal. In contrast, `A` would
+likely not have an implicit conversion to `B`, so they would not be
+considered equal.
+
+`GetHashCode` would be implemented by calling `GetHashCode` on each of
+the data members and these would be combined using a symmetric operation
+(like `xor`) to prevent ordering from affecting the results.
```
CyrusNajmabadi: is it expected that 'data' is unordered?  Above, you mentioned:

```Data *is* a collection of values with potentially heterogenous types. ```

Should that be "an unordered collection of values"?

**Comment by HaloFour** [comment]
When will an Issue be created for this proposal?  I imagine that some of these points will be quite contentious.  Just briefly scanning the "readonly initialization" section is enough to give me nightmares.

**Comment by gulshan** [comment]
How this proposal relates to Records #39 ? As I [commented there](https://github.com/dotnet/csharplang/issues/39#issuecomment-281687278), I am for just a primary constructor based short hand definition capability for all/any types, and then the equality, Hash, `with` and deconstructor generation on top of that when associated with some keyword like `data` (or attribute like `[Data]`). The same `data` keyword being proposed here with equality and hash generation. Shouldn't these two proposals be consolidated?

I think proposed readonly initialization changes current meaning of initializers in C#. With non-nullable references are coming, initialization of individual reference type fields will be a concern. I personally think, default values for primary constructor parameters are enough to deal with this. It also optionally provides names for fields and ignore the order of the arguments to some extent.

File: proposals/data-classes.md (hunk 2 above)
Comment (svick) [inline proposals/data-classes.md]: I think the primary result of this would be a bad hash code.

Trying to pretend that two different data classes are the same if they have the same members sounds like a bad idea to me.

File: proposals/data-classes.md [hunk 3]
```diff
@@ -0,0 +1,241 @@
...
+
+
+## Extensible data classes (data class subtyping)
+
+Like normal C# classes, data classes are not sealed by default and can be
+inherited from in base classes. If there are any readonly members without
+default initialization in the base class, the subclass is required to
+define a protected constructor. The constructor must assign all readonly
+members of the base class before the constructor ends, or an error is
+produced.
```
svick: Do I understand it correctly that because of this, it's not possible to make something like the following code work?

```c#
data class Person
{
    public string Name { get; }
}

data class Student : Person
{
    public int Grade { get; }
}

…

new Student { Name = "Jane Doe", Grade = 1 }
```

Or do these rules only apply to inheritors that are normal classes, not data classes?

gafter:
@agocke we figured out how to make equals be symmetric in the case of inheritance. I’ll send you notes tomorrow.

agocke:
@HaloFour Several members of the LDM requested a PR, since that's essentially an issue in GitHub, but it has better change tracking and you can leave comments on individual sections.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+`Equals` method with an identity conversion to the target type is preferred
+over the virtual `Equals(object)` method).
+
+Note that the equality defined in this manner is not symmetric in the case
+of record subtyping. This topic will be addressed in more detail later, but
+it's important to note for equality that `a.Equals(b)` may be true if `B`
+is a subtype of `A` (and therefore has an implicit conversion) and all of
+the shared members between `A` and `B` are equal. In contrast, `A` would
+likely not have an implicit conversion to `B`, so they would not be
+considered equal.
```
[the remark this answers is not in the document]
agocke: This is a really good point. The problem I ran into was: how do I prohibit it?

Making `.Equals` do a dynamic type check seems problematic. So does calling `GetHashCode` inside `Equals`.

However, Neal has sent me a great idea for producing a symmetric `Equals`:

```C#
class Person
{
    public string Name { get; }
 
    public Person(string name) { Name = name; }
 
    protected virtual Type EqualityContractOrigin => typeof(Person); // Note here
 
    public override bool Equals(object obj)
    {
        var that = obj as Person;
        return that != null &&
            that.EqualityContractOrigin == this.EqualityContractOrigin && // Note here
            that.Name.Equals(this.Name);
    }
}
```

I really like this solution.

proposals/data-classes.md
agocke: This sounds like you're restating the previous comment. If `Equals` is symmetric, do you have any other objection?

proposals/data-classes.md
agocke: This should only apply to normal classes. Data classes should inherit the requirements of their base data classes.

agocke:
@gulshan @CyrusNajmabadi If you're looking for an ultra-short syntax for data members, you may be interested in my forthcoming companion proposal for expanding tuples (I call them "named tuples").

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+
+**Classes/Structs**. This is by far the most common representation of data
+in C#. A canonical example looks something like this:
+
+```C#
+public class LoginResource
+{
+    public string Username { get; set; }
+    public string Password { get; set; }
+    public bool RememberMe { get; set; } = false;
```
jaredpar: Why have `= false` here? It's the default and having it put there makes me think it's relevant to the feature.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+var x = new LoginResource {
+    Username = "andy",
+    Password = password
+}
+```
+
+Unfortunately, there are still serious problems. There is no piecewise
+comparer implicitly defined for C# classes, so if you want simple data
+comparison, the real example looks like
+[this](https://gist.github.com/agocke/6ba6e64f77f1212ba7292bfd1f1000e5)
```
jaredpar: Rather than link to a GIST I would just inline the sample here. Yes it's a big sample but that's part of the point.

proposals/data-classes.md
agocke: I'll elaborate more on this. The idea is that the public contract should be seen as a collection of data. As always, there's some room to hide your dirty laundry in private implementation details if the real world is more complicated than that.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+
+Immutable data is also a problem. The object initializer syntax provides
+a simple name-based mechanism to create a data type. With `readonly`
+fields or properties, a constructor must be used instead. This creates
+another set of problems:
+
+1. A constructor must be manually defined.
+1. The constructor parameters are ordered, while the fields are not.
+   Consumers have now taken a dependency on the parameter ordering.
+1. The constructor must be maintained with any field changes.
```
jaredpar: Another issue: constructor parameter names don't necessarily line up with field / property names hence named argument passing doesn't have the same ease of use as object initializers.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+
+There is also no way to create a copy of a data structure with readonly
+fields with only one item changed. A new type must be constructed manually.
+
```
jaredpar: double blank line.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Proposal
+
+To resolve many of these issues, I propse a new modifier for classes and structs: `data`.
+`data` classes or structs are meant to satisfy the goals listed above by doing the
```
jaredpar: Can we have a sample of a `data` class usage in the proposal?

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+following things:
+
+1. Automatically generating `Equals`, `GetHashCode`, `ToString`, `==`, `!=`, and `IEquatable<T>`
```
jaredpar: How is `ToString` relevant here?

proposals/data-classes.md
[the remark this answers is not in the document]
svick: @agocke 

> Making `.Equals` do a dynamic type check seems problematic.

I think that's actually the right solution. Different types should not compare as equal, even if e.g. a derived type does not override methods it should override.

Why would it be problematic?

If it's about performance, I think code like `that.GetType() == this.GetType()` has the potential to be more efficient than two virtual calls (though the CLR could do a better job at optimizing it).

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+   Consumers have now taken a dependency on the parameter ordering.
+1. The constructor must be maintained with any field changes.
+
+There is also no way to create a copy of a data structure with readonly
+fields with only one item changed. A new type must be constructed manually.
+
+
+## Proposal
+
+To resolve many of these issues, I propse a new modifier for classes and structs: `data`.
```
jnm2: Typo `propse`

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+semantics with `readonly` semantics. Requiring initialization via constructor
+means that field order becomes a public API and requires careful versioning,
+which is not true of mutable `data` types and is a constraint that should
+be irrelevant to `readonly` semantics.
+
+One way to remove dependence on a constructor is simply not make the members
+`readonly` in metadata. The CLR treats `readonly` mostly as guidance -- it
+can easily be overriden using reflection anyway. Most of the safety of
+`readonly` members in C# is not provided by the runtime, but by C# safety
+rules. One way we could enfore compiler rules would be to generate public
```
jnm2: Typo `enfore`

proposals/data-classes.md
svick: @agocke No, this is a different issue. The problem with using a symmetric operation for combining hash codes is that it means that e.g. `new Point { X = 1, Y = 0 }.GetHashCode()` returns the same value as `new Point { X = 0, Y = 1 }.GetHashCode()`.

I believe that in some data sets, this could increase the number of collisions by a lot.

The comment about pretending that types are the same is because I assumed the reason for symmetry is that you wanted e.g. `new Contoso.Person { FirstName = "Andy", LastName = "Gocke" }.GetHashCode()` to be the same as `new AdventureWorks.Person { LastName = "Gocke", FirstName = "Andy" }.GetHashCode()`.

Sorry I was unclear.

jnm2:
Concerned about the public unspeakable members, but I'm thrilled about this proposal in general. Can hardly wait for this!

- > There is also no way to create a copy of a data structure with readonly fields with only one item changed. A new type must be constructed manually.
  
  I'm particularly interested in this since today I'm maintaining `With*` methods by hand. What would the   syntax be for creating a copy with only one item changed?
 If I wanted to suppress a certain `With*` method or add a custom `With*` method to require two fields to be set at once, perhaps with very simple validation, would I have to swap the entire class over to a normal class first?

- Would a manual override of ToString on a data type ever be considered? I write data types and override ToString for UI data binding to control the display of the object, choosing one or two of the fields to use in the display string.

- Would `readonly data struct` ever be considered? I do love that `readonly` guarantee. (`readonly class` would be sublime.)

proposals/data-classes.md
agocke: Ah, in that case the intention is: yes, fields are deliberately unordered. However, there is no structural equality -- the types must be nominally compatible for two types to be considered equal. It sounds like we're also centering around subtypes not being equal to supertypes, so that would imply the types are equal as well.

svick: @agocke In that case, I think the order of fields should matter specifically for the purpose of computing the hash code. Hash codes should not be preserved across executions, so I think it's fine if changing the order of fields changes the hash code.

(Even stronger example is .Net Core 2.1 `HashCode`. It uses randomization, which means hash codes produced by it change even if the types themselves don't. If that behavior is acceptable, then I think depending on the order of fields should be also acceptable, even if they are otherwise unordered.)

HaloFour:
@agocke 

> Several members of the LDM requested a PR, since that's essentially an issue in GitHub, but it has better change tracking and you can leave comments on individual sections.

I agree that PRs make it easier to discuss specific portions of a proposal, but it doesn't feel appropriate to have this one discussion segregated from every other discussion.  Once this proposal is merged the conversation is effectively lost since a user would have to know to look in a completely different place to find it.  The only thing worse than the shortcomings of the Issues is scattering the discussions all over the place.

agocke:
@dotnet/csharplangdesign See @HaloFour's comment above about proposal formatting.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+Support for `readonly` members in object initializers may seem like a small
+feature, but it's important that making a `data` type readonly not come with
+a lot of extra ceremony. The essence of a `data` type is a set of named fields
+and that should stay true regardless of whether or not the fields are 
+`readonly`. It may be tempting for implementation simplicity to try to use
+constructors instead, but this is a design smell that conflates positional
+semantics with `readonly` semantics. Requiring initialization via constructor
+means that field order becomes a public API and requires careful versioning,
+which is not true of mutable `data` types and is a constraint that should
+be irrelevant to `readonly` semantics.
```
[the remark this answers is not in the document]
HaloFour: The most unpalatable aspect of this would be the compiler surfacing public mutable fields with autogenerated unspeakable names with only a wink and a nudge to enforce their immutability.  There must be a significantly better way to accomplish this behavior without requiring private implementation details to be exposed which then every other compiler either has to ignore or has to respect.

What about creating a nested public `struct` to contain all of the data members and having the constructor or a static factory method accepting that struct?

```cs
public data class LoginResource {
    public string Username { get; }
    public string Password { get; }
}
var loginResource = new LoginResource {
    Username = "andy",
    Password = password
};

// translates into

public class LoginResource {
    private readonly DataMembers _data;

    public LoginResource(DataMembers data) => _data = data;
    
    public string Username { get => _data.Username; }
    public string Password { get => _data.Password; }

    public struct DataMembers {
        public string Username { get; set; }
        public string Password { get; set; }
    }
}
var loginResource = new LoginResource(new LoginResource.DataMembers {
    Username = "andy", Password = password
});
```

gulshan:
So, what about the Records #39 ? Abandoned?

agocke:
@gulshan You can see this as part of a broader "records" proposal. If the thing you were looking for in records was the syntax, that will be part of a follow-up proposal I'm calling "named tuples".

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+               RememberMe == other.RememberMe;
+    }
+
+    public override int GetHashCode()
+    {
+        var hashCode = -736459255;
+        hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Username);
+        hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Password);
+        hashCode = hashCode * -1521134295 + RememberMe.GetHashCode();
+        return hashCode;
```
CyrusNajmabadi: note: this is simpler now with HashCode.Combine.  (just to make the example code fair :)).

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+fields or properties, a constructor must be used instead. This creates
+another set of problems:
+
+1. A constructor must be manually defined.
+1. The constructor parameters are ordered, while the fields are not.
+   Consumers have now taken a dependency on the parameter ordering.
+1. The constructor must be maintained with any field changes.
+1. Constructor parameter names don't necessarily line up with field/property
+names hence named argument passing doesn't have the same ease of use as
+object initializers.
```
CyrusNajmabadi: Note: tehse are definitely cons of a constructor. But it might be nice to mention the pros as well.  Namely terse and sensible syntax for 'data' that has well-understood positional ordering.

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+
+## Proposal
+
+To resolve many of these issues, I propose a new modifier for classes and structs: `data`.
+`data` classes or structs are meant to satisfy the goals listed above by doing the
+following things:
+
+1. Automatically generating `Equals`, `GetHashCode`, `ToString`, `==`, `!=`, and `IEquatable<T>`
+   based on the member data of the type.
+1. Allow object initializers to also initialize readonly members.
```
CyrusNajmabadi: not sure if it's mentioned later.  But is it possible to override any of this?  For example, if you want to provide a better GethashCode impl?  Or if you want to provide your own specific ToString?

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+`LoginResource` class that people write today.
+
+The LoginResource class would now be defined as
+
+```C#
+public data class LoginResource
+{
+    public string Username { get; set; }
+    public string Password { get; set; }
+    public bool RememberMe { get; set; }
```
CyrusNajmabadi: very nervous about these being mutable.   can i make immutable if i want?

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+
+    protected virtual Type EqualityContractOrigin => typeof(LoginResource);
+
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            return EqualityComparer<string>.Default.GetHashCode(Username) +
+                EqualityComparer<string>.Default.GetHashCode(Password) +
+                RememberMe.GetHashCode();
```
CyrusNajmabadi: since you can only compare with other LoginResource's... is there a reason that hashing is unordered?

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+        }
+        set { ... }
+    }
+}
+```
+
+To support these cases and provide an easy escape hatch, I propose a
+new attribute, `DataMemberAttribute` with a boolean flag argument on the
+constructor. This allows users to override the normal behavior and include
+or exclude extra members in equality. The previous example would now read:
```
CyrusNajmabadi: this seems odd to me.  Given that it's a Data-Class or Data-Struct, my bias is that the public props/fields should be part of the data-ness.  It seems like if someone does not want that, the attribute should be to *opt-out*, instead of needing to opt-in.

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+}
+```
+
+Equality itself would be defined in terms of its data members. A `data` type
+is equal to another `data` type when there is an implicit conversion between
+the target type and the source type and each of the corresponding members
+are equal. The members are compared by `==` if it is available. Otherwise,
+the method `Equals` is tried according to overload resolution rules (st. an
+`Equals` method with an identity conversion to the target type is preferred
+over the virtual `Equals(object)` method).
```
CyrusNajmabadi: is this better than saying that you use ```EqualityComparer<X>.Default.Equals(a, b)```?  That shoudl do "the right thing" for basically all types, right?  Of course, the compiler could consider optimizing that for some things like primitive types.   but it seems odd to bake in the knowledge about == vs .Equals vs IEquatable.Equals.

proposals/data-classes.md
```diff
@@ -0,0 +1,372 @@
...
+static type of its containing type, i.e. `typeof(Containing)`. This means
+that sub-classes are not, by default, considered equal to their base classes,
+or vice versa. This also ensures that equality is commutative and `GetHashCode`
+matches the results of `Equals`. These methods are virtual, so they can be
+overridden, but then it is the user's responsibility to ensure that they
+abide by the appropriate contract.
+
+`GetHashCode` would be implemented by calling `GetHashCode` on each of
+the data members and these would be combined using a symmetric operation
+(like `+`) to prevent ordering from affecting the results.
```
CyrusNajmabadi: not clear ot me why/how ordering could/would matter.  

Since you can't compare against a different type... your orders will be the same.

agocke: You could certainly save the hash code and compare before/after reordering members.

proposals/data-classes.md
agocke: That's a great question that I don't know the answer to.

proposals/data-classes.md
agocke: *auto*-properties are part of data-ness, computed properties may not be. That's especially true if you have, say, a half dozen computed properties all reading from single piece of data that you could compare directly for equality. For instance, flags enums to boolean properties.

proposals/data-classes.md
agocke: Yup.

proposals/data-classes.md
agocke: Yup. We may want to give a warning/error if you override everything and `data` just means nothing.

proposals/data-classes.md
CyrusNajmabadi: This is interesting because now you're providing an even stronger guarantee.  That the hashcode is not only safe across reordering.  But also safe across versioning of the compiler/language.  I'm ok with such a guarantee.  but we would really have to be specific about it.  

Personally though, i don't like it.   :)
 
I would much rather us get behind a *weaker* guarantee.  That only within a session could you even be sure that hte hashcodes would be the same.  I would even do what some systems do and introduce some random value so that each session produces new hashcodes.  That way, if you actually tried to take a dependency on hashcode values, that you would break almost immediately.

proposals/data-classes.md
CyrusNajmabadi: > auto-properties are part of data-ness, computed properties may not be.

Agreed.  but my point is simply: when something "may not be", what side of hte fence do you decide they're on.  My argument is we shoudl just assume they *are* part of the data-ness.  And if you don't want that, you opt-out.

Note: i also think opt-out is a good idea because maybe i want an auto-prop and i do not want that auto-prop to be part of the data-ness.  Now, i'd have a consistent way for auto-props and computed-props to say that.  i.e.:

```c#
[NotData]
internal int LogLevel { get; set; }
```

I want this to be an auto-prop because i really don't need to compute anything.  But i don't want it participating in the data-ness.

proposals/data-classes.md
svick: @agocke [The documentation of `GetHashCode()`](https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode) strongly warns against doing that.

proposals/data-classes.md
CyrusNajmabadi: Well, i was hoping 'data' would solve the issue of:  

> +There is also no way to create a copy of a data structure with readonly
+fields with only one item changed. A new type must be constructed manually.

So i could imagine having a data class where i override everything.  But i still benefit from the fact that you provided me the easy way to generate a copy with only some pieces changed.

proposals/data-classes.md
CyrusNajmabadi: Ok good :)   Personally, in any docs, i think it might be good to start with the data being immutable. but then showing how you can opt-into mutability pretty easily.  but that' sa very minor point.

proposals/data-classes.md
agocke: We would not give the warning whenever there are read-only data members because someone may be using the object initializer support for them.

As far as `With`-ers goes, they will be generated too, but I don't have the details worked out yet. I think we may want to introduce an actual `with { }` expression to mirror the object initializer syntax to make handling read-only members easier (and not generate many intermediate objects).

proposals/data-classes.md
agocke: I've changed the example to read-only to show off the codegen hacks.

proposals/data-classes.md
CyrusNajmabadi: Basically, because collisions are far more likely.  frankly, i think from a language perspective we should say nothing more than: if you have the same member values in teh same session, you'll get teh same hashcode.    Then, from an impl perspective we should basically either call into HashCode.Combine (if it exists), or just emit the same GetHashCode code as we do for anonymous-types.

CyrusNajmabadi: Also: > is simpler

This isn't really a virtue (to me at least) for generated code.  We already have fairly good hashcodes for tuples/anonymous-types.   I would not want us to pick a poorer hashcode here (i.e. more likely to collide) and then regret that later on because customers are negatively impacted.

proposals/data-classes.md
CyrusNajmabadi: SGTM

CyrusNajmabadi:
@agocke Btw, i'm hugely supportive of design explorations in this space. So don't take my nitpicking to mean i'm trying to tear it down.  :)   I'm just trying to help make sure we can land in a great place across all these concepts!

proposals/data-classes.md
agocke: Sure. You've convinced me. We should use whatever hash code combine we think is fast and provides a good distribution. Hash.Combine is probably a decent starting point.

agocke:
@CyrusNajmabadi Totally understood, this is exactly what I hoped would happen. I think this proposal is getting better as we talk through all the details.

proposals/data-classes.md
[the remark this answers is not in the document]
agocke: @HaloFour I'm came up with this exact same idea! The problem is, what if you have a data struct. Now you have two structs for every struct. That seems like a lot of metadata bloat.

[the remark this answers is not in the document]
HaloFour: @agocke 

> That seems like a lot of metadata bloat.

Which, in my opinion, is a much more palatable issue to have than publicly-exposed secretly named writable "readonly" fields.  😄 

The builder pattern isn't an uncommon one to see, especially with readonly data classes.  To see it codified and accessible as a language feature would be a welcome addition, in my opinion.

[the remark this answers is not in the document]
CyrusNajmabadi: I, too, think this would be worth exploring.  

Though, i'm also amenable to either approach being taken.  I think @HaloFour (understandably) has a very visceral reaction to 'faked up' readonly.  I'm more ok with it.  But it would be nice if the underlying generation was actually just 'clean' and followed some sort of pattern that any language could consume/produce without having to understand hackery...

proposals/data-classes.md
CyrusNajmabadi: @333fred I think in that case, the best thing would be to just provide your own override of GetHashCode/Equals.  i.e. you know better.  So you can give the fastest impl.

CyrusNajmabadi:
Minor idea:  We're providing IEquatable/Equals/GetHashCode.  We could also potentially consider IComparable if all the values themselves are IComparable.  It would be nice since sorted-collections are a thing.  And this interface composes nicely over subcomponents.

proposals/data-classes.md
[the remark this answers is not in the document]
agocke: Not to reveal too many secrets but...

*all of F#'s supposedly thoroughly safe, immutable data structures are not marked initonly in metadata*

😉

This is not to imply F# is unsafe -- just the opposite. I think the truth is that language rules provide the vast majority of safety. This is also the position we took with the new `ref readonly` feature, where we allow `ref` variables to be taken directly to `readonly` fields and rely on the C# language rules to enforce the safety of `ref readonly` (which the CLR has no conception of).

agocke:
@CyrusNajmabadi This sounds like Swift's ["conditional conformance"](https://swift.org/blog/conditional-conformance/) and I'm all for it.

proposals/data-classes.md
agocke: @333fred That's what the proposal is talking about. You put `[DataMember(true)]` on the private field and it's considered in equality. The public computed properties are not.

I think @CyrusNajmabadi's approach is too heavy-handed.

CyrusNajmabadi: :D

Note: in fred's example, he jsut wants the private field used for equals/gethashcode.  Is there then a way to opt-out of something being a data member?  i.e. even if you have an auto-prop, can you put ```[DataMember(false)]``` on it? 

If so, i think i'm happy.  Basically, the language would have rules about if an unadorned member was a data-member or not (i.e. auto-props are, blah blah blah are not).  But you can always put on ```DataMember(true/false)``` on anything to explicitly opt-in/out.

Does that make sense ot you Andy?

proposals/data-classes.md
```diff
@@ -0,0 +1,381 @@
...
+    public string Username { get; }
+    public string Password { get; }
+    public bool RememberMe { get; } = false;
+}
+```
+
+and the use would be identical:
+
+```C#
+var x = new LoginResource {
```
huoyaoyuan: I'd like initializer becoming special calls to generated constructor, since I'm using
```csharp
var x = new LoginResource
(
    username: "andy",
    password: password
) 
```
, looks similar to object initializer, having benefits of `readonly` and preventing violate after initializer in old compilers.

proposals/data-classes.md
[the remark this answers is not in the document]
svick: @agocke I think relying on language rules with `private` or `internal` members is fine, but doing the same with `public` members is much more problematic. Does F# ever produce `public` non-`initonly` fields for its immutable data structures?

And `ref readonly` uses `modreq`, which means that some compiler could circumvent those rules, but it would have to be wilful.

Using unspeakable names offers questionable safety, especially since there is a major .Net language that can speak them, namely F# with its double backtick syntax.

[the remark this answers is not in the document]
HaloFour: @agocke 

*Warning:  I saw Hello World implemented in F# once so I'm totally qualified to comment on the language.*

F# also encodes a boatload of secret metadata in binary resources that is intended to only be understood and supported by F#.  Where you can expose metadata publicly either the rules are different or the magic breaks down.  An example might be F#'s extended generic type parameter constraints which most languages don't understand and don't enforce only to have the program explode at runtime if the generic type argument isn't "valid".  In my opinion C# should be a much better citizen in the .NET ecosystem of languages and should try to avoid relying on magic public behavior where it is possible.

I do agree that `readonly ref` does involve some magic, but I would argue that it's different in that it's relatively niche and that the magic mostly happens on the implementer side.  IIRC it would take someone writing some pathological code to simulate a `readonly ref` from another language which could abuse the lack of enforcement.  Data classes, on the other hand, are intended for a significantly wider audience to eliminate the boilerplate of creating these very common classes.  Having a leaky abstraction where any consumer can wreck the internal state of the class seems incredibly dangerous.

Anywho, the crux of my argument is that I would prefer that clean/safe approaches be considered before hacky ones.  In fact, I think support for a builder pattern would be nice for C# as an orthogonal feature and data classes can simply piggyback on that.

HaloFour:
How would the code gen work for `partial` classes?

CyrusNajmabadi:
One thing (important to me) came up in a gitter discussion on this topic.  Specifically, should/shouldn't positionality be part of the proposal.  I think i land in the camp where i want it.  Specifically, for both construction/deconstruction cases.  i.e. i would very much like to be able to write:

```c#
data struct Point {
    public int X { get; }
    public int Y { get; }
    public int Z { get; }
}

//...

var p = new Point(0, 0, 0); // <- this feels very idiomatic

// ...

var (x, y, z) = p; // <- data feels like it should be sensibly deconstructible

// ...

void TakesTuple((int x, int y, int z) t); //<-- seems like i could implicitly (or explicitly) cast to tuple to do this:
TakesTuple(p)
```

--

From our discussions we already seem to have acknowledge that hashing can be ordering dependent (and that's desirable).  

It feels like accepting and incorporating positionality isn't really problematic, and it very much helps bridge between any new world we've created and the existing world we have.  

Thoughts @agocke ?

HaloFour:
@CyrusNajmabadi 

> One thing (important to me) came up in a gitter discussion on this topic. Specifically, should/shouldn't positionality be part of the proposal. I think i land in the camp where i want it. Specifically, for both construction/deconstruction cases.

I disagree.  I think that `data` classes by themselves should not be positional.  They should be more or less synonymous with F# records, which are not positional.  I think that if you want to add in positionality that this is where primary constructors would come into play.  That gives you a constructor and the associated metadata necessary to generate a deconstructor.

```cs
data struct Point(int X, int Y, int Z);
```

I already think that there are challenges here with `partial` classes when it comes to codegen.  Having public members that depend on an order of the properties exasperates the problem by now exposing fragile APIs that are highly dependent on that order.  Given that I believe that `data` classes will be exceptionally useful with code generators like Entity Framework I think that `partial`+`data` classes will be a very common scenario.

CyrusNajmabadi:
@HaloFour's points are very interesting.  I do agree that positionality likely loses more appeal the larger the data type gets.  However, i still think there is a lot of merit in it for many areas i would use data.  For example, i could envision the entire roslyn syntax tree exposed through 'data classes' (if we had them at the start.  at this point, it probably wouldn't be changed).  And if we had, we would make those positional.  

Because it makes sense for some types, but not for others, i could see this being a knob somehow that a type could opt-into/out-of.  What controls that knob is open for discussion.  However, i think the discussion shoudl happen because i think a lot of *positional*-oriented data loses value here if there is no supported story around this.

Thanks!

CyrusNajmabadi:
>  I think that if you want to add in positionality that this is where primary constructors would come into play. 

That's def an interesting way to do things, and i'm amenable to it.  Just as a data-class/struct itself is an opinionated shorthand for generating a lot of boilerplate.  A data-class/struct *with* a primary constructor, is an even more opinionated shorthand.  

Seems like they're complimentary, and allow for an interesting way for the person declaring the type to make these sorts of opinionated decisions depending on their domain.

HaloFour:
@CyrusNajmabadi 

> I can see that. The important question i would have is: in those scenarios how common is it that you want **data** members spread over multiple parts. I can see other parts adding other members/functionality. But it's unclear to me if you'd want the other part to add data members.

You're probably right.  It's likely that the tooling would spit out the data members and the developer would supply helper methods or computed properties, etc., that don't necessarily participate with the other features of the `data` class.  The question is whether or not the compiler would enforce that rule and make it illegal to spread `data` members between `partial` classes.  I'm also very concerned that the tooling has to be *very* careful to be sure to always spit out the members in a specific order, which is something I guarantee won't happen.  This is going to be challenging enough to manage with `ToString` or `Equals` implementations that'll be "subtly" change based on order.  Having constructors/methods where the order of their arguments will change feels very dangerous.

I think positionality is quite useful and that there are numerous cases where it applies.  A `data struct Point` is a great example.  A `data` class representing a DB record with 20+ columns most of which are `string` much less so.  That's why I believe that this should be split across multiple individual language features that can be combined to provide the functionality, such as primary constructors, which are perfect for declaring positionality explicitly where it makes sense.

CyrusNajmabadi:
> I'm also very concerned that the tooling has to be very careful to be sure to always spit out the members in a specific order, which is something I guarantee won't happen.

I'm skeptical that that's true.  as long as this is ordered from the start.  Because if someone screws something up with tooling here, you literally break everything immediately :)

333fred:
@CyrusNajmabadi from initial LDM discussions, the intent for this PR is specifically for `data` classes to not be positional. @agocke will have an upcoming proposal for positional data classes, akin to something like a named tuple.

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+* Login info
+* Configuration options
+
+Data *is not*
+
+* A process
+* A computation
+* A conversation
+* Interactive
+* An object
```
chucker: This example I'm unclear about.

Inheritance is explicitly mentioned further down, so what makes it _not_ an object? Lack of a complex graph?

CyrusNajmabadi:
@333fred That's fine :)   I would like the PRs to exist at the same time though.  Because it's hard to comment on one without the other to look at and understand the relationship to.  Thanks! :)

proposals/data-classes.md
```diff
@@ -0,0 +1,241 @@
...
+
+For an object-oriented language this may seem strange, because isn't
+everything an object? In some sense, you can view data as a degenerate object
+-- fields with pure transparency. But this also misses the point. The point
+of object-oriented architecture is to bundle state and behavior and provide
+composable objects that can interactively respond to the system, like cells
+in an organism. There's value in this structure, but it also creates a
+binding between the data and the behavior. By creating data individually we
+allow the data to shift contexts and allow other components to define their
+own behaviors.
```
chucker: Right, so the argument here is data isn't an object because it has no behavior?

proposals/data-classes.md
chucker: FWIW, this also breaks serialization (in particular, _de_serialization), at least the way it's commonly implemented in e.g. Newtonsoft.Json.

chucker:
The main driving argument in this proposal seems to be:

* anonymous types and tuples are both inadequate for data
* classes get close, with some caveats:
  * implementing a comparer requires [a fair bit of boilerplate](https://gist.github.com/agocke/6ba6e64f77f1212ba7292bfd1f1000e5) there is no piecewise comparer implicitly defined for C# classes, so if you want simple data comparison, the real example looks like this
  * the object initializer syntax requires writeable properties (or fields)
  * a constructor, instead:
    * requires manual maintenance
    * implies/requires positioning, including for the caller
    * offers no support for `With()`

However, I see some disagreements in the comments. For example, to some, positionality in data classes is a feature, not a bug.

This proposal is arguably an attempt to slim down the broad #39, and I'm not convinced an end-all-be-all solution to a term as broad as "data" or "records" is (or needs to be) achievable.

**I wonder if we can instead approach the above goals piecemeal:**

1. much of the code in "C# data class example" revolves around equality and hashing. I would wager that a _lot_ of programmers writing C# skip on an implementation of `GetHashCode()`, or worse, implement it incorrectly. Making a C# feature available that provides, out of the box, reasonable (but perhaps not optimized, just good enough) implementations of `Equals`, `GetHashCode`, `==` and `!=` seems like a useful goal _in and of itself_ and completely separate from data classes. Model, ViewModel, Controller, Manager, whathaveyou classes can all benefit from such a language feature.
2. similarly, the object initializer syntax not allowing _private_ setters is generally a good feature, but given that specific initializer syntax does exist, and that writing to readonly setters is occasionally requested (such as [here](https://stackoverflow.com/questions/18385248/readonly-field-in-object-initializer) or [here](https://social.msdn.microsoft.com/Forums/en-US/23e2b51b-b82e-4149-aa2d-46db4f0f7597/allowing-object-initializers-to-set-readonly-fields?forum=vcsharp2008prerelease)), it stands to reason that this, too, would be in and of itself a useful language feature. Look ma, no constructor required!

These two features wouldn't depend on each other. They also wouldn't, unlike the `With()` syntax and deconstructors, feel rather unidiomatic for C#. They could, per se, already offer benefits.

And when taken together, they would solve many of the problems of data classes while _also_ letting you specify additional niceties. E.g., if you _do_ want positionality, just _do_ have a constructor. If you don't, use initializer syntax.

To solve **1**: introduce an `auto-equality` keyword on classes. This may require a more explicit annotation for what constitutes a "relevant" field, perhaps exactly as in this proposal's `[DataMember]` attribute.

While we're at it, a similar `auto-string` keyword for the `ToString()` impl. seen here, perhaps? This is similarly something C# programmers effectively often _skip on_ because it's a bit of boilerplate for what's perceived as little gain.

And to solve **2**, we do the same on properties, with a new `init` keyword on the setter, allowing _(only)_ object initialization to write _(once)_ to the property.

proposals/data-classes.md
jcouv: +1 on including `ToString`. It's handy for debugging.
A debugger display tip would be ok, as an alternative.

---
In reply to: [198207292](https://github.com/dotnet/csharplang/pull/1667#discussion_r198207292) [](ancestors = 198207292)

proposals/data-classes.md
```diff
@@ -0,0 +1,381 @@
...
+}
+```
+
+and the use would be identical:
+
+```C#
+var x = new LoginResource {
+    Username = "andy",
+    Password = password
+};
```
jcouv: We'll have to think about the intersection between `data` types and nullable feature. The nullable feature produces warnings if some non-null fields are left uninitialized by the constructor.
If you hadn't initialized `Username`, we'd probably want to warn for uninitialized non-null field...
Can you take a note of that somewhere to remind ourselves to think the issue through? Thanks

proposals/data-classes.md
```diff
@@ -0,0 +1,381 @@
...
+```
+
+Note that `RememberMe` must have an initializer to avoid a warning in the
+object initializer about an unset read-only property.
+
+However, the generated class code would look like:
+
+```C#
+public class LoginResource : IEquatable<LoginResource>
```
jcouv: >IEquatable [](start = 29, length = 10)

Looking at `System.Tuple` and `System.ValueTuple`, there are some more interfaces to consider: IEquatable<T>, IStructuralEquatable, IStructuralComparable, IComparable, IComparable<T>

jcouv: >LoginResource [](start = 13, length = 13)

I assume that we would tag this type somehow (maybe with an attribute), to identify it as a data class?

proposals/data-classes.md
```diff
@@ -0,0 +1,381 @@
...
+    public string <>Backing_Username;
+    public string Username => <>Backing_Username;
+    public string <>Backing_Password;
+    public string Password => <>Backing_Password;
+    public string <>Backing_RememberMe = false;
+    public bool RememberMe => <>Backing_RememberMe;
+
+    protected LoginResource() { }
+
+    public static LoginResource Init() => new LoginResource();
```
jcouv: >Init [](start = 32, length = 4)

It would be good to show what how caller code is transformed to use these APIs. I assume it's simply:

``` C#
var temp = LoginResource.Init();
temp.<>Backing_Username = "";
temp.<>Backing_Password = "";
...
```

I'm wondering if the `Init` method should also be unpronouncable. If someone wrote `var temp = LoginResource.Init();` they would escape the initialization checks (which members are left unset).

proposals/data-classes.md
```diff
@@ -0,0 +1,381 @@
...
+The LoginResource class now could be defined as
+
+```C#
+public data class LoginResource
+{
+    public string Username { get; }
+    public string Password { get; }
+    public bool RememberMe { get; } = false;
+}
+```
```
jcouv: It may be called out somewhere already, but one can add other members such as fields and methods (including some implementations that supercede the auto-implemented GetHashCode/Equals/ToString/...).

agocke:
I have posted my second of three proposals around the "simple data -> simple types" concepts: https://github.com/dotnet/csharplang/pull/1673

proposals/data-classes.md
agocke: Yes that should absolutely be unspeakable.

proposals/data-classes.md
agocke: Inheritance can be possible without changing the contract with the type, namely that it is simple data and does not represent an interactive process.

jnm2:
I'm happy to see the withers section is coming. Still curious about the ability to override ToString() (https://github.com/dotnet/csharplang/pull/1667#issuecomment-400388178). What would be the ramifications of allowing custom ToString overrides?

agocke:
@jnm2 In my proposal, if you define a method yourself, like ToString, it will automatically prevent the generated one from being created. So you could write your own ToString and the compiler will not generate one, but the rest of the items will still be generated (like Equals).

jnm2:
@agocke Oh, sorry I missed reading that!

jnm2:
@agocke Oh, sorry I missed reading that! I see that in the named tuples proposal but not this one.

proposals/data-classes.md
```diff
@@ -0,0 +1,392 @@
...
+            Debug.Assert(_field >= 0);
+            return _field;
+        }
+        set { ... }
+    }
+}
+```
+
+To support these cases and provide an easy escape hatch, I propose a
+new attribute, `DataMemberAttribute` with a boolean flag argument on the
```
HaloFour: `System.Runtime.Serialization.DataMemberAttribute` already exists in the BCL.

HaloFour:
I'd like to posit the following gist as the generated code for the example `data` class.  It's largely the same as what @agocke posted already but the biggest changes are around "readonly initialization".

https://gist.github.com/HaloFour/66c582deac779464371381410b66f80f

And the following demonstrates adding a primary constructor to enable positional construction and deconstruction:

https://gist.github.com/HaloFour/4ca895d3f4356dafdc19cfdde20b7c41

I've probably missed some bits and I've excluded "withers" as that hasn't been fleshed out yet.

CyrusNajmabadi:
@agocke From looking at @HaloFour 's proposal, all that is extra seems to be the following:

```c#
    public struct Builder
    {
        public string Username;
        public string Password;
        public bool RememberMe;
}
```

So this effectively equates to '1' extra type,  and 'N' extra fields.  It would be good to get a sense of if that would actually be a problem.  In my experience when we talk about metadata-bloat, it's usually for areas where we would actually reasonably expect someone could use a feature an genuinely add thousands and thousands of things.  That doesn't seem to be the case here.  Nearly all data types are going to map out at less than a dozen or so data members.   So having those small handful of additional field-members in another type doesn't seem like a large burden for these.

proposals/data-classes.md
agocke: Oops. DataMemberAttribute2 😄

agocke:
@HaloFour yeah, it's workable. Still seems like metadata bloat. I don't know how much to care.

CyrusNajmabadi:
I had a semi-serious, semi-awful though.  What if we did it effectively as: 

```c#

// pseudo code.  i know C# doesn't actually support this
using Builder = (string Username, string Password, bool RememberMe);

public sealed class LoginResource : IEquatable<LoginResource> {
    public static void Init(ref Builder builder)
    {
        builder.RememberMe = true;
    }
    
    private readonly Builder _builder;
    
    public LoginResource(Builder builder) => _builder = builder;
    
    public string Username => _builder.Username;
    public string Password => _builder.Password;
    public bool RememberMe => _builder.RememberMe;
```

Effectively, we don't need an actual 'Builder' type.  We just need a mutable struct.  We already have a perfectly good mutable struct in the language already: tuples.  

--

Note: my guess is that this would have some binary-compat issues.  With a real nominal type, you're safe to member reordering.  With a tuple, you likely wouldn't be.

--

But it's still interesting from the 'reduce metadata' perspective...

agocke:
Now you would be dependent on ordering of the builder tuple elements.

agocke:
You can also simplify your example by adding a Builder constructor, removing Init, and having the LoginResource constructor take the builder by `in`

HaloFour:
@CyrusNajmabadi 

> So this effectively equates to '1' extra type, and 'N' extra fields.

Why would it be 'N' extra fields?  Wouldn't using the `Builder` struct as the only field of `LoginResource` cause only 1 extra type and 1 extra field compared to the baseline of this proposal?

@agocke 

> make Builder unspeakable

I disagree with this in principal.  There isn't a single case (that I am aware of) where C# currently allows a generated "unspeakable" detail to leak publicly (at least not without having to break out reflection, which is inherently fragile anyway).  Doing so in something intended to be consumed publicly forces that name and how it is generated to be a part of the language specification otherwise no other language (or no other implementation of the C# compiler) can safely emit consuming code that can be expected to survive a recompile of a dependent assembly in which it is defined.

> and having the `LoginResource` constructor take the builder by `in`

I'm still toying with BenchmarkDotNet here but I did try this approach using both `in` and `ref` and it was slower than passing `Builder` by value, at least on RyuJITx64.  The difference wasn't that great, though, maybe ~10%.  I expected it to be faster due to saving a copy of the struct.  This was even the case when I tried structs with 20 fields.  But I have zero issue with that parameter being `in`/`ref`.

svick:
@chucker 

> I would wager that a *lot* of programmers writing C# skip on an implementation of `GetHashCode()`, or worse, implement it incorrectly.

I think this part is fairly decently covered already by .Net Core 2.1 [`HashCode`](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode?view=netcore-2.1) and by Roslyn's [`GetHashCode()` generator](https://github.com/dotnet/roslyn/tree/master/src/Features/Core/Portable/GenerateEqualsAndGetHashCodeFromMembers).

agocke:
@CyrusNajmabadi @HaloFour Both of you have brought up safety of use in other languages. I don't see how you enforce the mandatory initialization of readonly properties with this proposal in other languages.

jnm2:
Single public constructor receiving a builder struct?

jcouv:
@agocke `modreq`s seem the standard solution to prevent other languages/compilers from misusing a new construct that they may not understand properly.

agocke:
@jcouv That's the same as making it unspeakable. Other languages won't be able to interop without compiler changes.

HaloFour:
@agocke 

>  I don't see how you enforce the mandatory initialization of readonly properties with this proposal in other languages.

You mean properties that have initializers, such as `public bool RememberMe { get; } = false;`?  Or do you mean forcing the consumer to supply a value for those fields, like with non-nullable references?

agocke:
@HaloFour The latter.

Consider

``C#
data class MyClass
{
    public int Prop { get; }
}
```

The object initializer will force the user to provide a value for this property, or produce an error.

**Comment by HaloFour** [comment]
@agocke 

> The object initializer will force the user to provide a value for this property, or produce an error.

I don't know that you can force it, not without a constructor in the mix somewhere.  The public field method can't either.  That relies on other languages not being able to modify the field, but F# can already speak the unspeakable.

Is it really that big of a deal, though?  I mean, if you have a `data` struct it's not like you can prevent it from being `default`, even in C#.

@jcouv 

> `modreqs` seem the standard solution to prevent other languages/compilers from misusing a new construct that they may not understand properly.

This feature is supposed to simplify development of an *existing* construct, no?  Why would anyone want to adopt something that can only be used in one language and doesn't play well with any of the existing libraries that normally work with data classes?

**Comment by agocke** [comment]
> Is it really that big of a deal, though? I mean, if you have a data struct it's not like you can prevent it from being default, even in C#.

No, but then again, is skipping `readonly` for the backing fields such a big deal? 😉

**Comment by HaloFour** [comment]
@agocke 

> No, but then again, is skipping readonly for the backing fields such a big deal? 😉

Yes.

```cs
public struct Foo {
    public readonly string Bar;
}

Foo foo = default;
foo.Bar = "DEFEATED!"; // compiler error!
```

But public unspeakable is a much bigger deal.  It's a leaky abstraction and fragile.

I keep coming back to maybe primary constructors are a necessity for these kinds of classes.  They provide the positionality to generate a proper constructor (which C# could hide behind initializer syntax) which enforces assignment of these particular properties.

Either that or lexical ordering of properties becomes important which establishes the positionality to generate a proper constructor (which, again, C# could hide behind initializer syntax).  That's how records work in F#.  But that's a somewhat big break from how properties normally work in classes and it necessitates making it illegal to apply `data` to multiple parts of a `partial` class (which should probably be the case regardless).

HaloFour:
@agocke 

> Other languages setting unspeakable properties?

Safety is part of it, yes.  Marking a field as `readonly` sets an expectation which is now broken.  Sure, you could defeat that before via reflection, but reflection used to require full trust where all bets were off anyway.  I don't think that's reason enough to bin it.  I'm willing to contend that there are situations where the language can effectively enforce `readonly`-ness, but that only works if the field is never exposed outside of the assembly.

Much more than that it's the idea of making something public and unspeakable with a magickally generated name.  It creates this weird and unnecessary barrier to a feature that's intended to make life easier on the developer by eliminating boilerplate.

I feel like I'm just repeating myself about this particular point.  If you don't feel that it's sufficient reason to explore a less hacky approach that's fine, we'll have to agree to disagree.  Either way you're not going to stop other languages from doing the wrong thing.

agocke:
OK. I think the builder approach is also viable and something worth considering.

However, given that preventing dependency on positional layout was an explicit goal of this proposal, I'm going to reject any solutions that require that out of hand.

If that ends up being a requirement, we shouldn't take this proposal.

HaloFour:
@agocke

I know that I have been pretty vocal about very specific aspects of the implementation of this proposal but I think that the majority of it is great and I celebrate the desire to eliminate this very common boilerplate.

agocke:
@HaloFour No problem, there are many proposals in this area. We just have to work out which one everyone thinks is best.

HaloFour:
@agocke 

Just tossing some spaghetti against the wall but here's a method that I think might be able to enforce assignment of required fields at runtime, but because it involves adding properties for each of the fields in the builder struct if there is real concern about metadata bloat then this might be over the line.

https://gist.github.com/HaloFour/3a661fda9a9b1b131c1f65c25652a333

Of course there should also be metadata (attributes) on the properties of the data class so that the C# compiler and other supporting compilers can catch the missing fields at compile time.

proposals/data-classes.md
DweeberlyLoom: This _feels_ like a feature better suited to an external code generator rather than an addition to the language.  It is too complicated to be considered syntactic sugar, but doesn't add 'new' functionality, it just "rewraps" existing available features.  A code generator that accepts a data definition and generates the appropriate C# class would provide this without adding complexity to the language.

theunrepentantgeek: And such generators already exist - here's one: https://github.com/johnazariah/csharp-algebraictypes

agocke: There are multiple features that could not be implemented by a code generator in this proposal, including the use of object initializers to initialize read-only members.

proposals/data-classes.md
agocke: @CyrusNajmabadi Yeah, the boolean flag in the constructor was supposed to work both ways. To both opt-in and opt-out.

fanoI:
What is the difference between this and the Record proposal?

HaloFour:
@agocke

> This is not quite @HaloFour's proposal, because I still favor unspeakable initializers to allow compilers to enforce safety rules around required initialization.

In my opinion that makes them unnecessarily inaccessible from other languages or downlevel compilers.

proposals/data-classes.md
```diff
@@ -195,21 +194,66 @@ var x = new LoginResource {
...
 ```C#
 public class LoginResource : IEquatable<LoginResource>
 {
-    public string <>Backing_Username;
-    public string Username => <>Backing_Username;
-    public string <>Backing_Password;
-    public string Password => <>Backing_Password;
-    public string <>Backing_RememberMe = false;
-    public bool RememberMe => <>Backing_RememberMe;
+    public struct <>Initializer
```
CyrusNajmabadi: my personal preference would be that this just be a well known name that then supports this pattern 'by convention'.  Like so many patterns in the language, we would just be looking for a general accessible/public API-shape, and if found, would light up this functionality.  

So, in this case, having htis be called "Initializer" or "Builder" would be far preferred to some unnameable name like `<>Initializer`.

proposals/data-classes.md
```diff
@@ -372,6 +414,27 @@ unspeakable public "initialization" method. This method is called when using
...
+`readonly` members in object initializers may seem like a small feature, but
+it's important that making a `data` type readonly not come with a lot of
+extra ceremony. The essence of a `data` type is a set of named fields and
+that should stay true regardless of whether or not the fields are `readonly`.
+It may be tempting for implementation simplicity to try to use constructors
+instead, but this is a design smell that conflates positional semantics with
+`readonly` semantics. Requiring initialization via constructor means that
+field order becomes a public API and requires careful versioning, which is
+not true of mutable `data` types and is a constraint that should be
+irrelevant to `readonly` semantics.
```
CyrusNajmabadi: ok... i've read this a few times... and i feel like i'm not getting it.  Specifically, i don't see why it's necessary for the initializer name to be unspeakable.

HaloFour:
@CyrusNajmabadi 

I think the argument for unspeakability is that then the compiler can enforce that some members *must* be initialized.  If the initializer is simply based on a normal coding pattern then downlevel compilers can consume it directly and initialization of those required members can't be enforced.  

Given that these protections are at best suggestions which can't be enforced particularly with languages like F# that can speak unspeakables and that adding a new required member is very intentionally not a breaking charge and existing consumers would continue to not initialize that member I don't think it's a point worth making the type unspeakable.  The compiler enforcement could just as easily be accomplished through an analyzer on the builder member.

proposals/data-classes.md
```diff
@@ -0,0 +1,455 @@
...
+Aside from places where semantics are underspecified, this section is mostly
+for edification. The implementation of the feature is irrelevant to the
+language design itself. The implementation may change significantly as it is
+actually developed.
+
+That being said, the current proposal is that the generated class code would
+look like:
+
+```C#
+public class LoginResource : IEquatable<LoginResource>
```
alrz: Can we make IEquatable impl opt-in? e.g. only if spelled out in the source: `data class C : IEquatable<C>`

proposals/data-classes.md
```diff
@@ -0,0 +1,455 @@
...
+1. Automatically generating `Equals`, `GetHashCode`, `ToString`, `==`, `!=`, and `IEquatable<T>`
+   based on the member data of the type.
+1. Allow object initializers to also initialize readonly members.
+
+The LoginResource class now could be defined as
+
+```C#
+public data class LoginResource
+{
+    public string Username { get; }
```
alrz: Does it even make sense to make these non-public? Could it default to `public`?

agocke: Private fields and properties are perfectly fine. We could consider it, but right now I'm not wild about changing the default accessibility of members. That's a significant amount of extra complexity in the language, and peoples style guidelines often depend on it.

agocke:
@HaloFour That's one argument, but I also want to leave this entire initialization phase as implementation defined to allow for codegen changes later.

proposals/data-classes.md
CyrusNajmabadi: > You realize this is a ridiculously complicated pattern, right?

Not to me :)

> You have to have a public builder.

Not ridiculous.  Not complicated (IMO).

> You have to manually move your initializers from the class to the builder (the compiler can't help with or verify this step).

Neither ridiculous nor complicated :)

> The fields must have the exact same names as the properties. 

Neither ridiculous nor complicated :)

> What if the property has a setter? Is the pattern illegal now (allowing it would be a breaking change)? What message does the compiler produce?

I'm happy to discuss all those potential concerns.  But i would like to start with an open mind that this isn't defacto 'absurd' as you are making it out to be :)

It may be that there is enough cases to help drive the idea that this is better as something truly unnameable.  But i would prefer that be the *result* of a good discussion with enough informign cases, rather than haivng that be the presumed starting point.

Effectively, none of the above seemed at all strange or onerous to have as a defacto pattern.  Heck, it seems exceptionally tame compared to many other patterns we have in the language :)   (Just my 2c on these).

proposals/data-classes.md
alrz: to reduce the amount of possibly unused code generated. the opt-in solution would be synonymous to `derive` or `deriving` in rust and haskell.

proposals/data-classes.md
alrz: also related, I think it's actually useful to force properties to come first in the class body, otherwise it doesnt look like a "data class" anymore. sure it might be basically a style preference but it worth to consider.

agocke:
@CyrusNajmabadi @HaloFour I think you guys are misremembering how patterns work in C#. Patterns are for allowing user-written types to substitute for well-known framework types. So instead of just Task<T> being awaitable, anything that looks like Task<T> is awaitable.

There's no such thing as a "compiler-generated" pattern. If the pattern components are public and speakable, that means that regular C# code can interface with them, which means that they must be *real* and actually exist in source. If you can type it, then you need to be able to go-to-def on it, so the source must be there and navigable. The compiler, never, ever, alters source trees or produces source code. This is just a non-starter.

proposals/data-classes.md
alrz: I don't underestand what it buys you if it was a pattern? why you'd want to implement it manually? also I think @agocke's point about it being "complicated" is not the pattern itself,  the compiler should be able to extract semantical information from that pattern which could change as the feature involve in the future, defining such pattern can be quite compilicated to be future-proof. on the other hand, as a "lowering strategy" it will come after all the semnatic analysis and also can *change* which has a greater value imo.

svick:
@agocke 

> If you can type it, then you need to be able to go-to-def on it, so the source must be there and navigable.

What about properties of anonymous types or the `Invoke` method of a delegate type? You can type them, but you can't really go-to-def on them.

alrz:
@svick I think it means "at the declaration site."

agocke:
Well, the rule isn't exact: certainly the generated constructor of a class with no constructors is "real" and has a go-to-def target, but this isn't a "pattern". I actually have the same question for the "with" methods proposed in the records proposal. If you can call those methods and treat them as "real" methods, how do they appear to the compiler? Do we actually create fake method symbols to represent them? Where's their syntax?

CyrusNajmabadi:
> @CyrusNajmabadi @HaloFour I think you guys are misremembering how patterns work in C#. Patterns are for allowing user-written types to substitute for well-known framework types. 

I'm just 'pattern' in a generic sense.  i.e. just the general idea of "what the compiler would look for."

If what we lower to is something reasonable (which i feel like it totally could be0, i don't see any reason why i wouldn't want the compiler to be able to see such a thing in IL (which might have been produced by something else) and go: "ah, i know how to use that!  i'll let you use things like Object-Initializers, but then have it compile down to that Class *and* it's Builder class automatically for you". 

> So instead of just Task being awaitable, anything that looks like Task is awaitable.

Yes.  And that's *effectively* what i want.  Instead of 'object initializers' only working with classes/structs with mutable fields/properties, i want *anything* that looks 'initializable' to be initializable.  However, we would extend the 'initializable' pattern to *recognize* seeing a well known convention to enable this facility.

CyrusNajmabadi:
> The compiler, never, ever, alters source trees or produces source code. This is just a non-starter.

I don't know why that needed to be mentioned :)   There must be some confusion as i haven't ever expected the compiler to do anything of the sort :)

CyrusNajmabadi:
>  That's one argument, but I also want to leave this entire initialization phase as implementation defined to allow for codegen changes later.

I definitely get that.  And i think that's a reasonable position to take.  It's just one i disagree with.

I think we should spend hte cycles (even though they may not work out) to consider trying to make this *not* an implementation detail, but to strongly specify what the phase should look like.  

If you then created a `data class` you would get that specific lowering output.  However, *conversely*, if we saw that lowered 'pattern' (i'm not sure what a great term here is, so i'm going to keep using 'pattern'),  we would say: "oh, this is a data class!".    

That's all. :)

Note: i'm not saying: "we have to do this".  Or "it's ludicrous and stupid to do this another way".  I'm simply saying that we should examine and consider this and properly weigh out the pros/cons here before dismissing it out of hand right at the beginning.  I personally think there is a lot of merit in defining a convention-based approach to mutable/immutable data.  And i think if we're going to spend cycles looking into this space, we shouldn't be so quick to just immediately reject this idea.

HaloFour:
@agocke 

> I think you guys are misremembering how patterns work in C#. Patterns are for allowing user-written types to substitute for well-known framework types.

Are collector or indexer initializers not patterns?  If not, what are they?  I certainly don't want to get bogged down in an argument over what things are named.  I care about how they behave.  I'd rather a general purpose feature, one that enables me to then add said support to my own types or possibly even other types via extension.  Immutable object initializers are an orthogonal concern with other existing requests/proposals.  Having this proposal come up with a method that both can't be used elsewhere in the language and also can't be used well from other languages or downlevel compilers seems like a real shame.

CyrusNajmabadi:
I think i feel very similarly to @halofour.  It seems like a shame to go down a path that seems less capable and less enabling, when (my hope is) that there is a very cheap alternative that surfaces the same user experience, while also not preventing these other interesting scenarios.

HaloFour:
One question, which I don't think is explicitly addressed in the proposal: is the addition of a new required readonly property to a nominal record considered a breaking change?

gulshan:
Will positional and nominal records be mutually exclusive?

agocke:
@gulshan I don't think they have to be, no. But the intersection of the two could be very different than either alone.

HaloFour:
@agocke 

> It would produce a new warning at the minimum, so yes.

Ok.  Just mentally reconciling the gymnastics around initialization with the end result of being able to add new members without breaking existing consumers, at least until they're recompiled.  This feature seems closely related to records in F# which take a fairly different approach regarding initialization and consider the addition of members to be a hard breaking change which requires recompilation.

HaloFour:
I know that I've beaten this poor horse to death but I wanted to note the contention that language-specific unspeakable encoding of features is causing in CoreFX to the extent that they are recommending that such features not be used in public APIs:

https://github.com/dotnet/corefx/issues/33553

agocke:
@HaloFour You're not wrong. In fact, you've convinced me. The nail in the coffin for me is usage in older versions of the C# compiler. I don't want to completely lock out old compiler versions, so I think the builder should be speakable, but marked with [EditorBrowsable(false)] so it doesn't pollute IDE suggestions.

agocke:
I didn't mention it in the notes, but in the LDM discussion on this I actually proposed making the builder speakable.

proposals/data-classes.md
sharpninja: > @agocke No, this is a different issue. The problem with using a symmetric operation for combining hash codes is that it means that e.g. `new Point { X = 1, Y = 0 }.GetHashCode()` returns the same value as `new Point { X = 0, Y = 1 }.GetHashCode()`.

This is easily solvable by adding the name of the field to the hash.  
 
> The comment about pretending that types are the same is because I assumed the reason for symmetry is that you wanted e.g. `new Contoso.Person { FirstName = "Andy", LastName = "Gocke" }.GetHashCode()` to be the same as `new AdventureWorks.Person { LastName = "Gocke", FirstName = "Andy" }.GetHashCode()`.

Again, making the name of the field part of the hash solves the problem.

agocke:
I'm gonna close this out because I have a new proposal that I'll soon that doesn't require builders or marking anything unspeakable

PR #1683, opened 2018-07-02
readonly object initializers

jaredpar:
This proposal will extend object initializers to allow assignment of
readonly fields and get only auto-implemented properties.

### DIFF
--- a/proposals/readonly-struct-init.md
+++ b/proposals/readonly-struct-init.md
@@ -1,0 +1,161 @@
+# readonly object initializers 
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Below
+
+## Summary
+
+This feature will allow for assigning of `readonly` fields and get only auto-implemented properties
+in object initialezrs when the type being constructed is a `struct` or `ref struct`.
+
+## Motivation
+
+The intent of `readonly` in C# is to prevent the modification of a value after construction has 
+completed. That is once the code calling `new` has a reference to the value there should be no 
+further modification of `readonly` fields. Until the calling code has a reference to the object
+modifications can continue to be made via `this`.
+
+This makes it even easier to define usable `readonly struct` types in C#. The simplest form of
+declaring the fields now creates a fully usable value:
+
+``` csharp
+readonly struct Point { 
+    readonly int X;
+    readonly int Y; 
+}
+
+var p = new Point { X = 42; Y = 13 };
+```
+
+This means authoring `readonly struct` types in C# is just as easy as creating the less
+perferred mutable `struct` types. There is no longer a need to define a constructor which simply 
+re-lists every field as a parameter and then assigns them one at a time. The definition of the 
+field itself allows provides a mechanism for initialization
+
+## Detailed Design 
+
+### Language
+
+The language will extend the object initialization rules to allow for an accessible readonly field 
+or auto-implemented property to appear in the member initialization list. In such a context the 
+members will allow for assignment just as it would inside a constructor of the object. 
+
+The following restrictions will apply to this extension:
+
+1. The type being instantiated must be a `struct` or `ref struct`. 
+2. The constructor used in the initializer cannot have any `out` or `ref` parameters.
+
+The (2) restriction exists because such a parameter could smuggle out `this` from the object 
+constructor. That would make a reference to the constructed object visible to the code calling
+`new` and hence allow it to observe mutations to the value.
+
+### Compiler 
+
+The compiler will change how the backing field of an auto-implemented property is emitted in the 
+following ways:
+
+- The accessibility of the backing field will match the accessibility of the property.
+- The name of the field will be unspeakable (illegal C# identifier hence never imported) but it 
+will be possible to map such fields to the properties that defined them.
+
+Taken together this allows the C# compiler to translate a get only auto-implemented property into
+a simple field initialization when used as an object initializer. 
+
+``` csharp 
+struct Name {
+    // Backing field: public readonly string <FirstName>__BackingField;
+    public string FirstName { get; }
+    // Backing field: public readonly string <LastName>__BackingField;
+    public string LastNmae { get; }
+}
+
+var name = new Name { FirstName = "Jared", LastName = "Parsons" };
+
+// Emitted as 
+var name = new Name { <FirstName>__BackingField = "Jared", <LastName>__BackingField = "Parsons" };
+```
+
+This does mean that the feature will only apply to auto-implemented properties that are 
+re-compiled with a new compiler. The compiler today emits the fields as `private readonly` and 
+hence will be inaccessible to calling code. Upon recompile the backing fields will be emitted 
+as specified above.
+
+### Verification
+
+As is the case with several features added since C# 7.2 this feature will not pass any known 
+peverify implementation as it makes no real distinction between read only and mutable memory. 
+However it is still desirable for C#, and other languages, to emit IL that can be verified 
+with appropriate adjustments to the verification spec. Eventually allowing for projects like
+[ILVerify](https://github.com/dotnet/corert/tree/master/src/ILVerify) to take the place of 
+peverify today. 
+
+This rules could be adjusted to allow for mutation of a `readonly` field following an `.initobj` 
+or `.call .ctor` instruction if:
+
+1. The reciever of the call is a local 
+1. The type of the reciever is a `struct` or `ref struct`
+1. There are no `ref` locals which refer to that local
+
+The `readonly` fields would remain mutable until one of the following occured:
+
+1. The reciever was passed as any `ref` to another function except when being used as `this`
+1. A `ref` local was taken to the local or any of its members
+
+This may force conforming compilers to create extra locals in some cases to allow for a
+successful verification of the emitted code.
+
+## Considerations
+
+### Type Parameters
+
+The feature does not apply to type parameters constrained to a `struct`. There is just no case 
+where such a type would have a field (`readonly` or not) or an auto-implemented property. In the
+future if we allow for `struct` inheritance it's possible this could change and the decision 
+re-evaluated.
+
+## classes
+
+This can't apply to `class` instances as it's much easier to smuggle a reference out of the 
+constructor. The `this` value can be easily stored in a `static` making it accessible to the 
+calling code long before the object initialization has completed. 
+
+``` csharp
+static C Global;
+
+class C { 
+    public readonly int Field;
+
+    public C() { 
+        Global = this;
+    }
+}
+
+int SomeMethod() => Global.Field++;
+
+// The instance is observed by caller while mutations are still in progress.
+var c = new C() { Field = SomeMethod()}
+```
+
+A `struct` can't be smuggled out in such a way as it would require a direct copy to another 
+instance of the same type or by boxing to an interface / object. In either case it would be a 
+different value and not interesting here.
+
+### Future Considerations
+
+There is another proposal to allow for simple data types in the language via the `data` modifier.
+Such types would have auto-matic equality, hashing, ToString, etc ... Combined with this proposal
+and immutable data now has the simplest possible definition in C#:
+
+``` csharp
+public readonly data struct Point { 
+    public readonly int X;
+    public readonly int Y;
+}
+```
+
+This type now has initialazation via object initializers, equality, hashing and ToString. As the 
+type evolves by adding new fields these pieces of functionality will evolve with it.
+
+Note: this begins to look very much like F# record types. 


proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+
+Taken together this allows the C# compiler to translate a get only auto-implemented property into
+a simple field initialization when used as an object initializer. 
+
+``` csharp 
+struct Name {
+    // Backing field: public readonly string <FirstName>__BackingField;
+    public string FirstName { get; }
+    // Backing field: public readonly string <LastName>__BackingField;
+    public string LastNmae { get; }
```
davkean: "LastName"

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+will be possible to map such fields to the properties that defined them.
+
+Taken together this allows the C# compiler to translate a get only auto-implemented property into
+a simple field initialization when used as an object initializer. 
+
+``` csharp 
+struct Name {
+    // Backing field: public readonly string <FirstName>__BackingField;
+    public string FirstName { get; }
+    // Backing field: public readonly string <LastName>__BackingField;
```
davkean: What will these be marked as in metadata? Special named?

This would now be a breaking change to move from auto-property -> normal property.

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+    public string LastNmae { get; }
+}
+
+var name = new Name { FirstName = "Jared", LastName = "Parsons" };
+
+// Emitted as 
+var name = new Name { <FirstName>__BackingField = "Jared", <LastName>__BackingField = "Parsons" };
```
davkean: We're relying on the runtime not blocking assignment after construction?

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+further modification of `readonly` fields. Until the calling code has a reference to the object
+modifications can continue to be made via `this`.
+
+This makes it even easier to define usable `readonly struct` types in C#. The simplest form of
+declaring the fields now creates a fully usable value:
+
+``` csharp
+readonly struct Point { 
+    readonly int X;
+    readonly int Y;
```
davkean: Nit: These are private by default, right? Therefore below wouldn't compile unless within the same type?

proposals/readonly-struct-init.md
nguerrera: I think the breaking change aspect negates the whole point of properties. Might as well limit this feature to fields IMHO.

davkean: Also be a breaking change to compiler with an earlier compiler. Do we have many examples of that?

proposals/readonly-struct-init.md
nguerrera: This violates the CLI specification. A conforming implementation could reject the code as invalid. The spec just says that the runtime "need not" enforce this on unverifiable code, but not that it cannot enforce it. This is different from other cases, like overlapping object references, which are called out as unverifiable but valid. 

II.16.1.2
> In unverifiable code, the VES need not check whether initonly fields are mutated outside the constructors. The VES need not report any errors if a method changes the value of a constant. However, such code is not valid. 

I would also say this goes beyond other cases where I've supported deviating from the spec, like where it does not allow custom modifiers on type arguments. Those cases are spec bugs or omissions IMHO, where the consequences were likely unintended. Here, however, it is very clear that such code was never intended to be deemed valid.

proposals/readonly-struct-init.md
jaredpar: I will add the modifiers in. Generally when doing proposals I leave off accessibility modifiers to keep the samples concise. In this case though they're pretty relevant to the sample and I should add them.

proposals/readonly-struct-init.md
jaredpar: > What will these be marked as in metadata? Special named?

Today they are marked as `[CompilerGenerated]` and `DebuggerBrowsable(false)`. The `specialname` IL modifier isn't applied here. I don't believe C# does that for any field, but it does for methods. I don't know the history here and / or the implications of that. 

> This would mean that it be a breaking change to move from auto-property -> normal property. I think this should be called out.

I will add that. There are already subtle breaking change issues in this area. For example converting a prop to an auto-implemented property can break struct member-wise initialization. But this is a new one AFAICT and will call it out. 

> Might as well limit this feature to fields IMHO.

I considered that but feel like benefits outweigh the downsides here. Want to continue exploring it for now.

proposals/readonly-struct-init.md
jaredpar: Sigh ... 😦

proposals/readonly-struct-init.md
jaredpar: > We're relying on the runtime not blocking assignment after construction?

No. The fields are still marked as `readonly` hence they are blocked by default. Only compilers which understood the new verification rules would be able to participate in this. In theory at least. 

Note: this is mostly meant as an exercise of "if we wanted to verify this could we"? It's hard to be 100% certain of this until we have a working verifier. But for a number of features (ref locals in particular) we sketched out what we thought the rules would be to make sure a) we weren't ignoring this space and b) felt confident rules could be written on top of our feature that would be verifiable.

jaredpar:
@davkean, @nguerrera thank you for the feedback. Will get an updated document out sometime tomorrow.

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+var name = new Name { FirstName = "Jared", LastName = "Parsons" };
+
+// Emitted as 
+var name = new Name { <FirstName>__BackingField = "Jared", <LastName>__BackingField = "Parsons" };
+```
+
+This does mean that the feature will only apply to auto-implemented properties that are 
+re-compiled with a new compiler. The compiler today emits the fields as `private readonly` and 
+hence will be inaccessible to calling code. Upon recompile the backing fields will be emitted 
+as specified above.
```
jkotas: This will also have impact on the tools that reason about public surface (e.g. tooling used by the framework and documentation teams).

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
+# readonly object initializers 
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Below
+
+## Summary
+
+This feature will allow for assigning of `readonly` fields and get only auto-implemented properties
```
jkotas: Would it make sense to limit this to just assigning of `readonly` fields, and omit properties ?

You cannot freely convert back and forth between auto-implemented property with this. It makes the extra ceremony required by properties pointless and just unnecessary overhead.

jkotas: We may also need to review the framework and change auto-implemented properties affected by this to explicitly implemented properties. We do not necessary want to promise that all properties that happen to be implemented as auto-implemented properties today are going to be like that forever.

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+This makes it even easier to define usable `readonly struct` types in C#. The simplest form of
+declaring the fields now creates a fully usable value:
+
+``` csharp
+readonly struct Point { 
+    readonly int X;
+    readonly int Y; 
+}
+
+var p = new Point { X = 42; Y = 13 };
```
Neme12: should this be a comma instead?

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+
+    public C() { 
+        Global = this;
+    }
+}
+
+int SomeMethod() => Global.Field++;
+
+// The instance is observed by caller while mutations are still in progress.
+var c = new C() { Field = SomeMethod()}
```
Neme12: Can you please explain why this is a problem? It's not clear to me. So `SomeMethod` would be able to observe a different value of a readonly field, but that is already possible if we call `SomeMethod` from inside the constructor. Why is this form worse? We're essentially making it as if the object initializer was part of the constructor. That means it can mutate the fields and also observe the mutations.

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+
+class C { 
+    public readonly int Field;
+
+    public C() { 
+        Global = this;
+    }
+}
+
+int SomeMethod() => Global.Field++;
```
Neme12: `++` on a readonly field?

Neme12:
Based on the discussion, I don't think this is a good idea for 2 reasons:
1. As other have said, changing between regular and auto-implemented readonly properties would be a breaking change. This feels wrong. The name "auto-implemented" itself suggests that it should only affect the implementation.
2. It would also change the semantics of existing code with both readonly fields and auto-properties. Currently I can assume that noone is able to modify them outside this class.
```c#
    struct Restricted
    {
        public Restricted(int value)
        {
            if (value > 10)
                throw new Exception("invalid value!");

            this.Value = value;
        }

        public readonly int Value; // doesn't matter if this is a field or property

        // ok to assume Value will never be greater than 10
    }
```
Now I can easily break this invariant by using `new Restricted { Value = 50 }`. There has to be an opt-in for this, and the opt-in has to be more than "use a newer version of the compiler".

Neme12:
The fact that the opt-in needs to exist for both fields and properties would suggest it has to be a member modifier such as `data`. But that's a little ugly and would break the symmetry with mutable data. So I don't know how I feel about this proposal after all :confused:

HaloFour:
I don't think I need to reiterate my problems with the principal of this idea, and I'm sure that you've already read the back-and-forth on https://github.com/dotnet/csharplang/pull/1667, so I'll just link the start of my comments there for any reader who cares to know my opinion.

https://github.com/dotnet/csharplang/pull/1667#discussion_r198233743

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+    readonly int Y; 
+}
+
+var p = new Point { X = 42; Y = 13 };
+```
+
+This means authoring `readonly struct` types in C# is just as easy as creating the less
+perferred mutable `struct` types. There is no longer a need to define a constructor which simply 
+re-lists every field as a parameter and then assigns them one at a time. The definition of the 
+field itself allows provides a mechanism for initialization
```
jnm2: Missing end mark

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Detailed Design 
+
+### Language
+
+The language will extend the object initialization rules to allow for an accessible readonly field 
+or auto-implemented property to appear in the member initialization list. In such a context the 
+members will allow for assignment just as it would inside a constructor of the object.
```
jnm2: `the members will allow`…<code>just as *they* would</code>

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+members will allow for assignment just as it would inside a constructor of the object. 
+
+The following restrictions will apply to this extension:
+
+1. The type being instantiated must be a `struct` or `ref struct`. 
+2. The constructor used in the initializer cannot have any `out` or `ref` parameters.
+
+The (2) restriction exists because such a parameter could smuggle out `this` from the object 
+constructor. That would make a reference to the constructed object visible to the code calling
+`new` and hence allow it to observe mutations to the value.
```
jnm2: Is this fundamentally different from the existing situation where you can leak a reference to partially-constructed `this` from a constructor?

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+peverify implementation as it makes no real distinction between read only and mutable memory. 
+However it is still desirable for C#, and other languages, to emit IL that can be verified 
+with appropriate adjustments to the verification spec. Eventually allowing for projects like
+[ILVerify](https://github.com/dotnet/corert/tree/master/src/ILVerify) to take the place of 
+peverify today. 
+
+This rules could be adjusted to allow for mutation of a `readonly` field following an `.initobj` 
+or `.call .ctor` instruction if:
+
+1. The reciever of the call is a local
```
jnm2: Typo `reciever`

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+### Verification
+
+As is the case with several features added since C# 7.2 this feature will not pass any known 
+peverify implementation as it makes no real distinction between read only and mutable memory. 
+However it is still desirable for C#, and other languages, to emit IL that can be verified 
+with appropriate adjustments to the verification spec. Eventually allowing for projects like
+[ILVerify](https://github.com/dotnet/corert/tree/master/src/ILVerify) to take the place of 
+peverify today. 
+
+This rules could be adjusted to allow for mutation of a `readonly` field following an `.initobj`
```
jnm2: `These rules`?

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+peverify today. 
+
+This rules could be adjusted to allow for mutation of a `readonly` field following an `.initobj` 
+or `.call .ctor` instruction if:
+
+1. The reciever of the call is a local 
+1. The type of the reciever is a `struct` or `ref struct`
+1. There are no `ref` locals which refer to that local
+
+The `readonly` fields would remain mutable until one of the following occured:
```
jnm2: Typo `occured`

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+successful verification of the emitted code.
+
+## Considerations
+
+### Type Parameters
+
+The feature does not apply to type parameters constrained to a `struct`. There is just no case 
+where such a type would have a field (`readonly` or not) or an auto-implemented property. In the
+future if we allow for `struct` inheritance it's possible this could change and the decision 
+re-evaluated.
```
jnm2: `it's possible`…<code>the decision *could be* re-evaluated.</code>

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+
+A `struct` can't be smuggled out in such a way as it would require a direct copy to another 
+instance of the same type or by boxing to an interface / object. In either case it would be a 
+different value and not interesting here.
+
+### Future Considerations
+
+There is another proposal to allow for simple data types in the language via the `data` modifier.
+Such types would have auto-matic equality, hashing, ToString, etc ... Combined with this proposal
+and immutable data now has the simplest possible definition in C#:
```
jnm2: I don't think `auto-matic` is normally hyphenated.
`Combined with this proposal, immutable data` (no `and`)?

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+
+``` csharp
+public readonly data struct Point { 
+    public readonly int X;
+    public readonly int Y;
+}
+```
+
+This type now has initialazation via object initializers, equality, hashing and ToString. As the
```
jnm2: Typo `initialazation`

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
+# readonly object initializers 
+
+* [x] Proposed
```
jcouv: Nit: I suggest to move the checklist to the championed issue. Otherwise, it never gets updated.

proposals/readonly-struct-init.md
```diff
@@ -0,0 +1,161 @@
...
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Below
+
+## Summary
+
+This feature will allow for assigning of `readonly` fields and get only auto-implemented properties
+in object initializers when the type being constructed is a `struct` or `ref struct`.
+
+## Motivation
```
jcouv: I'm concerned about this feature breaking encapsulation. If the constructor has logic that safeguards the state of the object, then assigning to a field or property directly could violate it.
Note there is another proposal which is similar: ["data" types](https://github.com/dotnet/csharplang/pull/1667), which *requires* fields or properties to be initializable from a constructor initializer. But the choice is explicit (`data` modifier in type declaration).

If we move forward with the `data` types proposal, would you still push the present proposal forward?

proposals/readonly-struct-init.md
jaredpar: Will do.

proposals/readonly-struct-init.md
jaredpar: Note that you can't freely convert between auto-properties and regular properties today. There are situations where it causes breaking changes at the source level. This is indeed a new case though.

proposals/readonly-struct-init.md
jaredpar: >  If the constructor has logic that safeguards the state of the object, then assigning to a field or property directly could violate it.

How is this different than constructor logic that safeguards mutable fields?

proposals/readonly-struct-init.md
jaredpar: Indeed. I'm mixing up my F# and C# syntax here.

proposals/readonly-struct-init.md
jaredpar: > Why is this form worse? 

It's not worse. It's just demonstrating that you can observe the value changing after the constructor has completed. The cases you speak of apply from within the constructor only (and methods transitively called from it).

proposals/readonly-struct-init.md
jkotas: > you can't freely convert between auto-properties and regular properties today

Does it cause breaking changes in the public surface, or just in other places within the same compilation unit?

We have been freely converting between auto-properties and regular properties in the framework implementation a lot, and I do not remember ever encountering a case where it would cause a breaking change.

jaredpar: It can cause a break in the public surface area when it changes the shape of the backing fields. Consider:

``` csharp
struct S { 
  public int Value => 42; 
}
``` 

In version 2 if you change it to 

``` csharp
struct S {
  public int Value { get; } = 42;
}
```

This breaks any code that was doing member wise initialization of `S`.

jaredpar:
@Neme12 

> Currently I can assume that noone is able to modify them outside this class.

Not really though. A `struct` instance does not own it's storage. It's owned by the container of the `struct`. Unlike a `class` the contents of a `struct` can be pretty easily modified even when private (field offset is a typical example).

proposals/readonly-struct-init.md
jcouv: A constructor does not and cannot safeguard a (public) mutable field, just like a sieve does not and cannot hold water.

Currently, if you want to protect invariants with encapsulation, you make the mutable parts inaccessible or you put additional logic in the setters. But the present proposal opens up an accessible and unguarded setter. That worries me.

tannergooding: The contract of mutability. Constructor logic on mutable fields can be broken, and if you want that logic generally applicable you need to also place it in the property setter (and should not expose the field).

While `readonly` properties/fields today are believed to only be assignable from the constructor, so the logic only needs to exist there (such as saying that an integer must be positive).

tannergooding:
> Not really though. A struct instance does not own it's storage. It's owned by the container of the struct. Unlike a class the contents of a struct can be pretty easily modified even when private (field offset is a typical example).

Doesn't this require either `ExplicitLayout` or `Unsafe` code, both of which are `undefined` behavior when dealing with `readonly`

proposals/readonly-struct-init.md
jaredpar: I'm really not too worried about publicly exposed fields being mutated. Structs don't own their storage and hence don't have the same guarantees as classes do for their state. It's easy to mutate struct contents today without this feature.

jkotas: The same is true for classes. Mutating struct state to violate its internal invariant is about as hard as mutating class state to violate its internal invariant.

tannergooding:
> The behavior may be undefined in spec but in reality the behavior is well defined, understood and relied upon. It also doesn't necessarily require these. You can always mutate a struct by directly assigning over it. This means it's pretty easy to get a "validated" struct into an inconsistent state via states like tearing.

The same can be said for a lot of native code, until someone turns on the "make my code safe" instead of the "make my code fast" switch, and your `readonly` data is now placed in `readonly` memory and throws an `AV` when you try to write it 😄 

I think the general point is: If I recompile my code, the contract that I had in the previous version of the compiler is broken.

The contract was:
* "The readonly property `Value`, outside of `unsafe` code, is only assignable from the constructor."

The contract would now be:
* "The readonly property `Value`, outside of `unsafe` code, is assignable from the constructor or from an object initializer statement

As such, recompiling silently breaks a contract that was existed since v1.0 of the language

A new contextual keyword/attribute would alleviate that concern

proposals/readonly-struct-init.md
nguerrera: >> We're relying on the runtime not blocking assignment after construction?

> No. The fields are still marked as readonly hence they are blocked by default

I think David meant that the codegen for the initialization would be relying on the fact that the runtime doesn't enforce initonly rules (unless the verifier is required to run).

And that is relying on an implementation detail of the runtime since the spec says that code that does this is "not valid".  It also says that the runtime "need not" enforce it. but if this change goes in, then the runtime "must not" enforce it.

I think this is worth calling out: we wouldn't just be relying on code that is currently unverifiable, we'd be relying on code that is currently **invalid**.

jaredpar:
@tannergooding 

> As such, recompiling silently breaks a contract that was existed since v1.0 of the language

You're mixing up arguments here. Auto-implemented properties were added post v1.0 (IIRC 4.0). The idea of writing a `readonly` field post ctor *via assignment* is a contract from v1.0 but does not necessitate recompilation.

Issue #191, opened 2017-02-26
Champion "Compiler intrinsics (calli, ldftn, etc)"


gafter:
- [X] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

See also
- Proposal: https://github.com/dotnet/csharplang/blob/master/proposals/intrinsics.md
- https://github.com/dotnet/csharplang/blob/master/meetings/2017/LDM-2017-01-11.md
- https://github.com/dotnet/csharplang/blob/master/meetings/2018/LDM-2018-09-05.md#compiler-intrinsics
- https://github.com/dotnet/csharplang/blob/master/meetings/2018/LDM-2018-09-05.md#compiler-intrinsics
- https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-01-23.md

Unknown6656:
What are the advantages of compiler intrinsics for specific IL-opcodes instead of directly supporting inline-IL?

orthoxerox:
@Unknown6656 you don't have to modify the parser, only the emitter, so you get to reuse type checking.

tannergooding:
At least for `calli`, the `static delegate` proposal (https://github.com/dotnet/csharplang/issues/302) would cover a way of using it outside of intrinsic support. However, the intrinsic support would still be useful for cases where users can't or don't want to change their API surface.

MadsTorgersen:
Need more motivation to prioritize this

MichalStrehovsky:
> Need more motivation to prioritize this

@MadsTorgersen What kind of motivation? I can buy beers to anyone who implements it if they need to motivate that way. There's plenty of use cases for this - see all the people referencing the issue at dotnet/roslyn#11475.

Could implementing this be an up for grabs issue in the Roslyn repo?

4creators:
@MichalStrehovsky @MadsTorgersen 
Without any beer I could volunteer to implement it, however, due to my time limitations it could be only free time job.

jaredpar:
> Could implementing this be an up for grabs issue in the Roslyn repo?

Unlikely. We are still uncertain this is the design we want to take. I personally have a lot of reservations about the approach.

4creators:
@jaredpar 
Can you describe what kind of reservations C# team has about this proposal?

GeirGrusom:
IL as a language feature is a little bit of a cop-out I think.

[...]

Issue #302, opened 2017-03-21
Champion "Static Delegates"


tannergooding:
Superseded by https://github.com/dotnet/csharplang/blob/master/proposals/csharp-9.0/function-pointers.md

- [x] Proposal added: [proposals/static-delegates.md](https://github.com/dotnet/csharplang/blob/master/proposals/static-delegates.md)
- [x] Discussed in LDM (https://github.com/dotnet/csharplang/blob/master/meetings/2017/LDM-2017-05-16.md#static-delegates)
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

tannergooding:
FYI. @gafter. I think this one was skipped over after the proposal was merged. It is being championed by @jaredpar

AlgorithmsAreCool:
@jaredpar I took a look at the proposal and maybe this is beyond my understanding, but if this is geared at  interop i'm not really clear what the difference or synergy between this and an `extern` function is. Maybe that isn't even a sensible question...

DavidArno:
Could you please add some examples of how this might be used, and it's limitations? For example, in the proposal, it's stated that "Static Delegates would not work with existing APIs that use regular delegates". I'm reading that as I couldn't use this to improve the performance of:
```cs
int Foo(int p, Func<int, int> f) => f(p);
```
But could static delegates be used to provide a better performing version of `Foo`? If so, what would the signature look like?

tannergooding:
@AlgorithmsAreCool, think of this as a type of `value delegate` (in the same vein as `ValueTuple` compared to `Tuple`).

It can be used for interop, but it can also be used as a type of lightweight delegate.

tannergooding:
@DavidArno. It wouldn't be useable as a drop in replacement in the same way that you can't just use `ValueTuple` anywhere that you can use `Tuple`. One always exists on the heap, and the other could potentially exist on the stack.

You could convert from one to the other, so it could be 'used', but it would incur the cost of the allocation/etc.

AlgorithmsAreCool:
Could you provide a little code snippet showing the creation, passing (to a function) and invocation of a static delegate?

Also why is it "static". Wouldn't `struct delegate` be a better moniker?

tannergooding:
`static delegate` was just the name @jaredpar gave them in the original issue: https://github.com/dotnet/csharplang/issues/80. But, as everyone knows, computer programming only has two hard problems ;)

[...]

PR #1685, opened 2018-07-02
Compiler intrinsics

jaredpar:
This is an alternate proposal for the compiler intrinsicts feature:

https://github.com/dotnet/csharplang/issues/191
https://github.com/dotnet/roslyn/issues/11475

This alternate design proposal comes after reviewing a prototype implementation
of the original proposal by @msjabby as well as the use throughout a significant
code base.

This design was done with significant input from @mjsabby, @tmat and @jkotas.

### DIFF
--- a/intrinsics.md
+++ b/intrinsics.md
@@ -1,0 +1,183 @@
+# Compiler Intrinsics
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Below
+
+## Summary
+
+This proposal provides language constructs that expose low level IL opcodes that cannot currently
+be accessed efficiently, or at all: `ldftn`, `ldvirtftn`, `ldtoken` and `calli`. These low level op 
+codes can be important in high performance code and developers need an efficient way to access 
+them.
+
+## Motivation
+
+The motivations and background for this feature are described in the following isssue (as is a 
+potential implementation of the feature): 
+
+https://github.com/dotnet/csharplang/issues/191
+
+This alternate design proposal comes after reviewing a prototype implementation of the original
+proposal by @msjabby as well as the use throughout a significant code base. This design was done 
+with significant input from @mjsabby, @tmat and @jkotas.
+
+## Detailed Design 
+
+### Allow address of to target methods
+
+Method groups will now be allowed as arguments to an address-of expression. The type of such an 
+expression will be `void*`. 
+
+``` csharp
+class Util { 
+    public static void Log() { } 
+}
+
+// ldftn Util.Log
+void* ptr = &Util.Log; 
+```
+
+Given there is no delegate conversion here the only mechanism for filtering members in the method
+group is by static / instance access. If that cannot distinguish the members then a compile time
+error will occur.
+
+``` csharp
+class Util { 
+    public void Log() { } 
+    public void Log(string p1) { } 
+    public static void Log(int i) { };
+}
+
+unsafe {
+    // Error: Method group Log has more than one applicable candidate.
+    void* ptr1 = &Log; 
+
+    // Okay: only one static member to consider here.
+    void* ptr2 = &Util.Log;
+}
+```
+
+The addressof expression in this context will be implemented in the following manner:
+
+- ldftn: when the method is static 
+- ldvirtftn: when the method is an instance method.
+
+Restrictions of this feature:
+
+- Instance methods can only be specified when using an invocation expression on a value
+- Local functions cannot be used in `&`. The implementation details of these methods are
+deliberately not specified by the language. This includes whether they are static vs. instance or
+exactly what signature they are emitted with.
+
+### handleof
+
+The `handleof` contextual keyword will translate a field, member or type into their equivalent 
+`RuntimeHandle` using the `ldtoken` instruction.
+
+The arguments to `handleof` are identical to `nameof`. It must be a simple name, qualified name, 
+member access, base access with a specified member, or this access with a specified member. The 
+argument expression identifies a code definition, but it is never evaluated.
+
+The `handleof` expression is evaluated at runtime and has a return type of `RuntimeHandle`. This 
+can be executed in safe code as well as unsafe. 
+
+``` 
+RuntimeHandle stringHandle = handleof(string);
+```
+
+Restrictions of this feature:
+
+- Properties cannot be used in a `handleof` expression.
+- The `handleof` expression cannot be used when there is an existing `handelof` name in scope. For 
+example a type, namespace, etc ...
+
+### calli
+
+The compiler will add support for a new type of `extern` function that efficients translates into
+a `.calli` instruction. The exten attribute will be marked with an attribute of the following
+shape:
+
+``` csharp
+[AttributeUsage(AttributeTargets.Method)]
+public sealed class CallIndirectAttribute : Attribute
+{
+    public CallingConvention CallingConvention { get; set; }
+    public CallIndirectAttribute(CallingConvention callingConvention)
+    {
+        CallingConvention = callingConvention;
+    }
+}
+```
+
+This allows developers to define methods in the following form:
+
+``` csharp
+[CallIndirect(CallingConvention.Cdecl)]
+static extern int MapValue(string s, void *ptr);
+
+unsafe {
+    var i = MapValue("42", &int.Parse);
+    Console.WriteLine(i);
+}
+```
+
+## Considerations
+
+### Disambiguating method groups
+
+There was some discussion around features that would make it easier to disambiguate method groups
+passed to an addressof expression. For instance potentially adding signature elements to the 
+syntax:
+
+``` csharp
+class Util {
+    public static void Log() { ... }
+    public static void Log(string) { ... }
+}
+
+unsafe {
+    // Error: ambiguous Log
+    void *ptr1 = &Util.Log;
+
+    // Use Util.Log();
+    void *ptr2 = &Util.Log();
+}
+```
+
+This was rejected because a compelling case could not be made nor could a simple syntax be 
+envisioned here. Also there is a fairly straight forward work around: simple define another 
+method, possible local function, that is unambiguous and uses C# code to call into the 
+desired function. 
+
+``` csharp
+unsafe { 
+    static void LocalLog() => Util.Log();
+    void* ptr = &LocalLog;
+}
+```
+
+### LoadTypeTokenInt32
+
+The original proposal allowed for metadata tokens to be loaded as `int` values at compile time. 
+Essentially have `tokenof` that has the same arguments as `handleof` but is evaluated at 
+compile time to an `int` constant. 
+
+This was rejected as it causes significant problem for IL rewrites (of which .NET has many). Such 
+rewriters often manipulate the metadata tables in a way that could invalidate these values. There 
+is no reasonable way for such rewriters to update these values when they are stored as simple 
+`int` values.
+
+The underlying idea of having an opaque handle for metadata entries will continue to be explored 
+by the runtime team. 
+
+## Future Considerations
+
+### static local functions
+
+This refers to [the proposal](https://github.com/dotnet/csharplang/issues/1565) to allow the 
+`static` modifier on local functions. Such a function would be guaranteed to be emitted as 
+`static` and with the exact signature specified in source code. Such a function should be a valid
+argument to `&` as it contains none of the problems local functions have today.
+
--- a/intrinsics.md
+++ b/intrinsics.md
@@ -1,10 +1,5 @@
 # Compiler Intrinsics
 
-* [x] Proposed
-* [ ] Prototype: Not Started
-* [ ] Implementation: Not Started
-* [ ] Specification: Below
-
 ## Summary
 
 This proposal provides language constructs that expose low level IL opcodes that cannot currently
@@ -14,7 +9,7 @@ them.
 
 ## Motivation
 
-The motivations and background for this feature are described in the following isssue (as is a 
+The motivations and background for this feature are described in the following issue (as is a 
 potential implementation of the feature): 
 
 https://github.com/dotnet/csharplang/issues/191
@@ -61,8 +56,8 @@ unsafe {
 
 The addressof expression in this context will be implemented in the following manner:
 
-- ldftn: when the method is static 
-- ldvirtftn: when the method is an instance method.
+- ldftn: when the method is non-virtual.
+- ldvirtftn: when the method is virtual.
 
 Restrictions of this feature:
 
@@ -74,7 +69,12 @@ exactly what signature they are emitted with.
 ### handleof
 
 The `handleof` contextual keyword will translate a field, member or type into their equivalent 
-`RuntimeHandle` using the `ldtoken` instruction.
+`RuntimeHandle` type using the `ldtoken` instruction. The exact type of the expression will 
+depend on the kind of the name in `handleof`:
+
+- field: `RuntimeFieldHandle`
+- type: `RuntimeTypeHandle`
+- method: `RuntimeMethodHandle`
 
 The arguments to `handleof` are identical to `nameof`. It must be a simple name, qualified name, 
 member access, base access with a specified member, or this access with a specified member. The 
@@ -90,20 +90,20 @@ RuntimeHandle stringHandle = handleof(string);
 Restrictions of this feature:
 
 - Properties cannot be used in a `handleof` expression.
-- The `handleof` expression cannot be used when there is an existing `handelof` name in scope. For 
+- The `handleof` expression cannot be used when there is an existing `handleof` name in scope. For 
 example a type, namespace, etc ...
 
 ### calli
 
-The compiler will add support for a new type of `extern` function that efficients translates into
-a `.calli` instruction. The exten attribute will be marked with an attribute of the following
+The compiler will add support for a new type of `extern` function that efficiently translates into
+a `.calli` instruction. The extern attribute will be marked with an attribute of the following
 shape:
 
 ``` csharp
 [AttributeUsage(AttributeTargets.Method)]
 public sealed class CallIndirectAttribute : Attribute
 {
-    public CallingConvention CallingConvention { get; set; }
+    public CallingConvention CallingConvention { get; }
     public CallIndirectAttribute(CallingConvention callingConvention)
     {
         CallingConvention = callingConvention;
@@ -123,12 +123,25 @@ unsafe {
 }
 ```
 
+Restrictions on the method which has the `CallIndirect` attribute applied:
+
+- Cannot have a `DllImport` attribute.
+- Cannot be generic.
+
+## Open Issuess
+
+### CallingConvention
+
+The `CallIndirectAttribute` as designed uses the `CallingConvention` enum which lacks an entry for
+managed calling conventions. The enum either needs to be extended to include this calling convention
+or the attribute needs to take a different approach.
+
 ## Considerations
 
 ### Disambiguating method groups
 
 There was some discussion around features that would make it easier to disambiguate method groups
-passed to an addressof expression. For instance potentially adding signature elements to the 
+passed to an address-of expression. For instance potentially adding signature elements to the 
 syntax:
 
 ``` csharp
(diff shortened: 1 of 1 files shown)


intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+
+The compiler will add support for a new type of `extern` function that efficients translates into
+a `.calli` instruction. The exten attribute will be marked with an attribute of the following
+shape:
+
+``` csharp
+[AttributeUsage(AttributeTargets.Method)]
+public sealed class CallIndirectAttribute : Attribute
+{
+    public CallingConvention CallingConvention { get; set; }
```
nguerrera: Should be read-only. Design guidelines say that attribute arguments should be either positional or named, not both.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+    {
+        CallingConvention = callingConvention;
+    }
+}
+```
+
+This allows developers to define methods in the following form:
+
+``` csharp
+[CallIndirect(CallingConvention.Cdecl)]
```
jkotas: This enum does not have a value for regular managed calling convention. (@mjsabby use case needs that). 
I think we may need to have two attributes: 
- CallIndirectAttribute for regular managed calls that does not have CallingConvention. 
- UnmanagedCallIindirectAttribute for unmanaged calls that does have CallingConvention 

An alternative is to add a value to for the managed calling convention to CallingConvention enum. I have rejected it because of the new value would be invalid in all existing APIs that use CallingConvention enum today. All these APIs are for unmanaged calls.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+- ldftn: when the method is static 
+- ldvirtftn: when the method is an instance method.
+
+Restrictions of this feature:
+
+- Instance methods can only be specified when using an invocation expression on a value
+- Local functions cannot be used in `&`. The implementation details of these methods are
+deliberately not specified by the language. This includes whether they are static vs. instance or
+exactly what signature they are emitted with.
+
```
jkotas: We should productize the NativeCallableAttribute that will work together with this. This proposal together with NativeCallableAttribute will deliver complete solution for low-level managed/unmanaged interop for function pointers:
- address-of + NativeCallableAttribute can be used the get unmanager pointer of method implemented in C# 
- Calli indirect can be used to call unmanaged pointer

.NET Core supports the NativeCallableAttribute attribute internally, but the feature was not exposed in public contracts and documented.

@jeffschwMSFT @AaronRobinsonMSFT

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+    // Okay: only one static member to consider here.
+    void* ptr2 = &Util.Log;
+}
+```
+
+The addressof expression in this context will be implemented in the following manner:
+
+- ldftn: when the method is static 
+- ldvirtftn: when the method is an instance method.
```
mjsabby: Would we allow the emission of ldftn if we know that the method is non-virtual?

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+## Summary
+
+This proposal provides language constructs that expose low level IL opcodes that cannot currently
+be accessed efficiently, or at all: `ldftn`, `ldvirtftn`, `ldtoken` and `calli`. These low level op 
+codes can be important in high performance code and developers need an efficient way to access 
+them.
+
+## Motivation
+
+The motivations and background for this feature are described in the following isssue (as is a
```
jnm2: Typo `isssue`

sylveon:
How would the call indirect attribute mix with P/Invoke?

jkotas:
> How would the call indirect attribute mix with P/Invoke?

DllImportAttribute and CallIndirectAttribute do not mix. Having both should produce compile error.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
+# Compiler Intrinsics
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Below
```
jaredpar: Will move this to the issue only.

intrinsics.md
jaredpar: C# always uses `callvirt` instead of `call` and I was trying to mirror that here. I don't have any particular reason beyond that.

intrinsics.md
jaredpar: What does this attribute do today in the internal support?

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+This was rejected because a compelling case could not be made nor could a simple syntax be 
+envisioned here. Also there is a fairly straight forward work around: simple define another 
+method, possible local function, that is unambiguous and uses C# code to call into the 
+desired function. 
+
+``` csharp
+unsafe { 
+    static void LocalLog() => Util.Log();
```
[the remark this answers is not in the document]
jaredpar: It is. I was trying to keep the sample concise. I will make it clearer by using a plain old static.

intrinsics.md
jkotas: It adds unmanaged->managed->unmanaged transition to the method prolog/epilog. 

Marking the method with NativeCallableAttribute and taking its address is more efficient low-level equivalent of Marshal.GetFunctionPointerForDelegate. It avoids the delegate wrapping and unwrapping overhead. More details are in description for https://github.com/dotnet/coreclr/pull/1566

tannergooding:
The [Static Delegates](https://github.com/dotnet/csharplang/issues/302) proposal would be fully emulatable in managed code if this was implemented.

Perhaps it would be worth just implementing them simultaneously (assuming the additional overhead wasn't too great)?

jaredpar:
@tannergooding 

static delegates require further runtime work to be done in a safe fashion. There needs to be some way to prevent assembly unloading when a static delegate points inside it. This proposal side steps this problem by forcing all such operations to be `unsafe`.

tannergooding:
@jaredpar, you could just make `static delegates` require unsafe in v1 and then relax that when the runtime support gets in 😉 (but it's also no big deal, since I can just modify my binding generator to emulate the functionality for the time being)

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+
+- Instance methods can only be specified when using an invocation expression on a value
+- Local functions cannot be used in `&`. The implementation details of these methods are
+deliberately not specified by the language. This includes whether they are static vs. instance or
+exactly what signature they are emitted with.
+
+### handleof
+
+The `handleof` contextual keyword will translate a field, member or type into their equivalent 
+`RuntimeHandle` using the `ldtoken` instruction.
```
AaronRobinsonMSFT: Just for clarity - I was not familiar with `RuntimeHandle` - it technically is `RuntimeXXXXHandle` right? Want to make sure and not assume something new will be created. It might be beneficial to enumerate all of them at least once and then use the `RuntimeHandle` as the placeholder going forward.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+``` csharp
+class Util { 
+    public static void Log() { } 
+}
+
+// ldftn Util.Log
+void* ptr = &Util.Log; 
+```
+
+Given there is no delegate conversion here the only mechanism for filtering members in the method
```
333fred: Do we think that this is going to be a significant bar for existing libraries to use this feature, if they've already got a number of method overloads that they want to distinguish between?

333fred: Ah, I see you have a section on that below, hadn't gotten there yet.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+unsafe {
+    // Error: ambiguous Log
+    void *ptr1 = &Util.Log;
+
+    // Use Util.Log();
+    void *ptr2 = &Util.Log();
+}
+```
+
+This was rejected because a compelling case could not be made nor could a simple syntax be
```
333fred: Did we consider allowing a "cast"? ie, something like `&((Action<string>)Util.Log)`?

intrinsics.md
jaredpar: Correct.  I will clarify that here.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+    Console.WriteLine(i);
+}
+```
+
+## Considerations
+
+### Disambiguating method groups
+
+There was some discussion around features that would make it easier to disambiguate method groups
+passed to an addressof expression. For instance potentially adding signature elements to the
```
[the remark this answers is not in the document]
jaredpar: It's not code though, it's a language concept.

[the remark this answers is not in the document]
tannergooding: I think some people use it to highlight keywords as well 😄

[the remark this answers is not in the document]
tannergooding: (whether language keywords or words that are key to the sentence)

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+        CallingConvention = callingConvention;
+    }
+}
+```
+
+This allows developers to define methods in the following form:
+
+``` csharp
+[CallIndirect(CallingConvention.Cdecl)]
+static extern int MapValue(string s, void *ptr);
```
MichalStrehovsky: Is the C# compiler going to emit this method into the metadata, or will it disappear at source compilation time? The runtime typically won't load a type that has an `extern` method without a `DllImport` or some other special marking.

What happens if I try to reference this method in a non-call context (try to construct a delegate to this or get a RuntimeHandle of it)?

mjsabby: The current implementation generates method bodies for this reason, and at least for the managed calling convention the calli gets inlined. Or would you like roslyn to prevent the cases you're mentioning and not emit metadata?

intrinsics.md
MichalStrehovsky: Having a way to do both `ldvirtftn` and `ldftn` would let library writers determine whether a method has been overridden by a subclass. [This has been used](https://github.com/dotnet/coreclr/blob/95ae6621b1c7f176b518ad819f6eccd469e91ea1/src/System.Private.CoreLib/src/System/IO/Stream.CoreCLR.cs#L12) in performance optimization in the past, but it's unavailable to code that doesn't live in the runtime.

intrinsics.md
MichalStrehovsky: I was just wondering if it's going to be actually emitted as `extern`, or `extern` was just a spare keyword.

intrinsics.md
jkotas: BTW: There is nothing that says that the runtime has to return the same pointer from `ldvirtftn` all the time. .NET Framework / Core do actually return different pointers for the same method. 

If the pointer equality is used as an optimization, comparing function pointers will work but reliably.

If the pointer equality is used for correctness, comparing function pointers is not going to work well. (Your HasOverriddenBeginEndRead example is used as both performance optimization but also to achieve correctness.)

intrinsics.md
jaredpar: @MichalStrehovsky the `extern` keyword here indicates the method will not have a body that is supplied by the developer. Instead it will be supplied by the compiler, or the runtime depending on how it's omitted. In this case as @mjsabby indicates we'll be putting the calli implementation in the body here. 

This is still being fleshed out a bit and is more compiler centric vs. language centric. Hence I didn't dive too deep into that here.

intrinsics.md
MichalStrehovsky: > There is nothing that says that the runtime has to return the same pointer

We could add public API to compare function pointers for equality, couldn't we? Then this optimization would become available to library developers outside of CoreLib (provided we get a `ldvirtftn`/`ldftn` distinction in this proposal).

mjsabby: If we think we can get the CoreFX API approval then it makes sense to be able to distinguish. It won't be a lot more code I can't imagine. @jaredpar what do you think?

jaredpar: @mjsabby sorry I'm not quite sure what the suggestion is here. Is it to change which IL opcode we used based on whether the method is virtual vs. non-virtual?

MichalStrehovsky: > Is it to change which IL opcode we used based on whether the method is virtual vs. non-virtual?

The ask is to provide a way to control whether a `ldftn` or `ldvirtftn` is generated. I somewhat hijacked @mjsabby's thread.

jaredpar: Gotcha. I doubt we'd go much further than giving guarantees like the ones already in the proposal. It balances keeping the language pretty simple and gives the developers the lever necessary to get one or the other in the cases where they need it.

On the subject of whether we should use `ldvirtftn` vs `ldftn` on instance methods I'm very open to discussion here. This defaults were chosen just because they matched an existing pattern. Not for a reason specific to this scenario.

intrinsics.md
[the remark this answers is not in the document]
AaronRobinsonMSFT: If this is a language concept, then it should be hyphenated "address-of" (i.e. 'addressof' is not a word).

intrinsics.md
MichalStrehovsky: > On the subject of whether we should use `ldvirtftn` vs `ldftn` on instance methods I'm very open to discussion here

The difference between these two for non-virtual methods would only be about whether we get a `NullReferenceException` for a `null` `this`, right? Since the `this` part is not going to be captured in the returned function pointer, maybe the `&` operator shouldn't care what `this` is for non-virtual methods (and it should use `ldftn`).

It might be convenient to be able to write

```csharp
void* logPtr = &default(Util).Log;
```

jaredpar: @MichalStrehovsky 

> The difference between these two for non-virtual methods would only be about whether we get a NullReferenceException for a null this, right? 

For `call` and `callvirt` this also affects binary compatibility. Imagine a method in a utility library changes from non-virtual to virtual and a consuming assembly isn't recompiled. In C# the code in the consuming assembly will automatically begin calling the method as a virtual. 

I'm unsure if `ldftn` and `ldvirtftn` behave in a similar way. Been trying to track that down but haven't found any good dos on it. Assuming they do then it would be a consideration when deciding what code the address-of `&` operator generated.

MichalStrehovsky: > For `call` and `callvirt` this also affects binary compatibility

Isn't this just a side effect, as opposed from being a conscious design choice? When the compiler knows the `this` is going to non-null, it will happily generate a `call`:

```csharp
    public class Base
    {
        public void Frob()
        {
        }
    }

    class Derived : Base
    {
        public void Call()
        {
            // Emits call
            Frob();
        }

        public static void CallStatic(Derived d)
        {
            // Emits callvirt
            d.Frob();
        }
    }
```

jaredpar: > Isn't this just a side effect, as opposed from being a conscious design choice? 

It's hard to say whether in C# 2.0 (where callvirt became the default over call) whether binary compat was a concern. At this point though it's a known quantity of the language that we consider when making changes. It's not an unbreakable change by any means but it's a consideration we make.

jaredpar: I think we're over focusing on the callvirt thing here a bit. It's not a big blocking problem so much as it is a box that needs to be checked. I don't suspect there would be any substantial push back from LDM if we make a different decision for how `ldftn` / `ldvirtftn` were chosen.

jeffschwMSFT:
@luqunl

intrinsics.md
luqunl: We should dis-allow DllImport method with CallIndirect attribute.

intrinsics.md
```diff
@@ -0,0 +1,183 @@
...
+This alternate design proposal comes after reviewing a prototype implementation of the original
+proposal by @msjabby as well as the use throughout a significant code base. This design was done 
+with significant input from @mjsabby, @tmat and @jkotas.
+
+## Detailed Design 
+
+### Allow address of to target methods
+
+Method groups will now be allowed as arguments to an address-of expression. The type of such an 
+expression will be `void*`.
```
luqunl: curious:  why do we prefer "void *"  instead of "IntPtr"? If we use IntPtr, we don't need to use unsafe part?

intrinsics.md
luqunl: For calli, Will we support generic method? also named arguments?
such as Int MapValue<T1, T2>(T1 a1, T2 a2, void *p)?

intrinsics.md
luqunl: >Marking the method with NativeCallableAttribute and taking its address is more efficient low-level equivalent of Marshal.GetFunctionPointerForDelegate

They are not exactly equivalent. The first one doesn't support marshalling(you need to use blittable types), the second one support marshalling.

intrinsics.md
luqunl: > add a value to for the managed calling convention to CallingConvention enum

Add a new value in CallingConvention seems better than use two attributes. simpler and easy to understand.

AaronRobinsonMSFT: @luqunl I agree in principle that two attributes are more annoying - however, I agree with @jkotas that adding it to the enum is less than ideal.

Another approach that would work from the API perspective would be to make the default constructor for `CallIndirectAttribute` specify a sentinel value of sorts that indicates managed convention - a default that probably makes sense - and another constructor that takes a `CallingConvention` enum value. In this case the default is 'managed to managed' and if a calling convention is used the intent is to leave the managed world so let us know what is being called.

The nice aspect about this is that the sentinel value can be anything and doesn't require breaking the API boundary.

intrinsics.md
jaredpar: This is not a safe operation which is part of the motivation for returning a `void*` here.

intrinsics.md
jaredpar: Unsure about supporting a generic method. As long as it's supported in the runtime I think we can make it work fine in the language. 

As for named arguments: yes. Likely optional as well.

jkotas: Generic calli is not supported by the runtime.

MichalStrehovsky: I think @luqunl meant support for this:

```
.assembly foo { }

.method static !!0 Calli<TResult>(native int)
{
    ldarg.0
    calli !!0 ()
    ret
}

.method static int32 Test()
{
    ldc.i4 42
    ret
}

.method static int32 Main()
{
    .entrypoint
    ldftn int32 Test()
    call !!0 Calli<int32>(native int)
    ret
}
```

Rather than generating a `calli` whose signature is generic (something like `calli !!0<!!0>()`).

This seems to work fine.

Allowing this on generic methods would save quite a bit of typing so I would vote for not blocking this.

jkotas: I agree that it would nice to allow it - if we do the matching work in CoreCLR and other runtimes to support it. 

- For managed calling convention, it is fine and works.
- For unmanaged calling convention, it is going to hit https://github.com/dotnet/coreclr/issues/14524 in CoreCLR.

luqunl: > Allowing this on generic methods would save quite a bit of typing so I would vote for not blocking this.

Yes. Interop would love to support this. Currently MCG and System.Private.Interop will define/generic a lot calli helper methods, similar to the following

`
		internal static void HasThisCall__9<TArg0>( `
`					object __this, `
`					global::System.IntPtr pfn, `
`					TArg0 arg0)`
`		{`
`			// This method is implemented elsewhere in the toolchain`
`		}`

jkotas: > would save quite a bit of typing 

I do not think that the extra typing is a problem here. This is a low-level feature where saving typing is not important. Also, this would be typically auto-generated, not typed by humans. 

I would look at the potential support for generic unmanaged calli as second phase of this feature. Generic unmanaged calli would makes this feature significantly more complex because of it would dependent on a runtime feature that does not exist yet.

jaredpar: > I would look at the potential support for generic unmanaged calli as second phase of this feature.

FWIW: expanding this to include generic support at a later date is basically a bug fix level change. I agree a second phase is a good landing place for this. 

>  Generic unmanaged calli would make this feature significantly more complex because of it would dependent on a runtime feature that does not exist yet

It adds complexity yes but not significant complexity to the compiler. The compiler is already setup to look at `RuntimeFeatures` to enable / disable items we know depend on runtime changes. As long as the fix in CoreClr included an entry here we could depend on it. 

That being said I think we can move forward for now with generic disabled and revisit once we make a bit more progress. It won't really need language buy off at that point, just keyboard time to get the change in.

intrinsics.md
jaredpar: I will add a section explaining exactly why this is unsafe.

intrinsics.md
mjsabby: Alright, LGTM.

intrinsics.md
jaredpar: Chatted with @AaronRobinsonMSFT and I better understand how this could interact with the feature. Agree there are benefits. Seems easy to add if the runtime decides to productize this. Will add this to the "future consideration" section.

jaredpar:
Updated PR based on feedback.

Issue #1467, opened 2018-04-19
A comprehensive list of pattern-based language features?


tuespetre:
I am fascinated with the way some language features are interface-based whereas others are simply pattern-based.

Here is the list I have so far:

- `Main` (thanks @jnm2 and @yaakov-h)
- [foreach iterators](https://github.com/dotnet/csharplang/blob/master/spec/statements.md#the-foreach-statement)
- [query expressions](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#the-query-expression-pattern)
- [awaitable expressions (task types)](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#awaitable-expressions)
- [deconstruction](https://github.com/dotnet/roslyn/blob/master/docs/features/deconstruction.md) (thanks @yaakov-h and @zippec)
- [fixed statements](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.3/pattern-based-fixed.md) (thanks @jnm2)
- [async methods (builder types)](https://github.com/dotnet/roslyn/blob/master/docs/features/task-types.md#builder-type) (thanks @jnm2)
- [collection initializers](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#collection-initializers) (thanks @ufcpp)
- [using (very new/recent)](https://github.com/jaredpar/csharplang/blob/d2ce4cc3e17708e7e1d062bf40da0901a744fa3b/proposals/using.md) (thanks @jnm2)

Are there any more than this, and is there some kind of unambiguous term that can be used to refer to this type of construct as a whole (something that 'Googles better than' the term 'c# language pattern'?) Would it be useful in general if there were a section in the spec devoted to listing such constructs (even the interface-based ones) and discussing the merits of pattern-versus-interface-based language constructs?

yaakov-h:
Deconstruction is another.

zippec:
@tuespetre Tuple deconstruction is also pattern-based

HaloFour:
Depending on what you consider a "pattern" you could probably include tuples and extension methods.

Tuples require that the container type have the name `System.ValueTuple<...>` with a specific shape, specifically a constructor accepting the elements and those elements as public fields.  But it doesn't care where the type lies or how it's implemented beyond those requirements.

[...]

Issue #114, opened 2017-02-15
Proposal:  Implicitly scoped using statement


HaloFour:
See https://github.com/dotnet/roslyn/issues/5881 and https://github.com/dotnet/roslyn/issues/181

## Motivation (TL;DR edition):

Allow `using` keyword to be used to declare a variable that will be disposed at the end of the current block.  This avoids creating a new block with an additional level of indentation:

### Before:
```cs
public void Foo() {
    using (var connection = new SqlConnection(connectionString)) {
        connection.Open();
        using (var command = connection.CreateCommand()) {
            command.CommandText = "SELECT FOO FROM BAR";
            using (var reader = command.ExecuteReader()) {
                while (reader.Read()) {
                    ProcessRecord(reader);
                }
            }
        }
    }
}
```

### After:
```cs
public void Foo() {
    using var connection = new SqlConnection(connectionString));
    connection.Open();
    using var command = connection.CreateCommand());
    command.CommandText = "SELECT FOO FROM BAR";
    using var reader = command.ExecuteReader());
    while (reader.Read()) {
        ProcessRecord(reader);
    }
}
```

As an implementation detail, the variables would be disposed in the reverse order in which they are declared.  So given the above code, the order in which the resources would be disposed is `reader`, `command` and then `connection`.

scottdorman:
I think this has a potential to introduce a lot of unintended consequences and unexpected behavior. How is this any better than the "before" code, except I didn't need to type 2 extra sets of parens and 2 extra sets of curly braces?

`Using` already creates an implicit single-line block, in much the same way as a single line `if` statement. Dropping the curly braces has too much risk to introduce bugs in the code, in much the same way a single line `if` statement without curly braces can. (Adding what you think is a second statement in the if block, but because there aren't curly braces it's actually not part of the if block even if it's indented.)

Disposability is non-deterministic and there isn't anything the compiler can do to tell the runtime GC that things should be disposed of in a certain order.

HaloFour:
> How is this any better than the "before" code, except I didn't need to type 2 extra sets of parens and 2 extra sets of curly braces?

And extra indentation.  When your code doesn't even start until half-way off of the screen it's a big waste of real-estate.

> Disposability is non-deterministic and there isn't anything the compiler can do to tell the runtime GC that things should be disposed of in a certain order.

This has nothing to do with GC.  It is strictly syntax candy for `using` blocks, and it is strictly deterministic.  The reverse declaration disposal order is the expected behavior.

[...]

PR #1703, opened 2018-07-09
Using patterns and declarations

jaredpar:
Proposal to add `using` local variable declarations and the `using` pattern to the language.

### DIFF
--- a/proposals/using.md
+++ b/proposals/using.md
@@ -23,6 +23,79 @@ be applied.
 
 ## Detailed Design 
 
+### using declaration
+
+The language will allow for `using` to be added to a local variable declaration. Such a declaration
+will have the same effect as declarating the variable in a `using` statement at the same location.
+
+``` csharp
+if (...) { 
+   using FileStream f = new FileStream(@"C:\users\jaredpar\using.md");
+   // statements
+}
+
+// Equivalent to 
+if (...) { 
+   using (FileStream f = new FileStream(@"C:\users\jaredpar\using.md")) {
+    // statements
+   }
+}
+```
+
+The lifetime of a `using` local will extend to the end of the scope in which it is declared. The 
+`using` locals will then be disposed in the reverse order in which they are declared. 
+
+``` csharp
+{ 
+    using var f1 = new FileStream("...");
+    using var f2 = new FileStream("..."), f3 = new FileStream("...");
+    ...
+    // Dispose f3
+    // Dispose f2 
+    // Dispose f1
+}
+```
+
+There are no restrictions around `goto`, are any other control flow construct in the face of 
+`using` declaration. Instead the code acts just as it would for the equivalent `using` statement:
+
+``` csharp
+{
+    using var f1 = new FileStream("...");
+  target:
+    using var f2 = new FileStream("...");
+    if (someCondition) {
+        // Causes f2 to be disposed but has no effect on f1
+        goto target;
+    }
+}
+```
+
+A local declared in a `using` local declaration will be implicitly read-only. This matches the 
+behavior of locals declared in a `using` statement. 
+
+The language grammar for `using` declarations will be the following:
+
+```
+local-using-declaration:
+  using type using-declarators
+
+using-declarators:
+  using-declarator
+  using-declarators , using-declarator
+  
+using-declarator:
+  identifier = expression
+```
+
+Restrictions around `using` declaration:
+
+- May not appear directly inside a `case` label but instead must be within a block inside the
+ `case` label.
+- May not appear as part of an `out` variable declaration. 
+- Must have an initializer for each declarator.
+- The local type must be implicitly convertible to `IDisposable` or fulfill the `using` pattern.
+
 ### using pattern
 
 The language will add the notion of a disposable pattern: that is a type which has an accessible 
@@ -70,8 +143,6 @@ etc ... The code geneartion will be different only in that there will not be a c
 In order to fit the `using` pattern the Dispose method must be accessible, parameterless and have 
 a `void` return type. There are no other restrictions. This explicitly means that extension methods
 can be used here.
-
-### using declaration
 
 ## Considerations
 
--- a/proposals/using.md
+++ b/proposals/using.md
@@ -2,24 +2,87 @@
 
 ## Summary
 
-
+The language will add two need capabilities around the `using` statement in order to make resoure
+managemente simpler: recognize a `using` pattern in addition to `IDisposable` and add a `using`
+declaration to the language.
 
 ## Motivation
+
+The `using` statement is an effective tool for resource management today but it requires quite a 
+bit of ceremony. Methods that have a number of resources to manage can get syntactically bogged 
+down with a series of `using` statements. This syntax burden is enough that most coding style 
+guidelines explicitly have an exception around braces for this scenario. 
+
+The `using` declaration removes much of the ceremony here and gets C# on par with other languages
+that include resource management blocks. Additionally the `using` pattern lets a developers expand
+the set of tyse that can participate here. In many cases removing the need to create wrapper types 
+that only exist to allow for a values use in a `using` statement. 
+
+Together these features allow developers to simplify and expand the scenarios where `using` can
+be applied.
 
 ## Detailed Design 
 
 ### using pattern
 
-The language will add the notion of the disposable pattern: that is a type which has an accessible 
-`Dispose` method.
-The language will no longer require a type implement `IDisposable` in order to 
+The language will add the notion of a disposable pattern: that is a type which has an accessible 
+Dispose instance method. Types which fit the disposable pattern can participate in a `using` 
+statement or declaration without being required to implement `IDisposable`. 
 
-Restrictions:
+``` csharp
+class Resource { 
+    public void Dispose() { ... }
+}
 
-- The `Dispose` method must have a void return type.
-- The `Dispose` method must be parameterless. 
+using (var r = new Resource()) {
+    // statements
+}
+```
+
+This will allow developers to leverage `using` in a number of new scenarios:
+
+- `ref struct`: These types can't implement interfaces today and hence can't participate in `using`
+statements.
+- Extension methods will allow developers to augment types in other assemblies to participate 
+in `using` statements.
+
+In the situation where a type can be implicitly converted to `IDisposable` and also fits the
+`using` pattern, then `IDisposable` will be prefered. While this takes the opposite of approach
+of `foreach` (pattern prefered over interface) it is necessary for backwards compatibility.
+
+The same restrictions from a traditional `using` statement apply here as well: local variables 
+declared in the `using` are read-only, a `null` value will not cause an exception to be thrown, 
+etc ... The code geneartion will be different only in that there will not be a cast to 
+`IDisposable` before calling Dispose:
+
+``` csharp
+{
+	Resource r = new Resource();
+	try {
+		// statements
+	}
+	finally {
+		if (resource != null) resource.Dispose();
+	}
+}
+```
+
+In order to fit the `using` pattern the Dispose method must be accessible, parameterless and have 
+a `void` return type. There are no other restrictions. This explicitly means that extension methods
+can be used here.
+
+### using declaration
 
 ## Considerations
+
+### case labels without blocks
+
+A `using declaration` is illegal directly inside a `case` label due to complications around it's 
+actual lifetime. One potential solution is to simply give it the same lifetime as an `out var` 
+in the same location. It was deemed the extra complexity to the feature implementation and the 
+ease of the work around (just add a block to the `case` label) didn't justify taking this route.
+
+## Future Expansions
 
 ### fixed locals
 
--- a/proposals/using.md
+++ b/proposals/using.md
@@ -1,0 +1,30 @@
+# using patterns and locals
+
+## Summary
+
+
+
+## Motivation
+
+## Detailed Design 
+
+### using pattern
+
+The language will add the notion of the disposable pattern: that is a type which has an accessible 
+`Dispose` method.
+The language will no longer require a type implement `IDisposable` in order to 
+
+Restrictions:
+
+- The `Dispose` method must have a void return type.
+- The `Dispose` method must be parameterless. 
+
+## Considerations
+
+### fixed locals
+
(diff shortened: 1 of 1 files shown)


jaredpar:
CC @agocke, @fayrose

benaadams:
> Types which fit the disposable pattern can participate in a using statement or declaration without being required to implement IDisposable

So please this! 😄

proposals/using.md
```diff
@@ -0,0 +1,164 @@
+# using patterns and locals
+
+## Summary
+
+The language will add two need capabilities around the `using` statement in order to make resoure
+managemente simpler: recognize a `using` pattern in addition to `IDisposable` and add a `using`
```
jnm2: typos `resouce` `managemente`

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Motivation
+
+The `using` statement is an effective tool for resource management today but it requires quite a 
+bit of ceremony. Methods that have a number of resources to manage can get syntactically bogged 
+down with a series of `using` statements. This syntax burden is enough that most coding style 
+guidelines explicitly have an exception around braces for this scenario. 
+
+The `using` declaration removes much of the ceremony here and gets C# on par with other languages
+that include resource management blocks. Additionally the `using` pattern lets a developers expand
```
jnm2: `a developers`

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+the set of tyse that can participate here. In many cases removing the need to create wrapper types
```
jnm2: `tyse`

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+    using var f1 = new FileStream("...");
+    using var f2 = new FileStream("..."), f3 = new FileStream("...");
+    ...
+    // Dispose f3
+    // Dispose f2 
+    // Dispose f1
+}
+```
+
+There are no restrictions around `goto`, are any other control flow construct in the face of
```
jnm2: `are` → `or`

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+    using var f2 = new FileStream("..."), f3 = new FileStream("...");
+    ...
+    // Dispose f3
+    // Dispose f2 
+    // Dispose f1
+}
+```
+
+There are no restrictions around `goto`, are any other control flow construct in the face of 
+`using` declaration. Instead the code acts just as it would for the equivalent `using` statement:
```
jnm2: <code>*a* \`using\` declaration</code>

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+    using var f2 = new FileStream("...");
+    if (someCondition) {
+        // Causes f2 to be disposed but has no effect on f1
+        goto target;
+    }
+}
+```
+
+A local declared in a `using` local declaration will be implicitly read-only. This matches the 
+behavior of locals declared in a `using` statement.
```
jnm2: Now this is cool.

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+This will allow developers to leverage `using` in a number of new scenarios:
+
+- `ref struct`: These types can't implement interfaces today and hence can't participate in `using`
+statements.
+- Extension methods will allow developers to augment types in other assemblies to participate 
+in `using` statements.
+
+In the situation where a type can be implicitly converted to `IDisposable` and also fits the
+`using` pattern, then `IDisposable` will be prefered. While this takes the opposite of approach
+of `foreach` (pattern prefered over interface) it is necessary for backwards compatibility.
```
jnm2: Typo `prefered` twice, <code>opposite <s>of </s>approach</code>

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+
+The same restrictions from a traditional `using` statement apply here as well: local variables 
+declared in the `using` are read-only, a `null` value will not cause an exception to be thrown, 
+etc ... The code geneartion will be different only in that there will not be a cast to
```
jnm2: `geneartion`

proposals/using.md
```diff
@@ -0,0 +1,164 @@
...
+
+In order to fit the `using` pattern the Dispose method must be accessible, parameterless and have 
+a `void` return type. There are no other restrictions. This explicitly means that extension methods
+can be used here.
+
+## Considerations
+
+### case labels without blocks
+
+A `using declaration` is illegal directly inside a `case` label due to complications around it's
```
jnm2: `it's` → `its`

jaredpar:
@jnm2 

> I don't know why I started doing this. Is this actually helpful, or would it be nice if I stopped? 

Clearly based on your feedback, and the feedback of all my peers, I can use all the spelling help I can get. Especially when I'm jet lagged apparently. 😄

OlegKarasik:
@jaredpar 

Sorry if my question is dumb but is the purpose of auto-recognition of Disposable pattern is to avoid implicit cast to IDisposable where possible? 

Because if you have already implemented Disposable method what is the difficulty to implement IDisposable.

benaadams:
> Because if you have already implemented Disposable method what is the difficulty to implement IDisposable.

You can't implement an interface on a `ref struct`; so currently cannot take advantage of the `using` pattern with a `Dispose` method.

jnm2:
@OlegKarasik Also: extension methods can make types disposable which you don't own.

OlegKarasik:
@benaadams , @jnm2 
OK, I got it. So basically the idea is the same as for 'foreach' and GetEnumerator() thing.

Thanks for the clarification.

jnm2:
@OlegKarasik And some others, too: https://github.com/dotnet/csharplang/issues/1467

proposals/using.md
```diff
@@ -0,0 +1,164 @@
+# using patterns and locals
+
+## Summary
+
+The language will add two need capabilities around the `using` statement in order to make resource
```
Tornhoof: Typo: need -> needed

migueldeicaza:
Would love to have a using that acts as a defer, to avoid scenarios where I need to either use an IDisposable, or add a "Dispose" method to the object being used.    It could be an optional block after the expression:

```
using x = Marshal.AllocHGlobal (size) { Marshal.FreeHGlobal (x); }
}
```

Neme12:
Is there an issue for this where we can discuss?

proposals/using.md
Neme12: Why couple this two unrelated proposals together?

Neme12:
I'm not really a fan of this. Most of the time in practice, the disposal can happen much earlier than at the end of the scope and I feel like this will just turn people on the wrong path of declaring a "using" local and having its lifetime extended to the end of the scope (that could potentially also include `await`s) just to get the nicer syntax. At least most of the using statements I've written only had one statement inside - sometimes with a few other statement after that.

Even if you start in this perfect sweetspot of needing the disposal at the end of the scope, let's say you later want to add more statements at the end (which don't need the variable and prevent disposal). Are you going to change the using variable into a using statement? No, most people would not.

> The `using` statement is an effective tool for resource management today but it requires quite a  bit of ceremony

I'm sorry but I don't see a single pair of parentheses and sometimes braces as "quite a bit of ceremony.


Here's an example of C++ I've had to write way too many times:
```c++
statement();

{
    std::lock_guard<std::mutex> lock(mutex);
    statement();
    statement();
}

statement();
```
Is this really so nice? I essentially have to introduce a scope just for this variable and this is neither pretty or easy to understand at first glance. It's also too easy to see these braces as redundant. I always squirm when I have to write something like this in C++ and miss the explicitness of a using scope in C#.

Now maybe I'm missing the point and overreacting since a using statement will still be available. But as I've said, I think it will be too convenient to simply bolt `using` onto a variable and forget about thinking about the scope where you actually need the resource and reducing it.

There is also the fact that many times you do not actually need a variable, and in that case this feature would not help you either. All combined, I don't like how this introduces syntax for just 1 special case - 1. there is a variable and 2. the lifetime should exactly match the end of the scope.

jnm2:
@Neme12 Yes, you might want to move that to https://github.com/dotnet/csharplang/issues/1174 and read the original proposal, https://github.com/dotnet/csharplang/issues/114.

Neme12:
Thanks

proposals/using.md
jaredpar: This is about improving the usability of the `using` statement hence grouping all aspects of that together.

proposals/using.md
```diff
@@ -0,0 +1,164 @@
+# using patterns and locals
```
jcouv: The "using patterns" is confusing terminology. I think it should be "pattern-based using" instead (just like we do "pattern-base foreach", etc).

jcouv:
@jaredpar I made some minor edits for terminology ("pattern-based `using`", "disposable pattern") in d2ce4cc. I hope you don't mind.
Is the proposal ready to commit, since it was reviewed by LDM?

jcouv:
@jaredpar Is this ready to merge?

jaredpar:
@jcouv not yet. I need to update for the LDM meeting feedback.

chojrak11:
I hope it's a good place to discuss `using` declarations proposed for a future version of C#.

Here's the example I found:
```csharp
{ 
    using var f1 = new FileStream("...");
    using var f2 = new FileStream("..."), f3 = new FileStream("...");
    ...
    // Dispose f3
    // Dispose f2 
    // Dispose f1
}
```

What I understood is that the compiler will try to call `IDisposable.Dispose()`, then - if absent - it will discover the `void Dispose()` methods and call them upon exiting of the scope. Perfect.

My question would be: why we need the 'using' keyword in this scenario? Wouldn't the below work equally well? Couldn't the compiler do it for all declarations in the scope?

```csharp
{ 
    var f1 = new FileStream("...");
    var f2 = new FileStream("..."), f3 = new FileStream("...");
    ...
    // Dispose f3
    // Dispose f2 
    // Dispose f1
}
```

chojrak11:
@OlegKarasik 
> This sounds like RAII, C++ and automatic storage :)

Yes, it does! Or at least 'deterministic destruction.' `Dispose` in this scenario is just a strange name for C++-like destructor, but well...

How would an explicit `using` help with your example?

OlegKarasik:
@chojrak11 

The explicit `using` defines whether compile should generate call to `Dispose()` or not. 

Extending the previous example we could have the following:

``` csharp
void B()
{
  var fs = new FileStream("...");
  A(fs);

  using fs1 = new FileStream("...");

  // Dispose fs1
}
```

With explicit `using` it is obvious that `fs` shouldn't be disposed while `fs1` should be.

HaloFour:
@chojrak11 

What you are suggesting would change the meaning of existing programs in a breaking way.

svick:
@chojrak11 How would that help? If you had a class with `Destroy`, how do you write code that doesn't call it when the variable goes out of scope?

chojrak11:
By not defining it, I think. It's like asking 'how not to call the destructor in C++ when an object goes out of scope?' Maybe we're just realizing that C# needs automatic deterministic destruction, and this issue we're discussing is a good start to have that? Maybe `using` declaration is just a crippled version of it.

The `Destroy` name is not a good one, though. It'd need to be something that is not used in existing code.

HaloFour:
@chojrak11 

> Maybe we're just realizing that C# needs automatic deterministic destruction, and this issue we're discussing is a good start to have that?

That's not what this proposal is attempting to achieve.  Disposal is intentionally opt-in because there are plenty of scenarios where the current method would not want to dispose of the resource.  Ownership is not a concept that is defined in C#.

svick:
@chojrak11 

> It's like asking 'how not to call the destructor in C++ when an object goes out of scope?'

What you do in C++ is that you allocate it dynamically ("on the heap"), possibly using some smart pointer type so that you still get deterministic destruction, but one that's not tied to a single block scope. But there's no such option in C#.

msedi:
>  Maybe we're just realizing that C# needs automatic deterministic destruction, and this issue we're discussing is a good start to have that? 

I agree with with @chojrak11. We have a lot of occasions where we definitely know that the object/memory can go away.

I personally find the C++ CLI approach a little more initutive where there is a finalizer and a destructor.
So having a new/delete something would nice.

Nevertheless, this requirement doesn't fit in this discussion, just as a side note.

HaloFour:
@msedi 

> I personally find the C++ CLI approach a little more initutive where there is a finalizer and a destructor.
> So having a new/delete something would nice.

The "destructor" is just syntax candy for implementing `IDisposable` and `delete` just calls `Dispose`.  You don't have control over releasing the memory of the object.

What C++/CLI offers that is interesting is stack-like semantics for working with reference types.  The instances are still allocated on the managed heap and garbage collected, but you get copy semantics (requiring a copy constructor and assignment operator) and RAII (through `IDisposable`).

chojrak11:
@msedi remember that destruction/disposal is not only about memory deallocation. Memory is my least concern. I'm more concerned with database connections, files, network resources and so forth (this is what IDisposable was invented to help with).

I like the idea to open a file and have it automatically closed at the end of the scope. Having a single `using` keyword that signifies the intent in the declaration is an elegant and acceptable:

```csharp
void MyFunc(void) {
    using var myFile = File.Open(@"\path\to\file");
    // ... use the file

    // now it's automatically closed
}
```

while not having the `using` keyword lets me have a helper function that returns an open file:

```csharp
File MyOpenFile(string path) {
    var myFile = File.Open(path);   // no 'using' marker
    return myFile;
    // no automatic cleanup
}

void MyFunc(void) {
    using var myFile = MyOpenFile(@"\path\to\file");
    // ... use the file

    // now it's automatically closed
}
```

In this proposal I only miss Golang's `defer` stuff, that is, registration of delegates to be executed at the end of the scope, as in below pseudo-C# example - that could trivially be implemented in the language.

```csharp
void MyFunc(void) {
    using var myFile = File.Open(@"\path\to\file");
    defer () => Console.WriteLine("Finished processing the file.");
    defer () => Console.WriteLine("Another delegate.");
    // ... use the file

    // "Another delegate."
    // "Finished processing the file."
    // now the file is automatically closed
}
```

chojrak11:
As this proposal replaces end of `using` blocks with end of enclosing scope, I'm just trying to explore additional possibilities it may open.

The `defer` stuff would be useful and is quite easy to add to the language.

PR #1721, opened 2018-07-17
Update null-coalescing-assignment proposal with changes from LDM.

333fred:


proposals/null-coalescing-assignment.md
```diff
@@ -10,6 +10,8 @@
 
 Simplifies a common coding pattern where a variable is assigned a value if it is null.
 
+As part of this proposal, we will also loosen the type requirements on `??` to allow variables typed with the unconstrained type parameter to be used on the left-hand side.
```
gafter: >variables typed with the unconstrained type parameter [](start = 85, length = 53)

...an expression whose type is an unconstrained type parameter...

proposals/null-coalescing-assignment.md
```diff
@@ -37,7 +39,23 @@ assignment_operator
     ;
 ```
 
-Which follows the [existing semantic rules for compound assignment operators](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#compound-assignment). What that means is that an operation of the form `x ??= y` is processed as if written `x = (x ?? y)` (except `x` shall be computed once). However, we permit the assignment to `x` (of its existing value) to be elided in the case that `x` is not null.
+Which follows the [existing semantic rules for compound assignment operators](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#compound-assignment), except that we elide the assignment if the left-hand side is non-null. The rules for this feature are as follows.
+
+Given `a ??= b`, where `A` is the type of `a`, `B` is the type of `b`, and `A0` is the underlying value type of `A` if `A` is the nullable type.
```
gafter: >type [](start = 139, length = 4)

... otherwise the type of A.

gafter: We don't appear to need A0 in the specification.

---
In reply to: [203555762](https://github.com/dotnet/csharplang/pull/1721#discussion_r203555762) [](ancestors = 203555762)

proposals/null-coalescing-assignment.md
```diff
@@ -37,7 +39,23 @@ assignment_operator
...
+1. If `A` does not exist or is a non-nullable value type, a compile-time error occurs.
+2. If `B` is not implicitly convertible to `A` or `A0`, a compile-time error occurs.
+3. The type of `a ??= b` is `A`.
```
gafter: I think we just need it convertible to `A`. If it is convertible to `A0`, then that causes it to be convertible to `A`.

proposals/null-coalescing-assignment.md
```diff
@@ -37,7 +39,23 @@ assignment_operator
...
+4. `a ??= b` is evaluated at runtime as `a ?? (a = b)`, except that `a` is only evaluated once.
+
+Note that rule 4 means that if `B` is implicitly convertible to both `A` and `A0`, we will prefer the conversion to `A` over `A0`.
+
```
gafter: If we never mention `A0`, this would be implied.

333fred:
@gafter if you could take a quick look and make sure you agree with the updated language.

proposals/null-coalescing-assignment.md
```diff
@@ -10,6 +10,8 @@
 
 Simplifies a common coding pattern where a variable is assigned a value if it is null.
 
+As part of this proposal, we will also loosen the type requirements on `??` to allow an expression whose type is the unconstrained type parameter to be used on the left-hand side.
```
gafter: >the [](start = 113, length = 3)

"an expression whose type is **the** unconstrained type parameter" should be "an expression whose type is **an** unconstrained type parameter".

gafter:
Still looks good.

333fred:
Thanks @gafter

Issue #1710, opened 2018-07-12
Provide a way to specify that methods do not modify state.


tannergooding:
## Summary
[summary]: #summary

Provide a way to specify that methods do not modify state.

## Motivation
[motivation]: #motivation

Today, users have the ability to create `readonly struct` types which the compiler enforces that all fields are readonly (and by extension, that no methods modify the state). However, there are some scenarios where you have an existing API that exposes public fields, or needs to modify some small amount of internal state in select scenarios. Under these circumstances, you cannot mark the type as `readonly` (it would be a breaking change).

This normally doesn't have much impact, except in the case of `in` parameters. With `in` parameters for non-readonly structs, the compiler will make a copy of the parameter for each instance method call, since it cannot guarantee that the call does not modify internal state. This can lead to a multitude of copies and worse overall performance than if you had just passed the struct directly by value. For example: https://sharplab.io/#v2:CYLg1APgAgDABFAjAbgLACgNQMxwM4AuATgK4DGBcAagKYUD2RATBgN4ZycK4BmANvQCGlAB5p0XbnH5DKAT3GSOXHNIHC4AGRoA7AOYEAFgGUAjiUFEawZZ3YTJXPTQK3H9x54QB2OAAoROAAqOBEASjgwNy8YvzlguDkwxS8AXzd09EysXCgmOABhOA8VXnVKAFk/AEsdajoCRnyAN0E+EhoIks8oX1b2mgA6bX0jMwsrYEi4fo7h3QMTc0trFM5M1KA==

## Design
[design]: #design

I propose allowing a user to specify that a method itself is `readonly` and does not modify the state of the struct (with all the appropriate verification done by the compiler, of course). For example:

```C#
public struct Vector2
{
    public float x;
    public float y;

    public readonly float GetLength()
    {
        return MathF.Sqrt(LengthSquared);
    }

    public float LengthSquared
    {
        readonly get
        {
            return (x * x) +
                   (y * y);
        }
    }
}
```

The compiler would emit the method, as usual, and would additionally emit a compiler recognized attribute indicating that the method does not modify state.

This would allow the user to safely call said instance method without the compiler needing to make a copy.

## Drawbacks
[drawbacks]: #1094 

Using the `readonly` keyword is useful, since it is an existing keyword. However, it may get confusing to read if you have something like `public readonly ref readonly float GetRefToX()` (`const T* const` anyone 😄)

## Notes
[notes]: #notes

Using an attribute or another keyword may also be possible.

[...]

PR #1730, opened 2018-07-19
Adding the readonly-methods proposal document.

tannergooding:
FYI. @jaredpar who is championing the proposal

### DIFF
--- a/proposals/readonly-instance-members.md
+++ b/proposals/readonly-instance-members.md
@@ -1,0 +1,112 @@
+# Readonly Instance Members
+
+Championed Issue: https://github.com/dotnet/csharplang/issues/1710
+
+## Summary
+[summary]: #summary
+
+Provide a way to specify individual instance members on a struct do not modify state, in the same way that `readonly struct` specifies no instance members modify state.
+
+It is worth noting that `readonly instance member` != `pure instance member`. A `pure` instance member guarantees no state will be modified. A `readonly` instance member only guarantees that instance state will not be modified.
+
+All instance members on a `readonly struct` could be considered implicitly `readonly instance members`. Explicit `readonly instance members` declared on non-readonly structs would behave in the same manner. For example, they would still create hidden copies if you called an instance member (on the current instance or on a field of the instance) which was itself not-readonly.
+
+## Motivation
+[motivation]: #motivation
+
+Today, users have the ability to create `readonly struct` types which the compiler enforces that all fields are readonly (and by extension, that no instance members modify the state). However, there are some scenarios where you have an existing API that exposes accessible fields or that has a mix of mutating and non-mutating members. Under these circumstances, you cannot mark the type as `readonly` (it would be a breaking change).
+
+This normally doesn't have much impact, except in the case of `in` parameters. With `in` parameters for non-readonly structs, the compiler will make a copy of the parameter for each instance member invocation, since it cannot guarantee that the invocation does not modify internal state. This can lead to a multitude of copies and worse overall performance than if you had just passed the struct directly by value. For an example, see this code on [sharplab](https://sharplab.io/#v2:CYLg1APgAgDABFAjAbgLACgNQMxwM4AuATgK4DGBcAagKYUD2RATBgN4ZycK4BmANvQCGlAB5p0XbnH5DKAT3GSOXHNIHC4AGRoA7AOYEAFgGUAjiUFEawZZ3YTJXPTQK3H9x54QB2OAAoROAAqOBEASjgwNy8YvzlguDkwxS8AXzd09EysXCgmOABhOA8VXnVKAFk/AEsdajoCRnyAN0E+EhoIks8oX1b2mgA6bX0jMwsrYEi4fo7h3QMTc0trFM5M1KA==)
+
+Some other scenarios where hidden copies can occur include `static readonly fields` and `literals`. If they are supported in the future, `blittable constants` would end up in the same boat; that is they all currently necessitate a full copy (on instance member invocation) if the struct is not marked `readonly`.
+
+## Design
+[design]: #design
+
+Allow a user to specify that an instance member is, itself, `readonly` and does not modify the state of the instance (with all the appropriate verification done by the compiler, of course). For example:
+
+```C#
+public struct Vector2
+{
+    public float x;
+    public float y;
+
+    public float GetLengthReadonly() readonly
+    {
+        return MathF.Sqrt(LengthSquared);
+    }
+
+    public float GetLength()
+    {
+        return MathF.Sqrt(LengthSquared);
+    }
+
+    public float GetLengthIllegal() readonly
+    {
+        var tmp = MathF.Sqrt(LengthSquared);
+
+        x = tmp;    // Compiler error, cannot write x
+        y = tmp;    // Compiler error, cannot write y
+
+        return tmp;
+    }
+
+    public float LengthSquared
+    {
+        get readonly
+        {
+            return (x * x) +
+                   (y * y);
+        }
+    }
+}
+
+public static class MyClass
+{
+    public static float ExistingBehavior(in Vector2 vector)
+    {
+        // This code causes a hidden copy, the compiler effectively emits:
+        //    var tmpVector = vector;
+        //    return tmpVector.GetLength();
+        //
+        // This is done because the compiler doesn't know that `GetLength()`
+        // won't mutate `vector`.
+
+        return vector.GetLength();
+    }
+
+    public static float ReadonlyBehavior(in Vector2 vector)
+    {
+        // This code is emitted exactly as listed. There are no hidden
+        // copies as the `readonly` modifier indicates that the method
+        // won't mutate `vector`.
+
+        return vector.GetLengthReadonly();
+    }
+}
+```
+
+Some other syntax examples:
+* Expression bodied members: `public float ExpressionBodiedMember readonly => (x * x) + (y * y);`
+* Generic constraints: `public static void GenericMethod<T>(T value) readonly where T : struct { }`
+
+The compiler would emit the instance member, as usual, and would additionally emit a compiler recognized attribute indicating that the instance member does not modify state. This effectively causes the hidden `this` parameter to become `in T` instead of `ref T`.
+
+This would allow the user to safely call said instance method without the compiler needing to make a copy.
+
+The restrictions would include:
+* The `readonly` modifier cannot be applied to static methods, constructors or destructors.
+* The `readonly` modifier cannot be applied to delegates.
+* The `readonly` modifier cannot be applied to members of class or interface.
+
+## Drawbacks
+[drawbacks]: #drawbacks 
+
+Same drawbacks as exist with `readonly struct` methods today. Certain code may still cause hidden copies.
+
+## Notes
+[notes]: #notes
+
+Using an attribute or another keyword may also be possible.
+
+This proposal is somewhat related to (but is more a subset of) `functional purity` and/or `constant expressions`, both of which have had some existing proposals.


proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
+# Readonly Methods
```
tannergooding: This is basically identical to the issue: https://github.com/dotnet/csharplang/issues/1710, with a couple minor changes I've called out below.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
+
+## Summary
+[summary]: #summary
+
+Provide a way to specify individual methods on a struct do not modify state, in the same way that `readonly struct` specifies no instance method modifies state.
+
+It is worth noting that `readonly method` != `pure method`. A `pure` method guarantees no state will be modified. A `readonly` method only guarantees that instance state will not be modified.
```
tannergooding: Changed the big NOTE section I had at the top to this here.

I think it's still worth calling out so that people don't get the wrong idea 😄

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+All instance methods on a `readonly struct` could be considered implicitly `readonly methods`. Explicit `readonly methods` declared on non-readonly structs would behave in the same manner. For example, they would still create hidden copies if you called an instance method (on the current instance or on a field of the instance) which was itself not-readonly.
```
tannergooding: Probably worth checking that this statement matches what you'd expect. It is basically saying that all instance methods on a `readonly struct` are implicitly `readonly methods`.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Motivation
+[motivation]: #motivation
+
+Today, users have the ability to create `readonly struct` types which the compiler enforces that all fields are readonly (and by extension, that no methods modify the state). However, there are some scenarios where you have an existing API that exposes public fields, or needs to modify some small amount of internal state in select scenarios. Under these circumstances, you cannot mark the type as `readonly` (it would be a breaking change).
+
+This normally doesn't have much impact, except in the case of `in` parameters. With `in` parameters for non-readonly structs, the compiler will make a copy of the parameter for each instance method call, since it cannot guarantee that the call does not modify internal state. This can lead to a multitude of copies and worse overall performance than if you had just passed the struct directly by value. For example: https://sharplab.io/#v2:CYLg1APgAgDABFAjAbgLACgNQMxwM4AuATgK4DGBcAagKYUD2RATBgN4ZycK4BmANvQCGlAB5p0XbnH5DKAT3GSOXHNIHC4AGRoA7AOYEAFgGUAjiUFEawZZ3YTJXPTQK3H9x54QB2OAAoROAAqOBEASjgwNy8YvzlguDkwxS8AXzd09EysXCgmOABhOA8VXnVKAFk/AEsdajoCRnyAN0E+EhoIks8oX1b2mgA6bX0jMwsrYEi4fo7h3QMTc0trFM5M1KA==
```
Joe4evr: This could be changed to a markdown link tag so it doesn't look so ugly.

tannergooding: Will fix.

tannergooding: Fixed.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
+# Readonly Methods
+
+* [x] Proposed
+* [ ] Prototype: Not Started
+* [ ] Implementation: Not Started
+* [ ] Specification: Not Started
```
jaredpar: Please move the state tracking to the issue. This is how we've been handling recent proposals.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Summary
+[summary]: #summary
+
+Provide a way to specify individual methods on a struct do not modify state, in the same way that `readonly struct` specifies no instance method modifies state.
```
jaredpar: Instead of "methods" I'd use "instance members". The feature will be applicable to properties as well and never applicable to static members.

proposals/readonly-methods.md
jaredpar: I think somewhere we should add the more explicit language explanation. Essentially that marking an instance member as `readonly` causes the type of the `this` parameter to be `in T` instead of `ref T`.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+It is worth noting that `readonly method` != `pure method`. A `pure` method guarantees no state will be modified. A `readonly` method only guarantees that instance state will not be modified.
+
+All instance methods on a `readonly struct` could be considered implicitly `readonly methods`. Explicit `readonly methods` declared on non-readonly structs would behave in the same manner. For example, they would still create hidden copies if you called an instance method (on the current instance or on a field of the instance) which was itself not-readonly.
+
+## Motivation
+[motivation]: #motivation
```
jaredpar: Why do you add this?

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+Today, users have the ability to create `readonly struct` types which the compiler enforces that all fields are readonly (and by extension, that no methods modify the state). However, there are some scenarios where you have an existing API that exposes public fields, or needs to modify some small amount of internal state in select scenarios. Under these circumstances, you cannot mark the type as `readonly` (it would be a breaking change).
```
jaredpar: I think "expose public fields" is too limiting. Possibly use something along the lines of "exposes accessible fields and has a mix of mutating and non-mutating members".

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+This normally doesn't have much impact, except in the case of `in` parameters. With `in` parameters for non-readonly structs, the compiler will make a copy of the parameter for each instance method call, since it cannot guarantee that the call does not modify internal state. This can lead to a multitude of copies and worse overall performance than if you had just passed the struct directly by value. For an example, see this code on [sharplab](https://sharplab.io/#v2:CYLg1APgAgDABFAjAbgLACgNQMxwM4AuATgK4DGBcAagKYUD2RATBgN4ZycK4BmANvQCGlAB5p0XbnH5DKAT3GSOXHNIHC4AGRoA7AOYEAFgGUAjiUFEawZZ3YTJXPTQK3H9x54QB2OAAoROAAqOBEASjgwNy8YvzlguDkwxS8AXzd09EysXCgmOABhOA8VXnVKAFk/AEsdajoCRnyAN0E+EhoIks8oX1b2mgA6bX0jMwsrYEi4fo7h3QMTc0trFM5M1KA==)
```
jaredpar: `static readonly` fields are another case where this comes up a lot. 

Oh and literals. Consider the impact of blittable constants here. If we don't have the ability to mark methods as `readonly` then invoking a member on a blittable constant will necessitate a full copy 😦

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Design
+[design]: #design
+
+I propose allowing a user to specify that a method itself is `readonly` and does not modify the state of the struct (with all the appropriate verification done by the compiler, of course). For example:
```
jaredpar: Avoid "I" here. I think these are generally written without first person narrative.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+        get readonly
+        {
+            return (x * x) +
+                   (y * y);
+        }
+    }
+}
+```
+
+The compiler would emit the method, as usual, and would additionally emit a compiler recognized attribute indicating that the method does not modify state.
```
jaredpar: Possibly the right place to add "marking an instance member as readonly causes the type of the this parameter to be in T instead of ref T."

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+
+```C#
+public struct Vector2
+{
+    public float x;
+    public float y;
+
+    public float GetLength() readonly
+    {
+        return MathF.Sqrt(LengthSquared);
```
jaredpar: I think it would help to give a negative example that demonstrates exactly what this does. Essentially marking a method `readonly` and showing there is a compilation error when attempting to write to a mutable field.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public float LengthSquared
+    {
+        get readonly
+        {
+            return (x * x) +
+                   (y * y);
+        }
```
jaredpar: Should add examples for expression bodied members to demonstrate where the `readonly` syntax would go.

proposals/readonly-methods.md
CyrusNajmabadi: > For example, they would still create hidden copies if you called an instance method (on the current instance or on a field of the instance) which was itself not-readonly.

I feel like this will be surprising to people.  It think a lot of the motivation here is that "oh, if i have readonly methods, i can help avoid copies". When doing simple stuff causes copying, that seems like it woudl go against that intent.  Woudl it perhaps be better to instead have this error thus making it clear: these methods will will not copy.  If users hit this and then they want to actually perform that functionality that would copy, they can then realize the method will not be readonly, or, they can figure out what they need to update downstream to enable things to stay readonly.

If users really are ok with the copy, i would say that they should then write the copy explicitly themselves.  i.e.:

```c#
var copy = this;
copy.NonReadOnlyMethod();
```

This will make the code nice and explicit and make it much more clear waht it going on.

proposals/readonly-methods.md
jaredpar: Should add an example demonstrating where the `readonly` syntax is positioned on a generic instance method that has constraints.

proposals/readonly-methods.md
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+```
+
+The compiler would emit the method, as usual, and would additionally emit a compiler recognized attribute indicating that the method does not modify state.
+
+This would allow the user to safely call said instance method without the compiler needing to make a copy.
+
```
jaredpar: Restrictions:

- The `readonly` modifier cannot be applied to `static` methods, constructors or destructors.
- The `readonly` modifier cannot be applied to `delegates`.
- The `readonly` modifier cannot be applied to members of `class` or `interface`.

proposals/readonly-methods.md
tannergooding: It is in the template, and is what allows you to link to sub-sections of the markdown file.

proposals/readonly-methods.md
tannergooding: @CyrusNajmabadi, yes that would be ideal. Unfortunately, we already shipped `readonly structs` this way, and making `readonly method` behave differently may be more confusing or more error prone.

I believe the current answer is to provide an analyzer (or warning wave support) to help flag these. CC. @jaredpar

CyrusNajmabadi: That seems reasonable to me.  I would suggest whoever does this work also does hte work to make such an analyzer.  that was we know we have enough information in the compilation API to make it clear when copies are happening :)

jaredpar: The analyzers to detect silent struct copies already exist 😄 

https://www.nuget.org/packages/ErrorProne.NET.Structs/0.1.0
https://github.com/SergeyTeplyakov/ErrorProne.NET

CC @SergeyTeplyako

proposals/readonly-methods.md
orthoxerox: @jaredpar Why couldn't it be applied to members of `class` or `interface`? Not modifying the state of `this` is a valuable property. Or is it just a temporary restriction to limit the scope of the feature?

tannergooding: @orthoxerox, we don't have `readonly class` today, so it isn't relevant.

If/when we get `readonly class`, then (assuming it was after) `readonly methods` would naturally be considered to extend there as well.

tannergooding:
Will try to respond to the PR feedback today, have a few other items on my plate first however 😄

proposals/readonly-methods.md
jaredpar: @orthoxerox 

The use of `readonly` today provides a demonstrable benefit to developers: it avoids unnecessary `struct` copies. This can be a real benefit for high performance code. There is no equivalent motivating scenario for `class`, and `interface` to warrant adding this now.

Additionally I'm not sure it's the solution that we want for `class` and `interface`. There you likely want something closer to `const` in C++ (without all of the downsides like double method definitions). I have ideas on how to add notions of readonly behavior to reference types but it requires a bit more than just `readonly`.

SergeyTeplyakov:
Maybe I've missed it, but what about methods that implement interface members or methods that override methods from `System.Object`? Like, can I mark `ToString` method with `readonly` modifier?

jaredpar:
@SergeyTeplyakov 

> Maybe I've missed it, but what about methods that implement interface members or methods that override methods from System.Object? Like, can I mark ToString method with readonly modifier?

It's not explicitly called out in the doc but yes that would be legal. We should add this to the doc though to be clear here. There are similar cases (say named arguments) where the spec has to be clear on which method / override it considers for features.

tannergooding:
Working on fixing up the document now.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

proposals/readonly-methods.md
tannergooding: Fixed.

tannergooding:
@jaredpar, think I've responded to all the feedback so far, if you could review again when you get the chance.

jaredpar:
@tannergooding is there a reason you didn't add the restriction list?

- The readonly modifier cannot be applied to static methods, constructors or destructors.
- The readonly modifier cannot be applied to delegates.
- The readonly modifier cannot be applied to members of class or interface.

tannergooding:
>  is there a reason you didn't add the restriction list?

Yes, a very good one in fact...

... I missed it being called out in the slew of other fixes I had to make 😄

tannergooding:
Added the restrictions list.

Issue #313, opened 2017-03-22
Add a way to inherit documentation <inheritdoc />


gafter:
(moved from https://github.com/dotnet/roslyn/issues/67 by @dfkeenan)

Hi,

**Please add `<inheritdoc />` tag or similar:**

I think it would be really handy if the compiler(s) when outputting the documentation XML file if it could use inherited documentation where available/requested.  Though I am not sure what should be output if there is no documentation available.

**Please add the ability to add documentation comments for Namespace/Assembly:**

Currently if you want to do this for tools that generate documentation html etc. you have to add a dummy class and add the documentation comments to them. This one is a bit trickier because there aren't any viable options for doing them directly in code.  There is nothing you could put the assembly documentation against in code and namespaces are "declared" possibly in multiple files.  The only thing I could really think of is a separate XML file that gets merged into the documentation XML file that is output.

Cheers,
dfkeenan

P.S. A Diagnostic and Code Fix that generates document comments for me would also great :smiley:

daveaglick:
Should `<inheritdoc />` and namespace comments be tracked in separate issues?

Also, FWIW, I have an open pull request at https://github.com/dotnet/roslyn/pull/15494 that addresses namespace documentation comments as described in https://github.com/dotnet/roslyn/issues/15474 (which hasn't been closed or migrated yet).

gafter:
@daveaglick Yes, having separate issues would be great. If you create a new one for one of these and link to it here, I'll edit the OP here.

daveaglick:
@gafter :+1:

I created https://github.com/dotnet/csharplang/issues/315 which is just a copy of https://github.com/dotnet/roslyn/issues/15474 (so I guess that corresponding Roslyn issue can also be closed now)

terrajobst:
Is's worth mentioning that this tag can be (and already is) supported by documentation generators. By making it a Roslyn feature it makes it first class by making IntelliSense work (e.g. hovering over calls in the class overriding the member shows the documentation). Also, it should be implied to reduce boilerplate without causing a compiler warning for missing documentation.

sharwell:
I believe a good way to start here would be examining the features currently provided by SHFB for this documentation element, and then identifying ones which would be *impractical* from an implementation perspective or "not particularly needed" from an overall usage perspective.

[...]

PR #1749, opened 2018-07-30
Add specification for 'inheritdoc'

sharwell:
See #313

### DIFF
--- a/proposals/inheritdoc.md
+++ b/proposals/inheritdoc.md
@@ -41,9 +41,8 @@ The compiler is updated in the following ways to account for this element:
 * The compiler MUST allow the use of an `inheritdoc` element as a top-level element in XML documentation comments.
 * The compiler MUST allow the use of an `inheritdoc` element as an inline element in XML documentation comments.
 * The compiler SHOULD report a warning if the `inheritdoc` element appears without a `cref` attribute, and no candidate for inheriting documentation exists.
-* The compiler MUST evaluate the `cref` attribute value in the same manner as it does for the `see` element, and include the resolved documentation ID as the value of the attribute in the output documentation file.
+* The compiler SHOULD emit the documentation file with `inheritdoc` elements replaced by their inherited content. The compiler MAY support emitting the documentation file with `inheritdoc` elements not replaced; in this case the compiler SHOULD preserve the placement and form of the `inheritdoc` element in the XML documentation file, except with the `cref` attribute expanded to the documentation ID of the referenced member.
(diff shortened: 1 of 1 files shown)


sharwell:
@gafter This mostly exists so I can track changes in the proposed design as I work on a prototype. The issue made it difficult to see the progress over time (stuck between editing the comment with the proposal and losing the clear history, or making new comments and not having the final form easily viewable).

gafter:
I recommend you track your spec along with the implementation in `roslyn/tree/master/docs/features`

WarrenFerrell:
@sharwell any update on the status on this change?

proposals/inheritdoc.md
```diff
@@ -0,0 +1,201 @@
...
+This element may be placed either as a top-level element or as an inline element in a documentation comment.
+
+### Compilation
+
+The compiler is updated in the following ways to account for this element:
+
+* The compiler MUST allow the use of an `inheritdoc` element as a top-level element in XML documentation comments.
+* The compiler MUST allow the use of an `inheritdoc` element as an inline element in XML documentation comments.
+* The compiler SHOULD report a warning if the `inheritdoc` element appears without a `cref` attribute, and no candidate for inheriting documentation exists.
+* The compiler SHOULD emit the documentation file with `inheritdoc` elements replaced by their inherited content. The compiler MAY support emitting the documentation file with `inheritdoc` elements not replaced; in this case the compiler SHOULD preserve the placement and form of the `inheritdoc` element in the XML documentation file, except with the `cref` attribute expanded to the documentation ID of the referenced member.
```
gafter: >elements replaced by their inherited content [](start = 68, length = 44)

This seems like a bad idea for a number of reasons:
1. The compiler doesn't normally consume doc comments in any way.  This would force the compiler to do so.
2. It makes the compiler output documentation frozen (copy of the referenced doc) at the time the code was compiled, rather than being flexible when typos are corrected in the inherited documentation later.
3. It makes the determinism and build story somewhat more complicated (as the compilation may depend on the doc comments of referenced libraries)

I suggest instead that the compiler should simply copy the inheritdoc element into the output.  Consuming tools would be required to process them.

sharwell: Here are the two main comments covering this:
https://github.com/dotnet/csharplang/issues/313#issuecomment-322226397
https://github.com/dotnet/csharplang/issues/313#issuecomment-322227985

In general, preserving the elements would produce better results when using tools that support the elements, but would produce worse results with tools that have not yet been updated.

PR #1750, opened 2018-07-30
cref method groups spec

sharwell:


jnm2:
We need these so bad.

Issue #401, opened 2017-04-04
CS1584+CS1658/CS0081 when using constructed generic types in XML comments


paulomorgado:
This code:

```CSharp
/// <summary>
/// A <see cref="TaskCompletionSource{IReadOnlyDictionary{string, object}}" /> object that handles the completion of the state machine execution.
/// </summary>
private readonly TaskCompletionSource<IReadOnlyDictionary<string, object>> completion;
```

Generates 4 CS1658 warning caused by a CS0081 error in the comment line warnings both in VS2015 and VS2017 and an extra CS1584 warning in VS2017.

By the way, where is the documentation for these errors/warnings?

jnm2:
Constructed generic types are not supported in XML documentation, to my irritation.
Though in your example copying the entire type into the documentation does seem like a waste since people will be viewing the type twice in a row now, no matter where they are reading the documentation.

paulomorgado:
> to my irritation
Our irritation! 😄 

I'm trying to work out phrasing that inform unexperienced developers that asynchronous methods return a `Task` and everyone that the important part is the `Task<T>.Result`.

gafter:
I'm not sure if this is intended as a language proposal, a language question, or if it belongs in the Roslyn repo as a comment about the behavior of the compiler.

paulomorgado:
Neither do I!

Should that be wrong? What does the spec say about that?

aluanhaddad:
This is definitely a pain point. I would love to see support for this.

sharwell:
Currently the compiler does not allow a documentation comment to directly reference an instantiated generic type. For the purpose of *hyperlinking* in the resulting documentation, this seems to make sense. However, documentation includes more than just a simple hyperlink. I can certainly see value in the ability to reference an instantiated generic type which displays using the type as-referenced, but links to the original type definition.

From what I can tell, the spec neither requires nor forbids this ability. I believe it would make sense to formalize the desired behavior as a language proposal prior to implementing it within one specific compiler.

jnm2:
As far as hyperlinking goes, I'd assume it would hyperlink to the `.GetGenericTypeDefinition()` of the type.

sharwell:
@jnm2 Yes, that's what I meant by linking to the original type definition.

paulomorgado:
@sharwell,

I does allows `IReadOnlyDictionary{string, object}`. Just doesn't allow `TaskCompletionSource{IReadOnlyDictionary{string, object}}`.

[...]

PR #1751, opened 2018-07-30
cref instantiated types

sharwell:
See #401

### DIFF
--- a/proposals/cref-instantiated-types.md
+++ b/proposals/cref-instantiated-types.md
@@ -1,0 +1,41 @@
+﻿# `cref` instantiated types
+
+* [x] Proposed
+* [ ] Prototype: [Complete](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
+* [ ] Implementation: [In Progress](https://github.com/dotnet/roslyn/BRANCH_NAME)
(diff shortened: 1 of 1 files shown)


Joe4evr:
> One option is to allow `<see cref="Dictionary{,}"/>`

Would this also be supported partially, e.g. `<see cref="Dictionary{int,}"/>`?

proposals/cref-instantiated-types.md
```diff
@@ -0,0 +1,70 @@
...
+
+<!-- What other designs have been considered? What is the impact of not doing this? -->
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+<!-- What parts of the design are still undecided? -->
+
+* How should the open generic type be referenced if a name matching the type parameter appears in the scope of the reference?
+    * One option is to allow `<see cref="Dictionary{,}"/>`
```
Joe4evr: Would this also be supported partially, e.g. `<see cref="Dictionary{int,}"/>`?

jnm2: @Joe4evr That seems complex for consumers of the XML. Why would you want to represent that? It's not even representable in C#.

PR #1778, opened 2018-08-07
Add draft spec for pattern-matching in C# 7.

gafter:


gafter:
@dotnet/ldm Better late than never... C# 7 pattern-matching spec submitted for your review.

PR #1780, opened 2018-08-09
Update target-typed-new.md

alrz:
Updated according to the last LDM and added open issues.

/cc @jcouv

### DIFF
--- a/proposals/target-typed-new.md
+++ b/proposals/target-typed-new.md
@@ -59,7 +59,7 @@ All the other types that are not permitted in the *object_creation_expression* a
 
 > **Open Issue:** should we allow delegates and tuples as the target-type?
 
-Although both types are constructible, if the type is inferrable, an anonymous function or a tuple literal can already be used.
+The above rules include delegates (a reference type) and tuples (a struct type). Although both types are constructible, if the type is inferrable, an anonymous function or a tuple literal can already be used.
 ```cs
 (int a, int b) t = new(1, 2); // "new" is redundant
 Action a = new(() => {}); // "new" is redundant
@@ -74,9 +74,9 @@ var x = new(1, 2) == (1, 2) // "new" is redundant
 
 > **Open Issue:** should we allow `throw new()` with `Exception` as the target-type?
 
-We have `throw null` today, but not `throw default` (though it would have the same effect). On the other hand, `throw new()` could be actually useful as a shorthand for `throw new Exception(...)`.
+We have `throw null` today, but not `throw default` (though it would have the same effect). On the other hand, `throw new()` could be actually useful as a shorthand for `throw new Exception(...)`. Note that it is already allowed by the current specification. `Exception` is a reference type, and the specification for the throw statement says that the expression is converted to `Exception`.
 
-> **Open Issue:** should we allow usages of a target-typed `new` with comparison and arithmetic operators?
+> **Open Issue:** should we allow usages of a target-typed `new` with user-defined comparison and arithmetic operators?
 
 For comparison, `default` only supports equality (user-defined and built-in) operators. Would it make sense to support other operators for `new()` as well?
 
--- a/proposals/target-typed-new.md
+++ b/proposals/target-typed-new.md
@@ -24,7 +24,10 @@ Allow omitting the type when it can be inferred from usage.
 ```cs
 XmlReader.Create(reader, new() { IgnoreWhitespace = true });
 ```
-
+Instantiate an object without spelling out the type.
+```cs
+private readonly static object s_syncObj = new();
+```
 ## Detailed design
 [design]: #detailed-design
 
@@ -35,7 +38,9 @@ object_creation_expression
     | 'new' type object_or_collection_initializer
     ;
 ```
-The target-type is resolved through a conversion that only exists if and only if there exists a single best accessible constructor with regard to provided arguments. Target of the initialization is determined after the conversion succeeded.
+A target-typed `new` is convertible to any type. As a result, it does not contribute to overload resolution. This is mainly to avoid unpredictable breaking changes.
+
+The argument list and the initializer expressions will be bound after the type is determined.
 
 The type of the expression would be inferred from the target-type which would be required to be one of the following:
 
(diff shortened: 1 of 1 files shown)


jcouv:
Tagging @gafter for a second look.

alrz:
Currently `T v = e ? new(e) : new(e)` won't work unless if we target-type the whole ternary expression. Could we consider it as an open issue? (though it could be a separate feature altogether, but I think it's a little inconvenient with new() in particular).

jnm2:
@alrz Yes, I've been having that same thought but wanting to target-type delegates instead:

```cs
// Today
Foo(d == null ? null : new Action<Translation>(t => t.SetResult(d));

// Wish
Foo(d == null ? null : t => t.SetResult(d));
```

gafter:
Target-typing through a ternary expression would be an entirely separate issue.

proposals/target-typed-new.md
```diff
@@ -46,54 +51,51 @@ The type of the expression would be inferred from the target-type which would be
...
-- **Tuple types:** tuple literals should be used for such expressions.
 - **Array types:** arrays need a special syntax to provide the length.
+- **Struct default constructor**: this rules out all primitive types and most value types. If you wanted to use the default value of such types you could write `default` instead.
 
 All the other types that are not permitted in the *object_creation_expression* are excluded as well, for instance, pointer types.
 
-The default constructor for a value type (unless appeared explicitly in metadata) is excluded, which has the effect of excluding the primitive types and the default constructor for most value types. If you wanted to use the default value of such types you could write `default`.
-
-> **Open Issue:** What other types should be excluded?
+> **Open Issue:** should we allow delegates and tuples as the target-type?
```
gafter: A tuple type is an example of "any struct type", so it is allowed by the above.

proposals/target-typed-new.md
```diff
@@ -46,54 +51,51 @@ The type of the expression would be inferred from the target-type which would be
...
-Otherwise it would fail with an ambiguous call error.
 
-## Drawbacks
-[drawbacks]: #drawbacks
 
-Since we're relying on overload resolution of the target-type members, there is a high chance that any change in one of the participating members (target-type's constructors or call-site method overloads, if any) would result in breaking the compilation. For instance,
-```cs
-class Foo { }
-class Bar { }
+> **Open Issue:** should we allow `throw new()` with `Exception` as the target-type?
```
gafter: It already is allowed by the current specification. Exception is a reference type, and the specification for the throw statement says that the expression is converted to Exception.

proposals/target-typed-new.md
```diff
@@ -46,54 +51,51 @@ The type of the expression would be inferred from the target-type which would be
...
 
-void M(Foo foo) {}
+We have `throw null` today, but not `throw default` (though it would have the same effect). On the other hand, `throw new()` could be actually useful as a shorthand for `throw new Exception(...)`.
 
-M(new());
-```
-The invocation compiles until an overload like `void M(Bar bar) {}` is added alongside of the existing method.
+> **Open Issue:** should we allow usages of a target-typed `new` with comparison and arithmetic operators?
+
```
gafter: It is forbidden by the current spec because it uses the default constructor of a value type (for `1 + new()`). Probably worth stating.

alrz: We're talking about user-defined arithmetic operators. I'll mention it there.

proposals/target-typed-new.md
```diff
@@ -46,54 +51,51 @@ The type of the expression would be inferred from the target-type which would be
...
-- **Tuple types:** tuple literals should be used for such expressions.
 - **Array types:** arrays need a special syntax to provide the length.
+- **Struct default constructor**: this rules out all primitive types and most value types. If you wanted to use the default value of such types you could write `default` instead.
 
 All the other types that are not permitted in the *object_creation_expression* are excluded as well, for instance, pointer types.
 
-The default constructor for a value type (unless appeared explicitly in metadata) is excluded, which has the effect of excluding the primitive types and the default constructor for most value types. If you wanted to use the default value of such types you could write `default`.
-
-> **Open Issue:** What other types should be excluded?
+> **Open Issue:** should we allow delegates and tuples as the target-type?
```
jcouv: It's probably good to wait for the formal notes (@agocke should make them available next week), but here are the decisions I remember:
- tuple type: allow (many pros/cons, including discussion of `new(int, ..., int)` constraint and `(int, ..., int)` constructor for long tuple. Useful for folks migrating code between tuple and types)
- delegate type: somehow we missed that (didn't discuss)
- `throw new(...)`: disallow (guides users to heavily rely on `Exception` as opposed to dedicated types. That's undesirable)
- usage in binary operators: allow (that seems useful, unlike `i + default` and such)

proposals/target-typed-new.md
```diff
@@ -46,54 +51,51 @@ The type of the expression would be inferred from the target-type which would be
 with the following exceptions:
 
 - **Enum types:** not all enum types contain the constant zero, so it should be desirable to use the explicit enum member.
-- **Delegate types:** if the type is inferrable, an anonymous function can already be used.
 - **Interface types:** this is a niche feature and it should be preferable to explicitly mention the type.
-- **Tuple types:** tuple literals should be used for such expressions.
 - **Array types:** arrays need a special syntax to provide the length.
+- **Struct default constructor**: this rules out all primitive types and most value types. If you wanted to use the default value of such types you could write `default` instead.
```
alrz: The downside of this is that you can't use object initializer with default ctor, `(S)new() { x = 1 }`. `default` doesn't help here. Would that be enough reason to relax this? (note: now that we went with the late-filter, this restriction wouldn't make much difference in overload resolution).

jcouv: That's a good point. Let me raise the question of default constructor again.
I think the rationale for disallowing it was that you could just use `default`, but as your example shows, that's not the case (target-typed new would allow an initializer).

proposals/target-typed-new.md
jcouv: @alrz For the delegate type question, was there an example where that would be useful?

alrz: no, I couldn't think of any. method groups and anonymous funcs are already coverted through an implicit conversion.

proposals/target-typed-new.md
jcouv: Interestingly, I couldn't find notes for disallowing binding `new()` to default constructor. I'll ping Mads/Neal to confirm.

jcouv: Mads doesn't recall disallowing. Let's proceed under the assumption that it's allowed.

jcouv:
I'll merge and follow-up with some tune-ups in a separate PR.

Issue #881, opened 2017-09-05
Permit ternary operation with int? and double operands


gafter:
We would like to change the spec for https://github.com/dotnet/csharplang/issues/33 so that this proposed feature also permits the following:
``` c#
int? ni = 1;
double d = 4;
bool c = true;
var x1 = c ? ni : d;
```
This is currently an error because there is no implicit conversion between `int?` and `double` in either direction.

bondsbw:
To clarify, is the type of `x1` intended to be `double?` (as the greatest common type) or `int?` due to flow analysis?

eyalsk:
@bondsbw yes `x1` is intended to be `double?` as it can't be `int?` because there's no conversion between double to int or double? to int? and there's no flow analysis done here so the type of `var` is always one of the types you had to type yourself so the only option left here is `double?`.

MgSam:
@gafter I assume this implies lots of other conversions as well? 

- `short?` and `double`
- `short?` and `int`
- `int?` and `long`
- etc.

bondsbw:
How does this play out considering union/intersection types (#399)?

This proposal advocates that the result of a ternary expression is a union type, because `double?` is equivalent to `int? | double`.

But as @HaloFour commented [here](https://github.com/dotnet/csharplang/issues/33#issuecomment-279284400), an intersection type seems reasonable as well.  You should only be able to perform operations on the result which are common to both types.  The result of `var animal = condition ? new Dog() : new Cat()` should not allow `animal.Bark()` since it could be a `Cat`.

MgSam:
I'd say ternary should never make implicitly a union/intersection type but require you to instead be explicit. Implicit behavior is bound to lead to confusion and ambiguity for precisely the reasons you illustrate.

All academic, of course, since the team hasn't shown any interest in the feature.

bondsbw:
> All academic, of course, since the team hasn't shown any interest in the feature.

Perhaps, but implementing this proposal could impact future plans.

DavidArno:
Definitely with @bondsbw here (and presumably, by implication, @HaloFour too) that such changes as proposed here should be deferred until after any work (or formal rejection) of union and intersection types is complete.

MgSam:
@DavidArno So stop a nice feature we could have in a few months for the remote possibility of a feature in 5 or 10 years? Sounds like paralysis through analysis.

DavidArno:
@MgSam,

[...]

PR #1806, opened 2018-08-21
Add support for common type of `int` and `double?`.

gafter:
See https://github.com/dotnet/csharplang/issues/881

As requested by LDM about a year ago

### DIFF
--- a/proposals/nullable-enhanced-common-type.md
+++ b/proposals/nullable-enhanced-common-type.md
@@ -8,7 +8,7 @@
 ## Summary
 [summary]: #summary
 
-There is a situation in which the current common-type algorithm results are counter-intuitive, and results in the programmer adding what feels like a redundant cast to the code. With this change, an expression such as `condition ? 1 : null` would result in a value of type `int?`.
+There is a situation in which the current common-type algorithm results are counter-intuitive, and results in the programmer adding what feels like a redundant cast to the code. With this change, an expression such as `condition ? 1 : null` would result in a value of type `int?`, and an expression such as `condition ? x : 1.0` where `x` is of type `int?` would result in a value of type `double?`.
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


gafter:
@MadsTorgersen @dotnet/ldm Could someone please look at this and verify that it is as expected?

alrz:
There's a prototype over at https://github.com/alrz/roslyn/commit/67f9a3a119254d4aa965b71ed17cc6902e40ef50.

alrz:
@gafter Can you please take a look at the above commit to see if it's using the right approach. Thanks.

gafter:
I will look at it when back from vacation on Monday.

alrz:
HashSet can handle a `null` value, in the next iteration I'm going to use `null` to signify a null literal in the BestType algorithm. btw should this change include target-typing for `:?` and `??` expressions? If so I'd be happy to take a stab at it.

gafter:
@alrz I don't think the approach you are taking addresses the problem described in the title of this issue.  See also https://github.com/dotnet/csharplang/issues/881

If we target-type `??` and `?:` as we plan to, there will be less motivation for making this change.

alrz:
It's buried in the diff https://github.com/alrz/roslyn/blob/645c343114b9eec2c069242d231cfce95af1e40c/src/Compilers/CSharp/Test/Semantic/Semantics/NullableEnhancedCommonTypeTests.cs

 I think the compiler yields the correct result for both `int?/double => double?` and `double/null => double?` however, as I saw in that prototype (at least the way I wrote it), doing either of cases just does not fit into the current way of how things are done, we need to make a lot of exceptions and whatnot. 

>If we target-type ?? and ?: as we plan to, there will be less motivation for making this change.

makes sense. target-typing of those would have the same effect. 👍

PR #1843, opened 2018-09-08
Fixed rendering of OR in tables

svick:
Previously, OR operators were not rendering correctly in tables, because Markdown interprets the `|` symbol as cell separator (even between backticks).

This fixes the issue on GitHub preview. I'm not sure if it's going to improve rendering on docs.MS.


svick:
https://github.com/dotnet/csharplang/pull/1946 looks like a more comprehensive fix, so I think this PR is no longer needed.

PR #1916, opened 2018-10-07
Fix old mention of foreach await in async-streams

slang25:
Credit to Marc Gravell for spotting


CyrusNajmabadi:
tagging @jcouv

jcouv:
The current (tentative) syntax has not changed. It is still `foreach await (...)` as far as I know.
I think @stephentoub made a mistake in updating the proposal.

Closing. Thanks

stephentoub:
Ah, somehow I got it in my head that we changed that. Will fix.

CyrusNajmabadi:
👍 on changing the syntax :)   @jcouv knows my opinions here.

slang25:
Thanks all for the clarification

PR #1924, opened 2018-10-09
Add design notes

MadsTorgersen:


meetings/2018/LDM-2018-10-03.md
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+Here the `TryGetNext` method can be used to loop synchronously through all the readily available values. Only when it yields `false` do we need to fall back to an outer loop that awaits calls to `WaitForNextAsync()` until more data is available.
+
+In the degenerate case where `TryGetNext` always yields `false`, this is not faster: you fall back to the outer loop and always have two interface calls. However, the more often `TryGetNext` yields true, the more often we can stay in the inner loop and skip an interface call.
+
+The interface has several drawbacks, though:
+
+- It is meaningfully different from the synchronous `IAsyncEnumerator<T>`
```
paulomorgado: Shouldn't it be "synchronous `IEnumerable<T>`"?

meetings/2018/LDM-2018-10-03.md
```diff
@@ -0,0 +1,189 @@
...
+
+Directives would more likely allow editor-config integration.
+
+Attributes would maybe complicate the semantic model.
+
+Finally, "pseudo-attributes", recognized specially by the compiler, are a new concept to the language. Is it worth it? In practice, though, the seams are mostly not going to show: a user won’t generally need to worry that it’s not a regular attribute. Conversely, `#nonnull` would be a new directive: Is it worth it?
+
+## Conclusion
+
+Given all this, we are in favor of the directive approach. We will start from the strawman and refine over the coming weeks.
```
paulomorgado: Aren't attributes better for consumption? Or is this supposed to be source only and not reflected in consumption?

paulomorgado: I know it's not reflected in the runtime, but an attribute can indicate to consumers (other source code referencing the compiled code) information that compiler directives can't.

How does the compiler know that the reference assembly is NRT-aware? Or are there other annotations?

PR #1934, opened 2018-10-16
Update syntax to `await foreach`

jcouv:
FYI @stephentoub 
I'll go ahead and merge straight away.


NinoFloris:
Ooh, curious, could you point me/us to any relevant info why this was changed?

stephentoub:
> Ooh, curious, could you point me/us to any relevant info why this was changed?

@agocke has notes he'll be posting I believe.

jcouv:
@NinoFloris The notes likely won't be very detailed for that decision.
Between `async` and `await`, there is clear precedent for `await` indicating where some waiting happens. That's the "easy" part.
Between `await foreach/using` and `foreach/using await` there was a decision based on taste rather than a definitive rationale (that I can recall). Personally, I preferred `foreach await` a little.

jcouv:
@alrz Would you have a pointer for TS/ES? Are you saying they use `foreach await` or `await foreach`?

CyrusNajmabadi:
@jcouv see: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html

> Finally, ES2015 introduced the for..of statement as a means of iterating over an iterable. Similarly, the Async Iteration proposal introduces the for..await..of statement to iterate over an async iterable:

```ts
async function f() {
  for await (const x of g()) {
     console.log(x);
  }
}
```

alrz:
Thanks Cyrus. that being said, I think await foreach would definitely come off a as a surprise.

markusschaber:
In my eyes, `await foreach` looks like the whole foreach loop is run asynchroneously and awaited as a whole, while `foreach await` awaits "inside" the loop. :-)

YairHalberstadt:
@markusschaber
Exactly my thoughts

Issue #269, opened 2017-03-14
Proposal: Language support for Func & Action


TylerBrinkley:
Similar to what was done for `ValueTuple` in [roslyn#347](https://github.com/dotnet/roslyn/issues/347), `Func` and `Action` should be given language support in order to support naming parameters as well as making them more user friendly, especially for new .NET developers who are less acquainted with them. Named parameters can currently be specified when creating custom delegates but doing so is discouraged as they aren't assignable to the ubiquitous `Func` and `Action`. The table below shows what this feature might look like.

| Before | After |
| ------- | ----- |
| `Action` | `(() => void)` |
| `Action<decimal>` | `((decimal amount) => void)` |
| `Action<int, string>` | `((int index, string value) => void)` |
| `Func<DateTime>` | `(() => DateTime)` |
| `Func<DateTime, int>` | `((DateTime startDate) => int)` |
| `Func<int, int, int>` | `((int inclusiveMin, int exclusiveMax) => int)` |

The great thing about this is that existing API's can be updated to provide better documentation of the input parameters of these delegates without breaking compatibility.

## Syntax
The general syntax is `((inputParameters) => returnType)` where `inputParameters` are a comma separated list of the input parameters or empty when there are no input parameters.

### Enclosing parentheses are required
Enclosing parentheses are required to distinguish the alias as a type.

### Arrow
The arrow `=>` separates the input parameters from the return type. It was chosen as it mirrors lambda and expression bodied members syntax.

### No input parameters syntax
The syntax for no input parameters is a set of empty parentheses, `()`. The set of empty parentheses was chosen as it mirrors lambda's syntax for no input parameters.

### No return parameter syntax
The syntax for no return parameter, e.g. `Action` delegates, is `void` as this mirrors the existing method syntax.

### Parameter names
The parameter names are optional and would only be stored at API surface levels in an `Attribute` similar to how tuples work.

## Alias syntax within methods
Additionally this syntax could also be used within methods to declare variables so the following

```c#
Func<int, string> stringifier = a => a.ToString();
```

could become

```c#
((int value) => string) stringifier = a => a.ToString();
```

## Existing BCL Use Cases
To show how ubiquitous `Func` and `Action` are in the BCL, when going through the preliminary [.NET Standard 2.0 spec](https://raw.githubusercontent.com/dotnet/standard/master/docs/versions/netstandard2.0_ref.md) there was found around 544 references to `Func` and around 121 references to `Action`.

[...]

PR #1951, opened 2018-10-23
Function Pointer Proposal

jaredpar:
This is the function pointer feature reviewed in LDM on 10/15. Notes for that meeting are [here](https://github.com/dotnet/csharplang/blob/master/meetings/2018/LDM-2018-10-15.md). This is an alternate approach to [intrinsics](https://github.com/dotnet/csharplang/blob/master/proposals/intrinsics.md) for exposing the `calli` and `ldftn` instructions.

 This is the proposal as discussed in the meeting. Uploading this version so that the notes will make a bit more sense. Will upload a new version, based on the changes discussed in LDM, in the next few days.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -17,23 +17,24 @@ This is an alternate design propsoal to [compiler intrinsics](https://github.com
 ## Detailed Design 
 
 ### funcptr
-The language will allow for the declaration of function pointers using the `funcptr` syntax. This form closely 
-resembles the syntax form of `delegate`.
+The language will allow for the declaration of function pointers using the `funcptr` syntax. The declaration and usage
+of function pointers closely resemble that of `delegate`:
 
 ``` csharp
-delegate void DAction(int a);
-funcptr void FAction(int a);
+unsafe class Example {
+    delegate void DAction(int a);
+    funcptr void FAction(int a);
+
+    void Example(DAction d, FAction f) {
+        d(42);
+        f(42);
+    }
+}
 ```
 
 These types are represented using the function pointer type as outlined in ECMA-335. This means invocation
 of a `funcptr` will use `calli` where invocation of a `delegate` will use `callvirt` on the `Invoke` method.
 Syntactically though invocation looks no different:
-
-``` csharp
-void Example(FAction f) {
-    f(42);
-}
-```
 
 The `calli` instruction requires the calling convention be specified as a part of the invocation. The default 
 for `funcptr` will be managed. Alternate forms can be specified by adding the appropriate modifier after the 
@@ -44,44 +45,147 @@ for `funcptr` will be managed. Alternate forms can be specified by adding the ap
 funcptr cdecl int Square(int value);
 ```
 
+A `funcptr` type is a pointer type which means it has all of the capabilities and restrictions of a standard pointer
+type:
+- Only valid in an `unsafe` context.
+- Methods which contain a `funcptr` parameter or return type can only be called from an `unsafe` context.
+- Cannot be converted to `object`.
+- Cannot be used an a generic argument.
+- Can implicitly convert to and from `void*`
+
 Conversions between `funcptr` types is done based on their signature, not name. Hence when two `funcptr` 
-declarations have the same signature they have an identity conversion no matter what their name is:
+declarations have the same signature they have an identity conversion no matter what the name is:
 
 ``` csharp
-funcptr int Sum(int left, int right);
-funcptr int Add(int x, int y);
+unsafe class Example {
+    funcptr int Sum(int left, int right);
+    funcptr int Add(int x, int y);
+    funcptr int Echo(int x);
 
-void Conversions() {
-    Sum s = ...;
-    Add a = s; // okay
-    Console.WriteLine(a == s); // True
+    void Conversions() {
+        Sum s = ...;
+        Echo e = ...;
+
+        Add a1 = s; // okay
+        Add a1 = e; // error: incompatible signatures
+        Console.WriteLine(a1 == s); // True
+    }
 }
 ```
 
-The use of ECMA-335 function pointer types means instances are not convertible to `objec
+In addition to declaring a named `funcptr` type, as you declare a `delegate`, it is possible to use an unnamed 
+`funcptr` type directly. This type can be used anywhere a type declaration would occur:
 
-Invocations of a `funcptr` will use the `calli` instruction. 
-
-The `funcptr` type differs from a `delegate` in the following ways:
+``` csharp
+unsafe struct Example {
+    funcptr int (int) Field;
+    unsafe void UnnamedExample(funcptr int(int) ptr) {
+        int x = ptr(42);
+        Field = ptr;
+        ...
+    }
+}
+```
 
 Restrictions:
 
 - Cannot overload when the only difference in parameter types is the name of the function pointer. 
 - Custom attributes cannot be applied to a `funcptr` or any of its elements.
-- A `funcptr` type is not convertible to `object`. 
+- A `funcptr` parameter cannot be marked as `params`
+- A `funcptr` type has all of the restrictions of a normal pointer type.
 
-### Address of functions
+### Allow addresss-of to target methods
 
-### Metadata
+Method groups will now be allowed as arguments to an address-of expression. The type of such an 
+expression will be an unnamed `funcptr` which has the equivalent signature of the target method:
 
+``` csharp
+unsafe class Util { 
+    public static void Log() { } 
+
+    funcptr void Action();
+    funcptr int Func();
+    void Use() {
+        funcptr void() ptr1 = &Util.Log; 
+        Action ptr2 = &Util.Log;
+
+        // Error: type "funcptr void()" not compatible with "funcptr int()";
+        Func ptr3 = &Util.Log; 
+
+        // Okay. Conversion to void* is always allowed.
+        void* v = &Util.Log;
+   }
+}
+```
+
+The conversion of an address-of method group to `funcptr` has roughly the same process as method group to `delegate`  
+conversion. The only additional restriction is that only members of the method group that are marked as `static` will
+be considered. This means developers can depend on overload resolution rules to work in conjunction with the 
+address-of operator:
+
+``` csharp
+unsafe class Util { 
+    public static void Log() { } 
+    public static void Log(string p1) { } 
+    public static void Log(int i) { };
+
+    funcptr void Action1();
+    funcptr void Action2();
+
+    void Use() {
+        Action1 a1 = &Log; // Log()
+        Action2 a2 = &Log; // Log(int i)
+
+        // Error: ambiguous conversion from method group Log to "void*"
+        void* v = &Log; 
+    }
+```
+
+The address-of operator will be implemented using the `ldftn` instruction.
+
+Restrictions of this feature:
+
+- Only applies to methods marked as `static`.
+- Local functions cannot be used in `&`. The implementation details of these methods are
+deliberately not specified by the language. This includes whether they are static vs. instance or
+exactly what signature they are emitted with.
+
+### Better function member
+The better function member specification will be changed to include the following line:
+
+> A `funcptr` is more specific than `void*`
+
+This means that it is possible to overload on `void*` and a `funcptr` and still sensibly use the address-of operator.
 
 ## Open Issuess
 
-### Issue 1
-
-### Issue 2
+- Round tripping function pointer names, as well as parameter names, through metadata will require additional work. The
+function pointer type itself is natively supported by CLI but that does not include any names. This is not anticipated
+to be a big issue, just needs design work.
 
 ## Considerations
+
+### Don't require unsafe at declaration
+Instead of requiring `unsafe` at every use of a `funcptr`, only require it at the point where a method group is
+converted to a `funcptr`. This is where the core safety issues come into play (knowing that the containing assembly
+cannot be unloaded while the value is alive). Requiring `unsafe` on the other locations can be seen as excessive.
+
+This is how the design was originally intended. But the resulting language rules felt very awkward. It's impossible to
+hide the fact that this is a pointer value and it kept peeking through even without the `unsafe` keyword. For example 
+the conversion to `object` can't be allowed, it can't be a member of a `class`, etc ... The C# design is to require
+`unsafe` for all pointer uses and hence this design follows that.
+
+Developers will still be capable of preventing a _safe_ wrapper on top of `funcptr` values the same way that they do 
+for normal pointer types today. Consider:
+
+``` csharp
+unsafe struct Action {
+    funcptr void() _ptr;
+
+    Action(funcptr void() ptr) => _ptr = ptr;
+    public void Invoke() => _ptr();
+}
+```
 
 ### Using delegates
 Instead of using a new syntax element, `funcptr`, simply use exisiting `delegate` types with a `*` following the type:
@@ -106,7 +210,25 @@ must be different than the version defined in mscorlib.
 One option that was explored was emitting such a pointer as `mod_req(Func<int>) void*`. This doesn't 
 work though as a `mod_req` cannot bind to a `TypeSpec` and hence cannot target generic instantiations.
 
-### Unsafe
-Requiring the unsafe keyword 
+### No names altogether
+Given that a `funcptr` can be used without names why even allow names at all? The underlying CLI primitive doesn't have
+names hence the use of names is purely a C# invention. That ends up being a leaky abstraction in some cases (like
+not allowing overloads when `funcptr` differ by only names). 
+
+At the same time, `funcptr` look and feel so much like `delegate` types, not allowing them to be named would be seen
+as an enormous gap by customers. The leaky abstraction is wort the trade offs here.
+
+### Requiring names always
+Given that names are allowed for `funcptr` why not just require them always? Given that `funcptr` is an existing CLI
+type there are uses of it in the ecosystem today. None of those uses will have the metadata serialization format 
+chosen by the C# compiler. This means the feature would be using CLI function pointers but not interopting with any 
+existing usage.
 
 ## Future Considerations
+
+### static local functions
+
+This refers to [the proposal](https://github.com/dotnet/csharplang/issues/1565) to allow the 
+`static` modifier on local functions. Such a function would be guaranteed to be emitted as 
+`static` and with the exact signature specified in source code. Such a function should be a valid
+argument to `&` as it contains none of the problems local functions have today
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -16,11 +16,64 @@ This is an alternate design propsoal to [compiler intrinsics](https://github.com
 
 ## Detailed Design 
 
-### Item 1
+### funcptr
+The language will allow for the declaration of function pointers using the `funcptr` syntax. This form closely 
+resembles the syntax form of `delegate`.
+
+``` csharp
+delegate void DAction(int a);
+funcptr void FAction(int a);
+```
+
+These types are represented using the function pointer type as outlined in ECMA-335. This means invocation
+of a `funcptr` will use `calli` where invocation of a `delegate` will use `callvirt` on the `Invoke` method.
+Syntactically though invocation looks no different:
+
+``` csharp
+void Example(FAction f) {
+    f(42);
+}
+```
+
+The `calli` instruction requires the calling convention be specified as a part of the invocation. The default 
+for `funcptr` will be managed. Alternate forms can be specified by adding the appropriate modifier after the 
+`funcptr` keyword: `cdecl`, `fastcall`, `stdcall`, `thiscall` or `winapi`. Example:
+
+``` csharp
+// This method will be invoked using the cdecl calling convention
+funcptr cdecl int Square(int value);
+```
+
+Conversions between `funcptr` types is done based on their signature, not name. Hence when two `funcptr` 
+declarations have the same signature they have an identity conversion no matter what their name is:
+
+``` csharp
+funcptr int Sum(int left, int right);
+funcptr int Add(int x, int y);
+
+void Conversions() {
+    Sum s = ...;
+    Add a = s; // okay
+    Console.WriteLine(a == s); // True
+}
+```
+
+The use of ECMA-335 function pointer types means instances are not convertible to `objec
+
+Invocations of a `funcptr` will use the `calli` instruction. 
+
+The `funcptr` type differs from a `delegate` in the following ways:
 
 Restrictions:
 
-### Item 2
+- Cannot overload when the only difference in parameter types is the name of the function pointer. 
+- Custom attributes cannot be applied to a `funcptr` or any of its elements.
+- A `funcptr` type is not convertible to `object`. 
+
+### Address of functions
+
+### Metadata
+
 
 ## Open Issuess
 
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
+# Function Pointers
+
+## Summary
+This proposal privodes language constructs that expose IL opcodes that cannot currently be accessed efficiently,
```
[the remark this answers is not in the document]
jaredpar: Much nicer that I can now just approve everyone fixing my mispellings than having to do it all by hand myself. :)

jaredpar:
Merging so I can put good links in the LDM notes. Please continue to provide feedback.

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+        // Okay. Conversion to void* is always allowed.
+        void* v = &Util.Log;
+   }
+}
+```
+
+The conversion of an address-of method group to `funcptr` has roughly the same process as method group to `delegate`  
+conversion. There are two additional restrictions to the existing process:
+- Only members of the method group that are marked as `static` will be considered. 
+- Only a `funcptr` with a managed calling convention can be the target of such a conversion.
```
jkotas: This makes it unusable for interop that is one of the main scenarios for this.

Can we tie this with the `NativeCallableAttribute` to make it work for interop? 
- If the target method is not marked with `NativeCallableAttribute`, the result is function pointer with the managed calling convention
- If the target method is marked with `NativeCallableAttribute`, the result is function pointer with the unmanaged calling convention specified in the `NativeCallableAttribute`.

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+These types are represented using the function pointer type as outlined in ECMA-335. This means invocation
+of a `funcptr` will use `calli` where invocation of a `delegate` will use `callvirt` on the `Invoke` method.
+Syntactically though invocation is identical for both constructs.
+
+The `calli` instruction requires the calling convention be specified as a part of the invocation. The default 
+for `funcptr` will be managed. Alternate forms can be specified by adding the appropriate modifier after the 
+`funcptr` keyword: `cdecl`, `fastcall`, `stdcall`, `thiscall` or `winapi`. Example:
```
jkotas: We should have identifier for the default unmanaged calling convention - what you get using `default(CallingConvention)` today. Would `funcptr unmanaged` work?

winapi maps to default unmanaged calling convention, but it is not a very good name because if it is Windows-specific. It should be fine to not have it.

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+    funcptr void FAction(int a);
+
+    void Example(DAction d, FAction f) {
+        d(42);
+        f(42);
+    }
+}
+```
+
+These types are represented using the function pointer type as outlined in ECMA-335. This means invocation
```
jkotas: What is the named funcptr going to look like in the metadata? I know that ECMA-335 allows unnamed funcptrs, but I do not recall named funcptrs.

jkotas: Ok, I see that this is in the Open Issues.

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+exactly what signature they are emitted with.
+
+### Better function member
+The better function member specification will be changed to include the following line:
+
+> A `funcptr` is more specific than `void*`
+
+This means that it is possible to overload on `void*` and a `funcptr` and still sensibly use the address-of operator.
+
+## Open Issuess
```
jkotas: Typo: Issuess

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+potential implementation of the feature): 
+
+https://github.com/dotnet/csharplang/issues/191
+
+This is an alternate design propsoal to [compiler intrinsics](https://github.com/dotnet/csharplang/blob/master/proposals/intrinsics.md)
+
+## Detailed Design 
+
+### funcptr
+The language will allow for the declaration of function pointers using the `funcptr` contextual keyword. The
```
mikedn: Hmm, `funcptr`, kind of ugly for various reasons, including the fact that C#/.NET tends to deal in methods and not functions. And I'm no fan of the C/C++ syntax for function pointers but it's weird that C# data pointers are declared like C/C++ data pointers while C# function pointers, named or unnamed, require this distinct syntax.

And coupled with the names/no names issue, it's even more questionable. This introduces a very specific way to assign a name to an otherwise unnamed type even though such a feature (typedef basically) is requested from time to time. Not saying that C# should get typedef but this seems like a "let's rush things up and get this done" kind of thing.

Maybe `funcptr`, if added, should be just syntactic sugar for:
```
unsafe struct Action { 
    void (*_ptr)(); 
    Action(void (*ptr)()) => _ptr = ptr; 
    public void Invoke() => _ptr(); 
}
```
I guess this kind of shots down half of the proposal. But hey, just my 2 cents :)

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+> A `funcptr` is more specific than `void*`
+
+This means that it is possible to overload on `void*` and a `funcptr` and still sensibly use the address-of operator.
+
+## Open Issuess
+
+- Round tripping function pointer names, as well as parameter names, through metadata will require additional work. The
+function pointer type itself is natively supported by CLI but that does not include any names. This is not anticipated
+to be a big issue, just needs design work.
+- The address-of operator is limited to `static` methods in this proposal. It can be made to work with instance methods
```
jkotas: FWIW, this would not always work. Static methods and instance methods do not have the same calling convention. static function pointer to instance method cannot be manufactured by pre-pending "this" pointer.

proposals/function-pointers.md
jaredpar: That is pretty doable. We'd likely need to expose that attribute though and make it a supported API

proposals/function-pointers.md
jaredpar: Overall that seems fine. The proposal today though doesn't rely on the type `CallingConvention`. Is there a different way to calculate the default?

proposals/function-pointers.md
jaredpar: >Maybe funcptr, if added, should be just syntactic sugar for

In this case it can't really be. The function pointer concept is actually a CLI primitive value. The goal is to just directly expose that.

proposals/function-pointers.md
jaredpar: Where can I get some more info on how this does and doesn't work?

proposals/function-pointers.md
mikedn: > In this case it can't really be. The function pointer concept is actually a CLI primitive value. The goal is to just directly expose that.

Yep, I wasn't very clear. The idea was to expose that primitive value through a more primitive syntax - `void (*_ptr)();` or something similar, that doesn't need a keyword that tries to look like a delegate and ends up looking like a typedef.

xoofx:
I'm really glad to see this feature coming 👍 

Not a super fan about the introduction about a new keyword ` funcptr` while it's quite explicit, I would have preferred maybe to recycle something like `static delegate`... but I can see that it would not be as explicit (static referring to storage while here it would be slightly related but still different)... so yep, not sure if we could come up with a better declaration for it...

ltrzesniewski:
Why `static delegate` when you can write `unmanaged delegate`? 😉

mikedn:
> Why static delegate when you can write unmanaged delegate?

Hmm, that seems better indeed.

Lakritzator:
Some time ago I was discussing performance of calling native Win32 APIs, and was referred to the project [AdvanceDLSupport](https://github.com/Firwood-Software/AdvanceDLSupport), especially the documentation on [indirect calling](https://github.com/Firwood-Software/AdvanceDLSupport/blob/master/docs/indirect-calling.md).

Using AdvanceDLSupport did improve the performance but it also impacted the way I write my projects.
Can I use any of this to improve DllImport performance, or can this be used internally to assist?

ltrzesniewski:
@Lakritzator I did a couple benchmarks [here](https://github.com/ltrzesniewski/Benchmarks/tree/master/NativeInterop), and my conclusion is that `calli` isn't faster than P/Invoke when you do P/Invoke right (which means using `[SuppressUnmanagedCodeSecurity]` on the .NET framework for instance). The runtime still needs to do its stuff at each call.
Also, I found no way to tell it to marshal strings by pointer instead of copying them.

Anyway this proposal aims to let you access `calli` from C# without the need for external tools or run-time codegen.

jkotas:
> AdvanceDLSupport, especially the documentation on indirect calling.

Correct, the indirect calls via calli do not have any fundamental performance advantage over regular DllImport. The performance difference you have seen can be caused by:
- Security checks on .NET Framework. `SuppressUnmanagedCodeSecurityAttribute` will help.
- Using complex inefficient built-in marshaler (e.g. for non-blitable structs). You can fix this by using blitable types in your interop - blitable types are always as fast as they can be.
- AdvanceDLSupport taking shortcuts on implementing the same marshaling. From https://github.com/Firwood-Software/AdvanceDLSupport/blob/master/docs/indirect-calling.md: One of the main contributing factors to calli's speed improvements is that it skips certain types of call marshalling that would normally be performed. As such, you may see unexpected results if you rely on them.
- AdvanceDLSupport hitting path that results into broken (but faster) code: https://github.com/Firwood-Software/AdvanceDLSupport/blob/1b7394211a655b2f77649ce3b610a3161215cbdc/AdvancedDLSupport/ImplementationGenerators/Terminating/IndirectCallMethodImplementationGenerator.cs#L176

xoofx:
> Why static delegate when you can write unmanaged delegate?

yep, why not, I like it 👍

jaredpar:
Couple reasons for not using `unmanaged delegate`:

- The LDM notes essentially pushed us even further away from delegates by removing named function pointers altogether. It makes me even more wary of naming them anything `delegate` as it may set incorrect expectations.
- The keyword `unmanaged` is already used in the language and has a substantially different meaning.

The exact syntax for the declarations is still being worked out but I think it will look closer to C style function pointers than C# delegates in the end.

proposals/function-pointers.md
jkotas: Hmmm, the enum that describes calling conventions in signatures does not have neither `winapi` nor good value for default unmanaged calling convention:

https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Reflection.Metadata/src/System/Reflection/Metadata/Signatures/SignatureCallingConvention.cs#L22

Maybe we should consider adding value for the default unmanaged calling convention here? (Of course, it would work on new runtimes only...)

proposals/function-pointers.md
jkotas: Maybe we can reduce the number of keywords by using the calling convention enum to describe the unmanaged calling convention details? And use the variant without parameter to map to the default calling convention (if we do something about it)?

`funcptr unmanaged(CallingConvention.Cdecl) int F1()`
`funcptr unmanaged int F2()`

Note that there may be more unmanaged calling conventions coming in future: Windows have introduced vectorcall calling convention some time ago that .NET does not support yet.

proposals/function-pointers.md
svick: @jkotas 

> FWIW, this would not always work. Static methods and instance methods do not have the same calling convention. static function pointer to instance method cannot be manufactured by pre-pending "this" pointer.

Why not? What are the cases where it would not work?

jkotas: Methods with return buffers. For example:

```csharp
struct BigStruct { long a,b,c,d,e,f; }

class MyClass {
    BigStruct MyMethod();
}
```

On Windows x64, `MyMethod` takes `this` in `rcx` and return buffer pointer in `rdx`.

Static `BigStruct pfn(MyClass)` function pointed takes return buffer in `rcx` and `MyClass` argument in `rdx`.

svick:
First of all, why is this the first time we (the community) hear about this proposal? If C# is designed in the open, then I think there should be an issue to discuss each proposal *before* it goes to the LDM, especially since LDM notes are still sometimes published with a delay. As far as I can tell, there is still no issue for this proposal (this merged PR does not count and I think https://github.com/dotnet/csharplang/issues/191 does not either, since it's a very different proposal).

---

Second, I feel like this proposal is too heavyweight, especially when it comes to syntax, for such a niche feature. I would much rather have something that doesn't change the syntax much. I think the `funcptr void FAction(int a);` syntax does that reasonably well. I don't see the need for having calling convention keywords, when an attribute could be used as well (maybe reuse `[UnmanagedFunctionPointer]`). And I'm not sure syntax for unnamed function pointers is necessary, even if it's closer to CLI.

---

Third, how would this handle parameter variations that are encoded as attributes for methods and delegates? That includes tuples with names, `in`, `out` and `dynamic`.

---

Fourth, if you're going to add the unnamed syntax, have you considered the relationship of that with a unified syntax for delegates (https://github.com/dotnet/csharplang/issues/470, https://github.com/dotnet/csharplang/issues/269). On the surface, they look very similar, even if their goals and problems are very different. I wouldn't want to get into a situation in the future where we have one syntax for unnamed function pointers and a different syntax for unnamed delegates.

---

Fifth, I think the issues with requiring names always are solvable. The proposal doesn't say why exactly interop with existing unnamed function pointers would be an issue, but I think it's cases like overriding.

E.g. you have this code (implemented in IL):

```c#
class Base
{
    public virtual void F(funcptr int(int) ptr) {}
}
```

and you want to override the method without having access to the unnamed syntax. I think it should be possible to override it with a named function pointer like this:

```c#
class Derived : Base
{
    funcptr int FP(int arg);
    public override void F(FP ptr) {}
}
```

The language already sometimes allows overriding with a different type if there's an identity conversion between the types (if the difference is switching `dynamic` for `object`), so I think this would not be too much of a stretch.

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+Instead of requiring `unsafe` at every use of a `funcptr`, only require it at the point where a method group is
+converted to a `funcptr`. This is where the core safety issues come into play (knowing that the containing assembly
+cannot be unloaded while the value is alive). Requiring `unsafe` on the other locations can be seen as excessive.
+
+This is how the design was originally intended. But the resulting language rules felt very awkward. It's impossible to
+hide the fact that this is a pointer value and it kept peeking through even without the `unsafe` keyword. For example 
+the conversion to `object` can't be allowed, it can't be a member of a `class`, etc ... The C# design is to require
+`unsafe` for all pointer uses and hence this design follows that.
+
+Developers will still be capable of preventing a _safe_ wrapper on top of `funcptr` values the same way that they do
```
svick: What does it mean to "prevent a safe wrapper"? Did you mean "provide"?

proposals/function-pointers.md
```diff
@@ -0,0 +1,245 @@
...
+Func<object, object, bool>* ptr = &object.ReferenceEquals;
+```
+
+Handling calling convention can be done by annotating the `delegate` types with an attribute that specifies 
+a `CallingConvention` value. The lack of an attribute would signify the managed calling convention.
+
+Encoding this in IL is problematic. The underlying value needs to be represented as a pointer yet it also must:
+
+1. Have a unique type to allow for overloads with different function pointer types. 
+1. Be equivalent for OHI purposes across assembly boundaries.
```
svick: What's OHI? I found it in https://github.com/dotnet/roslyn/issues/20528, where it means "override, hiding, implementing", but I think it's not a familiar acronym and so it should be expanded here.

proposals/function-pointers.md
svick: @jkotas Thanks for the explanation.

jaredpar:
@svick 

> First of all, why is this the first time we (the community) hear about this proposal? If C# is designed in the open, then I think there should be an issue to discuss each proposal before it goes to the LDM,

Generally I like to post my proposals a couple days before LDM meetings for everyone to see. This time though we shuffled around the topics fairly last minute, and as a result I didn't finish the proposal till the morning of the meeting.

I published it in the exact form it had for the LDM discussion vs. adjusting for LDM feedback and then publishing. So there isn't any loss of information here, it will line up with the notes from the meeting and I'm still processing all the feedback left. 

> since LDM notes are still sometimes published with a delay

The delay these days is generally two days now that @agocke owns notes. The delay for this specific meeting was one week and that's just due to us trying to hit the Dev16 Preview 1 ship dates. 

> As far as I can tell, there is still no issue for this proposal (this merged PR does not count and I think #191 does not either, since it's a very different proposal).

I didn't file a new issue for this because in many ways it's still #191. This proposal is trying to solve the same set of problems: provide access to `calli` and `ldftn`. The mechanism is very different yes but it's still the same issue we're looking at. In general we keep the same issue open in cases like this until we settle on a design. 

Note: this is the third proposed implementation of compiler intrinsics, not the second. All three have used #191 as the issue.

> I don't see the need for having calling convention keywords, when an attribute could be used as well (maybe reuse [UnmanagedFunctionPointer])

There are several problems problems with the attribute approach:

1. Conversions are done based on method signatures and there is no place in C# today where attributes factor into such conversions. 
1. Function pointers can't have attributes as a rule. Adding an exception for this case would be strange. 
1. What would the semantic model return for such an attribute? It is not, and cannot, be an attribute in the metadata sense. It would be emitted as part of the signature element of the function pointer. Hence it' would be a very leaky abstraction.
1. The set of calling conventions is not actually fixed hence an enum isn't a good choice to represent the set of values. 

A simple contextual keyword is pretty light weight here. 

>I would much rather have something that doesn't change the syntax much

LDM disagrees here. The feedback from the second intrinsics proposal was specifically that we'd avoided pretty much any new syntax and as a result the feature felt "cute" vs. well designed. This proposal to give more syntax is a direct result of this previous meeting.

 > And I'm not sure syntax for unnamed function pointers is necessary, even if it's closer to CLI.

LDM disagrees here. The result of the 10/15 meeting is to remove all names and only have the unnamed form. 

> Third, how would this handle parameter variations that are encoded as attributes for methods and delegates? That includes tuples with names, in, out and dynamic

Most of these are TBD. I'm pretty sure we can get most of the features like in and out working. Those can be encoded as `modopt` and `modreq` and hence can fit into a CLI function pointer usage. The same could be done for  `dynamic` but I'm not sure we're going to find value in doing that. Have some meetings scheduled with CLR team to work out some of these details. 

> Fourth, if you're going to add the unnamed syntax, have you considered the relationship of that with a unified syntax for delegates

A little bit. I plan on looking at it more later now that the emphasis is on unnamed delegates.

> Fifth, I think the issues with requiring names always are solvable.

This is essentially what I suggested in LDM. Still the decision was against names of any kind hence it's a moot point. 😦

svick:
@jaredpar Thanks for the explanations.

I still think a separate issue to discuss this proposal would be nice. If you really want to reuse https://github.com/dotnet/csharplang/issues/191, then I think it's important that its top post is updated and also a comment added.

Right now, people who follow https://github.com/dotnet/csharplang/issues/191 or who read it today have almost no way of knowing that this proposal even exists.

> The set of calling conventions is not actually fixed hence an enum isn't a good choice to represent the set of values.

Doesn't that mean language keywords are an even worse choice? An `enum` can have different values for different target frameworks and you can even cast an `int` to it if you know it's a valid value not exposed by the standard library. You can't do any of that with a set of contextual keywords.

I guess with your arguments against attributes, that would point to @jkotas's proposal to use something like `unmanaged(CallingConvention.Cdecl)`.

jaredpar:
>  You can't do any of that with a set of contextual keywords.

That's intended here and a bonus. When emitting to IL the compiler has to determine which bits to set in the calling convention section for a given calling convention. That does not necessarily line up with the values in any particular enum. Additionally not all runtimes support all calling conventions hence allowing a cast to `int` gives users an opportunity to introduce unsupported behavior. A small set of contextual keywords solves both these problems:

- Compiler knows the mapping between calling convention and enconding.
- Compiler knows which runtimes support which calling conventions and can issue appropriate errors when there is a mismatch. 

> I still think a separate issue to discuss this proposal would be nice

I'll likely create a new issue once I get this updated based on LDM feedback.

HaloFour:
@jaredpar 

> There are several problems problems with the attribute approach

I don't understand this argument.  That's one of the reasons why (pseudo-)attributes exist and this is *exactly* how P/Invoke was designed.  That calling convention is embedded directly in IL metadata, not as a custom attribute.

I'm not taking a side on the argument, I just don't understand the position that attributes can't be used here when the language already demonstrates otherwise.

jkotas:
> A small set of contextual keywords solves both these problems

I would think that these problems can be solved with the calling convention enum equally well in the compiler. The code can be either look like this:

```csharp
switch (keyword)
{
case "cdecl":
   if (cdecl not supported) throw; 
   encode cdecl;
   break;
case "stdcall":
   if (stdcall not supported) throw; 
   encode stdcall
   break;
...
}
```

or look like this:

```csharp
switch (callconv)
{
case CallingConvention.CDecl:
   if (cdecl not supported) throw; 
   encode cdecl;
   break;
case CallingConvention.StdCall:
   if (cdecl not supported) throw; 
   encode stdcall
   break;
...
}
```

jaredpar:
@HaloFour 

> I don't understand this argument

Several points were raised. Which one did you not understand? 

>  That's one of the reasons why (pseudo-)attributes exist and this is exactly how P/Invoke was designed. 

P/Invoke doesn't have to deal with type identity / conversions issues. Function pointers must address that. There is no case today where user attributes affect type identity in the language. Also please consider the points about API and encoding

Bottom line for me: these would never be represented as attributes hence why use them? It's an explicit lie that we have to spend lots of time covering up.

CyrusNajmabadi:
> Bottom line for me: these would never be represented as attributes 

I think the different perspective comes from how one views an attribute.  It can be a very .net/metadata-centric view, in which case using attributes that don't end up as actual IL attributes seems weird.  *Or* one can have a very C#-centric view, where this is jsut a flexible language construct that can be used to annotate and configure code, without necessarily having that annotation ever be reflected (no pun intended) at runtime.  In hte latter view of the world, attributes serve as a very useful tool to help avoid a continual stream of 'one off' language/syntactic features around lanuage knobs.

HaloFour:
@CyrusNajmabadi 

> Or one can have a very C#-centric view, where this is jsut a flexible language construct that can be used to annotate and configure code, without necessarily having that annotation ever be reflected (no pun intended) at runtime. In hte latter view of the world, attributes serve as a very useful tool to help avoid a continual stream of 'one off' language/syntactic features around lanuage knobs.

This is where I'm coming from.  Granted, pseudo-attributes are fairly rare in .NET.  But I do understand the issues that are being raised here, and if you had to use attribute syntax to specify some of this metadata but you couldn't apply any other attributes that would certainly be strange.

pentp:
@jkotas's proposal to use `unmanaged(CallingConvention.Cdecl)` could be better for discoverability of calling conventions.

This proposal does not cover at least half of #191 - `handleof (ldtoken)`. Better to have a separate discussion thread for this proposal.

My only other concern is that this should leave open the future possibility of adding `ldftn` __and__ `ldvirtftn` for instance methods (e.g. `&this.VirtMethod` and `&VirtMethod`). At first look this seems to be the case, but I'm not a language designer 😄

proposals/function-pointers.md
jaredpar: OHI is the term to describe how method signature are compared when we are doing interface implementation, override and member hiding hookup. More specifically:

``` csharp
class Base {
  protected virtual void M(int i) { ... }
}
class Derived : Base { 
  protected override void M(int j) { ... }
}
```

The rules for determining if `Deride.M` is a proper override for `Base.M` is generally referred to as OHI.

proposals/function-pointers.md
jaredpar: Yep I meant provide

proposals/function-pointers.md
mjsabby: Is NativeCallableAttribute exposed in .NET Framework? This will make this feature not useful on .NET Framework or until we decide to put NativeCallableAttribute in .NET Framework. Any reason why we can't have this for all supported calling conventions with types already present on desktop and .net core?

jkotas: > Is NativeCallableAttribute exposed in .NET Framework

Nope. (In fact, it is not exposed in the public surface of any shipping runtime today.) 

> Any reason why we can't have this for all supported calling conventions with types already present on desktop and .net core?

What would you propose? I do not think that the runtime support required to make this work well exist in desktop.

PR #1987, opened 2018-11-06
Solution wide define constants

RUSshy:


NickCraver:
You can accomplish this with Directory.Build.props today - this is an MSBuild level problem, rather than a Roslyn one - as it runs per-project in general. Here’s how it works: 
https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2017#directorybuildprops-and-directorybuildtargets

PR #1999, opened 2018-11-13
Capture returned object

RUSshy:


YairHalberstadt:
Can I suggest that this might be better off discussed in a Feature Request?
You can create one by opening an issue on the csharplang repo.

RUSshy:
Oh sorry i didn't know that, i created it there https://github.com/dotnet/csharplang/issues/2000

Issue #49, opened 2017-02-09
Champion "Covariant Return Types"


gafter:
- [ ] Proposal added: [proposals/covariant-returns.md](https://github.com/dotnet/csharplang/blob/master/proposals/covariant-returns.md)
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

Joe4evr:
> This would apply to methods and properties, and be supported in classes and interfaces.

The part about support in interfaces, does this mean something like the following?
```cs
public interface IFoo { }
public interface IBar
{
    IFoo Get();
}
public interface IBaz : IFoo { }
public interface IQux : IBar
{
    override IBaz Get(); //or would it be 'override IBaz IBar.Get();'?
}
```
I guess it's lucky that discussions around Default Interface Implementations have [pushed back a little bit on using `override` as a keyword](https://github.com/dotnet/csharplang/blob/master/meetings/2017/LDM-2017-05-31.md#default-interface-members-overriding-or-implementing) for just that. Even if it wasn't, there probably could've been a way for the two to co-exist, but if you're talking about adding one new thing to interfaces, it would be good to consider the other things that are proposed to be added to interfaces, and make sure that one design choice doesn't preclude any other features entirely.

JonHanna:
I find myself wishing C# supported covariant returns very often, and often with code that isn't particularly "clever".

TylerBrinkley:
Covariant returns would be very useful for typed cloning methods, preventing the need to manually create unique protected cloning methods.

MohammadHamdyGhanem:
Or better:
```
public interface IQux : IBar
{
    this Get(); //or would it be 'override IBaz IBar.Get();'?
}

```
or better:
```
public interface IQux : IBar
{
    [IBaz] Get(); //or would it be 'override IBaz IBar.Get();'?
}


```

MohammadHamdyGhanem:
Why didn't this happen yet?
It seems popular enough.

HaloFour:
@MohammadHamdyGhanem 

> Why didn't this happen yet?
> It seems popular enough.

Because you haven't written it yet.

Seriously, there are a million "good" ideas, modifying the language is a very expensive process and the team has finite resources.  Things take time.

MohammadHamdyGhanem:
@HaloFour 
> Because you haven't written it yet.

[...]

PR #2007, opened 2018-11-15
Added Covariant Returns Exploration

YairHalberstadt:
I don't know if a pull request is the correct way to do this, so please let me know if there's some better way.

Following the [discussion](https://github.com/dotnet/csharplang/issues/49#issuecomment-420541217) I had with @gafter, I spent some time writing up an [exploration](https://github.com/YairHalberstadt/csharplang/blob/master/proposals/covariant-returns-exploration.md) of various designs for Covariant Returns in C#.

I would be interested in hearing any feedback the community has to offer on this.

Thank You


gafter:
This sort of document is best added as an issue, not checked in to the repository.

PR #2027, opened 2018-11-27
Update unsafe-code.md to include decimal type

BlackCentipede:
@tannergooding And I came across this that the table seems to omit the decimal type. Decimal type have a size constant of 128 bits as shown in the following documentations:

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/decimal

https://github.com/dotnet/csharplang/blob/98043cdc889303d956d540d7ab3bc4f5044a9d3b/spec/types.md#the-decimal-type

### DIFF
--- a/spec/unsafe-code.md
+++ b/spec/unsafe-code.md
@@ -667,6 +667,7 @@ The result of the `sizeof` operator is a value of type `int`. For certain predef
 | `sizeof(char)`   | `2`        |
 | `sizeof(float)`  | `4`        |
 | `sizeof(double)` | `8`        |
+| `sizeof(decimal)` | `16`        |
 | `sizeof(bool)`   | `1`        |
 
 For all other types, the result of the `sizeof` operator is implementation-defined and is classified as a value, not a constant.


BlackCentipede:
Also compiler emit ldc.i4.s 16 in place of sizeof(decimal) as a constant.

svick:
If this is updated in the spec, then I think [this comment](https://github.com/dotnet/roslyn/blob/547fbc0e/src/Compilers/Core/Portable/SpecialTypeExtensions.cs#L123-L125) in the Roslyn source could be removed.

BlackCentipede:
@svick I'm confused about that Roslyn Comment, isn't that already in the spec under SizeOf Operator in this document? 
<https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-334.pdf>

Here the relevant part:

spec

svick:
You're right, it's in the ECMA spec, but not here, in the Microsoft spec (the document this Pull Request is changing).

Which one is "the spec" is not quite clear right now. The long term plan is to update the ECMA spec to be more up-to-date (it's currently at C# 5.0) and then switch to using that one exclusively.

gafter:
We are not going to merge the ECMA spec changes into this spec line by line by taking pull requests. This is one of many items that will be resolved by the move to a common ECMA spec.

PR #2028, opened 2018-11-28
Add NRT spec

MadsTorgersen:
Draft - work in progress

### DIFF
--- a/proposals/nullable-reference-types-specification.md
+++ b/proposals/nullable-reference-types-specification.md
@@ -1,0 +1,266 @@
+﻿# Nullable Reference Types Specification
+
+*** This is a work in progress - several parts are missing or incomplete. ***
+
+# Syntax
+
+## Nullable reference types
+
+Nullable reference types have the same syntax `T?` as the short form of nullable value types, but do not have a corresponding long form.
+
+For the purposes of the specification, the current `nullable_type` production is renamed to `nullable_value_type`, and a `nullable_reference_type` production is added:
+
+```antlr
+reference_type
+    : ...
+    | nullable_reference_type
+    ;
+    
+nullable_reference_type
+    : non_nullable_reference_type '?'
+    ;
+    
+non_nullable_reference_type
+    : type
+    ;
+```
+
+The `non_nullable_reference_type` in a `nullable_reference_type` must be a non-nullable reference type (class, interface, delegate or array), or a type parameter that is constrained to be a non-nullable reference type (through the `class` constraint, or a class other than `object`).
+
+Nullable reference types cannot occur in the following positions:
+
+- as a base class or interface
+- as the receiver of a `member_access`
+- as the `type` in an `object_creation_expression`
+- as the `delegate_type` in a `delegate_creation_expression`
+- as the `type` in an `is_expression`, a `catch_clause` or a `type_pattern`
+- as the `interface` in a fully qualified interface member name
+
+A warning is given on a `nullable_reference_type` where the nullable annotation context is disabled.
+
+## Nullable class constraint
+
+The `class` constraint has a nullable counterpart `class?`:
+
+```antlr
+primary_constraint
+    : ...
+    | 'class' '?'
+    ;
+```
+
+## Null-forgiving expressions
+
+The post-fix `!` operator is called the null-forgiving operator.
+
+```antlr
+primary_expression
+    : ...
+    | null_forgiving_expression
+    ;
+    
+null_forgiving_expression
+    : primary_expression '!'
+    ;
+```
+
+The `primary_expression` must be of a reference type.  
+
+## Nullable compiler directives
+
+`#nullable` directives control the nullable annotation and warning contexts.
+
+```antlr
+pp_directive
+    : ...
+    | pp_nullable
+    ;
+    
+pp_nullable
+    : whitespace? '#' whitespace? 'nullable' whitespace nullable_action pp_new_line
+    ;
+    
+nullable_action
+    : 'disable'
+    | 'enable'
+    | 'restore'
+    | 'safeonly'
+    ;
+```
+
+`#pragma warning` directives are expanded to allow changing the nullable warning context, and to allow individual warnings to be enabled on even when they're disabled by default:
+
+```antlr
+pragma_warning_body
+    : ...
+    | 'warning' whitespace nullable_action whitespace 'nullable'
+    ;
+
+warning_action
+    : ...
+    | 'enable'
+    ;
+```
+
+Note that the new form of `pragma_warning_body` uses `nullable_action`, not `warning_action`.
+
+
+# Nullable contexts
+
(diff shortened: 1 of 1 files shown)


proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+    | 'enable'
+    ;
+```
+
+Note that the new form of `pragma_warning_body` uses `nullable_action`, not `warning_action`.
+
+
+# Nullable contexts
+
+Every line of source code has a *nullable annotation context* and a *nullable warning context*. These control whether nullable annotations have effect, and whether nullability warnings are given. The annotation context of a given line is either *disabled* or *enabled*. The warning context of a given line is either *disabled*, *safeonly* or *enabled*.
```
CyrusNajmabadi: it would be likely be good to explain how code relates to "line of source code".  Especially for things like an expression that might cross a PP boundary.  Are tehre rules that should be followed? i.e. using the starting position of an expression/statement?  the ending position?  something else entirely?

CyrusNajmabadi: >A `type` is deemed to occur in a given annotation context when the last token of the type is within that context.

sounds like it may be sufficient :)

proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+
+A given type can have one of four nullabilities: *Oblivious*, *nonnullable*, *nullable* and *unknown*. *Nonnullable* and *unknown* types may yield warnings if a potential `null` value is assigned to them. Values of *nullable* and *unknown* types may yield warnings when dereferenced without proper null checking.
+
+The kind of type and the nullable annotation context it occurs in determine its nullability:
+
+- A nonnullable value type `S` is always *nonnullable*
+- A nullable value type `S?` is always *nullable*
+- An unannotated reference type `C` in a *disabled* annotation context is *oblivious*
+- An unannotated reference type `C` in an *enabled* annotation context is *nonnullable*
+- A nullable reference type `C?` in a *disabled* annotation context is *nullable* (but a warning is yielded)
```
CyrusNajmabadi: the warning here seems unfortunate.  if i'm in the middle of updating things, but decide to disable things because it's turning out to be a lot of work, would i then get warning?  Or is there a recommendation on what combination of pp directives i should use in this case to allow me to suspend my work?

proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+
+`class?` is a new constraint denoting "possibly nullable reference type", whereas `class` denotes "nonnullable reference type".
+
+The nullability of a type argument or of a constraint does not impact whether the type satisfies the constraint, except where that is already the case today (nullable value types do not satisfy the `struct` constraint). However, if the type argument does not satisfy the nullability requirements of the constraint, a warning may be given.
+
+
+# Null state and null tracking
+
+Every expression in a given source location has a *null state*, which indicated whether it is believed to potentially evaluate to null. The null state is either "not null" or "maybe null". The null state is used to determine whether a warning should be given about null-unsafe conversions and dereferences.
+
```
CyrusNajmabadi: is there value in having a "definitely null" null-state?  for example:

```if (param != null) { ... } else { $param$... }```

if we know something to be 'definitely null', there could be specific warnings emitted that the user coudl then decide should be treated always as errors.

File: proposals/nullable-reference-types-specification.md [hunk 4]
```diff
@@ -0,0 +1,266 @@
...
+***Describe null state transitions similar to definite assignment***
+
+## Null state for expressions
+
+The null state of an expression is derived from its form and type, and from the null state of variables involved in it.
+
+## Literals
+
+The null state of a `null` literal is "maybe null". The null state of a `default` literal that is being converted to a type that is known not to be a nonnullable value type is "maybe null". The null state of any other literal is "not null".
+
```
CyrusNajmabadi: as per above, this is where 'definitely null' might be a useful 'null state'.  There will likely be times something might-be null, and i'm ok with a warning, but would like to update to 'error' if 'definitely null'.

proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+If a `simple_name` is not classified as a value, its null state is "not null". Otherwise it is a tracked expression, and its null state is its tracked null state at this source location.
+
+## Member access
+
+If a `member_access` is not classified as a value, its null state is "not null". Otherwise, if it is a tracked expression, its null state is its tracked null state at this source location. Otherwise, if the nullability of its type is nullable or unknown, the null state is "maybe null". Otherwise the null state is "not null".
+
+## Invocation expressions
+
+If an `invocation_expression` invokes a member that is declared with one or more attributes for special null behavior, the null state is determined by those attributes. Otherwise, if the nullability of the type of the expression is nullable or unknown, the null state is "maybe null". Otherwise the null state is "not null".
+
```
CyrusNajmabadi: will the C# langauge specify those attributes?  Or is that considered an open-ended implementation detail?

proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+    : simple_name
+    | this
+    | base
+    | tracked_expression '.' identifier
+    ;
+```
+
+Where the identifiers denote fields or properties.
+
+***Describe null state transitions similar to definite assignment***
```
CyrusNajmabadi: looking forward to this.

proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+
+If `B` denotes the base type of the enclosing type, `base.I` has the same null state as `((B)this).I` and `base[E]` has the same null state as `((B)this)[E]`.
+
+## Default expressions
+
+`default(T)` has the null state "non-null" if `T` is known to be a nonnullable value type. Otherwise it has the null state "maybe null".
+
+## Null-conditional expressions
+
+A `null_conditional_expression` has the null state "maybe null".
```
CyrusNajmabadi: is this appropriate if the 'null state' of the expression in `expr?.` is non-null?  I guess it is since the 'non-null' state we have may be incorrect, and this is a pretty explicit way for the user to be explicit that they can handle null and will get null back in that case.

proposals/nullable-reference-types-specification.md
jnm2: Would `git stash` be a good enough mitigation for this scenario?

proposals/nullable-reference-types-specification.md
```diff
@@ -0,0 +1,266 @@
...
+```antlr
+primary_constraint
+    : ...
+    | 'class' '?'
+    ;
+```
+
+## Null-forgiving expressions
+
+The post-fix `!` operator is called the null-forgiving operator.
```
jnm2: 'Null-forgiving' hints at smoothing over the effect of nulls. It sounds like an odd term for something that will blow up at runtime if there is a null.

yaakov-h:
What does “safeonly” mean? It’s listed as an option, but not described (yet).

333fred:
@yaakov-h that means you won't see the w warnings. As an example:
```C#
string s = null; // safeonly won't give you a warning here, as you're not actually doing something that can cause a null reference
AcceptsANonNullStringParameter(s); // You'd get a warning here, as you're doing something that could actually be unsafe.
```

PR #2096, opened 2018-12-20
Upate patterns.md - always use switch_expression

amis92:
There were some not updated usages of *match_expression*. I've replaced all of them with *switch_expression* as I guess was intended.


gafter:
Thank you!

PR #2111, opened 2018-12-28
Replace existing text with text from the C# 5 standard

BillWagner:
Fixes dotnet/docs#9752

The draft spec incorrectly includes "instance constructors" as one class of "static function members".

I replaced both bullet items with the text from the ECMA C# 5 standard.

/cc @MadsTorgersen @agocke

### DIFF
--- a/spec/expressions.md
+++ b/spec/expressions.md
@@ -923,8 +923,8 @@ This section describes the process that takes place at run-time to invoke a part
 
 For purposes of describing the invocation process, function members are divided into two categories:
 
-*  Static function members. These are instance constructors, static methods, static property accessors, and user-defined operators. Static function members are always non-virtual.
-*  Instance function members. These are instance methods, instance property accessors, and indexer accessors. Instance function members are either non-virtual or virtual, and are always invoked on a particular instance. The instance is computed by an instance expression, and it becomes accessible within the function member as `this` ([This access](expressions.md#this-access)).
+*  Static function members. These are static methods, static property accessors, and user-defined operators. Static function members are always non-virtual.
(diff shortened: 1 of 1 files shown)


spec/expressions.md
```diff
@@ -923,8 +923,8 @@ This section describes the process that takes place at run-time to invoke a part
...
 
-*  Static function members. These are instance constructors, static methods, static property accessors, and user-defined operators. Static function members are always non-virtual.
-*  Instance function members. These are instance methods, instance property accessors, and indexer accessors. Instance function members are either non-virtual or virtual, and are always invoked on a particular instance. The instance is computed by an instance expression, and it becomes accessible within the function member as `this` ([This access](expressions.md#this-access)).
+*  Static function members. These are static methods, static property accessors, and user-defined operators. Static function members are always non-virtual.
+*  These are instance methods, instance constructors, instance property accessors, and indexer accessors. Instance function members are either non-virtual or virtual, and are always invoked on a particular instance. The instance is computed by an instance expression, and it becomes accessible within the function member `this` ([This access](expressions.md#this-access)). For an instance constructor, the instance expression is taken to be the newly allocated object.
```
NextTurn:
```suggestion
*  Instance function members. These are instance methods, instance constructors, instance property accessors, and indexer accessors. Instance function members are either non-virtual or virtual, and are always invoked on a particular instance. The instance is computed by an instance expression, and it becomes accessible within the function member `this` ([This access](expressions.md#this-access)). For an instance constructor, the instance expression is taken to be the newly allocated object.
```

May be accidentally deleted?

agocke:
Note: this looks out of date with the C# 6 draft spec, so that'll be something we should reconcile when we merge the ECMA spec with the C# 6 spec.

Joe4evr:
:spaghetti: Is it by chance possible that the original text intended to say that *static* constructors are "static function members", or is that just as wrong (due to special handling)?

PR #2131, opened 2019-01-09
Update types.md

Rick-Anderson:
I'm sure this is wrong but we need to get "Open generics" in a H3 or higher so we can get some SEO. Google **c# open generics** and nothing from ms.docs comes up.

See https://github.com/aspnet/Docs/issues/8891 (moved to https://github.com/aspnet/Docs/issues/8933)

I need to link to open generic.

### DIFF
--- a/spec/types.md
+++ b/spec/types.md
@@ -543,7 +543,7 @@ type_argument
 
 In unsafe code ([Unsafe code](unsafe-code.md)), a *type_argument* may not be a pointer type. Each type argument must satisfy any constraints on the corresponding type parameter ([Type parameter constraints](classes.md#type-parameter-constraints)).
 
-### Open and closed types
+### Open and closed types (open generics)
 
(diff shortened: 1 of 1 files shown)


paulomorgado:
I think the "official" term is **open type**: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/types#open-and-closed-types

However, **open generic types** is used in the documentation of `typeof`: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/typeof#remarks

Rick-Anderson:
Maybe it would be better to have something like the following:

```
## Open and closed types
Open types are sometimes call **open generic types**.
```
But we need something to get SEO working on searches for open generic types. That's what most folks call them.

@BillWagner 	will know the best way to handle this.

@paulomorgado 	thanks for your valuable input.

gafter:
We are avoiding updating the spec until after we merge with the ECMA version.

PR #2144, opened 2019-01-15
Proposal: Simplified parameter null validation

jaredpar:
This is a proposal that was discussed by LDM on 2019/1/14. Will link to LDM notes when they're available.

Note: sorry for getting this out same day as LDM meeting. It's a small proposal, got tied up last week and didn't finish writing until shortly before our LDM meeting.

### DIFF
--- a/proposals/null-arg-checking.md
+++ b/proposals/null-arg-checking.md
@@ -1,21 +1,24 @@
 # Simplified Null Argument Checking
 
 ## Summary
-This proposal provides a simplified syntax for validating method arguments are not `null` and 
-throwing `ArgumentNullException` appropriately.
+This proposal provides a simplified syntax for validating method arguments are not `null` and throwing 
+`ArgumentNullException` appropriately.
 
 ## Motivation
-The work on designing nullable reference types has caused us to examine the code necessary for 
-null argument validation. Given that NRT doesn't affect code execution developers still must 
-add `if (arg is null) throw` boiler plate code even in projects which are fully null clean. This
-gave us the desire to explore a minimal syntax for argument `null` validation in the language. While
-anticipated to pair often with NRT the proposal is independent of it. 
+The work on designing nullable reference types has caused us to examine the code necessary for `null` argument 
+validation. Given that NRT doesn't affect code execution developers still must add `if (arg is null) throw` boiler 
+plate code even in projects which are fully `null` clean. This gave us the desire to explore a minimal syntax for 
+argument `null` validation in the language. 
+
+While this `null` parameter validation syntax is expected to pair frequently with NRT the proposal is fully independent
+of it. The syntax can be used independent of `#nullable` directives.
 
 ## Detailed Design 
 
 ### Null validation parameter syntax
-The bang operator, `!`, can be positioned after any identifier in a parameter list and this will 
-cause the C# compilet to emit standard `null` checking code for that parameter. For example:
+The bang operator, `!`, can be positioned after a parameter name in a parameter list and this will cause the C# 
+compiler to emit standard `null` checking code for that parameter. This is referred to as `null` validation parameter
+syntax. For example:
 
 ``` csharp
 void M(string name!) {
@@ -26,7 +29,7 @@ void M(string name!) {
 Will be translated into:
 
 ``` csharp
-void M(string name!) {
+void M(string name) {
     if (name is null) {
         throw new ArgumentNullException(nameof(name));
     }
@@ -34,18 +37,28 @@ void M(string name!) {
 }
 ```
 
-The generated `null` check will occur before any user authored code in the method. When multiple 
-parameters contain the `!` operator then the checks will occur in the same order as the parameters
-are declared.
-
-The check will be specifically for reference equality to `null`, it does not invoke `==` or any user 
-defined operators. This also means the `!` operator can only be added to parameters whose type can
-have the value `null`. Value types and type parameters not constrained to `class` or `interface`
-cannot be used here. 
+The generated `null` check will occur before any developer authored code in the method. When multiple parameters contain
+the `!` operator then the checks will occur in the same order as the parameters are declared.
 
 ``` csharp
-// Errro: Cannot use ! on parameter of type T. 
-void G<T>(T arg!) {
+void M(string p1, string p2) {
+    if (p1 is null) {
+        throw new ArgumentNullException(nameof(p1));
+    }
+    if (p2 is null) {
+        throw new ArgumentNullException(nameof(p2));
+    }
+    ...
+}
+```
+
+The check will be specifically for reference equality to `null`, it does not invoke `==` or any user defined operators. 
+This also means the `!` operator can only be added to parameters whose type can be tested for equality against `null`. 
+This means it can't be used on a parameter whose type is known to be a value type.
+
+``` csharp
+// Errro: Cannot use ! on parameters who types derive from System.ValueType
+void G<T>(T arg!) where T : struct {
 
 }
 ```
@@ -54,6 +67,8 @@ In the case of a constructor the `null` validation will occur before any other c
 
 - Chaining to other constructors with `this` or `base` 
 - Field initializers which implicitly occur in the constructor
+
+For example:
 
 ``` csharp
 class C {
@@ -115,12 +130,13 @@ definition.
 
 ### Extending is null
 The types for which the expression `is null` is valid will be extended to include unconstrained type parameters. This 
-will allow it to fill the intent of checking for `null` on all types which a `null` check is valid. That is types 
-which are not definitely known to be value types. Type parameters which are constrained to `struct` cannot be 
-used with this syntax.
+will allow it to fill the intent of checking for `null` on all types which a `null` check is valid. Specifically that
+is types which are not definitely known to be value types. For example Type parameters which are constrained to 
+`struct` cannot be used with this syntax.
 
 ``` csharp
 void NullCheck<T1, T2>(T1 p1, T2 p2) where T2 : struct {
+    // Okay: T1 could be a class or struct here.
     if (p1 is null) {
         ...
     }
@@ -141,7 +157,7 @@ Any parameter which has a `!` operator applied to it's name will start with the 
 true even if the type of the parameter itself is potentially `null`. That can occur with an explicitly nullable type, 
 such as say `string?`, or with an unconstrained type parameter. 
 
-When a `!` operator on parameters is combined with an explicitly nullable type on the parameter then a warning will
+When a `!` syntax on parameters is combined with an explicitly nullable type on the parameter then a warning will
 be issued by the compiler:
 
 ``` csharp
--- a/proposals/null-arg-checking.md
+++ b/proposals/null-arg-checking.md
@@ -1,0 +1,90 @@
+# Simplified Null Argument Checknig
+
+## Summary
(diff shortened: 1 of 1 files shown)


orthoxerox:
> Implicit property setter arguments

Could `!` be placed on `set` to indicate a null check?

```c#
public Foo Foo {get; set!;}
public Bar Bar {
    get { return bar; }
    set! { bar = value; DoSomethingElse(); }
}
```

orthoxerox:
@YairHalberstadt I don't think features that help violate LSP are a good idea.

YairHalberstadt:
@orthoxerox
CoreFX is full of NotSupportedExceptions. They violate LSP in exactly the same way.

jaredpar:
@orthoxerox 

> Could ! be placed on set to indicate a null check?

Let me think on that. There is nothing wrong with this but I'm worried if it will be obvious enough that it applies to `value`.

jaredpar:
@YairHalberstadt 

> When overloading a method that accepts a nullable value, it makes perfect sense to say that you intend to throw on a null value.

The scenario you outlined in the code is overriding though, not overloading. The NRT will issue a warning for that scenario as you're restricting the input set in the override, not expanding it. The language only allows for co / contravariant changes to nullable types in overrides.

YairHalberstadt:
@jaredpar
Sorry that was a typo. I meant overriding.

There are many libraries (including in CoreFX) that throw a NotSupportedException on certain overrides.

Throwing a Null argument exception on a null parameter on an override where the base accepts a nullable parameter is not all that different.

jaredpar:
@YairHalberstadt 

> Throwing a Null argument exception on a null parameter on an override where the base accepts a nullable parameter is not all that different.

I tend to agree with this. I wasn't in favor of issuing a warning for this case but LDM essentially decided the other way. If there is enough real world evidence this is common then LDM will reconsider their position.

CyrusNajmabadi:
@jaredpar @jcouv If/when this gets done, we should have the appropriate IDE features light up.  Off the top of my head:

1. "add null check" should be updated to use this form.
2. We should have an analyzer that looks for the idiomatic forms for throwing null, and convert to use this. 

For '2' that would be things like looking for:

```c#
    if (name is null) {
        throw new ArgumentNullException(...);
    }
```

As well as the current IDE recommended form of:

```c#
    someThing = name ?? throw new ArgumentNullException(...);
```

From the IDE perspective we can debate what can be in those `...` sections.i.e. if we need to strictly look for something, or if the pattern can loosely match.

tannergooding:
@jaredpar, has there been any consideration into having this emit a call to a throw helper (possibly on `PrivateImplementationDetails`) instead of directly emitting the throw (e.g. https://source.dot.net/#System.Private.CoreLib/shared/System/ThrowHelper.cs,170)?

Throwing the exception directly generates quite a bit of IL (and, in turn, assembly), which can hurt things like inlinability, codegen, and throughput.

orthoxerox:
@CyrusNajmabadi what about an analyzer that goes "you have a non-nullable reference type parameter in your public API, consider adding a null check to ensure it's never null"?

jaredpar:
@tannergooding 

> has there been any consideration into having this emit a call to a throw helper

No but it seems reasonable.

YairHalberstadt:
> I tend to agree with this. I wasn't in favor of issuing a warning for this case but LDM essentially decided the other way. If there is enough real world evidence this is common then LDM will reconsider their position.

`IDictionary<TKey, TValue>` might be an example of this. There are dictionaries which accept a null key, but `Dictionary` does not. So is it better to have IDictionary accept a null key, and Dictionary throw on null, or have IDictionary not accept a null key, or even have a new interface `INullableKeyDictionary` which inherits `IDictionary`?

YairHalberstadt:
Note, I'm not pushing for there to be no warning in such a case. I'm just trying to raise the issue so it is considered properly.

svick:
Will an `async` method that uses this feature throw an exception or return a faulted `Task`?

Directly throwing an exception would be kind of consistent with how constructors and iterators throw the exception as soon as possible, even sooner than any user code can. In the case of `async`, it would mean that the exception would be thrown even before the `Task` that's returned from the method is `await`ed.

svick:
@theunrepentantgeek If you have code like `async Task M() => throw new Exception();`, then it indeed runs synchronously. But it synchronously returns a faulted `Task`, it does not directly throw an exception.

So, if this feature directly threw an exception for `async` methods, it would *not* be the same as writing the checks manually  as the first lines of the `async` method.

alrz:
I think in a `#nullable` context, in the absence of `!` on the parameter, we should emit an Assert anyways, because NRT is never definite and can be hijacked in a lot of ways. in that scenario, it's still possible to end up with a NRE down the call stack which could be hard to reason about. This is along the same line of https://github.com/dotnet/csharplang/issues/2149 which suggests to emit an assertion via null-suppression operators.

theunrepentantgeek:
> we should emit an Assert anyways, because NRT is never definite and can be hijacked in a lot of ways.

This would upset the (very vocal) crowd who care about every wasted pico-second for performance reasons.

I like the idea of having a spec of syntax to write the null check for me - but it needs to be opt-in; I'd make extensive use of it at the borders of my libraries, but not internally.

HaloFour:
@theunrepentantgeek 

> This would upset the (very vocal) crowd who care about every wasted pico-second for performance reasons.

By using `Debug.Assert` (or a similar helper function) those checks would only apply to builds where the `DEBUG` preprocessor is defined.  You'd reclaim those lost picoseconds by switching to a release build.

HaloFour:
@theunrepentantgeek 

> When I specify that my method parameters are enforced, they better be enforced in production - for that's where it's most important to get a good stack trace to diagnose problems.

I agree.  That's what the `!` suffix on the parameter would do, add an always-enforced null-check.  I believe that @alrz is suggesting using these conditional helpers to throw for other NRTs situations, like a non-nullable parameter that *isn't* adorned with the `!` suffix, so that you can catch potential issues easier when debugging and testing.

Personally I'm on the fence about that as it can result in the application behaving differently depending on whether `DEBUG` is defined.

alrz:
it can result in the application behaving differently depending on whether DEBUG is defined.

It's not behavioural difference, it's just an elided check. we already use Debug.Assert in Roslyn extensively, and it hasn't been a source of confusion?

yaakov-h:
You could always define `DEBUG` in your production builds 😈

orthoxerox:
@HaloFour 

> Personally I'm on the fence about that as it can result in the application behaving differently depending on whether `DEBUG` is defined.

That's what already happens today when you use `Debug.Assert`.

HaloFour:
@orthoxerox 

> That's what already happens today when you use `Debug.Assert`.

I'm aware.  I've seen this bite some people when they try to assert on something like the return value of a method call which has side-effects and they didn't realize that their method call would also be eliminated.

@alrz 

> It's not behavioural difference, it's just an elided check.

Which becomes a behavioral difference when the method now allows a `null` through in production and that method does some percentage of business logic before it fails, assuming it fails at all.

> we already use Debug.Assert in Roslyn extensively, and it hasn't been a source of confusion?

It's not the people who develop on Roslyn that I would be worried about.  :)

YairHalberstadt:
> Which becomes a behavioral difference when the method now allows a null through in production and that method does some percentage of business logic before it fails, assuming it fails at all.

Presumably this never would have occurred in debug, or it would have hit the assert. And if it did hit the assert and they ignored it, there's little more the language can be expected to do to help.

So the issue might occur if once the production issue occurred they debug it and hit the assert. Well then great! They've found the issue!

mikernet:
@HaloFour The situation you're describing biting people isn't really relevant here as this assertion doesn't have side effects.

Regardless, if someone is failing assertions while debugging and ignoring them, well...what do they think is going to happen in production? I don't see how this is an issue.

HaloFour:
@mikernet 

> Regardless, if someone is failing assertions while debugging and ignoring them, well...what do they think is going to happen in production? I don't see how this is an issue.

Usually the assertion isn't happening during debugging, and it's missed by a unit test due some some bad assumptions.  The code in question could be multiple layers away from some POCO deserialized from JSON where a field described as required suddenly isn't.  These situations very easily make it into production, unfortunately.

mikernet:
@HaloFour Are you talking about an assertion with side effects or this null assertion? 

In the null assertion case it can only help catch the bug you are describing. If it's caught in debugging well then great, you caught the bug. If you didn't and it went to release then you missed the bug and it will be easier to catch when it hits your assertion as you deserialize the faulty data set in debug mode.

alrz:
>The code in question could be multiple layers away from some POCO deserialized from JSON where a field described as required suddenly isn't

That's actually a good example where such assertion could be greatly helpful.

mikernet:
As I mentioned in another thread somewhere, it would be awesome if nullable debug assertions in debug builds could be triggered by a compiler flag. Pretty sure if this was available, best practices would dictate that it be used everywhere possible and everyone would use it. It could either be on/off or configurable to be added for non-null parameter checks and/or when `!` is used in a method body to "de-null" a nullable value.

HaloFour:
@mikernet 

> As I mentioned in another thread somewhere, it would be awesome if nullable debug assertions in debug builds could be triggered by a compiler flag.

I'd be happy with a second preprocessor directive, `#define ENFORCE_NOT_NULL`

alrz:
Note: one could easily redefine the helper with desired ConditionalAttribute (or none at all) - the default impl could use something commonly preferred.

HaloFour:
@alrz 

That might depend on where they emit the helper method and how they call it.  I would imagine that other than flipping the `ConditionalAttribute`s applied to the method that there'd be little reason to define your own.

szalapski:
@mikernet I agree and commented such [here](https://github.com/dotnet/csharplang/issues/2145#issuecomment-575280778).  Perhaps we wouldn't need to mark anything with "!" to enable this.

jaroslavurban:
Not sure if I am late, but why adding new operator '!', instead of using existing '?'? My suggestion would be to use '?' that would define if the parameter shall be checked for ArgumentNullException. The check will be done at IL level and can be explicitly enabled/disabled within some compile setting.

theunrepentantgeek:
> ... can be explicitly enabled/disabled within some compile setting.

￼Computer flags like this introduce different dialects of the language - something the C# team is extremely reluctant to do. 

Keep in mind that each distinct compiler flag doubles the number of C# dialects - a dozen simple flags would mean 4096 subtly different dialects to document and test. Very much not fun. 

Null reference types reached the bar because of the enormous value of detection and elimination of null errors, but it's unlikely that anything else will.

MichaelPeter:
Could this also be extended to additional checks which are very frequent? 

I mean especally string empty checks, my code contains lots of:

```
Guard.CheckNotNullOrEmpty(nameof(myStringParam)), myStringParam); // first argument name, then value to check
or Guard.CheckNotNullOrEmpty(nameof(myStringParam)));
```

Also for things like `Guid.Empty `it could be helpfull but a exclusive priority on string checks woulld be enough.

This is now a bad example but maybe somebody has a better idea:
`void MyMethod([NotStringEmpty] string myStringParam!))`
`void MyMethod([NotStringEmptyOrWhitespace] string myStringParam!))`
`void MyMethod([NotGuidEmptyOrWhitespace] string myStringParam!))`

Or is this more a topic for `ISourceGenerator` ? And standard classes for it?

PR #2161, opened 2019-01-20
Eliminating new keyword and introduce functions

Viacheslav-Romanov:


Viacheslav-Romanov:
> If you have a suggestion for the csharp language, it's best to file an issue.
> 
> Also, you will need a bit more detail than your current proposal contains. What problem are you solving? How are you solving it? What are the syntax and semantics of your solution? What alternatives are there? You don't need to answer all of these necessarily, but definitely start thinking along these lines.

Hi Yair,

Thank for a prompt response.
Alternatives are the "most loved programming language" Kotlin, then Dart that recently also eliminated the "new" keyword and Python of course. Those are very popular and fast to prototype programming languages so far. In terms of high-level functions that's also a current trend now. You don't always need a class to implement simple logic, c# looks to bulky and overwhelmed with syntaxis sugar and unnecessary constructions. If I want to print "Hello world" why do I need to implement the class for that? The point is readability of the code should be improved as much as possible. It'll also help to keep you code more clean and easier to maintain. My eyes are hurting when I see the code with the variety of keywords which (new, async, await, public, private, protected, etc.) don't add up anything to the business logic but only make noise in your code.

Viacheslav-Romanov:
Hi

Sure, I'll create an issue. Thanks.

theunrepentantgeek:
Keep in mind that the C# language design folks are not willing to throw millions of developers and billions of existing lines of code under a bus - backward compatibility is _extremely_ important.

omariom:
> "most loved programming language" Kotlin

I would say: "most loved in JVM universe" :)

jcouv:
As discussed above this PR is not the right channel to make language change proposals. Closing it. Thanks

PR #2165, opened 2019-01-23
Function pointer updates

jaredpar:
This updates the function pointer proposal based on previous LDM meetings and PR feedback. 

Note: i'm merging this immediately so the proposal is updated for an LDM meeting later today. Please do though continue to leave feedback. I will incorporate it and feedback from the LDM meeting today in a later update of the proposal.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -247,19 +247,43 @@ must be different than the version defined in mscorlib.
 One option that was explored was emitting such a pointer as `mod_req(Func<int>) void*`. This doesn't 
 work though as a `mod_req` cannot bind to a `TypeSpec` and hence cannot target generic instantiations.
 
-### No names altogether
-Given that a `func*` can be used without names why even allow names at all? The underlying CLI primitive doesn't have
-names hence the use of names is purely a C# invention. That ends up being a leaky abstraction in some cases (like
-not allowing overloads when `func*` differ by only names). 
+### Named function pointers
+The function pointer syntax can be cumbersome, particlarly in complex cases like nested function pointers. Rather than
+have developers type out the signature every time the language could allow for named declarations of function pointers
+as is done with `delegate`. 
 
-At the same time, `func*` look and feel so much like `delegate` types, not allowing them to be named would be seen
-as an enormous gap by customers. The leaky abstraction is wort the trade offs here.
+``` csharp
+func* void Action();
 
-### Requiring names always
-Given that names are allowed for `func*` why not just require them always? Given that `func*` is an existing CLI
-type there are uses of it in the ecosystem today. None of those uses will have the metadata serialization format 
-chosen by the C# compiler. This means the feature would be using CLI function pointers but not interopting with any 
-existing usage.
+unsafe class NamedExample {
+    void M(Action a) {
+        a();
+    }
+}
+```
+
+Part of the problem here is the underlying CLI primitive doesn't have names hence this would be purely a C# invention 
+and require a bit of metadata work to enable. That is doable but is a significant about of work. It essentially requires
+C# to have a companion to the type def table purely for these names.
+
+Also when the arguments for named function pointers was examined we found they could apply equally well to a number of
+other scenarios. For example it would be just as convenient to declare named tuples to reduce the need to type out
+the full signature in all cases. 
+
+``` csharp
+(int x, int y) Point;
+
+class NamedTupleExample {
+    void M(Point p) {
+        Console.WriteLine(p.x);
+    }
+}
+```
+
+After discussion we decided to not allow named declaration of `func*` types. If we find there is significant need for
+this based on customer usage feedback then we will investigate a naming solution that works for function pointers, 
+tuples, generics, etc ... This is likely to be similar in form to other suggestions like full `typedef` support in
+the language.
 
 ## Future Considerations
 
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -288,12 +288,29 @@ the language.
 ## Future Considerations
 
 ### static local functions
-
 This refers to [the proposal](https://github.com/dotnet/csharplang/issues/1565) to allow the 
 `static` modifier on local functions. Such a function would be guaranteed to be emitted as 
 `static` and with the exact signature specified in source code. Such a function should be a valid
 argument to `&` as it contains none of the problems local functions have today
 
+### static delegates
+This refers to [the proposal](https://github.com/dotnet/csharplang/issues/302) to allow for the declaration of 
+`delegate` types which can only refer to `static` members. The advantage being that such `delegate` instances can be 
+allocation free and better in performance sensitive scenarios.
 
-*** static delegates https://github.com/dotnet/csharplang/blob/master/proposals/static-delegates.md
-*** PR feedback needs to be gone through.
+If the function pointer feature is implemented the `static delegate` proposal will likely be closed out. The proposed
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -11,122 +11,157 @@ potential implementation of the feature):
 
 https://github.com/dotnet/csharplang/issues/191
 
-This is an alternate design propsoal to [compiler intrinsics](https://github.com/dotnet/csharplang/blob/master/proposals/intrinsics.md)
+This is an alternate design propsoal to [compiler intrinsics]
```
tannergooding: nit: Maybe this should explicitly call out that it is an alternative for the `calli` intrinsic.

proposals/function-pointers.md
```diff
@@ -11,122 +11,157 @@ potential implementation of the feature):
...
+(https://github.com/dotnet/csharplang/blob/master/proposals/intrinsics.md)
 
 ## Detailed Design 
 
-### funcptr
-The language will allow for the declaration of function pointers using the `funcptr` contextual keyword. The 
-declaration and usage of function pointers closely resemble that of `delegate`:
+### Function pointers
+The language will allow for the declaration of function pointers using the `func*` syntax. The full syntax is described
```
tannergooding: I very much preferred `funcptr`.

I would think you are also less likely to encounter something called `funcptr` than something called `func` and that `func*` could be ambiguous/confusing in some scenarios (is it a function pointer or a pointer to a struct called `func`... Just think of `func* func*() f`).

proposals/function-pointers.md
```diff
@@ -11,122 +11,157 @@ potential implementation of the feature):
...
-of a `funcptr` will use `calli` where invocation of a `delegate` will use `callvirt` on the `Invoke` method.
+of a `func*` will use `calli` where invocation of a `delegate` will use `callvirt` on the `Invoke` method.
 Syntactically though invocation is identical for both constructs.
 
-The `calli` instruction requires the calling convention be specified as a part of the invocation. The default 
-for `funcptr` will be managed. Alternate forms can be specified by adding the appropriate modifier after the 
-`funcptr` keyword: `cdecl`, `fastcall`, `stdcall`, `thiscall` or `winapi`. Example:
+The ECMA-335 definition of method pointers includes the calling convention as part of the type signature (section 7.1).
+The default calling convention will be `managed `. Alternate forms can be specified by adding the appropriate modifier 
+after the `func*` syntax: `cdecl`, `fastcall`, `stdcall`, `thiscall` or `winapi`. Example:
```
tannergooding: Will explicitly specifying `managed` as the calling convention be supported?

proposals/function-pointers.md
```diff
@@ -11,122 +11,157 @@ potential implementation of the feature):
...
 }
 ```
 
-A `funcptr` type is a pointer type which means it has all of the capabilities and restrictions of a standard pointer
+A `func*` type is a pointer type which means it has all of the capabilities and restrictions of a standard pointer
 type:
 - Only valid in an `unsafe` context.
-- Methods which contain a `funcptr` parameter or return type can only be called from an `unsafe` context.
+- Methods which contain a `func*` parameter or return type can only be called from an `unsafe` context.
 - Cannot be converted to `object`.
```
tannergooding: We should validate it works with `Pointer.Box(void*, Type)`.

proposals/function-pointers.md
```diff
@@ -11,122 +11,157 @@ potential implementation of the feature):
...
+// Function pointer which has a cdecl calling convention, a cdecl return type and takes a single 
+// paramater of type cdecl;
+func* cdecl cdecl(cdecl);
+```
 
 ### Allow addresss-of to target methods
 
 Method groups will now be allowed as arguments to an address-of expression. The type of such an 
-expression will be an unnamed `funcptr` which has the equivalent signature of the target method and a managed 
+expression will be a `func*` which has the equivalent signature of the target method and a managed
```
tannergooding: What about taking the address of a method marked `extern` (`pinvokeimpl` in IL)? In this case, the method is known to be implemented in native code and may have an associated calling convention.

proposals/function-pointers.md
jaredpar: I went for `func*` over `funcptr` because this is a pointer type fundamentally. Having a pointer type without the `*` token but also have it respect all pointer rules seems like an odd inconsistency in the language. 

There is no ambiguity in the syntax though. It may be confusing at a glance but the language can disambiguate.

Note: I fully expect LDM to spend at least an hour on syntax today though. Doubt the syntax will survive as is.

proposals/function-pointers.md
jaredpar: Please read the section on calling conventions where this is called out.

proposals/function-pointers.md
jaredpar: See the section on `NativeCallbackAttribute`. That is fairly relevant to this scenario.

proposals/function-pointers.md
tannergooding: I saw that as I got further down. I was just expecting it to be listed here as well, since it is relevant 😄

proposals/function-pointers.md
tannergooding: I saw that section, after getting further down, as well. However, given that `DllImport` is not a custom attribute (at least from the perspective of the runtime/IL metadata) and that it encodes all the relevant information inline, it would be nice if C# just worked with the existing stuff.

For reference:
```csharp
[DllImport("a", BestFitMapping = false, CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode, EntryPoint = "foo", ExactSpelling = true, PreserveSig = true, SetLastError =true, ThrowOnUnmappableChar =false)]
public static extern void Foo();
```

becomes:
```ilasm
method public hidebysig static pinvokeimpl("a" as "foo" nomangle unicode lasterr cdecl) 
    void Foo () cil managed preservesig 
{
}
```

proposals/function-pointers.md
```diff
@@ -11,122 +11,157 @@ potential implementation of the feature):
...
-``` csharp
-unsafe struct Example {
-    funcptr int (int) Field;
-    unsafe void UnnamedExample(funcptr int(int) ptr) {
-        int x = ptr(42);
-        Field = ptr;
-        ...
+        p1 = p2; // okay Func1 and Func3 have compatible signatures
+        Console.WriteLine(p2 == p1); // True
+        p2 = p2; // error: calling conventions are incompatible
```
pentp: Should this be `p2 = p3`?

proposals/function-pointers.md
```diff
@@ -160,46 +192,104 @@ exactly what signature they are emitted with.
...
 The better function member specification will be changed to include the following line:
 
-> A `funcptr` is more specific than `void*`
+> A `func*` is more specific than `void*`
+
+This means that it is possible to overload on `void*` and a `func*` and still sensibly use the address-of operator.
+
+## Open Issues
+
+### NativeCallback Attribute
```
jkotas: Nit: The current name of the attribute is `NativeCallableAttribute`. (We can rename to NativeCallback.)

proposals/function-pointers.md
```diff
@@ -160,46 +192,104 @@ exactly what signature they are emitted with.
...
+This is an attribute used by the CLR to avoid the managed to native prologue when invoking. Methods marked by this
```
jkotas: > This is an attribute used by the CLR to avoid the managed to native prologue when invoking.

The attribute add native to managed prolog to the method. It is not used to avoid anything.

PR #2214, opened 2019-02-12
Add well-known member details

agocke:
This change reflects the recent CoreFX and language decisions about
naming and signature of the Index/Range types & members.


gafter:
Before we make changes to this spec, we need to merge with ECMA.

yaakov-h:
Even to proposal documents?

agocke:
@gafter This is the `ranges` speclet -- I think this is OK.

gafter:
Right - my mistake.

PR #2265, opened 2019-02-25
Update readonly-instance-members.md

RikkiGibson:
Updating the speclet based on more recent design changes.

In particular would appreciate feedback based on the auto-implemented properties section.


RikkiGibson:
I'm gonna run with this-- if anyone has feedback or concerns let's just update the speclet again.

PR #2271, opened 2019-02-26
Specify readonly members behavior for events

RikkiGibson:
Added detail to the speclet about how to handle `readonly` events. This design is meant to be as consistent as possible with what's allowed within a `readonly struct` declaration.

### DIFF
--- a/proposals/readonly-instance-members.md
+++ b/proposals/readonly-instance-members.md
@@ -159,6 +159,11 @@ public event Action<EventArgs> Event3
     readonly add { }
     readonly remove { }
 }
+public static readonly event Event4
+{
+    add { }
+    remove { }
+}
 ```
 
 Some other syntax examples:
--- a/proposals/readonly-instance-members.md
+++ b/proposals/readonly-instance-members.md
@@ -131,7 +131,7 @@ public int Prop3
 }
 ```
 
-Readonly can be applied to auto-implemented properties, but it won't have a meaningful effect. The compiler will treat all auto-implemented getters as readonly whether or not the `readonly` keyword is present.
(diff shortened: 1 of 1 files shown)


RikkiGibson:
FYI @jaredpar @tannergooding

proposals/readonly-instance-members.md
```diff
@@ -143,6 +143,23 @@ public readonly int Prop7 { get; set; }
...
+// Allowed
+public readonly event Action<EventArgs> Event1
+{
+    add { }
+    remove { }
+}
+
+// Not allowed
+public readonly event Action<EventArgs> Event2;
+public event Action<EventArgs> Event3
```
tannergooding: Why would this latter one not be allowed?

Say you have a partially mutable struct, where you support adding more event subscriptions, but not removing them. In which case `add` mutates, but `remove` does not.

tannergooding: As an example:
```csharp
public struct B
{
    private Action<EventArgs> _event;
    
    public event Action<EventArgs> Event1
    {
        add { _event += value; }
        readonly remove { }
    }
}
```

RikkiGibson: Mainly because it's easier to start by disallowing it and a less common case than property accessors. I'd suggest revisiting if we get feedback from people who would really benefit from being able to specify different `readonly`-ness on the add and remove accessors.

tannergooding: Fair enough.

I'm guessing supporting this is "more difficult" for event accessors, since they don't currently (afair) allow attached modifiers?

RikkiGibson: Yeah, event accessors don't allow any modifiers right now, and it's required to declare both `add` and `remove` accessors. It was feeling like the value added needed to be more clear in order to get past that friction.

PR #2278, opened 2019-02-28
Update LDM-2019-02-13.md

YairHalberstadt:
I believe `string? s = null` was meant to be  `string s = null`, as otherwise `s.ToString();` would definitely warn, since it remains possibly null.


meetings/2019/LDM-2019-02-13.md
```diff
@@ -144,7 +144,7 @@ Only allow `!` in simple `out` parameters with no declaration.
 See [dotnet/roslyn#30297](https://github.com/dotnet/roslyn/issues/30297)
 
 ```cs
-string? s = null;
+string s = null;
```
gafter: The correct fix should be
``` c#
string s = string.Empty;
```

YairHalberstadt: True. I missed that :-)

PR #2293, opened 2019-03-04
Efficient params and string formatting

jaredpar:
This proposal is being discussed at LDM on 3/4/2019.

### DIFF
--- a/proposals/format.md
+++ b/proposals/format.md
@@ -1,4 +1,4 @@
-# Efficent Formatting
+# Efficent Params and String Formatting
 
 ## Summary
 This combination of features will increase the efficiency of formatting `string` values and passing of `params` style
--- a/proposals/format.md
+++ b/proposals/format.md
@@ -5,21 +5,101 @@ This combination of features will allow for effecient and customizable formattin
 
 ## Motivation
 The allocation of `string` and `params` values dominates the performance of many text based applications like MSBuild. 
-In order to efficiently 
+In order to maintain efficiency such applications often need to abandon popular features like `params`, `string` 
+interpolation, etc ... This set of language features will enable applications to avoid the extra allocation overhead
+while continuing to use these features. 
 
 ## Detailed Design 
+There are a set of features that will be used here to achieve these results:
+
+- Expanding `params` to support more effecient types than array. 
+- Allowing for developers to customize how `string` interpolation is achieved. 
 
 ### params Span
+The language will allow for `params` in a method signature to have the types `Span<T>` and `ReadOnlySpan<T>`. The same
+rules will apply to `params Span<T>` that apply to `params T[]`:
+
+- Can't overload where the only difference is a `params` keyword.
+- Can invoke by passing a series of `T` arguments or a single `Span<T>` argument.
+- Must be the last parameter in a method signature.
+- Etc ... 
+
+The advantage this variant of `params` provides is it gives the compiler great flexbility in how it allocations the
+backing storage for the `Span<T>` value. With a `params T[]` the compiler must allocate a new `T[]` for every 
+invocation of a `params` method because it must assume the callee stored and reused the parameter. Given the 
+`Span<T>` and `ReadOnlySpan<T>` types are `ref struct` the callee cannot store the argument. Hence the compiler can
+safely re-use the value. 
+
+One such potential implementation is the following. Consider all `params` invocation of a given type in a method 
+body. The compiler could allocate an array which has a size equal to the largest `params` invocation and use that for
+all of the invocations by creating appropriately sized `Span<T>` instances over the array. For example:
+
+``` csharp
+static class OneAllocation {
+    static void Use(params Span<string> spans) {
+        ...
+    }
+
+    static void Go() {
+        Use("jaredpar");
+        Use("hello", "world");
+        Use("a", "longer", "set");
+    }
+}
+```
+
+The compiler could choose to emit the body of `Go` as follows:
+
+``` csharp
+    static void Go() {
+        var args = new string[3];
+        args[0] = "jaredpar";
+        Use(new Span<int>(args, start: 0, length: 1));
+
+        args[0] = "hello";
+        args[1] = "world";
+        Use(new Span<int>(args, start: 0, length: 2));
+
+        args[0] = "a";
+        args[1] = "longer";
+        args[2] = "set";
+        Use(new Span<int>(args, start: 0, length: 3));
+   }
+```
+
+This can siginficantly reduce the number of arrays allocated in an application. Allocations can be even further 
+reduced if the runtime provides utilities for smarter stack allocation of arrays.
+
+This optimization cannot always be applied though. Even though the callee cannot capture the `params` argument it can 
+still be captured in the caller when there is a `ref` or a `out / ref` parameter that is itself a `ref struct`
+type. 
+
+``` csharp
+static class SneakyCapture {
+    static ref int M(params Span<T> span) => ref span[0];
+
+    static void Oops() {
+        // This now holds onto the memory backing the Span<T> 
+        ref int r = ref M(42);
+    }
+}
+```
+
+These cases are statically dectable though. It potentially occurs whenever there is a `ref` return or a `ref struct`
+parameter passed by `out` or `ref`. In such a case the compiler must allocate a fresh `T[]` for every invocation. 
 
 ### params IEnumerable
 
 ### Customize interopolated strings
+existing behavior: interpolated strings have natural type of sttring but can target type to formattablestring
+
+change to ValueFormattableString
 
 ### Variant type
 
 ## Open Issuess
 
-### consideration 1
+### open issue1
 
 ## Considerations
 
--- a/proposals/format.md
+++ b/proposals/format.md
@@ -14,21 +14,34 @@ There are a set of features that will be used here to achieve these results:
 
 - Expanding `params` to support more effecient types than array. 
 - Allowing for developers to customize how `string` interpolation is achieved. 
+- Allowing for interpolated `string` to bind to more efficient `string.Format` overloads.
 
-### params Span
-The language will allow for `params` in a method signature to have the types `Span<T>` and `ReadOnlySpan<T>`. The same
-rules will apply to `params Span<T>` that apply to `params T[]`:
+### Extending params
+The language will allow for `params` in a method signature to have the types `Span<T>`, `ReadOnlySpan<T>` and 
+`IEnumerable<T>`. The same rules will apply to these new types that apply to `params T[]`:
 
 - Can't overload where the only difference is a `params` keyword.
-- Can invoke by passing a series of `T` arguments or a single `Span<T>` argument.
+- Can invoke by passing a series of arguments that are implicitly convertible to `T` or a single `Span<T>` / 
+`ReadOnlySpan<T>` / `IEnumerable<T>` argument.
 - Must be the last parameter in a method signature.
 - Etc ... 
 
-The advantage this variant of `params` provides is it gives the compiler great flexbility in how it allocations the
-backing storage for the `Span<T>` value. With a `params T[]` the compiler must allocate a new `T[]` for every 
-invocation of a `params` method because it must assume the callee stored and reused the parameter. Given the 
-`Span<T>` and `ReadOnlySpan<T>` types are `ref struct` the callee cannot store the argument. Hence the compiler can
-safely re-use the value. 
+The `Span<T>` and `ReadOnlySpa<T>` variants will be referred to as `Span<T>` below for simplicity. In cases where the 
+behavior of `ReadOnlySpan<T>` differs it will be called out. 
+
+The advantage the `Span<T>` variants of `params` provides is it gives the compiler great flexbility in how it allocates
+the backing storage for the `Span<T>` value. With a `params T[]` the compiler must allocate a new `T[]` for every 
+invocation of a `params` method because it must assume the callee stored and reused the parameter. This can lead to 
+a large inefficiency in methods with lots of `params` invocations.
+
+Given `Span<T>` variants are `ref struct` the callee cannot store the argument. Hence the compiler can optimize the 
+call sites by taking actions like re-using the argument. This can make repeated invocations very efficient. The 
+langauge though will make no specific guarantees about how such callsites are optimized. Only note that the compiler 
+is free to use values other than `T[]` when invoking a `Span<T>` method. 
+
+The `IEnumerable<T>` variant is a merely a covenience overload. It's useful in scenarios which have frequent uses of
+`IEnumeralbe<T>` but also have lots of `params` usage. When invoked in `T` argument form the backing storage will 
+be allocated as a `T[]` just as `params T[]` is done today.
 
 One such potential implementation is the following. Consider all `params` invocation of a given type in a method 
 body. The compiler could allocate an array which has a size equal to the largest `params` invocation and use that for
@@ -88,52 +101,7 @@ static class SneakyCapture {
 These cases are statically dectable though. It potentially occurs whenever there is a `ref` return or a `ref struct`
 parameter passed by `out` or `ref`. In such a case the compiler must allocate a fresh `T[]` for every invocation. 
 
-### params IEnumerable
-The language will allow for `params` in a method signature to have the type `IEnumerable<T>`. The same rules will apply 
-to `params IEnumerable<T>` that apply to `params T[]` (see above for listing of rules).
-
-The compiler will invoke a `params IEnumerable<T>` method exactly as it invokes a `params T[]` method. A new array will
-be allocated for every call site and passed to the callee.
-
-### params VariantCollection
-The language will allow for `params` in a method signature to have the type `VariantCollection`. This elements of a 
-`VariantCollection` are considered to be of type `Variant`. The same rules will apply to `params VariantCollection` 
-that apply to `params Variant[]` (see above for listing of rules).
-
-The `Variant` value type is a [suggested addition](https://github.com/dotnet/corefxlab/pull/2595) to CoreFX that has a
-conversion from any type. For the majority of common framework types (`int`, `double`, `TimeSpan`, etc ..) the 
-conversion is allocation free.
-
-Methods which take a `params` of hetrogeneous data today must declare as `params object[]`. This means that every 
-value type passed to the method incurs a boxing allocation (in addition to the array the compiler must allocate). The
-`Variant` and `VariantCollection` type allow such methods to avoid allocations in the most common paths. Invocations
-like `Console.WriteLine("hello {0] {1}", someInt, DateTime.Now)` can be invoked without any boxing or array allocations
-here.
-
-When emitting the code for a `params VariantCollection` invocation the compiler will look for an overload of 
-`VariantCollection.Create(in Variant v1, in Variant v2, ..., in Variant vn)` where `n` matches the number of arguments 
-being passed to the parameter. Each argument will be passed exactly as if the developer wrote the
-`VariantCollection.Create` call with the arguments. For example:
-
-``` csharp
-class VariantCall { 
-    static void Write(params VariantCollection collection) { ... } 
-
-    static void Use() {
-        Write(1, DateTime.UtcNow)
-
-        // Compiler will evaluate as 
-        Write(VariantCollection.Create(1, DateTime.UtcNow))
-
-        // Compiler will eventually emit as 
-        Write(VariantCollection.Create((Variant)1, (Variant)DateTime.UtcNow))
-    }
-}
-```
-
-In the case a `VariantCollection.Create` method with the correct number of parameters does not exist the compiler will
-attempt to invoke `Create(Variant[] array)` instead. A fresh `Variant[]` will be allocated for every invocation in this
-case. In the case `Create(Variant[] array)` does not exist the compiler will issue an error.
+Several other potential optimization strategies are discussed at the end of this document.
 
 ### params overload resolution changes
 This proposal means the language now has four variants of `params` where before it had one. It is also sensible for 
@@ -187,7 +155,7 @@ class ConsoleEx {
 class Program { 
     static void Main() { 
         ConsoleEx.Write(42);
-        ConsoleEx.Write("hello {DateTime.UtcNow}");
+        ConsoleEx.Write($"hello {DateTime.UtcNow}");
 
         // Translates into 
         ConsoleEx.Write(ValueFormattableString.Create((Variant)42));
@@ -202,6 +170,10 @@ Overload resolution rules will be changed to prefer `ValueFormattableString` ove
 interpolated string. This means it will be valuable to have overloads which differ only on `string` and 
 `ValueFormattableString`. Such an overload today with `FormattableString` is not valauble as the compiler will always
 prefer the `string` version (unless the developer uses an explicit cast). 
+
+### More efficient string.format
+
+Allow for interpolated strings to bind to more efficient `string.Format` overloads.
 
 ## Open Issuess
 
@@ -223,6 +195,10 @@ type in the `System` namespace this break seems like a reasonable compromise.
 ## Future Considerations
 
 CLR helper for stack allocating arrays 
+Lambdas and re-using arrays
+varargs won't work because of JIT and GC
+VariantCollection make it a ref struct and Span<Variant>
+Calling `Span<Variant>` more efficiently
 
 ## Misc
 Related issues
--- a/proposals/format.md
+++ b/proposals/format.md
@@ -2,32 +2,33 @@
 
 ## Summary
 This combination of features will increase the efficiency of formatting `string` values and passing of `params` style
-argumens.
+arguments.
 
 ## Motivation
-The allocation overhead of formatting `string` values can dominate the performance of many text applicatios: from the 
-boxing penalty of primitive types, the `object[]` allocation for `params` and the intermediate `string` allocations 
-during `string.Format` calls. In order to maintain efficiency such applications often need to abandon productivity 
-features such as `params` and `string` interpolation. 
+The allocation overhead of formatting `string` values can dominate the performance of many text based applications: 
+from the boxing penalty of `struct` types, the `object[]` allocation for `params` and the intermediate `string` 
+allocations during `string.Format` calls. In order to maintain efficiency such applications often need to abandon
+productivity features such as `params` and `string` interpolation and move to non-standard, hand coded solutions. 
 
-Consider MSBuild as an example. In one representative build sample MSBuild will generate 262MB of `string` allocation
+Consider MSBuild as an example. This is written using a lot of modern C# features by developers who are concious of 
+performance. Yet in one representative build sample MSBuild will generate 262MB of `string` allocation
 using minimal verbosity. Of that 1/2 of the allocations are short lived alloctaions inside `string.Format`. These 
 features would remove much of that on .NET Desktop and get it down to nearly zero on .NET Core due to the 
 
 The set of language features described here will enable applications to continue using these features, with very
-little churn to their application code base, while removing the unintended allocation overhead i n the majority of 
+little or no churn to their application code base, while removing the unintended allocation overhead in the majority of 
 cases.
 
 ## Detailed Design 
 There are a set of features that will be used here to achieve these results:
 
-- Expanding `params` to support more effecient types than array. 
+- Expanding `params` to support a broader set of collection types.
 - Allowing for developers to customize how `string` interpolation is achieved. 
 - Allowing for interpolated `string` to bind to more efficient `string.Format` overloads.
 
 ### Extending params
 The language will allow for `params` in a method signature to have the types `Span<T>`, `ReadOnlySpan<T>` and 
-`IEnumerable<T>`. The same rules will apply to these new types that apply to `params T[]`:
+`IEnumerable<T>`. The same rules for invocation will apply to these new types that apply to `params T[]`:
 
 - Can't overload where the only difference is a `params` keyword.
 - Can invoke by passing a series of arguments that are implicitly convertible to `T` or a single `Span<T>` / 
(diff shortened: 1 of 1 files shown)


proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+    }
+}
+```
+
+Overload resolution rules will be changed to prefer `ValueFormattableString` over `string` when the argument is an 
+interpolated string. This means it will be valuable to have overloads which differ only on `string` and 
+`ValueFormattableString`. Such an overload today with `FormattableString` is not valauble as the compiler will always
+prefer the `string` version (unless the developer uses an explicit cast). 
+
+## Open Issuess
```
Logerfo:
```suggestion
## Open Issues
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+
+For very simple cases, like one to two arguments, the C# compiler could always use `StackAlloc<T>` variant. This is 
+unlikely to significantly contribute to stack exhaustion in most cases. For other cases the compiler could choose to 
+use `MaybeStackAlloc<T>` instead and let the runtime make the call.
+
+How we choose will likely require a deeper investigation and examination of real world apps. But if these new intrinsics
+are available then it will give us this type of flexibility.
+
+### Why not varargs? 
+!!! Will flesh this out in a bit. Have to track down my notes!!!
```
jaredpar: FYI: i did not forget about this. I'm actually trying to track down my notes. Verdict though is it's not a good solution for C#.

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+}
+```
+
+These cases are statically dectable though. It potentially occurs whenever there is a `ref` return or a `ref struct`
+parameter passed by `out` or `ref`. In such a case the compiler must allocate a fresh `T[]` for every invocation. 
+
+Several other potential optimization strategies are discussed at the end of this document.
+
+The `IEnumerable<T>` variant is a merely a covenience overload. It's useful in scenarios which have frequent uses of
+`IEnumeralbe<T>` but also have lots of `params` usage. When invoked in `T` argument form the backing storage will
```
bdukes:
```suggestion
`IEnumerable<T>` but also have lots of `params` usage. When invoked in `T` argument form the backing storage will 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+as it becomes an implementation detail of other parts of the proposal. 
+
+### Efficient interpolated strings
+Interpolated strings are a popular yet innefecient feature in C#. The most common syntax, using an interpolated `string`
+as a `string`, translates into a `string.Format(string, params object[])` call. That will inccur boxing allocations for 
+all value types, intermediate `string` allocations as the implementation largely uses `object.ToString` for formatting
+as well as array allocations once the number of arguments exceeds the amount of parameters on the "fast" overloads of 
+`string.Format`. 
+
+The language will change it's interpolation lowering to consider alternate overloads of `string.Format`. It will
```
bdukes:
```suggestion
The language will change its interpolation lowering to consider alternate overloads of `string.Format`. It will
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+## Motivation
+The allocation overhead of formatting `string` values can dominate the performance of many text based applications: 
+from the boxing penalty of `struct` types, the `object[]` allocation for `params` and the intermediate `string` 
+allocations during `string.Format` calls. In order to maintain efficiency such applications often need to abandon
+productivity features such as `params` and `string` interpolation and move to non-standard, hand coded solutions. 
+
+Consider MSBuild as an example. This is written using a lot of modern C# features by developers who are concious of 
+performance. Yet in one representative build sample MSBuild will generate 262MB of `string` allocation
+using minimal verbosity. Of that 1/2 of the allocations are short lived alloctaions inside `string.Format`. These 
+features would remove much of that on .NET Desktop and get it down to nearly zero on .NET Core due to the
```
bdukes:
```suggestion
features would remove much of that on .NET Desktop and get it down to nearly zero on .NET Core due to the availability of `Span<T>`
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+The language will allow for `params` in a method signature to have the types `Span<T>`, `ReadOnlySpan<T>` and 
+`IEnumerable<T>`. The same rules for invocation will apply to these new types that apply to `params T[]`:
+
+- Can't overload where the only difference is a `params` keyword.
+- Can invoke by passing a series of arguments that are implicitly convertible to `T` or a single `Span<T>` / 
+`ReadOnlySpan<T>` / `IEnumerable<T>` argument.
+- Must be the last parameter in a method signature.
+- Etc ... 
+
+The `Span<T>` and `ReadOnlySpa<T>` variants will be referred to as `Span<T>` below for simplicity. In cases where the
```
bdukes:
```suggestion
The `Span<T>` and `ReadOnlySpan<T>` variants will be referred to as `Span<T>` below for simplicity. In cases where the 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+
+### CLR stack allocation helpers
+The CLR only provides only 
+[localloc](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.localloc?redirectedfrom=MSDN&view=netframework-4.7.2)
+ for stack allocation of contiguous memory. This instruction is limited in that it it only works for `unmanaged` types. 
+ This means it can't be used as a universal solution for efficientally allocating the backing storage for `params 
+ Span<T>`. 
+
+This limitation is not some fundamental restriction though but instead more an artifact of history. The CLR could choose
+to add new op codes / intrinsicts which provide universal stack allocation. These could then be used to allocate the
```
bdukes:
```suggestion
to add new op codes / intrinsics which provide universal stack allocation. These could then be used to allocate the
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+behavior of `ReadOnlySpan<T>` differs it will be explicitly called out. 
+
+The advantage the `Span<T>` variants of `params` provides is it gives the compiler great flexbility in how it allocates
+the backing storage for the `Span<T>` value. With a `params T[]` the compiler must allocate a new `T[]` for every 
+invocation of a `params` method. Re-use is not possible because it must assume the callee stored and reused the 
+parameter. This can lead to a large inefficiency in methods with lots of `params` invocations.
+
+Given `Span<T>` variants are `ref struct` the callee cannot store the argument. Hence the compiler can optimize the 
+call sites by taking actions like re-using the argument. This can make repeated invocations very efficient as compared
+to `T[]`. The langauge though will make no specific guarantees about how such callsites are optimized. Only note that
```
loic-sharma: `langauge` -> `language`

bdukes:
```suggestion
to `T[]`. The language though will make no specific guarantees about how such callsites are optimized. Only note that 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+
+1. `ReadOnlySpan<T>`
+1. `Span<T>`
+1. `T[]`
+1. `IEnumerable<T>`
+
+This order is the most to the least effecient for the general case.
+
+### Variant
+The CoreFX is introducing a new managed type `Variant`. This type is meant to be used in APIs which expect hetrogeneous
```
tannergooding: nit: `The CoreFX` reads weird. Maybe just use `CoreFX`

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+values but don't want the overhead brought on by using `object` as the parmeter. The `Variant` type provides universal 
+storage but avoids the boxing allocation for the most commonly used types. Using this type in APIs like `string.Format`
+can eliminate the boxing overhead in the majority of cases.
+
+This type itself is not necessarily special to the language. It is being introduced in this document separately though
```
tannergooding: Are you going to be putting this proposal up on CoreFX as well?

jaredpar: @JeremyKuhne owns that. Pretty sure it's one corefxlab already.

Logerfo: can't stop laughing about the typos they just keep coming

jaredpar: Don't worry they won't stop.

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+
+To facilitate this the language will introduce the following overload resolution tie breaking rule. When the candidate
+methods differ only by the `params` parameter then the canditates will be preferred in the following order:
+
+1. `ReadOnlySpan<T>`
+1. `Span<T>`
+1. `T[]`
+1. `IEnumerable<T>`
+
+This order is the most to the least effecient for the general case.
```
bdukes:
```suggestion
This order is the most to the least efficient for the general case.
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
+# Efficent Params and String Formatting
```
bdukes:
```suggestion
# Efficient Params and String Formatting
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+This combination of features will increase the efficiency of formatting `string` values and passing of `params` style
+arguments.
+
+## Motivation
+The allocation overhead of formatting `string` values can dominate the performance of many text based applications: 
+from the boxing penalty of `struct` types, the `object[]` allocation for `params` and the intermediate `string` 
+allocations during `string.Format` calls. In order to maintain efficiency such applications often need to abandon
+productivity features such as `params` and `string` interpolation and move to non-standard, hand coded solutions. 
+
+Consider MSBuild as an example. This is written using a lot of modern C# features by developers who are concious of
```
bdukes:
```suggestion
Consider MSBuild as an example. This is written using a lot of modern C# features by developers who are conscious of 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Motivation
+The allocation overhead of formatting `string` values can dominate the performance of many text based applications: 
+from the boxing penalty of `struct` types, the `object[]` allocation for `params` and the intermediate `string` 
+allocations during `string.Format` calls. In order to maintain efficiency such applications often need to abandon
+productivity features such as `params` and `string` interpolation and move to non-standard, hand coded solutions. 
+
+Consider MSBuild as an example. This is written using a lot of modern C# features by developers who are concious of 
+performance. Yet in one representative build sample MSBuild will generate 262MB of `string` allocation
+using minimal verbosity. Of that 1/2 of the allocations are short lived alloctaions inside `string.Format`. These
```
bdukes:
```suggestion
using minimal verbosity. Of that 1/2 of the allocations are short lived allocations inside `string.Format`. These 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+## Detailed Design 
+There are a set of features that will be used here to achieve these results:
+
+- Expanding `params` to support a broader set of collection types.
+- Allowing for developers to customize how `string` interpolation is achieved. 
+- Allowing for interpolated `string` to bind to more efficient `string.Format` overloads.
+
+### Extending params
+The language will allow for `params` in a method signature to have the types `Span<T>`, `ReadOnlySpan<T>` and 
+`IEnumerable<T>`. The same rules for invocation will apply to these new types that apply to `params T[]`:
```
CyrusNajmabadi: if we're going to go to this, wouldit be oke to also allow going to ICollection/IList/IReadOnlyList at teh same time (effectively, all the types that T[] already implements)?  It feels like it would be a prime time to just roll this in since we're working in this area.

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+
+        args[0] = "a";
+        args[1] = "longer";
+        args[2] = "set";
+        Use(new Span<int>(args, start: 0, length: 3));
+   }
+```
+
+This can siginficantly reduce the number of arrays allocated in an application. Allocations can be even further 
+reduced if the runtime provides utilities for smarter stack allocation of arrays.
```
CyrusNajmabadi: could you expand on this further?

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+        Use(new Span<int>(args, start: 0, length: 2));
+
+        args[0] = "a";
+        args[1] = "longer";
+        args[2] = "set";
+        Use(new Span<int>(args, start: 0, length: 3));
+   }
+```
+
+This can siginficantly reduce the number of arrays allocated in an application. Allocations can be even further
```
bdukes:
```suggestion
This can significantly reduce the number of arrays allocated in an application. Allocations can be even further 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+    static ref int M(params Span<T> span) => ref span[0];
+
+    static void Oops() {
+        // This now holds onto the memory backing the Span<T> 
+        ref int r = ref M(42);
+    }
+}
+```
+
+These cases are statically dectable though. It potentially occurs whenever there is a `ref` return or a `ref struct`
```
bdukes:
```suggestion
These cases are statically detectable though. It potentially occurs whenever there is a `ref` return or a `ref struct`
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+1. `T[]`
+1. `IEnumerable<T>`
+
+This order is the most to the least effecient for the general case.
+
+### Variant
+The CoreFX is introducing a new managed type `Variant`. This type is meant to be used in APIs which expect hetrogeneous
+values but don't want the overhead brought on by using `object` as the parmeter. The `Variant` type provides universal 
+storage but avoids the boxing allocation for the most commonly used types. Using this type in APIs like `string.Format`
+can eliminate the boxing overhead in the majority of cases.
```
CyrusNajmabadi: is there more info on Variant and how it works?   A link, perhaps, to the CoreFx side of tings?

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+    }
+}
+```
+
+These cases are statically dectable though. It potentially occurs whenever there is a `ref` return or a `ref struct`
+parameter passed by `out` or `ref`. In such a case the compiler must allocate a fresh `T[]` for every invocation. 
+
+Several other potential optimization strategies are discussed at the end of this document.
+
+The `IEnumerable<T>` variant is a merely a covenience overload. It's useful in scenarios which have frequent uses of
```
bdukes:
```suggestion
The `IEnumerable<T>` variant is a merely a convenience overload. It's useful in scenarios which have frequent uses of
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+### params overload resolution changes
+This proposal means the language now has four variants of `params` where before it had one. It is sensible for methods
+to define overloads of methods that differ only on the type of a `params` declarations. 
+
+Consider that `StringBuilder.AppendFormat` would certainly add a `params ReadOnlySpan<object>` overload in addition to
+the `params object[]`. This would allow it to substantially improve performance by reducing collection allocations 
+without requiring any changes to the calling code. 
+
+To facilitate this the language will introduce the following overload resolution tie breaking rule. When the candidate
+methods differ only by the `params` parameter then the canditates will be preferred in the following order:
```
bdukes:
```suggestion
methods differ only by the `params` parameter then the candidates will be preferred in the following order:
```

proposals/format.md
bdukes:
```suggestion
The CoreFX is introducing a new managed type `Variant`. This type is meant to be used in APIs which expect heterogeneous
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+1. `ReadOnlySpan<T>`
+1. `Span<T>`
+1. `T[]`
+1. `IEnumerable<T>`
+
+This order is the most to the least effecient for the general case.
+
+### Variant
+The CoreFX is introducing a new managed type `Variant`. This type is meant to be used in APIs which expect hetrogeneous
+values but don't want the overhead brought on by using `object` as the parmeter. The `Variant` type provides universal
```
bdukes:
```suggestion
values but don't want the overhead brought on by using `object` as the parameter. The `Variant` type provides universal 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+The CoreFX is introducing a new managed type `Variant`. This type is meant to be used in APIs which expect hetrogeneous
+values but don't want the overhead brought on by using `object` as the parmeter. The `Variant` type provides universal 
+storage but avoids the boxing allocation for the most commonly used types. Using this type in APIs like `string.Format`
+can eliminate the boxing overhead in the majority of cases.
+
+This type itself is not necessarily special to the language. It is being introduced in this document separately though
+as it becomes an implementation detail of other parts of the proposal. 
+
+### Efficient interpolated strings
+Interpolated strings are a popular yet innefecient feature in C#. The most common syntax, using an interpolated `string`
```
bdukes:
```suggestion
Interpolated strings are a popular yet inefficient feature in C#. The most common syntax, using an interpolated `string`
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+`System.ValueFormattableString`. All interpolated strings will have a target type conversion to this type. This will 
+be implemented by translating the interpolated string into the call `ValueFormattableString.Create` exactly as is done
+for `FormattableString.Create` today. The language will support all `params` options described in this document when
+looking for the most suitable `ValueFormattableString.Create` method. 
+
+``` csharp
+readonly struct ValueFormattableString {
+    public static ValueFormattableString Create(Variant v) { ... } 
+    public static ValueFormattableString Create(string s) { ... } 
+    public static ValueFormattableString Create(string s, params ReadOnlySpan<Variant> collection) { ... }
```
CyrusNajmabadi: would you need a ReadOnlySpan of hte string portions of hte interpolated string, along with the ReadOnlySPan of interpolation values?  Or am i misunderstanding how this would be used?  For example, if you have:

`$"foo{bar}baz"`

how is the ValueFormattableString created?  I would prefer a form that gave me `"foo"` and `"bar"` in a ReadOnlySpan, and something that gave me the Variant for `bar` in a ReadOnlySpan.

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+values but don't want the overhead brought on by using `object` as the parmeter. The `Variant` type provides universal 
+storage but avoids the boxing allocation for the most commonly used types. Using this type in APIs like `string.Format`
+can eliminate the boxing overhead in the majority of cases.
+
+This type itself is not necessarily special to the language. It is being introduced in this document separately though
+as it becomes an implementation detail of other parts of the proposal. 
+
+### Efficient interpolated strings
+Interpolated strings are a popular yet innefecient feature in C#. The most common syntax, using an interpolated `string`
+as a `string`, translates into a `string.Format(string, params object[])` call. That will inccur boxing allocations for
```
bdukes:
```suggestion
as a `string`, translates into a `string.Format(string, params object[])` call. That will incur boxing allocations for 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+bind to very efficient overloads like `string.Format(string format, params ReadOnlySpan<Variant> args)`. In many cases
+this will remove all intermediate allocations.
+
+### Customizable interpolated strings
+Developers are able to customize the behavior of interpolated strings with `FormattableString`. This contains the data
+which goes into an interpolated string: the format `string` and the arguments as an array. This though still has the 
+boxing and argument array allocation as well as the allocation for `FormattableString` (it's an `abstract class`). Hence
+it's of little use to applications which are allocation heavy in `string` formatting.
+
+To make interopolated string formatting efficient the language will recognize a new type:
```
bdukes:
```suggestion
To make interpolated string formatting efficient the language will recognize a new type: 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+class Program { 
+    static void Main() { 
+        ConsoleEx.Write(42);
+        ConsoleEx.Write($"hello {DateTime.UtcNow}");
+
+        // Translates into 
+        ConsoleEx.Write(ValueFormattableString.Create((Variant)42));
+        ConsoleEx.Write(ValueFormattableString.Create(
+            "hello {0}", 
+            VariantCollection.Create((Variant)DateTime.UtcNow));
```
CyrusNajmabadi: this approach seems unfortunate... because won't i now have to figure out how to break `"hello {0}"` up, even thouh the language already did that for me?  I'm coming from this from a JS/TS perspective where this is the natural way this data is packaged up and passed along to template processors.

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+    }
+}
+```
+
+Overload resolution rules will be changed to prefer `ValueFormattableString` over `string` when the argument is an 
+interpolated string. This means it will be valuable to have overloads which differ only on `string` and 
+`ValueFormattableString`. Such an overload today with `FormattableString` is not valauble as the compiler will always
```
bdukes:
```suggestion
`ValueFormattableString`. Such an overload today with `FormattableString` is not valuable as the compiler will always
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+prefer the `string` version (unless the developer uses an explicit cast). 
+
+## Open Issues
+
+### ValuableFormattableString breaking change
```
bdukes:
```suggestion
### ValueFormattableString breaking change
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+The change to prefer `ValueFormattableString` during overload resolution over `string` is a breaking change. It is
+possible for a developer to have defined a type called `ValueFormattableString` today and use it in method overloads
+with `string`. This proposed change would cause the compiler to pick a different overload once this set of features
+was implemented. 
+
+The possibility of this seems reasonably low. The type would need the full name `System.ValueFormattableString` and it 
+would need to have `static` methods named `Create`. Given that developers are strongly discouraged from defining any
+type in the `System` namespace this break seems like a reasonable compromise.
```
CyrusNajmabadi: It would be great if the Language/Compiler codified this (similer to how some languages carve out `__` names as being reserved for them).  If the language drew a line and said "we may change behavior for new types introduced in the `System...` space, and you should not be adding anything there yourself" then this gives both a reasonable path to add a lot of stuff in the future, while also being able to point users at some struct guidelines in case they do bad stuff.

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+already needs to manage a temporary per call and do clean up work after (even if in one case it's just marking 
+an internal temp as free). 
+
+Note: the `KeepAlive` function is only necessary on desktop. On .NET Core the method will not be available and hence
+the compiler won't emit a call to it.
+
+### CLR stack allocation helpers
+The CLR only provides only 
+[localloc](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.localloc?redirectedfrom=MSDN&view=netframework-4.7.2)
+ for stack allocation of contiguous memory. This instruction is limited in that it it only works for `unmanaged` types.
```
bdukes:
```suggestion
 for stack allocation of contiguous memory. This instruction is limited in that it only works for `unmanaged` types. 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+an internal temp as free). 
+
+Note: the `KeepAlive` function is only necessary on desktop. On .NET Core the method will not be available and hence
+the compiler won't emit a call to it.
+
+### CLR stack allocation helpers
+The CLR only provides only 
+[localloc](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.opcodes.localloc?redirectedfrom=MSDN&view=netframework-4.7.2)
+ for stack allocation of contiguous memory. This instruction is limited in that it it only works for `unmanaged` types. 
+ This means it can't be used as a universal solution for efficientally allocating the backing storage for `params
```
bdukes:
```suggestion
 This means it can't be used as a universal solution for efficiently allocating the backing storage for `params 
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+    static void Go() {
+        Span<T> span = RuntimeIntrinsic.StackAlloc<string>(length: 2);
+        span[0] = "hello";
+        span[1] = "world";
+        Use(span);
+    }
+}
+```
+
+While this approach is very heap efficient it does cause extra stack usage. In an algorthim which has a deep stack and
```
bdukes:
```suggestion
While this approach is very heap efficient it does cause extra stack usage. In an algorithm which has a deep stack and
```

proposals/format.md
```diff
@@ -0,0 +1,320 @@
...
+was implemented. 
+
+The possibility of this seems reasonably low. The type would need the full name `System.ValueFormattableString` and it 
+would need to have `static` methods named `Create`. Given that developers are strongly discouraged from defining any
+type in the `System` namespace this break seems like a reasonable compromise.
+
+## Considerations
+
+### Variant2 and Variant3
+The CoreFX team also has a non-allocating set of storage types for up to three `Variant` arguments. These are a single
```
CyrusNajmabadi: Was the '3' brought about by an sort of analysis of existing codebases?  i'd be really curious what the 'tail' looks like her.e  i.e. if 3 gets us 95% of the way, but 4 gets us to 99.9%, that would be interesting to know.

proposals/format.md
tannergooding: Why would you need to support the others, since they all implement `IEnumerable<T>`? Generally speaking, types that implement things like `ICollection` or `IList` can also efficiently implement their iterators (and do in the case of most `CoreFX` types).

tannergooding: I think that the currently proposed types cover most of the scenarios. That is, `Span<T>`/`ROSpan<T>` make sense since they provide allocation-less calls (and declare the immutability of the input data) and `IEnumerable<T>` allows easier interop with LINQ (it avoids the need to realize a query).

CyrusNajmabadi: > Why would you need to support the others, since they all implement IEnumerable<T>? 

Because i want to expose *this* api to users:

```c#
public static void Foo(params IReadOnlyList<T> whatever) { }
```

Allowing people to just pass their *existing* list to me.   Or, they can pass raw values, and it will be an array, an that doesn't need to be converted into an IReadOnlyList since it already is one.

proposals/format.md
bdukes: Leaving `System.` namespaces available for general use can allow third party libraries to implement features on older versions of the framework that aren't officially supported for that particular feature.  But having a compile warning may be appropriate.

proposals/format.md
jaredpar: I'll add it to the open issues. We didn't cover it in LDM.

proposals/format.md
jaredpar: The later sections go into a few of the possible scenarios here around stack allocation.

proposals/format.md
jaredpar: This is the approach that interpolated strings use today. The goal of this proposal is to make the existing process more efficient vs. opening up the door for text templating. That was brought up though in LDM and we're going to look into it a bit though. If you have some pointers to the JS / TS similar features I'd like to read up on them.

proposals/format.md
jaredpar: We've essentially drawn the line on types defined under `System.`. We're still pretty careful about introducing breaking changes here but if we feel we can improve things and it's a type under `System.` we'll accetp a level of breaking changes to do it.

proposals/format.md
jaredpar: I added a link to the PR that has the initial `Variant` code spit and that has a bit of conversation. I don't think there is an official design doc though. Will let @JeremyKuhne comment on that. 

https://github.com/dotnet/corefxlab/pull/2595

proposals/format.md
jaredpar: @JeremyKuhne can comment on that.

proposals/format.md
CyrusNajmabadi: Thanks so much!  I would understand if this didn't happen.  However, it's been one of those language niggles for ages.   If it's cheap, this would be nice to have :)  It would def clean up a few of my apis. 

Note: @tannergooding and i discussed this on Gitter.  It is definitely a strange thing that the language has *baked in* understanding that an array has an implicit reference conversion to IList and its ilk.  And, yet, for 'params' i *must* specify only a `T[]` type.  It seems like it would be pretty simple to just relax that to "anything an array has reference implicit conversion to".

proposals/format.md
CyrusNajmabadi: > This is the approach that interpolated strings use today

That's fair.  Though it would be nice if both forms could be supported.  Having to have clients manually parse strings is a PITA.   JS/TS info on template strings is here:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

Importantly: 

```
function template(strings, ...keys) {
  return (function(...values) {
    var dict = values[values.length - 1] || {};
    var result = [strings[0]];
    keys.forEach(function(key, i) {
      var value = Number.isInteger(key) ? values[key] : dict[key];
      result.push(value, strings[i + 1]);
    });
    return result.join('');
  });
}
```

This is how one could write a function that accepts an interpolated string.  From TypeScript's perspective, the signature is:

```ts
function template(literals: TemplateStringsArray, ...placeholders: string[]) {
```

TmplateStringsArray is also super interesting.  It is typed as:


```ts
interface TemplateStringsArray extends ReadonlyArray<string> {
    readonly raw: ReadonlyArray<string>;
}
```

Meaning, it actually represents *two* arrays of strings.  For example if i had:

```whatever `this is a \t { template } \t string literal`; ```

Then the first arg passed to 'whatever' would look like this:

```js
{
     [ "this is a \t ", " \t string literal"]
     raw: [ "this is a \\t ", " \\t string literal" ]
}
```

So JS gives both the original *raw* strings and the interpreted language strings.  I'm not asking for that myself, but at least having the separate strings makes it so much easier to do things.

proposals/format.md
CyrusNajmabadi: Great.  I wasn't certain if htat was something official, or more informal.  I was basically saying that it would be nice if really just stated it officially.

yaakov-h:
Sorry to bump this but is there a discussion thread that corresponds to this proposal? I can't find it.

PR #2301, opened 2019-03-05
fix two more broken links 

BillWagner:
The 8.0 pattern matching spec links to records.md, and vice-versa.

It looked like Deconstruct semantics are described in 8.0, in the "Positional pattern" section. I added an anchor tag to resolve to that section.

This would fix the build warnings in dotnet/docs#11026

/cc @jcouv @MadsTorgersen


jcouv:
Thanks

PR #2313, opened 2019-03-07
markdown lint for C# 7.1 proposals

BillWagner:
These had two sets of changes:

1. use `csharp` as the language identifier for code fences.
1. use relative links to other articles in the dotnet/csharplang repo that are being published on docs.microsoft.com so those links resolve to the article published there from that site.

/cc @mairaw @jcouv @agocke

### DIFF
--- a/proposals/csharp-7.1/async-main.md
+++ b/proposals/csharp-7.1/async-main.md
@@ -17,7 +17,8 @@ It is very common when learning C#, when writing console-based utilities, and wh
 to call and `await` `async` methods from Main.  Today we add a level of complexity here by forcing such `await`'ing to be
 done in a separate async method, which causes developers to need to write boilerplate like the following just to get
 started:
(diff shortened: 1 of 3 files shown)


proposals/csharp-7.1/async-main.md
```diff
@@ -17,7 +17,8 @@ It is very common when learning C#, when writing console-based utilities, and wh
 to call and `await` `async` methods from Main.  Today we add a level of complexity here by forcing such `await`'ing to be
 done in a separate async method, which causes developers to need to write boilerplate like the following just to get
 started:
-```C#
+
+```csharp
```
jcouv: just curious, does this make a difference?

BillWagner: It depends on the markdown engine. `csharp` is supported more widely.

PR #2351, opened 2019-03-18
Add an issue

Tvde1:


YairHalberstadt:
@Tvdeq
If you have a question about a proposed language feature, the correct thing to do is to ask about it in a GitHub issue.

In answer to your question, it will call the private member, as otherwise adding a new member to an interface with a default Implementation would silently break Implementing code which had defined a private member with the same name.

Hope that helps

Tvde1:
I think it shouldn't compile like that. Adding a new member should result in a compilation error. 
Also, how would one call `IA.M()`?

YairHalberstadt:
If so, feel free to open an issue.
However, as stated, doing so would cause a breaking change if a member is added to an interface with a default implementation. Doing so is unacceptable, given the entire purpose of DIM is to allow library authors to evolve interfaces.

PR #2397, opened 2019-04-05
Proposed changes to Index / Range

jaredpar:
This is a set of proposed changes to the `Index` and `Range` features based on customer feedback.

Note: when looking at the LDM notes from 4/1 the version of the proposal being discussed is captured [here](https://github.com/dotnet/csharplang/pull/2397/commits/8dd672612693932d86a4d065554bf4f97ac4a23e). The PR in it's current state reflects the changes requested by LDM team

### DIFF
--- a/proposals/index-range-changes.md
+++ b/proposals/index-range-changes.md
@@ -33,35 +33,13 @@ This proposal specifically does not want to change how `Index` and `Range` type 
 expressions continue to have the same syntax and types.
 
 ## Detailed Design 
-### Indexable types
-Any type which has an accessible getter property named `Length` or `Count` with a return type of `int` is considered
-Indexable. The language can make use of this property to convert an index expression into an `int` at the point of 
-the expression without the need to use the type `Index` at all. 
+### Countable Types
+Any type which has an property named `Length` or `Count` with an accessible getter adn a return type of `int` is 
+considered Countable. The language can make use of this property to convert an expression of type `Index` into an `int`
+at the point of the expression without the need to use the type `Index` at all. In the case both `Length` and `Count`
+are present then `Length` will be preferred. 
 
-Note: For simplicity going forward the proposal will use the name `Length` to represent `Count` or `Length`.
-
-For example it allows the following simplication:
-
-``` csharp
-Span<char> span = ...;
-char c = span[^1];
-
-// Can be translated to 
-Span<char> span = ...;
-char c = span[span.Length - 1];
-```
-
-Transforming an index expression to an `int` at the call site significantly reduces the burden of frameworks to adopt 
-`Index`. Vitrually any collection type will automatically work with `Index` now as the compiler can translate it to 
-`int` in all cases.
-
-Further this can improve performance by eliminating extra branching. The callee when accepting an `Index` parameter must
-do both test to see if the value is from the end, `Index.IsFromEnd`, and if the value is inside the bounds of the 
-collection. While a small check this can be important in performance sensitive areas.
-
-Doing the translation to `int` at the call site means the `IsFromEnd` check can often be eliminated. For example when 
-dealing with an `int` the compiler can pass the value through. Or in the cases where `^` is used the computation from
-end can be done directly without the additional branching.
+For simplicity going forward the proposal will use the name `Length` to represent `Count` or `Length`.
 
 ### Index and Range implementations are known
 The implementations of `Index` and `Range` are considered to be known and side effect free. Much like 
@@ -72,85 +50,145 @@ like the implicit conversion from `int` to `Index`.
 All arithemtic operations which are emitted will be done so using an `unchecked` context. That matches the context
 in which `Index` and `Range` are compiled in.
 
-### Index target type conversion
-Whenever an expression with type `Index` is used as an argument to an instance member invocation and the receiver is 
-Indexable then the expression will have a target type conversion to `int`. The member invocations applicable for this
-conversion include methods, indexers, properties, extension methods, etc ... Only constructors are excluded as they
-have no receiver. 
+### Implicit Index support
+The language will provide an instance indexer member with a single parameter of type `Index` for types which meet the 
+following criteria:
 
-The target type conversion will be implemented as follows on the index expression:
+- The type is Countable
+- The type has an accessible instance indexer which takes a single `int` as the argument.
+- The type does not have an accessible instance indexer which takes a single `Index` as the parameter.
 
-- When the expression is `^expr` and the type is `int` it will be translated to `receiver.Length - expr`.
-- Else it will be translated as `expr.GetOffset(receiver.Length)` where `expr` is the expression typed as `Index`.
+For such types the language will act as if there is an index member of the form `T this[Index index]` where `T` is 
+the return type of the `int` based indexer including any `ref` style annotations. The new member will have the same 
+`get` and `set` members with matching accessibilty as the `int` indexer. 
 
-The receiver will be spilled as appropriate to ensure he side effects of obtaining the receiver are only executed
+The new indexer will be implemented by converting the argument of type `Index` into an `int` and emitting a call 
+to the `int` based indexer. For discussion purposes lets use the example of `receiver[expr]`. The conversion of 
+`expr` to `int` will occur as follows:
+
+- When the argument is in the form `^expr2` and the type of `expr2` is `int` it will be translated to 
+`receiver.Length - expr2`.
+- Else it will be translated as `expr.GetOffset(receiver.Length)`.
+
+This allows for developers to use the `Index` feature on existing types without the need for modification. For example:
+
+``` csharp
+List<char> list = ...;
+var value = list[^1]; 
+
+// Gets translated to 
+var value = list[list.Length - 1]; 
+```
+
+The `receiver` and `Length` expressions will be spilled as appropriate to ensure the side effects are only executed 
 once. For example:
 
 ``` csharp
+class Collection {
+    private int[] _array = new[] { 1, 2, 3 };
+
+    int Length {
+        get {
+            Console.Write("Length ");
+            return _array.Length;
+        }
+    }
+
+    int this[int index] => _array[index];
+}
+
 class SideEffect {
-    int[] Get() {
+    Collection Get() {
         Console.Write("Get ");
-        return new [] { 1, 2 , 3};
+        return new Collection();
     }
 
     void Use() { 
-        int i = Get()[^1]
+        int i = Get()[^1];
         Console.WriteLine(i);
     }
 }
 ```
 
-This code will print "Get 3". 
+This code will print "Get Length 3". 
 
-When a range expression is used as an argument to an instance member invocation then the target type conversion to 
-`int` extends to both `Index` operands. In the case either of the `Index` members are omitted then the the appropriate
-start or end value will be inserted using `0` or `receiver.Length` as appropriate.
+### Implicit Range support
+The language will provide an instance indexer member with a single parameter of type `Range` for types which meet the 
+following criteria:
 
-``` csharp
-class RangeTargettype {
-    void Example() { 
-        var array = new[] { 1, 2, 3 };
-        Console.WriteLine(array[1..]);
+- The type is Countable.
+- The type has an accessible member named `Slice` indexer which has two parameters of type `int`.
+- The type does not have an instance indexer which takes a single `Range` as the parameter.
 
-        // Becomes
-        Console.WriteLine(array[new Range(1, array.Length));
-    }
-}
-```
+For such types the language will bind as if there is an index member of the form `T this[Range range]` where `T` is 
+the return type of the `Slice` method including any `ref` style annotations. The new member will also have matching
+accessibility with `Slice`. 
 
-### Indexing on Range
-When binding an index member on a Indexable type where the single argument is of type `Range` the language will 
-attempt to translate it to a `Slice` call. The arguments to slice will be both `Index` values of the range converted
-to `Index` using their target typed conversion described in the previous section. If this translation is not succesful
-then normal index binding will occur.
+When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by convering the `Range` 
+expression into two values that are then passed to the `Slice` method. For discussion purposes lets use the example of 
+`receiver[expr]`.
+
+The first argument of `Slice` will be obtained by converting the typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int` then it will be
+emitted as `expr1`.
+- When `expr` is in the form `^expr1..expr2` (where `expr2` can be omitted) then it will be emitted as 
+`receiver.Length - expr1`
+- When `expr` is in the form `..expr2` (where `expr2` can be omitted) then it will be emitted as `0`
+- Else it will be emitted as `expr.Start.GetOffset(receiver.Length)` 
+
+This value will be re-used in the calculation of the second `Slice` argument. When doing so it will be referred to as
+`start`. The second argument of `Slice` will be obtained by converting the range typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int` then it will be 
+emitted as `expr2 - start`.
+- When `expr` is in the form `expr1..^expr2` (where `expr1` can be omitted) then it will be emitted as 
+`(receiver.Length - expr2) - start`
+- When `expr` is in the form `expr1..` (where `expr1` can be omitted) then it will be emitted as `receiver.Length`
+- Else it will be emitted as `expr.End.GetOffset(receiver.Length) - start` 
+
+The `receiver`, `Length` and `expr` expressions will be spilled as appropriate to ensure the side effects are only
+executed once. For example:
 
 ``` csharp
 class Collection {
-    public int Length { get; }
-    public int[] this[Range range] => ...;
+    private int[] _array = new[] { 1, 2, 3 };
+
+    int Length {
+        get {
+            Console.Write("Length ");
+            return _array.Length;
+        }
+    }
+
+    int[] Slice(int start, int length) { 
+        var slice = new int[length];
+        Array.Copy(_array, start, slice, 0, length);
+        return slice;
+    }
 }
 
-class Slice {
-    void Example(Span<int> span, Collection collection) {
-        Span<int> slicedSpan = span[2..]
-        int[] slicedCollection = collection[2..];
+class SideEffect {
+    Collection Get() {
+        Console.Write("Get ");
+        return new Collection();
+    }
 
-        // Translated to 
-        Span<int> slicedSpan = span.Slice(2, span.Length);
-        int[] slicedCollection = collection[new Range(2, collection.Count)v;
+    void Use() { 
+        var array = Get()[0..2];
+        Console.WriteLine(array.length);
     }
 }
 ```
 
-The `Slice` method can instance or extension so long as it is accessible has types that are convertible from `int`.
+This code will print "Get Length 2". 
 
-The compiler will special case the following receiver types binding to `Slice`:
+The language will special case the following known types: 
 
-- `string`: instead of `Slice` the method `Substring` will be used. 
-- array: the runtime helper for array slicing will be used.
+- `string`: the method `Substring` will be used instead of `Slice`.
+- `array`: the method `System.Reflection.CompilerServices.GetSubArray` will be used instead of `Slice`.
 
 ## Open Issues
-
 
 ## Considerations
 
@@ -192,9 +230,68 @@ would've been ideal for slicing but the concept didn't exist when types were add
 
 Thus `Slice` being the sole example it was chosen as the name.
 
+### Index target type conversion
+Another way to view the `Index` transformation in an indexer expression is as a target type conversion. That is instead
+of binding as if there is a member of the form `return_type this[Index]` the language instead assigns a target 
+typed conversion to `int`. 
+
+This concept could be generalized to all member access on Countable types. Whenever an expression with type `Index` is 
+used as an argument to an instance member invocation and the receiver is Countable then the expression will have a 
+target type conversion to `int`. The member invocations applicable for this conversion include methods, indexers, 
+properties, extension methods, etc ... Only constructors are excluded as they have no receiver. 
+
+The target type conversion will be implemented as follows for any expression which has a type of `Index`. For
+discussion purposes lets use the example of `receiver[expr]`:
+
+- When `expr` is in the form `^expr2` and the type of `expr2` is `int` it will be translated to 
+`receiver.Length - expr2`.
+- Else it will be translated as `expr.GetOffset(receiver.Length)`.
+
+The `receiver` and `Length` expressions will be spilled as appropriate to ensure the side effects are only executed
+once. For example:
+
+``` csharp
+class Collection {
+    private int[] _array = new[] { 1, 2, 3 };
+
+    int Length {
+        get {
+            Console.Write("Length ");
+            return _array.Length;
+        }
+    }
+
+    int GetAt(int index) => _array[index];
+}
+
+class SideEffect {
+    Collection Get() {
+        Console.Write("Get ");
+        return new Collection();
+    }
+
+    void Use() { 
+        int i = Get().GetAt(^1);
+        Console.WriteLine(i);
+    }
+}
+```
+
+This code will print "Get Length 3". 
+
+This feature would be beneficial to any member which had a parameter that represented an index. For example 
+`List<T>.InsertAt`. This also has the potential for confusion as the language can't give any guidance as to whether
+or not an expression is meant for indexing. All it can do is convert any `Index` expression to `int` in when invoking
+a member on a Countable type. 
+
+Restrictions:
+
+- This conversion is only applicable when the expression with type `Index` is directly an argument to the member. It 
+would not apply to any nested expressions.
 
 ## Related Issues
 - https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.cs
 - https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md
 
 ## Design Meetings
+- https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-04-01.md
--- a/proposals/index-range-changes.md
+++ b/proposals/index-range-changes.md
@@ -12,22 +12,79 @@ The change is not to the syntax but rather how the langauge maps the syntax to A
 - Reduce the API burden for making existing libraries Index / Range aware
 
 ## Detailed Design 
-
-- Index / Range have known implementations
+- Does not change the existing type logic. 
 
 ### Indexable types
+Any type which has an accessible getter property named `Length` or `Count` with a return type of `int` is considered
+Indexable. The language can use this property to efficiently convert an `Index` expression into an `int` at the 
+point of the expression without relying on the implementation of receiver to do so. 
+
+Note: For simplicity going forward the proposal will use the name `Length` to represent `Count` or `Length`.
+
+For example it allows the following simplication:
+
+``` csharp
+Span<char> span = ...;
+char c = span[^1];
+
+// Can be translated to 
+Span<char> span = ...;
+char c = span[span.Length - 1];
+```
+
+Transforming an `Index` to an `int` at the call site significantly reduces the burden of frameworks to adopt `Index`. 
+Vitrually any collection type will automatically work with `Index` now as the compiler can translate it to `int` in 
+all cases.
+
+Further this can improve performance by eliminating extra branching. The callee when accepting an `Index` parameter must
+do both test to see if the value is from the end, `Index.IsFromEnd`, and if the value is inside the bounds of the 
+collection. While a small check this can be important in performance sensitive areas.
+
+Doing the translation to `int` at the call site means the `IsFromEnd` check can often be eliminated. For example when 
+dealing with an `int` the compiler can pass the value through. Or in the cases where `^` is used the computation from
+end can be done directly without the additional branching.
+
+### Index target type conversion
+Whenever an expression with type `Index` is used as an argument to an instance member invocation and the receiver is 
+Indexable then the expression will have a target type conversion to `int`. The member invocations applicable for this
+conversion include methods, indexers, properties, extension methods, etc ... Only constructors are excluded as they
+have no receiver. 
+
+The target type conversion will be implemented as follows:
+
+- When the expression is `^expr` and the type is `int` it will be translated to `receiver.Length - expr`. This 
+arithmetic will be implemented in an unchecked fashion.
+- Else it will be translated as `expr.GetOffset(receiver.Length)` where `expr` is the expression typed as `Index`.
+
+The receiver will be spilled as appropriate to ensure the side effects of obtaining the receiver are only executed
+once. For example:
+
+``` csharp
+class SideEffect {
+    int[] Get() {
+        Console.Write("Get ");
+        return new [] { 1, 2 , 3};
+    }
+
+    void Use() { 
+        int i = Get()[^1]
+        Console.WriteLine(i);
+    }
+}
+```
+
+This code will print "Get 3". 
+
+
+### Index and Range implementations are known
+The implementations of `Index` and `Range` are considered to be known and side effect free. Much like 
+`ValueTuple<T1, T2>` the language can assume a standard implementation and optimize around it.
 
 ### Detailed Decision Point 1
 
 ## Open Issues
 
-### open Issue 1
-
-### Detect on Count or Length
-The property name `Count` was chosen to identify Indexable types as it's the standard over `Length` in collection 
-types. The design could be extended to use `Count` or `Length` (preferring `Count` if both are present). This 
-complicates the design a bit though and given the overwelming majority of collections use `Count`, typically via
-`ICollection.Count`, the simpler design was chosen.
+### Extension method for Indexer
 
 ## Considerations
 
@@ -47,11 +104,23 @@ as the language does this in other areas (`foreach` lowering, constants, etc ...
 is more concerning as it's special casing an entire class of types. The get labeled as Indexable if they simply have
 a property named `Count` with a return type of `int`. 
 
-After consideration the design was normalized to say that any type which has a property `Count` with a return type of
-`int` is Indexable. That still requires array and `string` to be special cased. That is more palatable though as it's
-a fixed set of type rather than an entire class.
+After consideration the design was normalized to say that any type which has a property `Count` / `Length` with a 
+return type of `int` is Indexable. That removes all special casing, even for `string` and arrays.
 
-### Use Slice in
+### Detect just Count
+Detecting on the property names `Count` or `Length` does complicate the design a bit. Picking just one to standardize
+though is not sufficient as it ends up excluding a large number of types:
+
+- Use `Length`: excludes pretty much every collection in System.Collections and sub-namespaces. Those tend to derive 
+from `ICollection` and hence prefer `Count` over length.
+- Use `Count`: excludes `string`, arrays, `Span<T>` and most `ref struct` based types
+
+The extra complication on the initial detection of Indexable types is outweighed by it's simplification in other 
+aspects.
+
+### Bind to a
+
+### Choice of Slice as a anme
 
 ### Consideration 2
 
--- a/proposals/index-range-changes.md
+++ b/proposals/index-range-changes.md
@@ -7,17 +7,36 @@ customer feedback: particularly from the CoreFX team and their experiences addin
 The change is not to the syntax but rather how the langauge maps the syntax to APIs.
 
 ## Motivation
-- Abstraction penalty
-- Performance
-- Reduce the API burden for making existing libraries Index / Range aware
+The level of API churn necessary to adopt `Index` and `Range` is quite high today. Every collection type must have at
+least two new indexers added for `Index` and `Range` respectively. The implementation of `Index` is exactly the same 
+for virtually every collection in .NET. The implementation for `Range` is likewise often similar by simply deferring to 
+the underlying collection type.
+
+This work must be done by the collection author. The reliance on indexers as the primary use case means developers can't
+update existing collections using extension methods. This means the adoption of the feature will be slowed by waiting
+for library authors to update their types.
+
+The use of `Index` does come with a small performance hit. Typically a consumer of `Index` will pay an extra branch
+check (`IsFromEnd` and bounds) as they would with a `int` parameter (bounds only). Moving to `Range` doubles this cost
+as it occurs on each `Index`.
+
+This penalty is small and insignificant in the vast majority of code. For highly optimized code though this extra 
+branching can be unacceptable. It simply can't be thought of as a syntactic transformation from a call to 
+`span.Slice(start, end)` to `span[start..end]`. Instead the performance implications must be evaluated on each
+usage. This will likely lead to the feature being banned in certain portions of code bases.
+
+This proposals seeks to remove the abstraction penalties around `Index` and `Range`: both at an API and performance 
+layer. The goal being that most collection types just work today with minimal change and the feature can be adopted
+in existing code without fear of lost performance.
+
+This proposal specifically does not want to change how `Index` and `Range` type binding occurs. The index and range
+expressions continue to have the same syntax and types.
 
 ## Detailed Design 
-- Does not change the existing type logic. 
-
 ### Indexable types
 Any type which has an accessible getter property named `Length` or `Count` with a return type of `int` is considered
-Indexable. The language can use this property to efficiently convert an `Index` expression into an `int` at the 
-point of the expression without relying on the implementation of receiver to do so. 
+Indexable. The language can make use of this property to convert an index expression into an `int` at the point of 
+the expression without the need to use the type `Index` at all. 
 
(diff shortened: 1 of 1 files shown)


proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+This proposals seeks to remove the abstraction penalties around `Index` and `Range`: both at an API and performance 
+layer. The goal being that most collection types just work today with minimal change and the feature can be adopted
+in existing code without fear of lost performance.
+
+This proposal specifically does not want to change how `Index` and `Range` type binding occurs. The index and range
+expressions continue to have the same syntax and types.
+
+## Detailed Design 
+### Countable Types
+Any type which has an property named `Length` or `Count` with an accessible getter adn a return type of `int` is
```
CyrusNajmabadi: adn

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+All arithemtic operations which are emitted will be done so using an `unchecked` context. That matches the context
+in which `Index` and `Range` are compiled in.
+
+### Implicit Index support
+The language will provide an instance indexer member with a single parameter of type `Index` for types which meet the 
+following criteria:
+
+- The type is Countable
+- The type has an accessible instance indexer which takes a single `int` as the argument.
+- The type does not have an accessible instance indexer which takes a single `Index` as the parameter.
```
CyrusNajmabadi: any concerns about params in an indexer?

jaredpar: You mean in the original indexer?

proposals/index-range-changes.md
jaredpar: This is how you know it's authentic and I don't have a ghost writer.

tannergooding:
CC. @eerhardt, @ericstj, @danmosemsft, @KrzysztofCwalina, @terrajobst

jaredpar:
@stephentoub

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+```
+
+This code will print "Get Length 3". 
+
+### Implicit Range support
+The language will provide an instance indexer member with a single parameter of type `Range` for types which meet the 
+following criteria:
+
+- The type is Countable.
+- The type has an accessible member named `Slice` indexer which has two parameters of type `int`.
```
BillWagner: I think this was a copy paste error?

```suggestion
- The type has an accessible member named `Slice` which has two parameters of type `int`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+- The type does not have an instance indexer which takes a single `Range` as the parameter.
+
+For such types the language will bind as if there is an index member of the form `T this[Range range]` where `T` is 
+the return type of the `Slice` method including any `ref` style annotations. The new member will also have matching
+accessibility with `Slice`. 
+
+When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by convering the `Range`
```
BillWagner:
```suggestion
When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by converting the `Range` 
```

proposals/index-range-changes.md
CyrusNajmabadi: yes.  say i had an indexer like so:  

```c#
int this[int i] { get; }
int this[Index x, int y = 0, int z = 0] { get; }
```
 
based on the above, it sounds like you'd generate `int this[Index] { get; }`  Which would then supercede my 'index-indexer'.  Maybe something we just don't care about.  But it's somehting i always notice when we say "takes N parameters".  There's always that bit of ambiguity if it means exactly that, or if it means "not counting optional/params parameters".  

Thanks!

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+Initially this proposal required that types implement `ICollection` in order to qualify as Indexable. That though
+required a number of special cases:
+
+- `ref struct`: these cannot implement interfaces yet types like `Span<T>` are ideal for index / range support. 
+- `string`: does not implement `ICollection` and adding that `interface` has a large cost.
+
+This means to support key types special casing is already needed. The special casing of `string` is less interesting 
+as the language does this in other areas (`foreach` lowering, constants, etc ...). The special casing of `ref struct`
+is more concerning as it's special casing an entire class of types. The get labeled as Indexable if they simply have
```
BillWagner:
```suggestion
is more concerning as it's special casing an entire class of types. They get labeled as Indexable if they simply have
```

proposals/index-range-changes.md
jaredpar: That's a good point. The intent of this is to provide an `Index` indexer if one wasn't available. This would certainly be applicable for the scenario hence probably shouldn't override it.

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+- Else it will be translated as `expr.GetOffset(receiver.Length)`.
+
+This allows for developers to use the `Index` feature on existing types without the need for modification. For example:
+
+``` csharp
+List<char> list = ...;
+var value = list[^1]; 
+
+// Gets translated to 
+var value = list[list.Length - 1];
```
HaloFour: `list.Count - 1`

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+though is not sufficient as it ends up excluding a large number of types:
+
+- Use `Length`: excludes pretty much every collection in System.Collections and sub-namespaces. Those tend to derive 
+from `ICollection` and hence prefer `Count` over length.
+- Use `Count`: excludes `string`, arrays, `Span<T>` and most `ref struct` based types
+
+The extra complication on the initial detection of Indexable types is outweighed by it's simplification in other 
+aspects.
+
+### Choice of Slice as a anme
```
HaloFour: name

danmosemsft:
@jaredpar are you going to follow up separately with @joshfree @tarekgh for the CoreFX changes you mentioned to me this week you would be bringing? I assume they are related to this.

jaredpar:
@danmosemsft 

> are you going to follow up separately with @joshfree @tarekgh for the CoreFX changes you mentioned to me this week you would be bringing? I assume they are related to this.

Yes. Basically once this is implemented CoreFX can delete a number of APIs they added in a previous change. Been trying to lock down the schedule on my side. Think I have it in good shape now.

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+For such types the language will act as if there is an index member of the form `T this[Index index]` where `T` is 
+the return type of the `int` based indexer including any `ref` style annotations. The new member will have the same 
+`get` and `set` members with matching accessibilty as the `int` indexer. 
+
+The new indexer will be implemented by converting the argument of type `Index` into an `int` and emitting a call 
+to the `int` based indexer. For discussion purposes lets use the example of `receiver[expr]`. The conversion of 
+`expr` to `int` will occur as follows:
+
+- When the argument is in the form `^expr2` and the type of `expr2` is `int` it will be translated to
```
poke:
```suggestion
- When the argument is of the form `^expr2` and the type of `expr2` is `int`, it will be translated to 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+``` csharp
+List<char> list = ...;
+var value = list[^1]; 
+
+// Gets translated to 
+var value = list[list.Length - 1]; 
+```
+
+The `receiver` and `Length` expressions will be spilled as appropriate to ensure the side effects are only executed
```
poke:
```suggestion
The `receiver` and `Length` expressions will be spilled as appropriate to ensure any side effects are only executed 
```
(To avoid assuming that there are always side effects)

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Implicit Range support
+The language will provide an instance indexer member with a single parameter of type `Range` for types which meet the 
+following criteria:
+
+- The type is Countable.
+- The type has an accessible member named `Slice` which has two parameters of type `int`.
+- The type does not have an instance indexer which takes a single `Range` as the parameter.
+
+For such types the language will bind as if there is an index member of the form `T this[Range range]` where `T` is
```
poke:
```suggestion
For such types, the language will bind as if there is an index member of the form `T this[Range range]` where `T` is 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+- The type is Countable.
+- The type has an accessible member named `Slice` which has two parameters of type `int`.
+- The type does not have an instance indexer which takes a single `Range` as the parameter.
+
+For such types the language will bind as if there is an index member of the form `T this[Range range]` where `T` is 
+the return type of the `Slice` method including any `ref` style annotations. The new member will also have matching
+accessibility with `Slice`. 
+
+When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by converting the `Range`
```
poke:
```suggestion
When the `Range` based indexer is bound on an expression named `receiver`, it will be lowered by converting the `Range` 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+- The type is Countable.
+- The type has an accessible member named `Slice` which has two parameters of type `int`.
+- The type does not have an instance indexer which takes a single `Range` as the parameter.
+
+For such types the language will bind as if there is an index member of the form `T this[Range range]` where `T` is 
+the return type of the `Slice` method including any `ref` style annotations. The new member will also have matching
+accessibility with `Slice`. 
+
+When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by converting the `Range` 
+expression into two values that are then passed to the `Slice` method. For discussion purposes lets use the example of
```
poke:
```suggestion
expression into two values that are then passed to the `Slice` method. For discussion purposes, lets use the example of 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+the return type of the `Slice` method including any `ref` style annotations. The new member will also have matching
+accessibility with `Slice`. 
+
+When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by converting the `Range` 
+expression into two values that are then passed to the `Slice` method. For discussion purposes lets use the example of 
+`receiver[expr]`.
+
+The first argument of `Slice` will be obtained by converting the typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int` then it will be
```
poke:
```suggestion
- When `expr` is of the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int`, then it will be
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+emitted as `expr1`.
+- When `expr` is in the form `^expr1..expr2` (where `expr2` can be omitted) then it will be emitted as
```
poke:
```suggestion
- When `expr` is of the form `^expr1..expr2` (where `expr2` can be omitted), then it will be emitted as 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+expression into two values that are then passed to the `Slice` method. For discussion purposes lets use the example of 
+`receiver[expr]`.
+
+The first argument of `Slice` will be obtained by converting the typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int` then it will be
+emitted as `expr1`.
+- When `expr` is in the form `^expr1..expr2` (where `expr2` can be omitted) then it will be emitted as 
+`receiver.Length - expr1`
+- When `expr` is in the form `..expr2` (where `expr2` can be omitted) then it will be emitted as `0`
```
poke:
```suggestion
- When `expr` is of the form `..expr2` (where `expr2` can be omitted), then it will be emitted as `0`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+When the `Range` based indexer is bound on an expression named `receiver` it will be lowered by converting the `Range` 
+expression into two values that are then passed to the `Slice` method. For discussion purposes lets use the example of 
+`receiver[expr]`.
+
+The first argument of `Slice` will be obtained by converting the typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int` then it will be
+emitted as `expr1`.
+- When `expr` is in the form `^expr1..expr2` (where `expr2` can be omitted) then it will be emitted as 
+`receiver.Length - expr1`
```
poke:
```suggestion
`receiver.Length - expr1`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+the return type of the `int` based indexer including any `ref` style annotations. The new member will have the same 
+`get` and `set` members with matching accessibilty as the `int` indexer. 
+
+The new indexer will be implemented by converting the argument of type `Index` into an `int` and emitting a call 
+to the `int` based indexer. For discussion purposes lets use the example of `receiver[expr]`. The conversion of 
+`expr` to `int` will occur as follows:
+
+- When the argument is in the form `^expr2` and the type of `expr2` is `int` it will be translated to 
+`receiver.Length - expr2`.
+- Else it will be translated as `expr.GetOffset(receiver.Length)`.
```
poke:
```suggestion
- Otherwise, it will be translated as `expr.GetOffset(receiver.Length)`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+`receiver[expr]`.
+
+The first argument of `Slice` will be obtained by converting the typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int` then it will be
+emitted as `expr1`.
+- When `expr` is in the form `^expr1..expr2` (where `expr2` can be omitted) then it will be emitted as 
+`receiver.Length - expr1`
+- When `expr` is in the form `..expr2` (where `expr2` can be omitted) then it will be emitted as `0`
+- Else it will be emitted as `expr.Start.GetOffset(receiver.Length)`
```
poke:
```suggestion
- Otherwise, it will be emitted as `expr.Start.GetOffset(receiver.Length)`.
```

tarekgh:
@jaredpar please let me know when you have the final schedule so I can know when I can do the work. I hope both the compiler and libraries schedule will be aligned here.

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+emitted as `expr1`.
+- When `expr` is in the form `^expr1..expr2` (where `expr2` can be omitted) then it will be emitted as 
+`receiver.Length - expr1`
+- When `expr` is in the form `..expr2` (where `expr2` can be omitted) then it will be emitted as `0`
+- Else it will be emitted as `expr.Start.GetOffset(receiver.Length)` 
+
+This value will be re-used in the calculation of the second `Slice` argument. When doing so it will be referred to as
+`start`. The second argument of `Slice` will be obtained by converting the range typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int` then it will be
```
poke:
```suggestion
- When `expr` is of the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int`, then it will be 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+`receiver.Length - expr1`
+- When `expr` is in the form `..expr2` (where `expr2` can be omitted) then it will be emitted as `0`
+- Else it will be emitted as `expr.Start.GetOffset(receiver.Length)` 
+
+This value will be re-used in the calculation of the second `Slice` argument. When doing so it will be referred to as
+`start`. The second argument of `Slice` will be obtained by converting the range typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int` then it will be 
+emitted as `expr2 - start`.
+- When `expr` is in the form `expr1..^expr2` (where `expr1` can be omitted) then it will be emitted as
```
poke:
```suggestion
- When `expr` is of the form `expr1..^expr2` (where `expr1` can be omitted), then it will be emitted as 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+- When `expr` is in the form `..expr2` (where `expr2` can be omitted) then it will be emitted as `0`
+- Else it will be emitted as `expr.Start.GetOffset(receiver.Length)` 
+
+This value will be re-used in the calculation of the second `Slice` argument. When doing so it will be referred to as
+`start`. The second argument of `Slice` will be obtained by converting the range typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int` then it will be 
+emitted as `expr2 - start`.
+- When `expr` is in the form `expr1..^expr2` (where `expr1` can be omitted) then it will be emitted as 
+`(receiver.Length - expr2) - start`
```
poke:
```suggestion
`(receiver.Length - expr2) - start`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+- When `expr` is in the form `expr1..` (where `expr1` can be omitted) then it will be emitted as `receiver.Length`
```
poke:
```suggestion
- When `expr` is of the form `expr1..` (where `expr1` can be omitted), then it will be emitted as `receiver.Length`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+- Else it will be emitted as `expr.End.GetOffset(receiver.Length) - start`
```
poke:
```suggestion
- Otherwise, it will be emitted as `expr.End.GetOffset(receiver.Length) - start`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+`start`. The second argument of `Slice` will be obtained by converting the range typed expression in the following way:
+
+- When `expr` is in the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int` then it will be 
+emitted as `expr2 - start`.
+- When `expr` is in the form `expr1..^expr2` (where `expr1` can be omitted) then it will be emitted as 
+`(receiver.Length - expr2) - start`
+- When `expr` is in the form `expr1..` (where `expr1` can be omitted) then it will be emitted as `receiver.Length`
+- Else it will be emitted as `expr.End.GetOffset(receiver.Length) - start` 
+
+The `receiver`, `Length` and `expr` expressions will be spilled as appropriate to ensure the side effects are only
```
poke:
```suggestion
The `receiver`, `Length` and `expr` expressions will be spilled as appropriate to ensure any side effects are only
```
(To avoid assuming that there are always side effects)

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+## Open Issues
+
+## Considerations
+
+### Detect Indexable based on ICollection
+The inspiration for this proposal was collection initializers. Using the structure of a type to convey that it had
+opted into a feature. In the case of collection initializers types can opt into the feature by implementing the 
+interface `IEnumerable` (non generic).
+
+Initially this proposal required that types implement `ICollection` in order to qualify as Indexable. That though
```
poke:
```suggestion
This proposal initially required that types implement `ICollection` in order to qualify as Indexable. That required a number of special cases though:
```
(suggestion includes next line)

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Detect just Count
+Detecting on the property names `Count` or `Length` does complicate the design a bit. Picking just one to standardize
+though is not sufficient as it ends up excluding a large number of types:
+
+- Use `Length`: excludes pretty much every collection in System.Collections and sub-namespaces. Those tend to derive 
+from `ICollection` and hence prefer `Count` over length.
+- Use `Count`: excludes `string`, arrays, `Span<T>` and most `ref struct` based types
+
+The extra complication on the initial detection of Indexable types is outweighed by it's simplification in other
```
poke:
```suggestion
The extra complication on the initial detection of Indexable types is outweighed by its simplification in other 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+The extra complication on the initial detection of Indexable types is outweighed by it's simplification in other 
+aspects.
+
+### Choice of Slice as a anme
+The name `Slice` was chosen as it's the de-facto standard name for slice style operations in .NET. Starting with 
+netcoreapp2.1 all span style types use the name `Slice` for slicing operations. Prior to netcoreapp2.1 there really
+aren't any examples of slicing to look to for an example. Types like `List<T>`, `ArraySegment<T>`, `SortedList<T>`
+would've been ideal for slicing but the concept didn't exist when types were added. 
+
+Thus `Slice` being the sole example it was chosen as the name.
```
poke:
```suggestion
Thus, `Slice` being the sole example, it was chosen as the name.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Index target type conversion
+Another way to view the `Index` transformation in an indexer expression is as a target type conversion. That is instead
```
poke:
```suggestion
Another way to view the `Index` transformation in an indexer expression is as a target type conversion. Instead
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+of binding as if there is a member of the form `return_type this[Index]` the language instead assigns a target
```
poke:
```suggestion
of binding as if there is a member of the form `return_type this[Index]`, the language instead assigns a target
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+This concept could be generalized to all member access on Countable types. Whenever an expression with type `Index` is 
+used as an argument to an instance member invocation and the receiver is Countable then the expression will have a 
+target type conversion to `int`. The member invocations applicable for this conversion include methods, indexers, 
+properties, extension methods, etc ... Only constructors are excluded as they have no receiver. 
+
+The target type conversion will be implemented as follows for any expression which has a type of `Index`. For
+discussion purposes lets use the example of `receiver[expr]`:
+
+- When `expr` is in the form `^expr2` and the type of `expr2` is `int` it will be translated to
```
poke:
```suggestion
- When `expr` is of the form `^expr2` and the type of `expr2` is `int`, it will be translated to 
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+used as an argument to an instance member invocation and the receiver is Countable then the expression will have a 
+target type conversion to `int`. The member invocations applicable for this conversion include methods, indexers, 
+properties, extension methods, etc ... Only constructors are excluded as they have no receiver. 
+
+The target type conversion will be implemented as follows for any expression which has a type of `Index`. For
+discussion purposes lets use the example of `receiver[expr]`:
+
+- When `expr` is in the form `^expr2` and the type of `expr2` is `int` it will be translated to 
+`receiver.Length - expr2`.
+- Else it will be translated as `expr.GetOffset(receiver.Length)`.
```
poke:
```suggestion
- Otherwise, it will be translated as `expr.GetOffset(receiver.Length)`.
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+The `receiver` and `Length` expressions will be spilled as appropriate to ensure the side effects are only executed
```
poke:
```suggestion
The `receiver` and `Length` expressions will be spilled as appropriate to ensure any side effects are only executed
```
(as per above)

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+        Console.WriteLine(i);
+    }
+}
+```
+
+This code will print "Get Length 3". 
+
+This feature would be beneficial to any member which had a parameter that represented an index. For example 
+`List<T>.InsertAt`. This also has the potential for confusion as the language can't give any guidance as to whether
+or not an expression is meant for indexing. All it can do is convert any `Index` expression to `int` in when invoking
```
poke:
```suggestion
or not an expression is meant for indexing. All it can do is convert any `Index` expression to `int` when invoking
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+For simplicity going forward the proposal will use the name `Length` to represent `Count` or `Length`.
+
+### Index and Range implementations are known
+The implementations of `Index` and `Range` are considered to be known and side effect free. Much like 
+`ValueTuple<T1, T2>` the language can assume a standard implementation and emit code inline which represents the 
+implementation of methods on `Index` and `Range`. This implementation includes methods like `GetOffset` or conversions
+like the implicit conversion from `int` to `Index`.
+
+All arithemtic operations which are emitted will be done so using an `unchecked` context. That matches the context
```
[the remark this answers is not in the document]
jaredpar: Ouch. Three on one line.

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+// Gets translated to 
+var value = list[list.Length - 1]; 
+```
+
+The `receiver` and `Length` expressions will be spilled as appropriate to ensure any side effects are only executed 
+once. For example:
+
+``` csharp
+class Collection {
```
lindexi: In [corefx/coding-style.md at master · dotnet/corefx](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md ) NO.1, we use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
+# Index and Range Changes
+
+## Summary
+This proposes to make several changes to the 
+[index and range design](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md) based on 
+customer feedback: particularly from the CoreFX team and their experiences adding Index / Range support to .NET Core.
+The change is not to the syntax but rather how the langauge maps the syntax to APIs.
```
stephentoub: Thank you for driving this, @jaredpar.

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+```
+
+This code will print "Get Length 2". 
+
+The language will special case the following known types: 
+
+- `string`: the method `Substring` will be used instead of `Slice`.
+- `array`: the method `System.Reflection.CompilerServices.GetSubArray` will be used instead of `Slice`.
```
stephentoub: Nit: System.Reflection.CompilerServices.GetSubArray => System.Runtime.CompilerServices.RuntimeHelpers.GetSubArray

It's not a big deal, but I also still question why we defined this method on RuntimeHelpers rather than just making it a static method on Array, along with all the other static generic helper methods on Array.  Seems arbitrary to bury it here when a developer could reasonably call it directly.
cc: @tarekgh, @terrajobst

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+- This conversion is only applicable when the expression with type `Index` is directly an argument to the member. It 
+would not apply to any nested expressions.
+
+## Related Issues
+- https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.cs
+- https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md
+
+## Design Meetings
+- https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-04-01.md
```
stephentoub: @tarekgh, @terrajobst, we should review all of the Index/Range-based members we added to decide what to remove and what to add, if anything.  At a minimum we should remove all of the Index/Range-based indexers that were added.

proposals/index-range-changes.md
djcparker: Should this be list.Length/Count -2 ?
The related spec referred to the end as ^0 which implies that ^1 is one from the end?

marksmeltzer:
@djcparker: Excellent catch!

Regards,

________________________________

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Considerations
+
+### Detect Indexable based on ICollection
+The inspiration for this proposal was collection initializers. Using the structure of a type to convey that it had
+opted into a feature. In the case of collection initializers types can opt into the feature by implementing the 
+interface `IEnumerable` (non generic).
+
+This proposal initially required that types implement `ICollection` in order to qualify as Indexable. That required a number of special cases though:
+required a number of special cases:
```
poke:
```suggestion
```
Guess I shouldn’t have written that one using a suggestion there that you can just apply to a single line 😅

HaloFour:
@djcparker @marksmeltzer 

> Should this be list.Length/Count -2 ?
> The related spec referred to the end as ^0 which implies that ^1 is one from the end?

When indexing, `^0` is after the end and `^1` is the last element.  It provides a symmetry with `Length - x` and is also consistent with languages that allow for negative indexing:

```python
>>> list = [1, 2, 3]
>>> list[-1]
3
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    void Use() { 
+        int i = Get()[^1];
+        Console.WriteLine(i);
+    }
+}
+```
+
+This code will print "Get Length 3".
```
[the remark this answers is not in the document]
yaakov-h: ^1 will return the final element, which has a value of 3.
^0 would be an out of bounds exception.

djcparker:
@HaloFour (and @yaakov-h  )

thank you both for the explanation!

> When indexing, ^0 is after the end and ^1 is the last element. It provides a symmetry with Length - x and is also consistent with languages that allow for negative indexing:

I don't have c#8 installed/enabled at the moment so can't try it, but based on what you say these would produce the same result?

```
var _array = [1,2,3,4,5]
_array[0]   // 1
_array[1..1]  // 1
```
 
as would these:

```
 var _array = [1,2,3,4,5]
 _array[4]   //5
 _array[^1]  //5
 _array[^1..]  //5
```
Reading this article suggests that ranges don't work in that way, maybe the article needs correcting :):
[take-c-8-0-for-a-spin](https://devblogs.microsoft.com/dotnet/take-c-8-0-for-a-spin/)

> But you can also create an Index with a new ^ operator, meaning “from end”. So ^1 is one from the end
> 
> foreach (var name in names[1..^1])
> 
> This lobs off an element at each end of the array, producing an array with the middle three elements.
> 
> Range expressions can be open at either or both ends. ..^1 means the same as 0..^1. 1.. means the same as 1..^0. And .. means the same as 0..^0: beginning to end. Try them all out and see! Try mixing and matching “from beginning” and “from end” Indexes at either end of a Range and see what happens.
> 

Either way, this is an excellent addition to the language.

proposals/index-range-changes.md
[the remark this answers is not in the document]
djcparker: @yaakov-h 

the spec for [ranges](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md) currently states this:

> ^0 does not throw, but it translates to the length of the collection/enumerable it is supplied to.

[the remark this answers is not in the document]
stephentoub: ^0 itself doesn't throw; it's just c.Length. But as with c.Length, passing it to an indexer will generally throw, since there's no element there.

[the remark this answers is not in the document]
djcparker: Good point! Thanks.

HaloFour:
> _array[1..1]  // 1

That would produce an empty slice.  You'd want `_array[0..1]`, the right bound is exclusive so it stops short of that element.  That's also why `_array[0..^0]` succeeds, producing a slice of the entire array.

marksmeltzer:
Thanks for the explanation. Makes sense now ??

Regards,

________________________________

vadimkantorov:
Is indexing from the last dim also supported if `list` is 3-d: `array_3dim[.., ^1]` ? (like Python's `numpy_3dim_array[..., :-1]`) Or would it work only for 2-dim arrays?

yaakov-h:
I don't see why it wouldn't work for multi-dimensional arrays.

As for your second question, I believe you could declare an indexer such as:

```csharp
public object this[params Range[] ranges]
{
    // ...
}
```

proposals/index-range-changes.md
```diff
@@ -0,0 +1,297 @@
...
+
+- Use `Length`: excludes pretty much every collection in System.Collections and sub-namespaces. Those tend to derive 
+from `ICollection` and hence prefer `Count` over length.
+- Use `Count`: excludes `string`, arrays, `Span<T>` and most `ref struct` based types
+
+The extra complication on the initial detection of Indexable types is outweighed by its simplification in other 
+aspects.
+
+### Choice of Slice as a anme
+The name `Slice` was chosen as it's the de-facto standard name for slice style operations in .NET. Starting with
```
tarekgh: Will the compiler special case the string here as it doesn't have Slice but it has Substring?

CyrusNajmabadi: @tarekgh Yup!  that is called out explicitly elsewhere in the doc.

tarekgh: The following statement in the document could be confusing then:

```
After consideration the design was normalized to say that any type which has a property `Count` / `Length` with a 
return type of `int` is Indexable. That removes all special casing, even for `string` and arrays.
```

proposals/index-range-changes.md
tarekgh: https://github.com/Microsoft/dotnet-blog/blob/master/2019/03-Mar/dotnet-core3-preview3/dotnetcore3-preview3.md has all stuff we exposed in preview 3. before then we added only the Span indexers.

proposals/index-range-changes.md
tarekgh: I think having it in System.Runtime.CompilerServices.RuntimeHelpers is implicitly saying only the compiler suppose to call it. why you think the developers will need to call it if they can get the same result in other way?

stephentoub: > why you think the developers will need to call it if they can get the same result in other way?

Does VB expose this range-based array slicing functionality?  F#?

tarekgh: I am not aware of any support of VB or F# for array slicing, @jaredpar may know if there is any work there. But if they do, wouldn't they use System.Runtime.CompilerServices.RuntimeHelpers.GetSubArray too?

stephentoub: That's my attempt at a point :)  Assuming there is no built-in language support, it'd be nice if F# and VB could just call Array.Slice rather than calling System.Runtime.CompilerServices.RuntimeHelpers.GetSubArray.

tarekgh: That make sense. @agocke replied on the other issue that change will be a real breaking change. Earlier version of Roslyn will fail to compile against the newer framework, and vice versa

jaredpar: @cartermp for F# discussion

lostmsu:
I don't know if this was discussed outside of the LDM, but this proposal looks very unclean to me from both user perspective and the compiler developer perspective.

First, it puts quite a bit of work on the compiler to mingle between integers and `Index`.
Second, the way it tries to avoid performance hit caused by `IsFromEnd` checks is quite magical.
Third, it does not include the proposal on [generic `Range`](https://github.com/dotnet/csharplang/issues/2449), and I suspect it will be hard to expand the current proposal to support it in a backwards-compatible way.

To not look like I am just criticizing, here's how one could make it work without the performance penalties:
1. Have a generic `Range<TStart, TEnd>` type.
2. Replace `Index` with `IndexFromEnd<T>`, that does not have the `bool` field
3. Have collections, that want to support range indexing implement `this[Range<int, int>]`, `this[Range<IndexFromEnd<int>, int>]`, `this[int, Range<IndexFromEnd<int>>]`, and `this[Range<IndexFromEnd<int>, Range<IndexFromEnd<int>>]`. Yes, more verbose, but without any magic.

Or, alternatively,

3. Let them only implement `this[Range<int, int>]`, and replace `collection[1..^2]` with `collection[1..new IndexFromEnd<int>(2).Index(collection)]` under the hood when corresponding Range overload is missing. Then have an extension method 

```csharp
int Index<T>(this IndexFromEnd<int> index, ICollection<T> collection)
    => collection.Count - index.Value - 1;
```

providing default implementation.

marksmeltzer:
@lostmsu if they are structs, I kind of like this approach. Or at least, it would be nice to have this baked in. 

As for your suggested "under the hood" conversion, that could be just an implicit conversion operator. No compiler magic required since it could be inlined using normal means.

Problem is that standard collection interfaces won't support these. Not an issue for array, span, string, etc., but there's not a good way for retrofit any new collection operators back into the old interfaces.

I sometimes have a similar operation implemented using something like your range proposal struct plus additional generic structs that also accepts a list, collection, array, etc. Eg. `ArrayRange<T>`, etc., which implement implicit conversion operations and support enumeration, and implement IReadOnlyCollection, IReadOnlyList, etc.

marksmeltzer:
Most of my code like that was from before Span and related types were introduced, which are better suited for arrays. But we still don't have anything built in for slicing up collections, enumerations, or lists.

lostmsu:
@marksmeltzer it can't be an implicit conversion operator, as the conversion is context-dependent: it needs to know the length of the collection.

marksmeltzer:
@lostmsu sure it can. Just allow the range to support negative numbers. Then the range can be used to resolve the actual range boundaries.

So index from the end just means a negative number.

marksmeltzer:
@lostmsu then range could be non genetic and we could also have LongRange that uses longs. I don't see the need for generics here.

lostmsu:
@marksmeltzer having a negative number in the `Range` has a performance penalty similar to the boolean flag in the current `Index` design: you have to add a conditional check to decide how to index on every indexing operation.

verelpode:
@marksmeltzer
> But we still don't have anything built in for slicing up collections, enumerations, or lists.

I too would appreciate slicing functionality, especially advanced scenarios such as the following example with a `CutAndSplice` operation.  In this example, it calculates a moving average with overlapping slicing.  It smooths-out the average by overlapping the segments/slices.  See also the section titled "Other slicing operations" in ["Advanced Extents & Segments Proposal" #2510](https://github.com/dotnet/csharplang/issues/2510).
```
using System.Linq.Enumerable;
static IEnumerable<double> MovingAverage(Segment<int> inData)
{
    foreach (Segment<int> seg in inData.CutAndSplice(inPieceLength: 20, inOverlapLength: 10))
    {
        yield return seg.Average();
    }
}
```

Note I have provided the source code for the CutAndSplice method and some other slicing operations.

@lostmsu
> using negative numbers in Range has a performance penalty

I agree.  More research and real-world practical examples are needed in order to determine whether the added complexity and performance penalty of the negative indexes is counterproductive.  Provided that the total length is supplied via the design/API/syntax, then any range with negative indexes can be converted to a simple extent consisting of offset + length.  The question remains open whether or not it is genuinely worthwhile in real-world scenarios to give `System.Range` the ability to represent negative indexes independently, instead of the simpler alternative of supplying the total length and converting to a simple extent (offset + length).  This question is not very simple to immediately answer.  I think the only way of properly answering it is to collect a number of real-world non-academic examples and convert them to use the proposed Range design and see whether or not the negative indexes are productive or counterproductive.

Some people say that `^0` (`System.Index.End`) has the advantage of being able to represent the end even when the total length is unknown, but `int.MaxValue` is a simpler way of achieving the same thing.  The proposal also supports `^1`, `^2` and so forth, but I haven't seen non-academic examples that demonstrate that this added complexity and performance penalty of negative indexing is not ultimately counterproductive and unnecessary.

lostmsu:
@verelpode majority of the reason for this feature to exist is academic scenarios.

verelpode:
@lostmsu
> majority of the reason for this feature to exist is academic scenarios.

It does look that way (re the negative indexing), therefore I'd suggest that the proposal be reconsidered before being released.  This proposal needs to respect the principle that new language features begin at minus 100 points and then the real-life practicality of the idea/theory needs to demonstrated through the collection of a good quantity of real-world examples, not only academic scenarios.

In medical research, someone produces a new theory and then people can test whether it is true by performing biological experiments.  If the theory is about how bacteria divide, then they can run various experiments to observe how real-life bacteria divide.  This is helpful to avoid becoming divorced from reality.

In contrast, in some other fields including software engineering, there is a higher risk of producing an academic design that isn't genuinely helpful and practical in real-life situations.   When I invent software ideas and theories, I try to make the effort to ask myself:  I designed this nifty thing, but is it genuinely practical in a real software products?  Sometimes it really is.  Sometimes it is little more than academic.  Sometimes I just need to revise the idea and produce a new version.

HaloFour:
@verelpode 

As I've already mentioned countless times, negative indexing is not academic nor controversial.  Quite a few existing languages already support them, and numerous others do through extensions.  They are useful in scenarios of string parsing and the like.  The fact that you do not like this feature does not make it all of these things you accuse it to be and you've approached this entire conversation very emotionally charged and accusatory.  This feature has been under design for _four_ years.  It has already been prototyped and implemented in master and exposed in a preview.  Is it possible that the language team might decide to pull it?  Sure.  Is it likely?  Not remotely.  And a few individuals railing against it with nothing to add except that they don't like it while calling them sloppy will certainly not achieve that result.

PR #2410, opened 2019-04-08
Fix duplicate part from suggestion

poke:
This duplicate sentence was missed when #2397 got merged (/cc @jaredpar).


jaredpar:
Thanks for cleaning that up (and all the other suggestions).

Issue #2435, opened 2019-04-14
Recommend <remarks> rather than <remark>


carlreinke:
The C# documentation recommends the [`<remark>`](https://github.com/dotnet/csharplang/blob/eafa3a5f513559449765bf4928d529455a788a76/spec/documentation-comments.md#remark) tag, but everyone uses `<remarks>`.

* Visual Studio Intellisence \
![image](https://user-images.githubusercontent.com/11667451/56099609-6b6aa000-5ecc-11e9-80d1-2f5a47a1fe15.png)
* [Sandcastle](https://ewsoftware.github.io/XMLCommentsGuide/html/3671b61e-0695-4f0f-bcb5-7cf9108dd642.htm) -- `<remarks>`
* [doxygen](http://www.doxygen.nl/manual/xmlcmds.html) -- `<remarks>`
* [Microsoft C# Programming Guide](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/remarks) -- `<remarks>`

sharwell:
It looks like this was incorrectly modified from the orginal text in [ECMA-334](https://ecma-international.org/publications/files/ECMA-ST/ECMA-334.pdf) §D.3.12.

PR #2436, opened 2019-04-15
Recommend <remarks> rather than <remark>

scottdorman:
Fixes https://github.com/dotnet/csharplang/issues/2435 and updates `<remark>`  to `<remarks>`, per the original text in [ECMA-334](https://ecma-international.org/publications/files/ECMA-ST/ECMA-334.pdf) §D.3.12.

### DIFF
--- a/spec/documentation-comments.md
+++ b/spec/documentation-comments.md
@@ -1,4 +1,4 @@
-﻿# Documentation comments
+# Documentation comments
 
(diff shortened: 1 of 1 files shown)


jcouv:
Tagging @MadsTorgersen @gafter for review since this is updating the spec.

gafter:
Thank you for your contribution!

PR #2445, opened 2019-04-17
fix fixed_size_buffer_declaration syntax

johnnypolychronidis:
I added the "fixed_size_buffer_declarators" production because the current syntax does not require the 'fixed_size_buffer_declarator' to be separated by a ',' token.


jcouv:
Tagging @MadsTorgersen @gafter for review.

PR #2362, opened 2019-03-25
Fix Default Interface Methods proposal link to wikipedia definition of `Trait`

dpoeschl:
Super simple link syntax correction.

PR #2465, opened 2019-04-24
Fix wiki link

jakub-klima:


jakub-klima:
Duplicates https://github.com/dotnet/csharplang/pull/2362 therefore closing.

PR #2523, opened 2019-05-13
Return storage that exists

BillWagner:
The sample returns a field that is not defined in the sample. I assume the code should return the referenced element in the array.

Fixes dotnet/docs#12301

### DIFF
--- a/proposals/csharp-7.2/readonly-ref.md
+++ b/proposals/csharp-7.2/readonly-ref.md
@@ -332,7 +332,7 @@ struct ImmutableArray<T>
     public ref readonly T ItemRef(int i)
     {
         // returning a readonly reference to an array element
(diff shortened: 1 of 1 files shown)


proposals/csharp-7.2/readonly-ref.md
```diff
@@ -332,7 +332,7 @@ struct ImmutableArray<T>
     public ref readonly T ItemRef(int i)
     {
         // returning a readonly reference to an array element
-        return ref this.r1;
+        return ref this.arrary[i];
```
DanFTRX: arrary... You had one job

BillWagner:
```suggestion
        return ref this.array[i];
```

Issue #2672, opened 2019-07-19
[WIP] Correctify Unicode-related errors and omissions in "Lexical structure"


srutzky:
{ This is a work in progress.. I will be updating this to add more content and examples over the next hour }
==

Background
--

Large portions of the Unicode info in the "[Lexical structure](https://github.com/dotnet/csharplang/blob/master/spec/lexical-structure.md)" documentation is either missing, incorrect, or misleading. The main area of concern is with Unicode escape sequences, especially related to supplementary characters. And, it has been this way since the very beginning (almost 20 years now?). It has not helped that there was no example of the "\U" escape sequence anywhere, and the only description was in the ["String Escape Sequences" section of the C# Strings documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#string-escape-sequences) which said, "for surrogate pairs.", which is not exactly correct, but definitely misleading. That wording, coupled with the lack of any examples to guide interpretation, leads one to think that you specify the two surrogate code points as there are 8 hex digits available and that's what is required for two UTF-16 surrogate code points.

Cha cha cha changes to this page
--

1. Broke `unicode_escape_sequence` token into two tokens: `unicode_bmp_escape_sequence` and `unicode_supplementary_escape_sequence`. This not only accurately reflects reality, but it also helps distinguish in the ANTLR grammar (not just in text comments) the set of Unicode characters that can be used in character literals and identifiers vs those that are invalid for those two cases.

2. Specify in ANTLR grammar the constrained usage of "\U" that creates BMP characters and is thus equivalent to "\u" where "\u" is valid to use.

3. Specify in ANTLR grammar how to create supplementary characters using both "\u" to specify an actual surrogate pair, and "\U" to specify a Unicode code point / UTF-32 code unit (they are synonymous). Along with the updates I recently made to the ["Escape Sequences" section of the C# Strings documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#string-escape-sequences) to show usable range and an example, this should prevent any more misinterpreting of "\U".

[...]

Issue #1742, opened 2018-07-26
Champion: "Permit surrogate pairs and wide Unicode-escaped code points in identifiers"


gafter:
- [ ] Proposal added:
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

The C# [language specification](https://github.com/dotnet/csharplang/blob/master/spec/lexical-structure.md#identifiers) already requires that we accept Unicode letters (etc) in identifiers. But the Roslyn compiler doesn't support Unicode letters higher than U+10000 (https://github.com/dotnet/roslyn/issues/9731)(https://github.com/dotnet/roslyn/issues/13474), or those that are represented as a Unicode escape sequence for such a code point (https://github.com/dotnet/roslyn/issues/13560). We could treat it as a bug and fix it, but that fix probably needs to be tied to a language version change to prevent a backward breaking change. This championed issue is to track the fix (as a language change) in some language version.

svick:
How would allowing this break backwards compatibility?

alrz:
@svick since only an earlier version of the compiler would be able to parse those identifiers, this should be tied to a lang version to not silently break code in older versions (i.e. this is an observable "bug fix")

gafter:
@svick A source file that successfully compiles using language version 6 on a recent compiler should also compile using an older version of the compiler in the same language version. Similarly, APIs described in an assembly that was compiled with a recent compiler using language version 6 should be usable by an older compiler using language version 6. Both of these would be violated by permitting a newer compiler to accept identifiers that were rejected by previous compilers.

ufcpp:
@svick 
https://github.com/dotnet/roslyn/blob/master/docs/compilers/CSharp/Unicode%20Version.md
Unicode sometimes has braking changes.

Joe4evr:
As a prime example: Ask Jon Skeet about the Mongolian Vowel Separator. :upside_down_face:

ufcpp:
@ghord 
Do you want any characters to be supported? or do you have any rule?
Many languages - C#, Java, Go, Python, Elixir, etc. - support Unicode identifiers as defined in the [Unicode Annex #31](http://unicode.org/reports/tr31/).

ghord:
@ufcpp 
[...]

PR #2675, opened 2019-07-21
Correctify Unicode-related errors and omissions: escape sequences

srutzky:
Large portions of the Unicode info here are either missing, incorrect, or misleading. The changes made this time, outlined below, deal mostly with Unicode escape sequences, but also touch on identifiers in general.

Details on why these changes are being made, why the ANTLR grammar and text descriptions are being specified in the new way, example code where applicable, and links to Unicode.org as reference / support where applicable, are posted in the following issue: #2672 

1. Broke `unicode_escape_sequence` token into two tokens: `unicode_bmp_escape_sequence` and `unicode_supplementary_escape_sequence`. This not only accurately reflects reality, but it also helps distinguish in the ANTLR grammar (not just in text comments) the set of Unicode characters that can be used in character literals and identifiers vs those that are invalid for those two cases.

2. Specify in ANTLR grammar the constrained usage of "\U" that creates BMP characters and is thus equivalent to "\u" where "\u" is valid to use.

3. Specify in ANTLR grammar how to create supplementary characters using both "\u" to specify an actual surrogate pair, and "\U" to specify a Unicode code point / UTF-32 code unit (they are synonymous). Along with the updates I recently made to the ["Escape Sequences" section of the C# Strings documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/strings/#string-escape-sequences) to show usable range and an example, this should prevent any more misinterpreting of "\U".

4. Introduce the concept of "Supplementary Characters". That term has never been used in this document, which I believe has lead some to mistakenly use the term "surrogate pairs" (which was mentioned here only once, though not incorrectly) to mean "supplementary characters". Surrogate Pairs is a UTF-16 encoding-specific construct. It does not refer to the set of characters, but instead to how they are physically encoded, but only in UTF-16. Neither UTF-8 nor UTF-32 have surrogate pairs, yet they both certainly support all supplementary characters.

5.  Clean-up some inconsistent terminology and formatting:
    1. "code point" vs "value" 
    2. U+1234 vs `U+1234`
    3. U+1234 vs 0x1234
  &nbsp;

6. Provide an accurate and thorough description of BMP range/characters vs Supplementary range/characters, what constitutes a valid surrogate pair, that surrogate pairs are an encoding construct specific to UTF-16, and how all of that relates to `char` vs `string` literals and usage of "`\u`" and "`\U`".

7. Clarify that only BMP characters (direct or via `unicode_bmp_escape_sequence` ) can be used in identifiers. Previously, the ANTLR grammar did not prevent supplementary characters from being used, and there was not text description stating that they couldn't be used. However, they cannot be used (at least not in LINQPad 5 or VS 2015 -- related request: https://github.com/dotnet/csharplang/issues/1742 -- though "gmcs 4.6.2" via Mono does seem to allow them, possibly due to incomplete supplementary character definition here).

8. Added new token for "`hex_digit_except_zero`" to be used in the new "`unicode_supplementary_escape_sequence`" token, else the definition for "unicode_supplementary_escape_sequence" would have been rather messy / harder to read.

9. Adjusted definition of "`hex_digit`" to use new "`hex_digit_except_zero`" token. All previous uses of "`hex_digit`" are unaffected.

10. Some English grammatical improvements to "Note:" at beginning of **Character literals** section.

11. Updated ANTLR grammar for **Character literals** to specify only BMP characters can be used. Previously the grammar did not prevent supplementary characters from being used, and it was only noted in the text description.

12. Removed the "A Unicode character escape sequence ... must be in the range U+0000 to U+FFFF." sentence from the **Character literals** section as it is now doubly redundant, given the sentence about "code points above U+FFFF will error", _and_ the ANTLR grammar finally restriction the values to only the allowable range.

13. For clarity and to reduce any potential for confusion, stated that supplementary characters are allowed in string literals.

---

For more info on this project, please see:
     [Unicode Escape Sequences Across Various Languages and Platforms (including Supplementary Characters)](https://sqlquantumleap.com/2018/09/28/native-utf-8-support-in-sql-server-2019-savior-false-prophet-or-both/#csharp)

### DIFF
--- a/spec/lexical-structure.md
+++ b/spec/lexical-structure.md
@@ -195,12 +195,28 @@ A Unicode character escape sequence represents a Unicode character. Unicode char
 
 ```antlr
 unicode_escape_sequence
+    : unicode_bmp_escape_sequence
+    | unicode_supplementary_escape_sequence
+    ;
+
+unicode_bmp_escape_sequence
     : '\\u' hex_digit hex_digit hex_digit hex_digit
-    | '\\U' hex_digit hex_digit hex_digit hex_digit hex_digit hex_digit hex_digit hex_digit
+    | '\\U0000' hex_digit hex_digit hex_digit hex_digit
+    ;
+
+unicode_supplementary_escape_sequence
+    : '\\uD' ( '8' | '9' | 'A' | 'a' | 'B' | 'b' ) hex_digit hex_digit
+      '\\uD' ( 'C' | 'c' | 'D' | 'd' | 'E' | 'e' | 'F' | 'f' ) hex_digit hex_digit
+    | '\\U00' ( '10' | '0' hex_digit_except_zero ) hex_digit hex_digit hex_digit hex_digit
     ;
 ```
 
-A Unicode escape sequence represents the single Unicode character formed by the hexadecimal number following the "`\u`" or "`\U`" characters. Since C# uses a 16-bit encoding of Unicode code points in characters and string values, a Unicode character in the range U+10000 to U+10FFFF is not permitted in a character literal and is represented using a Unicode surrogate pair in a string literal. Unicode characters with code points above 0x10FFFF are not supported.
+A Unicode escape sequence represents the single Unicode character formed by the hexadecimal number following the "`\u`" or "`\U`" characters. Since C# uses UTF-16 — the 16-bit Unicode encoding — to encode string values, each `char` object represents a single code point in the range of U+0000 to U+FFFF. This range of code points is the Basic Multilingual Plane (BMP). Supplementary characters are code points in the range of U+10000 to U+10FFFF, which make up the supplementary planes. In UTF-16, supplementary characters are represented by two surrogate code points — a high-surrogate code point (U+D800 to U+DBFF) followed by a low-surrogate code point (U+DC00 to U+DFFF) — known as a surrogate pair. Therefore:
+
+* Unicode characters with code points above U+FFFF are represented internally by two characters, used to form the surrogate pair, and are permitted in a string literal, but not in a character literal.
+* The "`\u`" escape sequence cannot directly handle code points above U+FFFF, but it can be used in combination with two surrogate code points to produce a supplementary character.
+* The "`\U`" escape sequence can directly handle the entire range of Unicode code points, from U+0000 to U+10FFFF.
+* Unicode does not have any code points greater than U+10FFFF, and attempting to use "`\U`" with a value greater than "`0010FFFF`" is invalid and will result in a compile-time error.
 
 Multiple translations are not performed. For instance, the string literal "`\u005Cu005C`" is equivalent to "`\u005C`" rather than "`\`". The Unicode value `\u005C` is the character "`\`".
 
@@ -231,6 +247,8 @@ class Class1
 
 The rules for identifiers given in this section correspond exactly to those recommended by the Unicode Standard Annex 31, except that underscore is allowed as an initial character (as is traditional in the C programming language), Unicode escape sequences are permitted in identifiers, and the "`@`" character is allowed as a prefix to enable keywords to be used as identifiers.
 
+Note that supplementary characters are invalid for identifiers, whether specified directly or by a Unicode escape sequence.
+
 ```antlr
 identifier
     : available_identifier
@@ -259,28 +277,28 @@ identifier_part_character
     ;
 
 letter_character
-    : '<A Unicode character of classes Lu, Ll, Lt, Lm, Lo, or Nl>'
-    | '<A unicode_escape_sequence representing a character of classes Lu, Ll, Lt, Lm, Lo, or Nl>'
+    : '<A Unicode BMP character of classes Lu, Ll, Lt, Lm, Lo, or Nl>'
+    | '<A unicode_bmp_escape_sequence representing a character of classes Lu, Ll, Lt, Lm, Lo, or Nl>'
     ;
 
 combining_character
-    : '<A Unicode character of classes Mn or Mc>'
-    | '<A unicode_escape_sequence representing a character of classes Mn or Mc>'
+    : '<A Unicode BMP character of classes Mn or Mc>'
+    | '<A unicode_bmp_escape_sequence representing a character of classes Mn or Mc>'
     ;
 
 decimal_digit_character
-    : '<A Unicode character of the class Nd>'
-    | '<A unicode_escape_sequence representing a character of the class Nd>'
+    : '<A Unicode BMP character of the class Nd>'
+    | '<A unicode_bmp_escape_sequence representing a character of the class Nd>'
     ;
 
 connecting_character
-    : '<A Unicode character of the class Pc>'
-    | '<A unicode_escape_sequence representing a character of the class Pc>'
+    : '<A Unicode BMP character of the class Pc>'
+    | '<A unicode_bmp_escape_sequence representing a character of the class Pc>'
     ;
 
 formatting_character
-    : '<A Unicode character of the class Cf>'
-    | '<A unicode_escape_sequence representing a character of the class Cf>'
+    : '<A Unicode BMP character of the class Cf>'
+    | '<A unicode_bmp_escape_sequence representing a character of the class Cf>'
     ;
 ```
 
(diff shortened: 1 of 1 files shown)


gafter:
The spec in this repository is intended to describe C# 6.  Any significant update other than typographical will be handled by ECMA.  The process for that does not yet exist.  In the meantime we want to save these proposed changes for consideration for the next iteration of the ECMA specification.

srutzky:
> 1. what impact will these changes have on the set of programs that are _accepted_ or _rejected_ by the C# language (not compiler)?  i.e. will these changes allow for programs that were previously accepted by the language to now be rejected?  Or, vice versa?
> 2. will this language change mean that a previously accepted program will still be accepted but may change meaning as far as the language is concerned?  i.e. will a sequence of characters now be interpreted differently?

Hi @CyrusNajmabadi . Sorry for the delay in responding, but for a variety of reasons I've not had a chance to respond until now.

The changes I am proposing represent absolutely no functional differences. The goal of my proposed changes is merely to bring the C\# specification to a state of being explicitly clear about what it is allowing, and how the compiler is behaving. Of course, this assumes that determining what code is _acceptable_ is based on both the ANTLR grammar _and_ the text descriptions. If _accepted_ vs _rejected_ is based on the ANTLR grammar alone, then there is a _theoretical_ chance of currently accepted programs being rejected (due to the ANTLR grammar not being specific enough), but such programs wouldn't exist in the real world because any code that satisfies the current ANTLR grammar but fails on my updated ANTLR grammar would never compile due to being invalid code. For example, the ANTLR grammar currently allows for a Unicode escape sequence in the form of `\UD83DDE3A`, which is a valid surrogate pair, but is an invalid use of the `\U` escape sequence. Of course, the current wording of the specification does not provide any guidance indicating that the proper usage is instead `\U0001F63A` (for the same character), but this is clearly a) what was intended, and b) how the compiler has always behaved (along with being a nearly universal usage of `\U` across a variety of languages).

Again, my proposed changes merely remove ambiguity without altering behavior. So, to directly answer your two questions:

1. There is no impact to what code is either accepted or rejected.
1. There is no potential for changes in interpretation of any given sequence of characters.

That all being said, I will mention, in all fairness, that there is some disagreement on the subject of Identifiers. My proposed changes include clarification that only _BMP_ characters (of the currently defined General Categories) are valid for Identifiers (whether literal or via escape sequence). I am very confident that this is the correct interpretation of a) the original intention of the C\# language specification, _and_ b) the past and current behavior of the compiler, based on all available evidence. Yes, I realize that the specification does not exclude supplementary characters, but it did not need to as there were not any such characters defined. In fact, the Unicode Standard, version 3.0, published in 2000, in section 2.2, in the middle of page 12, states:

> None of these surrogate pairs has been assigned in this version of the standard.

It was not until the next version, 3.1, published in 2001, that any supplementary characters were defined. Please also note the use of the term "surrogate pairs" to refer to supplementary characters. The term "supplementary characters" was officially chosen in late 2000, and has been used ever since version 3.1 when the first supplementary characters were introduced. This shows that the terminology and behavior was set based on Unicode version 3.0 (which makes sense as that is the version stated in the specification), and nobody has ever gone back to update definitions or behavior.

It is quite unfortunate that the original writers of the C\# specification were not more explicit in defining that "any Unicode character of X general category" was really "any Unicode _BMP_ character..." because the resulting ambiguity opened the door for the current disagreement about whether supporting supplementary characters in Identifiers is a bug or intended behavior. It is only by looking backwards in time, now that supplementary characters do exist and some have been designated as valid for Identifiers, that it's possible to interpret the phrase "any Unicode character..." as intending to include supplementary characters, thereby making the current compiler's lack of support for them in Identifiers appear to be a bug instead of an area to be improved.

CyrusNajmabadi:
> The goal of my proposed changes is merely to bring the C# specification to a state of being explicitly clear about what it is allowing, and how the compiler is behaving. 

Would my suggestion not do that?

The biggest concern i have is in validating that hte current impl matches your suggested spec changes.  I don't know how i would actually go about doing that TBH.  Whereas, i can trivially validate the impl against the grammar i've put forth. 

> It is only by looking backwards in time, now that supplementary characters do exist and some have been designated as valid for Identifiers, 

When you say 'designated as valid for Identifiers', who do you mean by?  The C# LDM?  Or someone related to the unicode standard?

CyrusNajmabadi:
Note.  When i say "Would my suggestion not do that?" i mean the suggestion here: https://github.com/dotnet/csharplang/issues/2698#issuecomment-516193124

Basically, your posts seem to bring a *large* amount of complexity to them that doesn't seem to be providing significant value.  I get that there is a concern about the spec being underspecified.  But where i'm getting lost is why we can't just tweak things slightly here to simply document what the actual defacto compiler does, and leave it at that.

srutzky:
> The spec in this repository is intended to describe C# 6. Any significant update other than typographical will be handled by ECMA. The process for that does not yet exist. In the meantime we want to save these proposed changes for consideration for the next iteration of the ECMA specification.

Hi @gafter . Sorry for the delay in responding, but for a variety of reasons I've not had a chance to respond until now.

Yes, the goal of my proposed changes is to remove all ambiguities such that the specification in this repository will finally describe all current and past versions of C\#. The specification has never been correct in terms of how it expresses the intended behavior. There are no behavioral changes being proposed here. This is why I left out any mention of the [obsolete definitions of valid Identifier characters](https://github.com/dotnet/csharplang/issues/2698) and handled that issue separately.

And yes, it would be nice to find some way to propose these changes to ECMA because the published specification has the same overly broad definitions, and lack of information regarding the `\U` escape sequence, that only cause confusion.

The only changes that are candidates for not merging into the specification are the ones concerning valid Identifier characters. Given that the desire is for C\# to fully support supplementary characters, there is likely no need to copy the clarification of them being BMP characters only, even if that does accurately describe current behavior of the compiler.

srutzky:
> > It is only by looking backwards in time, now that supplementary characters do exist and some have been designated as valid for Identifiers,
>
> When you say 'designated as valid for Identifiers', who do you mean by? The C# LDM? Or someone related to the unicode standard?

Hi @CyrusNajmabadi . By "designated as valid..." I simply mean included in the "DerivedCoreProperties.txt" file, in one of the following lists:

* `ID_Start`
* `ID_Continue`
* `XID_Start`
* `XID_Continue`


> Note. When i say "Would my suggestion not do that?" i mean the suggestion here: [#2698 (comment)](https://github.com/dotnet/csharplang/issues/2698#issuecomment-516193124)
> 
> Basically, your posts seem to bring a _large_ amount of complexity to them that doesn't seem to be providing significant value. I get that there is a concern about the spec being underspecified. But where i'm getting lost is why we can't just tweak things slightly here to simply document what the actual defacto compiler does, and leave it at that.

Regarding your suggestion, I would say no, that does not actually work. I will respond to that with explanation in that issue.

Regarding complication, I think you are confusing the Identifier issue ( #2698 ) with what is being proposed here. They are separate. My proposal here is not adding any complication as it is not changing anything. It is merely fixing the ANTLR grammar and related text description such that they explicitly state what has only been implied / assumed for nearly 20 years now.

The changes proposed here (with the possible exception of the updates to the Identifiers section) are completely independent of any changes made (or not made) to how valid Identifier characters are determined. None of my changes here depend on the Identifier rule changes, nor are affected by them or their absence.

I am simply trying to bring the spec up to where it wanted to be from the very beginning, and where it will continue to be with or without these changes. I included the Identifier changes (clarifying that they were meant to be BMP only) for accuracy, and for the idea that we should start with where we are today (and have always been), and modify the definition of Identifiers to include supplementary characters are part of the several issues that have been opened requesting such support. However, if it is causing too much confusion to maintain that ideal here, I can remove the Identifier references from this PR (and update the related issue: https://github.com/dotnet/csharplang/issues/2672 ) as those will be short-lived changes anyway (given the desire to support supplementary characters). But everything else here stays the same, and again, merely describes the behavior as it currently is and always has been.

CyrusNajmabadi:
> Regarding your suggestion, I would say no, that does not actually work. I will respond to that with explanation in that issue.

I'm surprised by that, given that it's just speaking what's in code out loud.  But i look forward to your response :)

gafter:
Most of this change has no normative effect on the meaning of the specification, and therefore I'd prefer not to accept it.  The one normative change is the addition of

> Note that supplementary characters are invalid for identifiers, whether specified directly or by a Unicode escape sequence.

A normative sentence like this would not normally be introduced by "Note that", which is usually used to clarify an implication of the rest of the specification.  In any case this has not been accepted as a language change (despite it describing every existing implementation).  If you want that change to the specification, please propose it separately.

PR #2717, opened 2019-08-08
Update recordsv2.md

leotsarev:
Seems according to prev. text that "final generated code" should have `[WithConstructor]` attribute on `With` method. It's omission, isn't it?


leotsarev:
@agocke ?

agocke:
LGTM, thanks

PR #2754, opened 2019-08-28
Removed Enum and Delegate from prohibited type constraints

### LINKED ISSUE #2753: Incorrect class_type requirements in the specification
From [the specification](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#type-parameter-constraints):

> A class_type constraint must satisfy the following rules:
>
> * The type must not be one of the following types: System.Array, System.Delegate, System.Enum, or System.ValueType.

This is incorrect since C# 7.3 which allows to specify `Enum` and `Delegate` as a class type constraint.

YohDeadfall:
Fixes #2753.

### DIFF
--- a/spec/classes.md
+++ b/spec/classes.md
@@ -308,7 +308,7 @@ A *class_type* constraint must satisfy the following rules:
 
 *  The type must be a class type.
 *  The type must not be `sealed`.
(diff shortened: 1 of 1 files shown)


YohDeadfall:
Closing because the spec about C# 6.0.

PR #2797, opened 2019-09-10
fold ranges changes into initial spec

BillWagner:
See https://github.com/dotnet/docs/issues/12772#issuecomment-501451870

This will read better as a coherent story for the Range and Index features as delivered.

Notes on changes:

1. index-range-changes is deleted. 
1. content from index-range-changes.md was moved into ranges.md
1. Fix markdown lint issues.

None of the semantics of the proposal was changed.

One question for reviewers: Should I change the verb tense from future to present tense on this proposal?

Supports dotnet/docs#14344


proposals/csharp-8.0/ranges.md
```diff
@@ -133,13 +133,153 @@ var multiDimensional = list[1..2, ..]   // list[Range.Create(1, 2), Range.All]
...
+- When `expr` is of the form `expr1..expr2` (where `expr2` can be omitted) and `expr1` has type `int`, then it will be emitted as `expr1`.
+- When `expr` is of the form `^expr1..expr2` (where `expr2` can be omitted), then it will be emitted as `receiver.Length - expr1`.
+- When `expr` is of the form `..expr2` (where `expr2` can be omitted), then it will be emitted as `0`.
+- Otherwise, it will be emitted as `expr.Start.GetOffset(receiver.Length)`.
+
+This value will be re-used in the calculation of the second `Slice` argument. When doing so it will be referred to as `start`. The second argument of `Slice` will be obtained by converting the range typed expression in the following way:
+
+- When `expr` is of the form `expr1..expr2` (where `expr1` can be omitted) and `expr2` has type `int`, then it will be emitted as `expr2 - start`.
+- When `expr` is of the form `expr1..^expr2` (where `expr1` can be omitted), then it will be emitted as `(receiver.Length - expr2) - start`.
+- When `expr` is of the form `expr1..` (where `expr1` can be omitted), then it will be emitted as `receiver.Length`.
```
pkulikov: In all other cases `start` is subtracted to get the number of elements to slice.
```suggestion
- When `expr` is of the form `expr1..` (where `expr1` can be omitted), then it will be emitted as `receiver.Length - start`.
```

BillWagner: @pkulikov I think the original is correct. Consider:

```csharp
var slice = someArray[5..];
```

The range would be from `5` to `someArray.Length`, not from `5` to `someArray.Length-5`.

Did I analyze it incorrectly?

pkulikov: @BillWagner as far as I understand, the second argument of the `Slice` method is the *number of elements* in the range, not the end of the range. Both the end and start of the range are used to get that number.

In the case of arrays, there is no `Slice` method, but the [RuntimeHelpers.GetSubArray](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimehelpers.getsubarray?view=netstandard-2.1) method is used. In its [implementation](https://github.com/dotnet/corefx/blob/3c3b9ceee6e1416db6f1ddbc1ab0d0ffb67c16fa/src/Common/src/CoreLib/System/Runtime/CompilerServices/RuntimeHelpers.cs#L19-L51) one of the first lines gets `offset` (which would be `5` in your example) and `length` (which would be `someArray.Length - 5`)

BillWagner: Thanks @pkulikov 

After a second read, you are correct. Updated.

proposals/csharp-8.0/ranges.md
```diff
@@ -133,13 +136,149 @@ var multiDimensional = list[1..2, ..]   // list[Range.Create(1, 2), Range.All]
 
 Moreover, `System.Index` should have an implicit conversion from `System.Int32`, in order not to need to overload for mixing integers and indexes over multi-dimensional signatures.
 
-## Workarounds
+## Adding Index and Range support to library types
```
agocke: I would clean this up a bit by changing the language from proposal to spec. There's a lot of discussion about the design problems with the previous system which is not really relevant when the new spec is incorporated. In other words, I don't think the historical account is useful for the full spec.

BillWagner: Right. I was chatting about this with @MadsTorgersen today. I do like the idea of making these more like formal specs as they move to completion.

BillWagner: In this instance, I made a smaller set of changes to simplify the reading here.

As another PR, I'd like to rework this (and other proposals) to use more spec like language. Your comment above could be said for many of the proposals. In all cases, I'd like to stage the changes in two phases:

1. Get the technical details correct.
1. Update the language from spec to proposal.

We should discuss which order is better in general, but since this is in progress, I'd like to start with combining these two documents, and then address the language throughout the document in a separate PR.

PR #2806, opened 2019-09-16
Use string interpolation and small fixes

Youssef1313:
- Leave blank lines around snippets.
- Follow code style guidelines.
- Used string interpolation ( Fixes dotnet/docs#14455 )

I noticed that there are no metadata for this `.md` file. Is it missing or there is a reason for that ?

### DIFF
--- a/spec/introduction.md
+++ b/spec/introduction.md
@@ -21,17 +21,21 @@ using System;
 
 class Hello
 {
-    static void Main() {
+    static void Main()
+    {
         Console.WriteLine("Hello, World");
     }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gafter:
We are not likely to accept stylistic changes until after we have integrated with the ECMA specification and are working on the next version of the specification.

jcouv (approved):
LGTM Thanks (iteration 2). Tagging @MadsTorgersen for approval since affects the `specs` folder.

333fred (changes requested):
We should not be changing the specs folder. Spec changes should happen in dotnet/csharpstandard.

Youssef1313:
Going to close for now. Will possibly do a pass over csharpstandard to be consistent everywhere.

PR #2812, opened 2019-09-19
Updates to the function pointers spec.

333fred:
@jkotas @jaredpar for review. Addressed offline comments.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -56,7 +56,7 @@ unsafe class Example {
         func* managed int(int, int) p2 = ...;
         func* cdecl int(int, int) p3 = ...;
 
-        p1 = p2; // okay Func1 and Func3 have compatible signatures
+        p1 = p2; // okay p1 and p2 have compatible signatures
         Console.WriteLine(p2 == p1); // True
         p2 = p3; // error: calling conventions are incompatible
     }
@@ -86,16 +86,14 @@ funcptr_type =
     '(' funcptr_type ')' ;
 
 calling_convention = 
-    'managed' |
-    'unmanaged' |
     'cdecl' |
-    'winapi' | 
-    'fastcall' | 
+    'managed' |
     'stdcall' | 
-    'thiscall' ;
+    'thiscall' |
+    'unmanaged' ;
 ```
 
-The `unmanaged` calling convention represents the default calling convention for native code on the current platform.
+The `unmanaged` calling convention represents the default calling convention for native code on the current platform, and is encoded as winapi.
 
 When there is a nested function pointer, a function pointer which has or returns a function pointer, parens can be 
 optionally used to disambiguate the signature. Though they are not required and the resulting types are equivalent.
@@ -185,7 +183,7 @@ The address-of operator will be implemented using the `ldftn` instruction.
 
 Restrictions of this feature:
 - Only applies to methods marked as `static`.
-- Local functions cannot be used in `&`. The implementation details of these methods are
+- Non-`static` local functions cannot be used in `&`. The implementation details of these methods are
 deliberately not specified by the language. This includes whether they are static vs. instance or
 exactly what signature they are emitted with.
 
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -198,17 +196,17 @@ This means that it is possible to overload on `void*` and a `func*` and still se
...
 It's possible for the runtime and language to work together to fully support this. The language could choose to treat
-address-of `static` members with a `NativeCallback` attribute as a `func*` with the specified calling convention.
+address-of `static` members with a `NativeCallable` attribute as a `func*` with the specified calling convention.
 
 ``` csharp
-unsafe class NativeCallbackExample {
-    [NativeCallback(CallingConvention.CDecl)]
+unsafe class NativeCallableExample {
+    [NativeCallable(CallingConvention.CDecl)]
     static extern bool CloseHandle(IntPtr p);
```
jkotas: The NativeCallable methods are typically going to have a body in C#. A more realistic example may look like this:

```suggestion
    static void Close(IntPtr p) => Marshal.FreeHGlobal(p);
```

jkotas: If you go with my suggestions, `bool` needs to be changed to `void` in the following lines.

proposals/function-pointers.md
```diff
@@ -222,11 +220,11 @@ unsafe class NativeCallbackExample {
...
 
-- Flag any managed calls to a method tagged with `NativeCallback` as an error. Given the function can't be invoked from
+- Flag any managed calls to a method tagged with `NativeCallable` as an error. Given the function can't be invoked from
 managed code the compiler should prevent developers from attempting such an invocation.
-- Prevent method group conversions to `delegate` when the method is tagged with `NativeCallback`. 
+- Prevent method group conversions to `delegate` when the method is tagged with `NativeCallable`. 
 
-This is not necessary to support `NativeCallback` though. The compiler can support the `NativeCallback` attribute as is
+This is not necessary to support `NativeCallable` though. The compiler can support the `NativeCallable` attribute as is
 using the existing syntax. The runtime would simply need to cast to `void*` before casting to the correct `func*`
```
jkotas:
```suggestion
using the existing syntax. The program would simply need to cast to `void*` before casting to the correct `func*` 
```

jkotas:
cc @AaronRobinsonMSFT

jkotas:
Any thoughts what would be the best way to attach additional metadata to the function pointer - to implement https://github.com/dotnet/coreclr/pull/26458#issuecomment-526910943 ?

tannergooding:
Have we come up with a solution of how to handle native functions which are both `thiscall` and `stdcall`?

For example, COM interfaces are C++ "instance" members and therefore take the first parameter as an `implicit this` but they can also be annotated with another calling convention (typically `stdcall`) which can be impactful in some scenarios (such as struct returns).

This is an important scenario for some modern COM interop scenarios, such as DirectX.

Looking at the ECMA-335 table for `StandAloneMethodSig`, it looks like it has all the relevant bits to encode this:
image

That is, you could set both `EXPLICITTHIS` (which requires `HASTHIS` to also be set and which can only be used by `MethodDefSig` marked as `FNPTR`) and the appropriate calling convention (`STDCALL`).

The C# syntax could make use of the existing "extension member" syntax for declaring `this` parameters to allow the appropriate (and familiar) declaration. For example:
```cpp
    MIDL_INTERFACE("7b7166ec-21c7-44ae-b21a-c9ae321ae369")
    IDXGIFactory : public IDXGIObject
    {
    public:
        virtual HRESULT STDMETHODCALLTYPE EnumAdapters( 
            /* [in] */ UINT Adapter,
            /* [annotation][out] */ 
            _COM_Outptr_  IDXGIAdapter **ppAdapter) = 0;

        // More methods
    };
```

Could be exposed in C# as:
```csharp
func* stdcall int(this IDXGIFactory pThis, uint Adapter, IDXGIAdapter** ppAdapter);
```

jkotas:
> Have we come up with a solution of how to handle native functions which are both thiscall and stdcall?

This problem is mentioned in [Extensible set of unmanaged calling conventions](https://github.com/dotnet/csharplang/blob/master/proposals/function-pointers.md#extensible-set-of-unmanaged-calling-conventions) open issue paragraph in the spec. There are many unmanaged calling conventions that this does not cover.

tannergooding:
> This problem is mentioned in Extensible set of unmanaged calling conventions

Ah, ok. I had seen that more of a problem as to how to encode new calling conventions that may come up in the future. Not necessarily in marking the calling convention and also indicating that it is a call to an instance member (I apparently skimmed over the bullet for `StdCall with explicit this` 😄).

jkotas:
The `C` function pointers have the same problem. You cannot make "stdcall with explicit this" function pointer in `C` today. Let's to theoretical exercise what the fix for `C` would be if they were to fix it. My bet would be that they would add a new declspec calling convention. It makes me think about the "stdcall with explicit this" as just another calling convention.

PR #2829, opened 2019-09-26
Span safety samples

jaredpar:
Include same samples and explanations for why several of the `Span<T>`
and `ref struct` rules exist.

### DIFF
--- a/proposals/csharp-7.2/span-safety.md
+++ b/proposals/csharp-7.2/span-safety.md
@@ -157,7 +157,7 @@ An expression whose type is not a `ref struct` type is *safe-to-return* from the
 
 An lvalue designating a formal parameter is *ref-safe-to-escape* (by reference) as follows:
 - If the parameter is a `ref`, `out`, or `in` parameter, it is *ref-safe-to-escape* from the entire method (e.g. by a `return ref` statement); otherwise
-- If the parameter is the `this` parameter of a struct type, it is *ref-safe-to-escape* to the top-level scope of the method (but not from the entire method itself);
+- If the parameter is the `this` parameter of a struct type, it is *ref-safe-to-escape* to the top-level scope of the method (but not from the entire method itself); [Sample](#struct-this-escape)
 - Otherwise the parameter is a value parameter, and it is *ref-safe-to-escape* to the top-level scope of the method (but not from the method itself).
 
 An expression that is an rvalue designating the use of a formal parameter is *safe-to-escape* (by value) from the entire method (e.g. by a `return` statement). This applies to the `this` parameter as well.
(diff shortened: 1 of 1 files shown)


proposals/csharp-7.2/span-safety.md
```diff
@@ -290,6 +288,93 @@ We wish to ensure that no `ref` local variable, and no variable of `ref struct`
...
+{
+    return ref p.Get();
+}
+```
+
+Consider the case where `T` is instantiated as a `struct`. If the `this` parameter is ref-safe-to-escape then the return of `p.Get` could point to the stack (specifically it could be a field inside of the instantiated type of `T`). That means the language could not allow this sample to compile as it could be returning a `ref` to a stack location. On the other hand if `this` is not ref-safe-to-escape then `p.Get` cannot refer to the stack and hence it's safe to return. 
+
+This is why the escapability of `this` in a `struct` is really all about interfaces. It can absolutely be made to work but it has a trade off. The design eventually came down in favor of making interfaces more flexible. 
+
+There is potential for us to relax this in the future though.
```
VSadov: Interfaces allow the caller to be oblivious whether it works with classes or structs. That pushes the safety rules into the calee, where we know the kind of the containing type.

`ref struct` does not implement interfaces, so in theory could treat `this` as ordinary `ref` parameter. I am not 100% sure, but I think we decided against specialcasing `ref` structs vs ordinary structs.

This may not be as easy to relax as it seems. Primarily because you still need something for structs. You can only move where you enforce it - at the caller or at and callee, but such move would be compat breaking.

VSadov:
Nice to see the examples!! 👍

proposals/csharp-7.2/span-safety.md
jaredpar: I think there is actually a really easy way to relax this. Consider if we change the rules as follows:

> The `this` parameter in a `struct` instance member is ref-safe-to-escape if a) if the member does not implement an interface member and b) does not have any `out` or `ref` parameters. 
> Else it is ref-safe-to-escape to the method body only.

That would cover all the cases I'm aware of and be compatible with existing code.

VSadov: Then on the caller side you would need to consider a possibility of the method ref-returning `this`.  That could break things.

```cs
    public ref int M1()
    {
        Span<int> s = new int[]{42} ;

        // the following is currently allowed, 
        // but if indexer can do `return ref Length`, shouldn't this become an error?
        return ref s[0];
    }
```

jaredpar: Hmm that is true as today we consider everything to be heap lifetime if it comes from a `struct` with no other returnable values. That would suggest we have to provide some manner of opt in for this to be allowed. That's unfortunate.

Issue #435, opened 2017-04-13
Champion "Native-Sized Number Types"


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

/cc @KrzysztofCwalina @jaredpar

jkotas:
Proposal: https://github.com/dotnet/corefxlab/issues/1471

CyrusNajmabadi:
```nuint```... :)

jnm2:
Actually, could C# add `nint`, `nunit`, and `nfloat`?

CyrusNajmabadi:
As i mentioned in the other proposal, i don't see a lot of value in C# adding any keywords here.  IntN, FloatN and the like are totally reasonable names to just use as is.  And if you really want a all-lower name, then just add "using nint = System.IntN".

jnm2:
Do you foresee people opting to use `IntN` over `int` as iteration variables?

CyrusNajmabadi:
@jnm2 i'm not sure what you mean?  Can you give an example?  I don't generally think about iteration as being related to native sized ints... so i'm not sure what the connection is.  Thanks!

jnm2:
@CyrusNajmabadi Just a plain `for` loop for example. Everyone defaults to `int` for the iteration variable type, but would `IntN` potentially eke out more performance than `int` on x64?

CyrusNajmabadi:
I would really hope not :)   But i'll let the CLR guys weigh in on that.

jkotas:
> would IntN potentially eke out more performance than int on x64?

native int is more performant than int32 for loops that iterate over memory on 64-bit platforms. int32 tend to require int32->native int casts that are unnecessary overhead. E.g. Check how some methods on `Span<T>` have been optimized to avoid the extra overhead: https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/SpanHelpers.byte.cs#L76

zippec:
I wonder, should nint/nuint be allowed as underlying type of enums as well?

bbarry:
@zippec that would be a breaking change in struct layouts wouldn't it?

tannergooding:
@bbarry, how would that be a breaking change? Only new enums would support being subclassed from nint/nuint and the runtime already supports said functionality.

The runtime also supports native sized constants (although they are actually 32-bit constants that are implicitly converted)

[...]

PR #2833, opened 2019-09-27
Native-sized integers proposal

cston:


proposals/NativeInt.md
```diff
@@ -0,0 +1,247 @@
...
+[design]: #design
+
+New contextual keywords `nint` and `nuint` represent native signed and unsigned integer types.
+The identifiers are only treated as keywords when used in a type context and when the identifier does not otherwise bind to a symbol at that program location.
+
+The types `nint` and `nuint` are essentially aliases to underlying types `System.IntPtr` and `System.UIntPtr`, where the compiler can surface additional conversions and operations for native ints.
+
+### Constants
+
+There is no direct syntax for native int literals. Explicit casts of other integral constant values can be used instead: `(nint)42`.
```
tannergooding: Given that there is an implicit conversion from `int` to `nint`, `nint x = 42` and `nint x = (nint)42` would both be valid, correct?

I presume `const nint x = 42` would also work?

cston: Yes, `const nint x = 42` would work.

---
In reply to: [329172165](https://github.com/dotnet/csharplang/pull/2833#discussion_r329172165) [](ancestors = 329172165)

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+The motivation is for interop scenarios and for low-level libraries.
+
+## Design
+[design]: #design
+
+New contextual keywords `nint` and `nuint` represent native signed and unsigned integer types.
+The identifiers are only treated as keywords when used in a type context and when the identifier does not otherwise bind to a symbol at that program location.
+
+The types `nint` and `nuint` are essentially aliases to underlying types `System.IntPtr` and `System.UIntPtr`, where the compiler can surface additional conversions and operations for native ints.
```
danmosemsft: Presumably a secondary motivation is to reduce confusion? I have always found it confusing that `IntPtr` is often not representing a pointer.

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+`typeof(nint)` is `typeof(IntPtr)`.
+
+`sizeof(nint)` is supported and does not require compiling in an unsafe context. The value is not a compile-time constant. `sizeof(nint)` is implemented as `sizeof(IntPtr)` rather than `IntPtr.Size`.
+
+Compiler diagnostics for type references involving `nint` or `nuint` report `nint` or `nuint` rather than `IntPtr` or `UIntPtr`.
+
+### Metadata
+
+`nint` and `nuint` are represented in metadata as `System.IntPtr` and `System.UIntPtr`.
```
danmosemsft: Do I understand correctly that there is no longer any reason to use `IntPtr` and `UIntPtr`? Even public API can be changed, it seems. We could search and replace our entire codebase?

tannergooding: There may be some reasoning to continue using `IntPtr` if users want to clearly differentiate something that is meant to be an "opaque handle" vs something that is meant to be a "native integer". Something that is meant to be a "pointer" could fall into either category, depending on what operations you need to perform on it.

danmosemsft: An opaque handle is often not a pointer, even under the covers. So the confusion still exists (possibly in my mind only)

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+The only constructor for `nint` or `nuint` is the parameter-less constructor.
+
+The members of `System.IntPtr` and `System.UIntPtr` other than the constructors and operators are available from `nint` and `nuint`.
+For `nint` the following `System.IntPtr` members are available. The members available on `nuint` are similar.
+
+```C#
+public static readonly IntPtr Zero;
+public static int Size { get; }
+public static IntPtr Add(IntPtr pointer, int offset);
```
tannergooding: Is it worth calling out that these members have the behavior of that for `IntPtr`, so some (like `Add` and `Subtract`) may differ in behavior from the `nint` operators?

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+The `native int` aliases and the underlying types are considered equivalent for overriding, hiding and implementing.
+
+Overloads cannot differ by `native int` and underlying type alone.
+Overrides and implementations may differ by `native int` and underlying type alone.
+Methods hide other methods that differ by `native int` and underlying type alone.
+
+### Miscellaneous
+
+`native int` types cannot be used as enum underlying types.
```
tannergooding: Why not? Given the constraints for `constants` (values must be less than `int.MaxValue` and greater than `int.MinValue`) and that the runtime explicitly supports `native int` enums, it seems like this would be beneficial for some interop code?

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+Language support for a native-sized signed and unsigned integer types.
+
+The motivation is for interop scenarios and for low-level libraries.
+
+## Design
+[design]: #design
+
+New contextual keywords `nint` and `nuint` represent native signed and unsigned integer types.
+The identifiers are only treated as keywords when used in a type context and when the identifier does not otherwise bind to a symbol at that program location.
```
gafter: >when the identifier does not otherwise bind to a symbol at that program location [](start = 77, length = 80)

This needs to be specified as a modification to the spec for name lookup.  There is no existing spec concept of when an identifier "does not otherwise bind" to a symbol.

proposals/NativeInt.md
gafter: >essentially aliases [](start = 33, length = 19)

The first half of this sentence and the second half contradict each other (if it is an alias then it has precisely the same set of operations).  I think you want to say that they are *represented by* the underling types in the implementation.

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+The types `nint` and `nuint` are essentially aliases to underlying types `System.IntPtr` and `System.UIntPtr`, where the compiler can surface additional conversions and operations for native ints.
+
+### Constants
+
+There is no direct syntax for native int literals. Explicit casts of other integral constant values can be used instead: `(nint)42`.
```
gafter: >direct syntax for native int literals [](start = 12, length = 37)

Do you intend that constants of these new types exist?  If so, the spec for "constant expressions" would need to be amended.

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+`nint` constants are in the range [ `int.MinValue`, `int.MaxValue` ].
+
+`nuint` constants are in the range [ `uint.MinValue`, `uint.MaxValue` ].
+
+There are no `MinValue` or `MaxValue` fields on `nint` or `nuint` because, other than `nuint.MinValue`, those values cannot be emitted as constants.
+
+Constant folding is supported for all operators: { (unary)`-`, `~`, `+`, `-`, `*`, `/`, `%`, `==`, `!=`, `<`, `<=`, `>`, `>=`, `&`, `|`, `<<`, `>>` }.
+Constant folding operations are evaluated with `Int32` and `UInt32` operands rather than native ints for consistent behavior regardless of compiler platform.
```
gafter: >are evaluated with `Int32` and `UInt32` operands [](start = 28, length = 48)

In other words, they produce different values than are produced at runtime on some runtime platforms.

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+| `nuint` | `char` |   | `conv.u2` / `conv.ovf.u2.un` |
+| `nuint` | `float` | Y | `conv.r.un conv.r4` |
+| `nuint` | `double` | Y | `conv.r.un conv.r8` |
+
+Conversions between `decimal` and `nint` or `nuint` are not supported because there are no operators on `decimal` that use `native int`.
+There are operators on `decimal` to convert to and from `long` or `ulong` but those are not optimal on 32-bit platforms.
+
+### Operators
+
+The following operators are supported.
```
gafter: >are supported [](start = 24, length = 13)

Do you mean to say that these are built-in operators considered when seeking to determine the semantics of an operation in source, like the built-in ones described in [https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#addition-operator](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#addition-operator) ?

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+| `%` | `native` | `native` | `native` | `rem` / `rem.un` |
+| `==` | `native` | `native` | `native` | `beq` / `ceq` |
+| `!=` | `native` | `native` | `native` | `bne` |
+| `<` | `native` | `native` | `native` | `blt` / `clt` / `blt.un` / `clt.un` |
+| `<=` | `native` | `native` | `native` | `ble` / `ble.un` |
+| `>` | `native` | `native` | `native` | `bgt` / `cgt` / `bgt.un` / `cgt.un` |
+| `>=` | `native` | `native` | `native` | `bge` / `bge.un` |
+| `&` | `native` | `native` | `native` | `and` |
+| `\|` | `native` | `native` | `native` | `or` |
+| `<<` | `native` | `int` | `native` | `shl` |
```
gafter: >`<<` [](start = 2, length = 4)

What are the semantics of the shift operators?  For comparison, the spec in [https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#shift-operators](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#shift-operators) has text that begins "For the predefined operators, the number of bits to shift is computed as follows"

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+And interop with existing native int code that uses `IntPtr` would be more difficult. 
+
+Another alternative is to add more native int support for `IntPtr` in the framework but without any specific compiler support.
+Any new conversions and arithmetic operations would be supported by the compiler automatically.
+But the language would not provide keywords, constants, or `checked` operations.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still undecided?
```
gafter: >What parts of the design are still undecided [](start = 0, length = 44)

What is the common type between a value of type `nint` and a value of type `IntPtr`?  The common type between `G<nint>` and `G<IntPtr>`?  When both are bounds in type inference, which wins?

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+
+`nuint` constants are in the range [ `uint.MinValue`, `uint.MaxValue` ].
+
+There are no `MinValue` or `MaxValue` fields on `nint` or `nuint` because, other than `nuint.MinValue`, those values cannot be emitted as constants.
+
+Constant folding is supported for all operators: { (unary)`-`, `~`, `+`, `-`, `*`, `/`, `%`, `==`, `!=`, `<`, `<=`, `>`, `>=`, `&`, `|`, `<<`, `>>` }.
+Constant folding operations are evaluated with `Int32` and `UInt32` operands rather than native ints for consistent behavior regardless of compiler platform.
+
+### Conversions
+Types that differ only by `nint` and `IntPtr` and by `nuint` and `UIntPtr` are considered equivalent. That applies to the primitive types as well as arrays, `Nullable<>`, constructed types, and tuples.
```
gafter: >are considered equivalent [](start = 75, length = 25)

I presume you mean there is an identity conversion between them.

proposals/NativeInt.md
tannergooding: Right, opaque handles are often something you never want to perform arithmetic with; and so they are good candidates to stay as `IntPtr`.

Things which are actually `pointers` you may want to make `nint`, if performing arithmetic is something needed.

Things which are actually `numbers` you probably want to make `nint`, since you want to treat them as numbers.

danmosemsft: Right, hence we're continuing to recommend something that contains `Ptr` in the name for representing things that are not necessarily pointers. That's a shame, but perhaps it isn't a problem this set out to solve.

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+| `/` | `native` | `native` | `native` | `div` / `div.un` |
+| `%` | `native` | `native` | `native` | `rem` / `rem.un` |
+| `==` | `native` | `native` | `native` | `beq` / `ceq` |
+| `!=` | `native` | `native` | `native` | `bne` |
+| `<` | `native` | `native` | `native` | `blt` / `clt` / `blt.un` / `clt.un` |
+| `<=` | `native` | `native` | `native` | `ble` / `ble.un` |
+| `>` | `native` | `native` | `native` | `bgt` / `cgt` / `bgt.un` / `cgt.un` |
+| `>=` | `native` | `native` | `native` | `bge` / `bge.un` |
+| `&` | `native` | `native` | `native` | `and` |
+| `\|` | `native` | `native` | `native` | `or` |
```
gafter: >`\|` [](start = 2, length = 4)

This should be `<code>&#124;</code>` to get it to render properly

gafter:
I don't have any further comments (Iteration 5).

gafter:
From https://github.com/dotnet/csharplang/issues/435#issuecomment-446967104 : please specify that these types are atomic.

proposals/NativeInt.md
```diff
@@ -0,0 +1,263 @@
...
+    public sealed class NativeTypeAttribute : System.Attribute
+    {
+        public NativeTypeAttribute() { }
+        public NativeTypeAttribute(byte[] flags) { }
+    }
+}
+```
+
+The optional attribute argument contains a bit for each primitive type in the type reference.
+If there is a single primitive type, the parameter-less constructor can be used.
```
marek-safar: I think this needs better wording to be clear for cases like `(IntPtr, nint) B`

proposals/NativeInt.md
marek-safar: I find that confusing. What is the rationale for `Add` not to return native type?

nietras:
Good to see a proposal for adding this to C# :+1:

Just to plug my own exercises in futility you might find https://github.com/DotNetCross/NativeInts interesting with regards to API. This lists all operators and includes tests that can be viewed for inspiration.

proposals/NativeInt.md
pentp: Such combinations that could produce different values should be disallowed without explicit `unchecked` context like `int.MaxValue + 1` is disallowed today for `int` constants.

proposals/NativeInt.md
tannergooding: These are existing members on Sustem.IntPtr with a particular behavior, changing that behavior is a breaking change.

GSPP:
Will the `IntPtr` not be touched at all then? If I recall correctly there were proposals to upgrade the API with various miscellaneous improvements. Although that becomes less important with the arrival of `nint` it seems like a good time to take a look at making `IntPtr` nicer and more fully features.

proposals/NativeInt.md
```diff
@@ -0,0 +1,264 @@
...
+nint y = x + x; // ok
+dynamic d = x;
+nint z = d + x; // RuntimeBinderException: '+' cannot be applied 'System.IntPtr' and 'System.IntPtr'
+```
+
+### Type members
+
+The only constructor for `nint` or `nuint` is the parameter-less constructor.
+
+The members of `System.IntPtr` and `System.UIntPtr` other than the constructors and operators are available from `nint` and `nuint`.
```
GSPP: So all future members will automatically become available? That's a good thing. Maybe the spec should not say what is available but rather what members are replaced/added by synthetic members. All other members are just pass-through to `IntPtr`.

proposals/NativeInt.md
```diff
@@ -0,0 +1,264 @@
...
+
+Constant folding is supported for all operators: { (unary)`-`, `~`, `+`, `-`, `*`, `/`, `%`, `==`, `!=`, `<`, `<=`, `>`, `>=`, `&`, `|`, `<<`, `>>` }.
+Constant folding operations are evaluated with `Int32` and `UInt32` operands rather than native ints for consistent behavior regardless of compiler platform.
+
+### Conversions
+Types that differ only by `nint` and `IntPtr` and by `nuint` and `UIntPtr` are considered equivalent. That applies to the primitive types as well as arrays, `Nullable<>`, constructed types, and tuples.
+
+_Are these "identity" conversions between equivalent types represented in the Roslyn API?_
+
+The following numeric conversions are supported.
```
gafter: >The following numeric conversions are supported [](start = 0, length = 47)

Please (also) describe as an adjustment to [https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-numeric-conversions](https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-numeric-conversions) and [https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#explicit-numeric-conversions](https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#explicit-numeric-conversions).  Include the conversion classifications please (identity, implicit numeric, explicit numeric, enum conversion, etc).

Please say if any other conversions exist.  For example, is there an explicit conversion from an enum type to nint?

proposals/NativeInt.md
```diff
@@ -0,0 +1,264 @@
...
+| `nuint` | `char` |   | `conv.u2` / `conv.ovf.u2.un` |
+| `nuint` | `float` | Y | `conv.r.un conv.r4` |
+| `nuint` | `double` | Y | `conv.r.un conv.r8` |
+
+Conversions between `decimal` and `nint` or `nuint` are not supported because there are no operators on `decimal` that use `native int`.
+There are operators on `decimal` to convert to and from `long` or `ulong` but those are not optimal on 32-bit platforms.
+
+### Operators
+
+The following operators are supported.
```
gafter: >The following operators are supported [](start = 0, length = 37)

Please add a specification in the form suitable for use in the spec, e.g. give the declarations of the implicit operators to be considered during overload resolution, like `bool operator >=(nint x, nint y);`

Is there no unary `+` on nint?

tannergooding:
How will this apply to array indexing?

Namely, you can index an array with `int`, `uint`, `long`, or `ulong` today.

I imagine (given that IL for ldelem directly takes `int32` or `native int`) indexing an array with `nint` and `nuint` will also be possible?

tannergooding:
Basically the same question exists for stackalloc and creating an array, given that the IL instructions can directly take `nint`/`nuint`.

proposals/NativeInt.md
tannergooding: I believe the decision in the last LDM meeting was that `nint x = int.MaxValue + 1` would be illegal, even with `unchecked` due to the platform specific differences.

Instead, constant folding would only be allowed if the operation stayed within the 32-bit range limitation put forth by the compiler.

It would probably be worth explaining this limitation or elaborating on how folding works if I'm misremembering.

proposals/NativeInt.md
```diff
@@ -148,14 +163,18 @@ In cases where there are two overloads, one for `nint` and one for `nuint`, the
...
+(see [ECMA-335](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf) III.1.5 Operand type table).
 But the set of operand types supported by C# is limited for simplicity and for consistency with existing operators in the language.
 
 Lifted versions of the operators, where the arguments and return types are `nint?` and `nuint?`, are supported.
 
 Compound assignment operations `x op= y` where `x` or `y` are native ints follow the same rules as with other primitive types with pre-defined operators.
 Specifically the expression is bound as `x = (T)(x op y)` where `T` is the type of `x` and where `x` is only evaluated once.
 
+The shift operators should mask the number of bits to shift appropriately
+(see [shift operators](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#shift-operators) in C# spec).
```
tannergooding: Tomorrow we should be reviewing https://github.com/dotnet/corefx/issues/20256, part of which is exposing a new member on `IntPtr` and `UIntPtr` that would allow this. Such a member (currently going to propose `ShiftMask`, as I've not gotten any better suggestions) would be a runtime (but not a compile time) constant.

This would allow the compiler to emit `nintValue << (shiftCount & nint.ShiftMask)`.

The alternative would be for the compiler to emit `nintValue << (shiftCount & ((sizeof(nint) * 8) - 1))` which does the same thing but doesn't put a dependency on the framework exposing a new API.

proposals/NativeInt.md
```diff
@@ -172,34 +191,29 @@ nint z = d + x; // RuntimeBinderException: '+' cannot be applied 'System.IntPtr'
...
 public int ToInt32();
 public long ToInt64();
 public void* ToPointer();
-public override string ToString();
-public string ToString(string format);
 ```
 
-Some members above use `IntPtr` in the signature even when used from `nint`.
-
+The remaining members of `System.IntPtr` and `System.UIntPtr` _are implicitly included_ in `nint` and `nuint`. For .NET Framework 4.7.2:
```
tannergooding: The implicit list may change tomorrow with https://github.com/dotnet/corefx/issues/20256. If we have a corresponding section for .NET Core, it might be worth updating at that time.

proposals/NativeInt.md
```diff
@@ -0,0 +1,309 @@
...
+There are no `MinValue` or `MaxValue` fields on `nint` or `nuint` because, other than `nuint.MinValue`, those values cannot be emitted as constants.
+
+Constant folding is supported for all unary operators { `+`, `-`, `~` } and binary operators { `+`, `-`, `*`, `/`, `%`, `==`, `!=`, `<`, `<=`, `>`, `>=`, `&`, `|`, `<<`, `>>` }.
+Constant folding operations are evaluated with `Int32` and `UInt32` operands rather than native ints for consistent behavior regardless of compiler platform.
+
+### Conversions
+There are identity conversions between native ints and the underlying types in both directions.
+There are identity conversions between compound types that differ by native ints and underlying types only: arrays, `Nullable<>`, constructed types, and tuples.
+
+The tables below conver the conversions between special types.
```
jkotas:
```suggestion
The tables below cover the conversions between special types.
```

proposals/NativeInt.md
cston: Constant expressions that exceed 32 bits will be reported as errors.

---
In reply to: [330750070](https://github.com/dotnet/csharplang/pull/2833#discussion_r330750070) [](ancestors = 330750070)

proposals/NativeInt.md
```diff
@@ -0,0 +1,326 @@
...
+- otherwise invalid.
+
+### Operators
+
+The following operators are provided by the compiler.
+These operators are considered during overload resolution based on normal rules for implicit conversions of arguments.
+
+(The IL for each operator includes the variants for `unchecked` and `checked` contexts if different.)
+
+| Unary | Operator Signature | IL |
```
gafter: Per the shape of the language specification, this should be something like
```csharp
nint operator +(nint value); // nop
nuint operator +(nuint value); // nop
nint operator -(nint value); // neg
nint operator ~(nint value); // not
nuint operator ~(nuint value); // not
```

proposals/NativeInt.md
```diff
@@ -0,0 +1,326 @@
...
+- otherwise invalid.
+
+Conversion from `Nullable<A>` to `Nullable<B>` is:
+- an identity conversion if there is an identity conversion from `A` to `B`;
+- an explicit nullable conversion if there is an implicit or explicit numeric conversion from `A` to `B`;
+- otherwise invalid.
+
+### Operators
+
+The following operators are provided by the compiler.
```
gafter: >The following operators are provided by the compiler [](start = 0, length = 52)

The predefined operators are as follows.

proposals/NativeInt.md
```diff
@@ -0,0 +1,326 @@
...
+| `nuint` | `short` | ExplicitNumeric | `conv.i2` / `conv.ovf.i2` |
+| `nuint` | `ushort` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2` |
+| `nuint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4` |
+| `nuint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4` |
+| `nuint` | `long` | ExplicitNumeric | `conv.i8` / `conv.ovf.i8` |
+| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.ovf.u8` |
+| `nuint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
+| `nuint` | `float` | ImplicitNumeric | `conv.r.un conv.r4` |
+| `nuint` | `double` | ImplicitNumeric | `conv.r.un conv.r8` |
+| `nuint` | `decimal` | ExplicitNumeric | `ulong decimal.op_Explicit(decimal) UIntPtr UIntPtr.op_Explicit(ulong)` |
```
gafter: >ExplicitNumeric [](start = 24, length = 15)

Since the conversion from `ulong` to `decimal` is an implicit conversion, this should be too.

proposals/NativeInt.md
```diff
@@ -0,0 +1,326 @@
...
+| `nint` | `short` | ExplicitNumeric | `conv.i2` / `conv.ovf.i2` |
+| `nint` | `ushort` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2` |
+| `nint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4` |
+| `nint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4` |
+| `nint` | `long` | ImplicitNumeric | `conv.i8` / `conv.ovf.i8` |
+| `nint` | `ulong` | ExplicitNumeric | `conv.i8` / `conv.ovf.i8` |
+| `nint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2` |
+| `nint` | `float` | ImplicitNumeric | `conv.r4` |
+| `nint` | `double` | ImplicitNumeric | `conv.r8` |
+| `nint` | `decimal` | ExplicitNumeric | `long decimal.op_Explicit(decimal) IntPtr IntPtr.op_Explicit(long)` |
```
gafter: >ExplicitNumeric [](start = 23, length = 15)

Since the conversion from `long` to `decimal` is an implicit conversion, this should be too.

proposals/NativeInt.md
```diff
@@ -0,0 +1,345 @@
...
+
+`nuint` constants are in the range [ `uint.MinValue`, `uint.MaxValue` ].
+
+There are no `MinValue` or `MaxValue` fields on `nint` or `nuint` because, other than `nuint.MinValue`, those values cannot be emitted as constants.
+
+Constant folding is supported for all unary operators { `+`, `-`, `~` } and binary operators { `+`, `-`, `*`, `/`, `%`, `==`, `!=`, `<`, `<=`, `>`, `>=`, `&`, `|`, `^`, `<<`, `>>` }.
+Constant folding operations are evaluated with `Int32` and `UInt32` operands rather than native ints for consistent behavior regardless of compiler platform.
+
+### Conversions
+There are identity conversions between native ints and the underlying types in both directions.
```
gafter: >between native ints and the underlying types [](start = 31, length = 44)

Please make these explicit.  Specifically, please say that "There is an identity conversion between nint and IntPtr, and between nuint and UIntPtr."

proposals/NativeInt.md
```diff
@@ -0,0 +1,345 @@
...
+
+`nint` and `nuint` expressions used as array indices are emitted without conversion.
+```C#
+static object GetItem(object[] array, nint index)
+{
+    return array[index]; // ok
+}
+```
+
+`nint` and `nuint` can be used as an `enum` base type.
```
gafter: >enum [](start = 38, length = 4)

What is the reason for this change?  How it is helpful to users?

crowo:
Also there is need to specify FieldOffset in explicit StructLayout classes/structs which contain pointer sized fields. So we need compiler support  to specify word size's in attributes and/or constants. also something like const void* invalidHandle = (void*)-1; should be valid as IntPtr isnt treated like primitive.

proposals/NativeInt.md
Zenexer: > 
> 
> Constant expressions that exceed 32 bits will be reported as errors.
> 
> In reply to: [330750070](https://github.com/dotnet/csharplang/pull/2833#discussion_r330750070) [](ancestors = 330750070)

I don't see how unary `~` could ever be valid given that constraint.

proposals/NativeInt.md
Zenexer: From a future-proofing perspective, I don't think that makes sense.  `long` and `decimal` both have fixed sizes, and a `long` can always be represented as a `decimal` without any loss in precision--that's a strong guarantee.  That isn't the case for `nint`/`nuint` -> `decimal`.  If .NET Core ever lands on a platform with 128-bit native ints, that decision is going to be hard to correct.

proposals/NativeInt.md
Zenexer: https://github.com/dotnet/csharplang/pull/2833/files#r386614583

jcouv:
@cston Can this be merged, so that I can link to it?

proposals/NativeInt.md
cston: Good catch, thanks.

---
In reply to: [386612190](https://github.com/dotnet/csharplang/pull/2833#discussion_r386612190) [](ancestors = 386612190)

proposals/NativeInt.md
cston: The corresponding `nint operator +(nint, nint)` operator is included in the Operators section above.

---
In reply to: [330888788](https://github.com/dotnet/csharplang/pull/2833#discussion_r330888788) [](ancestors = 330888788)

proposals/NativeInt.md
cston: The decision at [https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-10-23.md](https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-10-23.md) was to allow as `enum` underlying type if it is "not too much implementation work".

---
In reply to: [365039199](https://github.com/dotnet/csharplang/pull/2833#discussion_r365039199) [](ancestors = 365039199)

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+`nint` constants are in the range [ `int.MinValue`, `int.MaxValue` ].
+
+`nuint` constants are in the range [ `uint.MinValue`, `uint.MaxValue` ].
+
+There are no `MinValue` or `MaxValue` fields on `nint` or `nuint` because, other than `nuint.MinValue`, those values cannot be emitted as constants.
+
+Constant folding is supported for all unary operators { `+`, `-` } and binary operators { `+`, `-`, `*`, `/`, `%`, `==`, `!=`, `<`, `<=`, `>`, `>=`, `&`, `|`, `^`, `<<`, `>>` }.
+Constant folding operations are evaluated with `Int32` and `UInt32` operands rather than native ints for consistent behavior regardless of compiler platform.
+If the operation results in a constant value in 32-bits, constant folding is performed at compile-time.
+Otherwise the operation is executed at runtime and not considered a constant. (The unary operator `~` in particular cannot be used in constant expressions.)
```
gafter, line 36: >The unary operator `~` in particular cannot be used in constant expressions [](start = 79, length = 75)

I think this is not correct.  `~` can be used in constant expressions for all `nint` values (e.g. `~(nint)0)` is `(nint)-1`; in general `~(nint)(int)N` is `(nint)(~(int)N)`). This sentence is correct for `nuint`, as the result value depends on `sizeof(nuint)`.

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+
+Conversion from `Nullable<A>` to `Nullable<B>` is:
+- an identity conversion if there is an identity conversion from `A` to `B`;
+- an explicit nullable conversion if there is an implicit or explicit numeric conversion from `A` to `B`;
+- otherwise invalid.
+
+### Operators
+
+The predefined operators are as follows.
+These operators are considered during overload resolution based on normal rules for implicit conversions of arguments.
```
gafter, line 134: >based on normal rules for implicit [](start = 58, length = 34)

These operators are only considered if one of the operands is of type `nint` or `nuint`.

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+For some binary operators, the IL operators support additional operand types
+(see [ECMA-335](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf) III.1.5 Operand type table).
+But the set of operand types supported by C# is limited for simplicity and for consistency with existing operators in the language.
+
+Lifted versions of the operators, where the arguments and return types are `nint?` and `nuint?`, are supported.
+
+Compound assignment operations `x op= y` where `x` or `y` are native ints follow the same rules as with other primitive types with pre-defined operators.
+Specifically the expression is bound as `x = (T)(x op y)` where `T` is the type of `x` and where `x` is only evaluated once.
+
+The shift operators should mask the number of bits to shift appropriately
```
gafter, line 190: >appropriately [](start = 60, length = 13)

You should clarify that "appropriately" means to five bits if `sizeof(nint)` is 4, and to six bits of `sizeof(nint)` is 5. We cannot get that from the quoted section of the spec.

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Design
+[design]: #design
+
+The identifiers `nint` and `nuint` are new contextual keywords that represent native signed and unsigned integer types.
+The identifiers are only treated as keywords when name lookup does not find a viable result at that program location.
+```C#
+nint x = 3;
+var y = nameof(nuint);
+var z = nint.Zero;
```
gafter, line 18: >Zero [](start = 13, length = 4)

Disagrees with line 214, which says that `Zero` is not a member.

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+
+The remaining members of `System.IntPtr` and `System.UIntPtr` _are implicitly included_ in `nint` and `nuint`. For .NET Framework 4.7.2:
+```C#
+public override bool Equals(object obj);
+public override int GetHashCode();
+public override string ToString();
+public string ToString(string format);
+```
+
+Interfaces implemented by `System.IntPtr` and `System.UIntPtr` _are implicitly included_ in `nint` and `nuint`.
```
gafter, line 231: >included_ [](start = 79, length = 9)

...with occurrences of the underlying type replaced by the corresponding native type.

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+
+`nint` and `System.IntPtr`, and `nuint` and `System.UIntPtr`, are considered equivalent for overriding, hiding, and implementing.
+
+Overloads cannot differ by `nint` and `System.IntPtr`, and `nuint` and `System.UIntPtr`, alone.
+Overrides and implementations may differ by `nint` and `System.IntPtr`, or `nuint` and `System.UIntPtr`, alone.
+Methods hide other methods that differ by `nint` and `System.IntPtr`, or `nuint` and `System.UIntPtr`, alone.
+
+### Miscellaneous
+
+`nint` and `nuint` expressions used as array indices are emitted without conversion.
```
gafter, line 245: >emitted without conversion [](start = 57, length = 26)

It isn't clear what "emitted without conversion" means. The specification for array indexing requires a conversion, and the array indexing IL instructions have specific requirements for the type of the index.

proposals/native-integers.md
```diff
@@ -0,0 +1,342 @@
...
+nint x = 3;
+var y = nameof(nuint);
+var z = nint.Zero;
+```
+
+The types `nint` and `nuint` are represented by the underlying types `System.IntPtr` and `System.UIntPtr` with compiler surfacing additional conversions and operations for those types as native ints.
+
+### Constants
+
+There is no direct syntax for native int literals. Explicit casts of other integral constant values can be used instead: `(nint)42`.
```
gafter, line 25: >direct syntax for native int literals [](start = 12, length = 37)

Do you intend that constant expressions can be of these new types?  If so, you need to explicitly say so, as the spec for "constant expressions" excludes these types from the list of types for which constants exist.

gafter:
Finished reviewing (Commit 17)

proposals/native-integers.md
tannergooding: > array indexing IL instructions have specific requirements for the type of the index

IL instructions like `ldelem` and `ldelema` take either a `native int` or an `int32`

gafter:
Open question: What is the common type between a value of type nint and a value of type IntPtr?  The common type between G<nint> and G<IntPtr>?  When both are bounds in type inference, which wins?

PR #2859, opened 2019-10-03
Delete unnecessary descriptions

patricka2:


333fred:
This description isn't unnecessary, this is the official supported scenario. We'll leave this as is.

patricka2:
> This description isn't unnecessary, this is the official supported scenario. We'll leave this as is.

Yes, but isn't necessary in this page. this page is only include C# and Visual Studio version that shipped with, this unnecessary  description deprecate Uniformity with other C# version in this page and its hurts OCD people like me

patricka2:
I made new pull request (#2862 ) that fix grammar please merge new one @333fred

Issue #275, opened 2017-03-15
Proposal: Use static keyword for lambdas to disallow capturing locals and parameters (VS 16.8, .NET 5)


mattwar:
Allow `Func<int, string> = static i => i.ToString();`.

- [x] Proposal added ([here](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-9.0/static-lambdas.md))
- [x] Discussed in LDM 2019-10-21
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

----

To avoid accidentally capturing any local state when supplying lambda functions for method arguments, prefix the lambda declaration with the static keyword.  This makes the lambda function just like a static method, no capturing locals and no access to this or base.

```cs
int y = 10;
someMethod(x => x + y); // captures 'y', causing unintended allocation.
```

with this proposal you could the static keyword to make this an error.

```cs
int y = 10;
someMethod(static x => x + y); // error!
```

```cs
const int y = 10;
someMethod(static x => x + y); // okay :-)
```

LDM history:
- 10/21/2019

mattwar:
Related issues/proposals:
Lambda Capture lists: https://github.com/dotnet/roslyn/issues/117
Non-Capture Lambdas with ==> syntax: https://github.com/dotnet/roslyn/issues/11620

MgSam:
I like this better than capture lists, which I think are way too heavyweight.

DavidArno:
I don't see the point to this, but it's got a bunch of up-votes. So rather than just down-vote, I'll assume I'm ignorant of the issue. Why is this (and/or capture lists) needed? Is accidental closure capture a problem for people?

eyalsk:
@MgSam Too heavyweight for whom? people want to have control over _what_ gets captured and sometimes _how_ it gets captured and this locks the language into a specific behaviour which seems too limited.

People can always have this as a syntactic sugar to capture lists but doing this first would be a shame.

_I have a **bad feeling** about capture lists, they won't make it._ 😆

jnm2:
I'm in the same place as @DavidArno. If there is a special case where you need to disrupt the syntax to be absolutely sure that you aren't capturing anything, is it so bad to refactor it to a static method?

yaakov-h:
Does this need to be a language feature or is this something you could reasonably achieve with an Analyzer?

alrz:
[...]

PR #2899, opened 2019-10-21
Proposed speclet for static lambdas

jcouv:
Championed issue https://github.com/dotnet/csharplang/issues/275
Relates to PR https://github.com/dotnet/roslyn/pull/39121

### DIFF
--- a/proposals/static-lambdas.md
+++ b/proposals/static-lambdas.md
@@ -1,0 +1,32 @@
+# Static lambdas
+
+## Summary
+
+Support lambdas that disallow capturing state from the enclosing scope.
+
+## Motivation
+
+Avoid unintentionally capturing state from the enclosing context.
+
+## Detailed design
+
+A lambdas with `static` cannot capture state from the enclosing scope.
+As a result, locals, parameters, and `this` from the enclosing scope are not available within a `static` lambda.
+
+A `static` lambda cannot reference instance members from an implicit or explicit `this` or `base` reference.
+
+A `static` lambda may reference `static` members from the enclosing scope.
+
+A `static` lambda may reference `constant` definitions from the enclosing scope.
+
+`nameof()` in a `static` lambda may reference locals, parameters, or `this` or `base` from the enclosing scope.
+
+Accessibility rules for `private` members in the enclosing scope are the same for `static` and non-`static` lambdas.
+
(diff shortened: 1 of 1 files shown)


jcouv:
Tagging @cston @cyrusnajmabadi @gafter for review.

proposals/static-lambdas.md
```diff
@@ -0,0 +1,32 @@
...
+A `static` lambda may reference `static` members from the enclosing scope.
+
+A `static` lambda may reference `constant` definitions from the enclosing scope.
+
+`nameof()` in a `static` lambda may reference locals, parameters, or `this` or `base` from the enclosing scope.
+
+Accessibility rules for `private` members in the enclosing scope are the same for `static` and non-`static` lambdas.
+
+No guarantee is made as to whether a `static` lambda definition is emitted as a `static` method in metadata. This is left up to the compiler implementation to optimize.
+
```
CyrusNajmabadi: i agree with this.  Should we consider something additional like:

> Implementations are recommended to emit a static lambda with a singleton allocation for improved performance

?

jcouv: I'm fine to leave that as an implementation detail

alrz: wouldn't that be useful with function pointers?
```cs
unsafe class Instance {
    void Use() {
        delgate*<Instance, string> f = &static i => i.ToString();
        f(this);
    }
}
```
In the proposal, the recommended workaround for instance methods is to use static local functions. I think a major advantage of using static lambdas over static local functions here is type inference and avoiding typing out the whole signature so it could be of help in this scenario as well.

proposals/static-lambdas.md
```diff
@@ -0,0 +1,32 @@
...
+## Motivation
+
+Avoid unintentionally capturing state from the enclosing context.
+
+## Detailed design
+
+A lambdas with `static` cannot capture state from the enclosing scope.
+As a result, locals, parameters, and `this` from the enclosing scope are not available within a `static` lambda.
+
+A `static` lambda cannot reference instance members from an implicit or explicit `this` or `base` reference.
```
gafter: I suspect you intend that such a reference is a compile-time error.  Please say that, and identify the sections of the ECMA specification that require modifications.

Issue #111, opened 2017-02-15
Discard for lambda parameters (_, _) =>


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

See
- https://github.com/dotnet/roslyn/issues/16255

DavidArno:
Please, please, please could this apply to methods too? Sometimes, the parameters must be there, but just aren't used. By way of example, [this Unit type](https://github.com/DavidArno/SuccincT/blob/master/SuccincT/Functional/Unit.cs) has to have both R# and `CA1801` suppressions for the operator overloads:
````cs
public static bool operator ==(Unit u1, Unit u2) => true;
public static bool operator !=(Unit u1, Unit u2) => false;
````
Being able to write those with discards to reassure those suppressions that the parameters must be there, but aren't used, would clean up such code nicely:
````cs
public static bool operator ==(Unit _, Unit _) => true;
public static bool operator !=(Unit _, Unit _) => false;
````

**Comment by jnm2** [comment]
@DavidArno I believe the .NET Frameowkr design guidelines have always been to use the names `left` and `right` specifically.

**Comment by DavidArno** [comment]
@jnm2,

Oh, I didn't know that. I'll give that a try to see if it makes the compiler happier. Would still like to be able to use discards though.

**Comment by jnm2** [comment]
@DavidArno It'll only make the compiler happier if you have code quality being analyzed.

**Comment by acple** [comment]
this proposal also permit in nested lambda expression `_ => _ => true` ?
```csharp
Enumerable.Range(0, 3)
    .SelectMany(_ => Enumerable.Range(0, 3)
        .SelectMany(_ => Enumerable.Range(0, 3))); // i don't need parameters in nested
```

gafter:
@acple Possibly not, unless we also do something like https://github.com/dotnet/roslyn/issues/15793 at the same time.

acple:
@gafter Thank you. in my opinion, lambda parameter shadowing is a bit dangerous but, discarded parameters are safe to overcover. so that should support only with discarded parameters. is this shadowing issue? or discarding issue?

gafter:
@acple This is a shadowing issue. When there is only one lambda parameter named `_`, it is a normal identifier that declares a parameter. It must be so for backward compatibility.

acple:
[...]

PR #2901, opened 2019-10-22
Create speclet for lambda discard parameters

jcouv:
Relates to championed issue https://github.com/dotnet/csharplang/issues/111


proposals/lambda-discard-parameters.md
```diff
@@ -0,0 +1,21 @@
...
+- `(int _, int _) => 0`
+- `delegate(int _, int _) { return 0; }`
+
+## Motivation
+
+Unused lambda parameters do not need to be named. The intent of discards is clear, i.e. they are unused/discarded.
+
+## Detailed design
+
+In a lambda or delegate declaration with more than one parameter named `_`, those parameters are discard parameters.
```
gafter: This specification does not describe the scoping implications.  Is there any hiding?

gafter: >delegate declaration [](start = 15, length = 20)

Did you really mean "delegate declaration", that is, a declaration of a delegate type?

proposals/lambda-discard-parameters.md
```diff
@@ -0,0 +1,21 @@
...
+
+Discard parameters are not in scope in the body of the lambda or delegate. So they cannot be accessed.
```
gafter: >parameters are not in scope [](start = 8, length = 27)

Can you pose this as a set of changes to the language specification?

jcouv: I took a stab at this:

In the section on [method parameters](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#method-parameters), I would say that discard parameters do not add any names in scope.
In the section on [corresponding parameters](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#corresponding-parameters), I would say that named arguments do not correspond to discard parameters.

proposals/discard-parameters.md
```diff
@@ -0,0 +1,26 @@
...
+- indexers, constructors
+- delegate types: `delegate void MyDelegate(int _, int _);`
+
+## Motivation
+
+Unused parameters do not need to be named. The intent of discards is clear, i.e. they are unused/discarded.
+
+## Detailed design
+
+In a parameter list with more than one parameter named `_`, such parameters are discard parameters.
```
gafter: > more than one parameter named `_` [](start = 24, length = 34)

The spec currently makes this an error. What spec changes are proposed to make it not be an error.

proposals/discard-parameters.md
```diff
@@ -0,0 +1,26 @@
...
+Note: if a single parameter is named `_` then it is a regular parameter for backwards compatibility reasons.
+
+Discard parameters do not introduce any names to any scopes.
```
gafter: >do not introduce any names to any scopes [](start = 19, length = 40)

What name lookup changes are proposed to implement this intent in the specification?

proposals/discard-parameters.md
jcouv: "It is an error for two members of a method declaration space to have the same name."  ([method parameters](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#method-parameters)).

I'll update speclet to say that discard parameters don't have a name (so the above rule doesn't apply).

proposals/discard-parameters.md
jcouv: Which section of the spec do you refer to for "name lookup"?
I think I'm proposing that we don't change name lookup, but rather we change what names are added to what scopes.
"Names are introduced into this declaration space by the type parameter list and the formal parameter list of the method and by local variable declarations in the block of the method." ([method parameters](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#method-parameters)) would be affected by the rule that discard parameters don't introduce a name in any scope.

proposals/lambda-discard-parameters.md
```diff
@@ -0,0 +1,20 @@
...
+## Motivation
+
+Unused parameters do not need to be named. The intent of discards is clear, i.e. they are unused/discarded.
+
+## Detailed design
+
+In the parameter list of a lambda or anonymous method with more than one parameter named `_`, such parameters are discard parameters.
+Note: if a single parameter is named `_` then it is a regular parameter for backwards compatibility reasons.
+
+Discard parameters do not introduce any names to any scopes.
```
gafter: >introduce any names to any scopes [](start = 26, length = 33)

Please identify the intended change to 8.7 **Scopes**.

You will need to modify the name lookup rules to cause this intent to be reified in the specification.  For example, we have the following text in ECMA's **12.7.3 Simple Names** (giving the meaning of a simple name as an expression):

> - If `K` is zero and the *simple-name* appears within a *block* and if the *block*’s (or an enclosing *block*’s) local variable declaration space (§8.3) contains a local variable, parameter or constant with name `I`, then the *simple-name* refers to that local variable, parameter or constant and is classified as a variable or value.

Note that the discard parameter is in the declaration space by 8.3, and is a parameter named `_` (see line 16 above), so without a change to the quoted paragraph it is indeed found during name lookup.  Please identify the intended changes to the specification.

jcouv:
@gafter I'll go ahead and merge. Feel free to provide further comments. I'll address separately. Thanks

PR #2922, opened 2019-11-01
Update introduction.md

Youssef1313:


spec/introduction.md
```diff
@@ -21,17 +21,21 @@ using System;
...
 
-C# source files typically have the file extension `.cs`. Assuming that the "Hello, World" program is stored in the file `hello.cs`, the program can be compiled with the Microsoft C# compiler using the command line
+C# source files typically have the file extension *.cs*. Assuming that the "Hello, World" program is stored in the file *hello.cs*, the program can be compiled with the Microsoft C# compiler using the command line:
+
 ```console
 csc hello.cs
 ```
-which produces an executable assembly named `hello.exe`. The output produced by this application when it is run is
+
+which produces an executable assembly named *hello.exe*. The output produced by this application when it is run is:
```
theunrepentantgeek: Switching from `code` style to *italic* for all these file names (and parts of file names) seems like a significant backward step to me.

spec/introduction.md
```diff
@@ -109,18 +119,21 @@ class Test
     }
 }
 ```
-If the program is stored in the file `test.cs`, when `test.cs` is compiled, the `acme.dll` assembly can be referenced using the compiler's `/r` option:
+
+If the program is stored in the file *test.cs*, when *test.cs* is compiled, the *acme.dll* assembly can be referenced using the compiler's `-r` option:
 
 ```console
-csc /r:acme.dll test.cs
+csc -r:acme.dll test.cs
```
theunrepentantgeek: Since this is a demo, perhaps use the full form of the flag for clarity: `-reference:acme.dll`
(Having `-r` is pretty cryptic.)

spec/introduction.md
```diff
@@ -855,16 +929,18 @@ using System;
...
 class Test
 {
-    static void Main() {
+    static void Main() 
+    {
         Entity.SetNextSerialNo(1000);
-        Entity e1 = new Entity();
-        Entity e2 = new Entity();
+        var e1 = new Entity();
+        var e2 = new Entity();
```
theunrepentantgeek: I'm of two minds on this - while **var** is idomatic, many teams forbid its use. Also, there's a significant fraction of new C# developers who associate it with *untyped* variables (similar to C#'s **dynamic**, Visual Basics **variant**, or practically anything in JavaScript). Since this doc is an introduction to the language, being explicit with the variable type may reduce confusion.

spec/introduction.md
```diff
@@ -1046,21 +1142,27 @@ public class List<T> {
...
         get { return count; }
     }
-    public int Capacity {
-        get {
+    public int Capacity
+    {
+        get
+        {
             return items.Length;
         }
```
theunrepentantgeek: Inconsistent formatting here - the **get** for `Count`, above, is entirely on a single line; this one should be as well.

spec/introduction.md
```diff
@@ -1069,34 +1171,42 @@ public class List<T> {
...
 
     // Indexer...
-    public T this[int index] {
-        get {
+    public T this[int index]
+    {
+        get
+        {
             return items[index];
         }
```
theunrepentantgeek: This should also be a single line:

``` csharp
        get { return items[index]; }
```

spec/introduction.md
HaloFour: Sounds like a good reason to demonstrate `var` and its type-safe behavior early on.  If the introduction has to cater to all of the different opinions regarding code style it's going to only end up incredibly long and incredibly confusing.

Youssef1313: @theunrepentantgeek, This change follows the [coding style of coreFX](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md).

spec/introduction.md
Youssef1313: @theunrepentantgeek, that's the current style in the dotnet/docs

theunrepentantgeek: Fair enough.

spec/introduction.md
theunrepentantgeek: Fair call. It may then be worth explicitly pointing out the use of **var** in the paragraph below.

gafter:
I don't think we want to take stylistic changes before we have a chance to integrate with the ECMA specification.

Youssef1313:
Closing for now, will probably do a pass on dotnet/csharpstandard in future.

Issue #2917, opened 2019-10-29
Function Pointer Syntax


333fred:
### Current State

Currently, we have the following syntax proposal for function pointers (grammar slightly corrected from the proposal):

```antlr
pointer_type
    : ...
    | funcptr_type
    ;

funcptr_type = 
    'func' '*' calling_convention? return_type '(' (type (',' type)* )? ')' |
    '(' funcptr_type ')' ;

calling_convention = 
    'cdecl' |
    'managed' |
    'stdcall' | 
    'thiscall' |
    'unmanaged' ;
```

This grammar has a few different problems, mainly around parsing:

1. `func` can be a variable bound to an existing construct, and this is impossible to figure out at parsing time. This means that we would potentially need to use semantic analysis to inform how the tree should parse. `var` has this problem, but to a lesser extent, as we can parse just the single token and reinterpret it later. A function pointer has many tokens as part of it, and if we determine that the `func` should have been bound to a local later in the process, we currently do not have the facilities to handle this.
1. Error recovery of specific scenarios requires an exponential number of parsing passes. Consider the following scenario:
   ```cs
   func* func* invalidcallingconvention (int, int)(string)()
   ```
   The proper interpretation of this can be clarified with these parens:
   ```cs
   func* (func* invalidcallingconvention (int, int)(string))()
   ```
   However, in order to interpret it in this manner, we have to parse the inner function pointer a minimum of twice. The first time we parse, we interpret as `func* invalidcallingconvention(int, int)`, and then later notice that there's a trailing `(`. We then have to retry parsing of the inner function pointer, this time being greedy and assuming that `invalidcallingconvention` is in fact that, and not a return type itself. They're somewhat unlikely scenarios, but handling this correctly without blowing up the compiler is tricky and each successive case requires an exponentially larger number of parse attempts.

### New Syntax Proposal

In order to handle these issues, I propose a new syntax for defining a function pointer, inspired by the `Func<T>` type we all know and love today:

```antlr
pointer_type
    : ...
    | funcptr_type
    ;

funcptr_type
    : calling_convention '*' '<' (type ',')* return_type '>'
    ;

calling_convention
    : 'func'
    | 'cdecl'
    | 'managed'
    | 'stdcall'
    | 'thiscall'
    | 'unmanaged'
    ;
```

[...]

PR #2923, opened 2019-11-01
Update function pointer proposal with LDM changes

333fred:


proposals/function-pointers.md
```diff
@@ -1,60 +1,61 @@
...
+Conversions between `delegate*` types is done based on their signature including the calling convention.
 
 ``` csharp
 unsafe class Example {
     void Conversions() {
-        func* int(int, int) p1 = ...;
-        func* managed int(int, int) p2 = ...;
-        func* cdecl int(int, int) p3 = ...;
+        delegate*<int, int, int> p1 = ...;
+        delegate* managed<int, int, int> p2 = ...;
```
Perksey: In my opinion, the calling convention appearing before the `delegate*` makes more sense, both gramatically and visually, as it's saying it's a `managed delegate pointer` rather than a `delegate pointer managed`

alrz: I think the reason for this is mostly easy disambiguation. See https://github.com/dotnet/csharplang/issues/2917 for discussion.

tannergooding: The latter also makes sense if you read it as `delegate pointer [of type] managed`

Perksey: My main issue is the type is usually the last thing that appears in a declaration, and in my opinion `delegate* managed<int>` implies you can just have a `managed<int>`, which you can't. The calling convention can be thought of as an attribute - attributes don't come after the type itself :P 

Same goes if you think about it like the `in`, `out`, and `ref` modifiers. It's unusual for the language to put the modifier after the type specifier, given that everywhere else the modifier is specified before the type.

333fred:
@gafter @MadsTorgersen @stephentoub @jkotas, did you have any comments?

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
+delegate* managed<string, int>;
+delegate*<delegate* managed<string, int>, delegate*<string, int>>;
 ```
 
 ### Allow address-of to target methods
 
-Method groups will now be allowed as arguments to an address-of expression. The type of such an 
-expression will be a `func*` which has the equivalent signature of the target method and a managed 
+Method groups will now be allowed as arguments to an address-of expression. The type of such an
+expression will be a `delegate*` which has the equivalent signature of the target method and a managed
```
gafter: >the target method [](start = 71, length = 17)

I don't know what "the target method" means when we're talking about a method group.

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
 
 ### Better function member
+
 The better function member specification will be changed to include the following line:
 
-> A `func*` is more specific than `void*`
+> A `delegate*` is more specific than `void*`
 
-This means that it is possible to overload on `void*` and a `func*` and still sensibly use the address-of operator.
+This means that it is possible to overload on `void*` and a `delegate*` and still sensibly use the address-of operator.
```
gafter: >sensibly [](start = 82, length = 8)

I don't see how.  You haven't defined a conversion from an "address of a method group" to `void*`.

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
-- Prevent method group conversions to `delegate` when the method is tagged with `NativeCallable`. 
+- Prevent method group conversions to `delegate` when the method is tagged with `NativeCallable`.
 
 This is not necessary to support `NativeCallable` though. The compiler can support the `NativeCallable` attribute as is
-using the existing syntax. The program would simply need to cast to `void*` before casting to the correct `func*` 
+using the existing syntax. The program would simply need to cast to `void*` before casting to the correct `delegate*`
 signature. That would be no worse than the support today.
 
 ``` csharp
 void* v = &CloseHandle;
```
gafter: >void* v = &CloseHandle [](start = 0, length = 22)

I don't see any rules that make this work.

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
+        // Error: type "delegate*<void>" not compatible with "delegate*<int>";
+        delegate*<int> ptr2 = &Util.Log;
 
         // Okay. Conversion to void* is always allowed.
         void* v = &Util.Log;
    }
 }
 ```
 
-The conversion of an address-of method group to `func*` has roughly the same process as method group to `delegate`
```
gafter: >address-of method group [](start = 21, length = 23)

I think you may need to introduce and define the concept of an "address-of method group" if you want to say things about it.

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
+        delegate*<int> ptr2 = &Util.Log;
 
         // Okay. Conversion to void* is always allowed.
         void* v = &Util.Log;
    }
 }
 ```
 
-The conversion of an address-of method group to `func*` has roughly the same process as method group to `delegate`  
+The conversion of an address-of method group to `delegate*` has roughly the same process as method group to `delegate`
```
gafter: >address-of method group [](start = 21, length = 23)

I think you may need to introduce and define the concept of an "address-of method group" if you want to say things about it.

gafter: >The conversion [](start = 0, length = 14)

It feels like you're defining a new conversion.  If so, can you please say that?

proposals/function-pointers.md
```diff
@@ -253,83 +253,87 @@ conventions is very rich.
...
+Additionally any other type which is defined with the name `System.Func<T>` in an assembly that is not mscorlib
 must be different than the version defined in mscorlib.
 
-One option that was explored was emitting such a pointer as `mod_req(Func<int>) void*`. This doesn't 
+One option that was explored was emitting such a pointer as `mod_req(Func<int>) void*`. This doesn't
 work though as a `mod_req` cannot bind to a `TypeSpec` and hence cannot target generic instantiations.
 
 ### Named function pointers
+
 The function pointer syntax can be cumbersome, particularly in complex cases like nested function pointers. Rather than
```
gafter: >The [](start = 0, length = 3)

Please say something to make it clear that this is an alternative to, and not part of, the spec.

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
+    | funcptr_type
+    ;
+
+funcptr_type
+    : 'delegate' '*' calling_convention? '<' (funcptr_parameter_modifier? type ',')* funcptr_return_modifier? return_type '>'
+    ;
+
+calling_convention
+    : 'cdecl'
+    | 'managed'
```
jkotas: In the original proposal, naming the `managed` calling convention was required for disambiguation in some case. Are there still ambiguous cases with this proposal where it is needed? 

If there are no such cases, should `managed` be dropped from here? It is better to only have one way to write something. Otherwise, people need to create coding convention to codify which one of the multiple ways is the correct one for the given project.

333fred: It's no longer required, but I'd prefer to keep it in. Just like we have defaults for accessibility modifiers and also allow you to specify them, I think it's good to allow to specify `managed` even if it's the default.

proposals/function-pointers.md
333fred: This is bigger section titled `Considerations`, with a bunch of different future proposals.

---
In reply to: [344939001](https://github.com/dotnet/csharplang/pull/2923#discussion_r344939001) [](ancestors = 344939001)

333fred:
@gafter for your comments, I agree that we need to nail down the concepts of conversion for method groups in function pointers, but I think that's out of scope for this PR. I'm just updating the syntax here, so I'm going to resolve your comments and come back to them at a later time.

proposals/function-pointers.md
333fred: OTOH, for existing uses of `delegate` today `delegate` is the keyword, followed by the delegate definition. This is the same, where you have `delegate*` followed by the definition. We don't think that this is inconsistent with C# today.

alrz: @333fred except that it's for an expression or declaration, the syntax for pointers is the whole type followed by an asterisk token.

proposals/function-pointers.md
```diff
@@ -63,174 +64,173 @@ unsafe class Example {
...
-- Can explicitly convert from `void*` to `func*`.
+- Can implicitly convert `delegate*` to `void*`.
+- Can explicitly convert from `void*` to `delegate*`.
 
 Restrictions:
-- Custom attributes cannot be applied to a `func*` or any of its elements.
-- A `func*` parameter cannot be marked as `params`
-- A `func*` type has all of the restrictions of a normal pointer type.
+
+- Custom attributes cannot be applied to a `delegate*` or any of its elements.
```
tannergooding: Custom attributes can't be applied today but we will likely want the ability to expose them in the future (to support `GCSuppressTransition`, other calling convention designations, etc).

Presumably, the compiler is currently planning on emitting these types directly as `FNPTR` type signatures. However, the runtime doesn't currently have any way to attach attributes to such type signatures.

Has this been taken into consideration since that would make it a breaking change to go from the `inline function pointer` syntax (the only thing available today) to change to some syntax that allows attributes?

CC. @333fred, @gafter, @jkotas

333fred:
Filed https://github.com/dotnet/roslyn/issues/39865 to track the open questions raised in this pr.

proposals/function-pointers.md
jaredpar: > Custom attributes can't be applied today but we will likely want the ability to expose them in the future 

Is there a design stating how such support would be implemented in the runtime?

> Has this been taken into consideration since that would make it a breaking change to go from the inline function pointer syntax (the only thing available today) to change to some syntax that allows attributes?

No it hasn't because the underlying runtime support for this does not exist and would likely require changes to `calli` or the method in which the signature is provided to `calli`. Can't really take it into consideration if we don't know how the feature would be implemented in the runtime.

tannergooding: Asked someone from the runtime team to comment on the relevant issue: <https://github.com/dotnet/roslyn/issues/39865#issuecomment-596800197>

PR #2967, opened 2019-11-15
Update DIM proposal to forbid class declared within variant interface.

gafter:


proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -40,6 +42,15 @@ Static and private methods permit useful refactoring and organization of code us
 
 A method override in an interface must use the explicit interface implementation syntax.
 
+It is an error to declare a class type, struct type, or enum type within the scope of a type parameter that was declared with a *variance_annotation*.  For example, the declaration of `C` below is an error.
```
AlekseyTs: >It is an error to declare a class type, struct type, or enum type within the scope of a type parameter that was declared with a *variance_annotation*. [](start = 0, length = 150)

Is it worth mentioning that signatures of local functions and lambdas used within interfaces are always considered variant-safe?

gafter: I do not think that would affect the legality or meaning of any program, as variance safety only affects interface methods.

PR #2970, opened 2019-11-17
Fluent class definition

alanhauge:
Shortly put, it is a new way of writing code with code in the fluent coding style.
Instead of writing almost identical code multiple times, it would be better to import a code template.
 
A natural extension of this is to import whole classes into others.


YairHalberstadt:
@alanhauge

Suggesting ideas for how to improve C# is great. However proposals for new ideas are best done as issues rather than PRs.

Please could you create an issue containing the information in the PR and close this?

Thanks

PR #3011, opened 2019-12-07
Fixed links to inactive and rejected folders

richardtallent:
Noticed that the relative links here as rendered at GitHub were leading to 404s in the blob folder.


RikkiGibson:
Apologies, this PR was closed automatically as part of renaming master->main. You are welcome to recreate it if you wish.

PR #3031, opened 2019-12-16
First set of changes for tuples from feature spec and LDM notes.

BillWagner:
Fixes dotnet/docs#15046

This PR updates the C# 7.0 proposal document so that it could be used as a proposed delta to the C# standard.

To make these edits, I did the following:
1. Update the language to conform to the standard language.
1. Update the changes based on C# LDM meeting notes in this repo, and the dotnet/roslyn repo. Note that I only incorporated changes that were released in C# 7.0. Later changes to tuples (such as equality) are not included.
1. Edit and review.
1. Test against current compiler set to `<LangVersion>7.0</LangVersion>`.

This is the first PR for this larger task. I have notes for reviewers, and questions about how to continue proceeding. 

Questions / notes for reviewers:

1. Are the proposals useful in their current form? This PR updates the tuples proposal in place, assuming the proposal has served its purpose.
1. Are the notes about where changes / additions to the standard go clear? If not, does anyone have a better formatting suggestion?
1. This PR removes language about how the IDE would treat tuples, or made them informative note text. Is that the best strategy?
1. I removed any text that communicated a decision to *not* implement a behavior. Are there cases where we should do that? 
1. I'm not sure the content on overloading / overriding and natural type is in the correct location. Should it move to be in "overload resolution" in general?

Thoughts on proceeding with other proposals:

This will help keep the C# language reference section of docs concise. I will keep making progress on these (although slowly). After finishing each release, I think a second set of edits should re-organize the changes to the standard by major clause rather than feature. (Updates to **Lexical structure(**, updates to **Basic concepts**, **Types**, **Variables**, etc). Thoughts on that work?

/cc @MadsTorgersen @gafter @jcouv @agocke @jaredpar

### DIFF
--- a/proposals/csharp-7.0/tuples.md
+++ b/proposals/csharp-7.0/tuples.md
@@ -4,6 +4,8 @@ This proposal specifies the changes required to the [C# 6.0 (draft) Language spe
 
 ## Changes to [Lexical structure](../../spec/lexical-structure.md)
 
+### Literals
+
 > The grammar for [Literals](../../spec/lexical-structure.md#literals) is extended to include `tuple_literal`:
 
 ```antlr
@@ -16,7 +18,15 @@ literal
     | null_literal
     | tuple_literal // new
     ;
+```
 
+> Add the following section after [The null literal](../../spec/lexical-structure.md#The-null-literal):
+
+#### Tuple literals
+
+A tuple literal consists of two or more tuple literal elements, each of which is optionally named.
+
+```antlr
 tuple_literal
     : '(' tuple_literal_element_list ')'
     ;
@@ -31,7 +41,49 @@ tuple_literal_element
     ;
 ```
 
-Note that because it is not a constant expression, a tuple literal cannot be used as default value for an optional parameter.
+A tuple literal is *target typed* whenever possible; that is, its type is determined by the context in which it is used.
+
+**ISSUE:** If the term target-typed is specific to this context, we probably need a complete definition somewhere.
+
+\[Example:
+```csharp
+var t1 = (0, 2);              // infer tuple type from values
+var t2 = (sum: 0, count: 1);  // infer tuple type from names and values
+```
+end example\]
+
+A tuple literal has a "conversion from expression" to any tuple type having the same number of elements, as long as each of the element expressions of the tuple literal has an implicit conversion to the type of the corresponding element of the tuple type.
+
+\[Example:
+```csharp
+(string name, byte age) t = (null, 5); // OK: the expressions null and 5 convert to string and byte, respectively
+```
+end example\]
+
+In cases where a tuple literal is not part of a conversion, it acquires its *natural type*, which means a tuple type where the element types are the types of the constituent expressions, in lexical order. Since not all expressions have types, not all tuple literals have a natural type either:
+
+**ISSUE:** Is natural type specific to tuples? Need we say more about this term?
+
+\[Example:
+```csharp
+var t = ("John", 5); // OK: the type of t is (string, int)
+var t = (null, 5);   // Error: null doesn't have a type
+```
+end example\]
+
+If a tuple literal includes element names, those names become part of the natural type: 
+
+\[Example:
+```csharp
+var t = (name: "John", age: 5); // The type of t is (string name, int age)
+```
+end example\]
+
+A tuple literal is *not* a [constant expression](../../spec/expressions.md#Constant-expressions). As such, a tuple literal cannot be used as the default value for an optional parameter.
+
+**ISSUE:** Consider removing the second sentence above, as it sounds like just one example of usage prohibition (rather than a spec requirement); there likely are others. If that is true, either completely omit that sentence, or delete it from here and add an example showing that such a usage fails.
+
+For a discussion of tuple literals as tuple initializers, see [Tuple types](../../spec/types.md#Tuple-types).
 
 ## Changes to [Types](../../spec/types.md)
 
@@ -47,7 +99,19 @@ value_type
     | enum_type
     | tuple_type // new
     ;
+```
 
+## Additions to [Types](../../spec/types.md)
+
+> Add the following sections after [Nullable types](../../spec/types.md#nullable-types) (at the end of the current *Value types* section.)
+
+### Tuple types
+
+#### General
+
+A tuple type is declared using the following syntax:
+
+```antlr
 tuple_type
     : '(' tuple_type_element_list ')'
     ;
@@ -62,164 +126,113 @@ tuple_type_element
     ;
 ```
 
-## Additions to [Types](../../spec/types.md)
+A ***tuple*** is an anonymous data structure type that contains an ordered sequence of two or more ***elements***. Each element is public. Each unique, ordered combination of element types designates a distinct tuple type.
 
-> The following sections should be added after [Nullable types](../../spec/types.md#nullable-types) (at the end of the current *Value types* section.)
+An element in a tuple is accessed using the [member-access operator `.`](../../spec/expressions.md#Member-access).
 
-### Tuple types
-
-Tuple types are declared with the following syntax:
+Given the following,
 
 ```csharp
-public (int sum, int count) Tally(IEnumerable<int> values) { ... }
-
-var t = Tally(myValues);
-Console.WriteLine($"Sum: {t.sum}, count: {t.count}");
- ```
-
-The syntax `(int sum, int count)` indicates an anonymous data structure with public fields of the given names and types, referred to as *tuple*.
-
-Tuple values can be created using *tuple literals*:
-
-```csharp
-var t1 = (sum: 0, count: 1);
-var t2 = (0, 1);     // field names are optional
+(int code, string message) pair1 = (3, "hello");
+System.Console.WriteLine("first = {0}, second = {1}", pair1.code, pair1.message);
 ```
 
-Tuples cannot be created with the `new` operator. The `new` operator can be used to initialize arrays of tuples, or nullable tuples:
+the syntax `(int code, string message)` declares a tuple type having two elements, each with the given type and name.
+
+As shown, a tuple can be initialized using a [tuple literal](../../spec/lexical-structure.md#Tuple-literals).
+
+An element need not have a name.
+
+The type of a tuple is the ordered set of types of its elements along with their names. An element without a name is unnamed. If a tuple declarator contains the type of all the tuple's elements, that set of types cannot be changed or augmented based on the context in which it is used; otherwise, element type information shall be inferred from the usage context. Likewise for element names.
+
+A tuple's type can be declared explicitly. Consider the following declarations:
 
 ```csharp
-var array = new (int x, int y)[10];
-var nullable = new (int x, int y)?();
+(int, string) pair2 = (2, "Goodbye"); 
+(int code, string message) pair3 = (2, "Goodbye");
+(int code, string) pair4 = (2, "Goodbye"); 
+(int, string message) pair5 = (2, "Goodbye"); 
+(int code, string) pair6 = (2, message: "Goodbye");	// Warning: can't give a name to the second element
+(int code, string) pair7 = (newName: 2, "Goodbye");	// Warning: can't change the name of element code
 ```
 
-Tuples can be created for a known target type:
+The type of `pair2` is `(int, string)` with unknown element names. Similarly, the type of `pair3` is `(int, string)` but this time with the element names `code` and `message`, respectively. For `pair4` and `pair5`, one element is named, the other not.
+
+In the case of `pair6`, the second element is declared as being unanmed, and any attempt to provide a name in an initializing context shall be ignored. Likewise for any attempt to change an element's name, as in the case of `pair7`.
+
+A tuple's type can be wholely inferred from the context in which it is used. Consider the following declarations:
 
 ```csharp
-public (int sum, int count) Tally(IEnumerable<int> values)
-{
-    var s = 0; var c = 0;
-    foreach (var value in values) { s += value; c++; }
-    return (s, c); // target typed to (int sum, int count)
-}
+var pair10 = (1, "Hello"); 
+var pair11 = (code: 1, message: "Hello");
+var pair12 = (code: 1, "Hello");
+var pair13 = (1, message: "Hello");
 ```
 
-Specifying field names is optional. Duplicate names are disallowed.
+The type of `pair10` is inferred from the initializer's tuple literal, as `(int, string)` with unknown element names. Similarly, the type of `pair11` is inferred from the initialer's tuple literal, as `(int, string)` but this time with the element names `code` and `message`, respectively. For `pair12` and `pair13`, the element types are inferred, and one element is named, the other not.
 
+Elements within a tuple type shall have distinct names or be unnamed.
+
+\[Example:
 ```csharp
-var t1 = (sum: 0, count: 1); // OK, all fields are named
-var t2 = (sum: 0, 1);       // Ok, implies (int sum, int);
-var t3 = (sum: 0, sum: 1);  // error! duplicate names.
+(int e1, float e1) t = (10, 1.2);					// Error: both elements have the same name
+(int e1, (int e1, int e2) e2) t = (10, (20, 30));	// OK: element names in each tuple type are distinct
 ```
+end example\]
 
-### Tuple literals
+The name of any element in a partial type declaration shall be the same for an element in the same position in any other partial declaration for that type.
 
-```csharp
-var t1 = (0, 2);              // infer tuple type from values
-var t2 = (sum: 0, count: 1);  // infer tuple type from names and values
-```
-
-A tuple literal is "target typed" whenever possible. The tuple literal has a "conversion from expression" to any tuple type, as long as the element expressions of the tuple literal have an implicit conversion to the element types of the tuple type.
-
-```csharp
-(string name, byte age) t = (null, 5); // Ok: the expressions null and 5 convert to string and byte
-```
-
-In cases where the tuple literal is not part of a conversion, it acquires its "natural type", which means a tuple type where the element types are the types of the constituent expressions. Since not all expressions have types, not all tuple literals have a natural type either:
-
-```csharp
-var t = ("John", 5); // Ok: the type of t is (string, int)
-var t = (null, 5); //   Error: null doesn't have a type
-```
-
-A tuple literal may include names, in which case they become part of the natural type:
-
-```csharp
-var t = (name: "John", age: 5); // The type of t is (string name, int age)
-```
-
-Tuple literals may be deconstructed directly:
-
-```csharp
-(string x, byte y, var z) = (null, 1, 2);
-(string x, byte y) t = (null, 1);
-```
-
-Or for deconstructing assignment:
-
-```csharp
-string x;
-byte y;
-
-(x, y) = (null, 1);
-(x, y) = (y, x); // swap!
-```
-
-The evaluation order of deconstruction assignment expressions is "breadth first":
-
-1. Evaluate the LHS: Evaluate each of the expressions inside of it one by one, left to right, to yield side effects and establish a storage location for each.
-1. Evaluate the RHS: Evaluate each of the expressions inside of it one by one, left to right to yield side effects
-1. Convert each of the RHS expressions to the LHS types expected, one by one, left to right.
-1. Assign each of the conversion results from 3 to the storage locations found in 
-
-> **Note to reviewers**: I found this in the LDM notes for July 13-16, 2016. I don't think it is still accurate:
-
-A deconstructing assignment is a *statement-expression* whose type could be `void`.
-
-### Duality with underlying type
-
-Tuples map to underlying types of particular names.
-
-```csharp
-System.ValueTuple<T1, T2>
-System.ValueTuple<T1, T2, T3>
-...
-System.ValueTuple<T1, T2, T3,..., T7, TRest>
-```
-
-Tuple types behave exactly like underlying types. The only additional enhancement is the more expressive field names given by the programmer.
-
-```csharp
-var t = (sum: 0, count: 1);
-t.sum   = 1;        // sum   is the name for the field #1
-t.Item1 = 1;        // Item1 is the name of underlying field #1 and is also available
-
-var t1 = (0, 1); // tuple omits the field names.
-t.Item1 = 1;     // underlying field name is still available
-
-t.ToString();     // ToString on the underlying tuple type is called.
-
-System.ValueTuple<int, int> vt = t; // identity conversion
-(int moo, int boo) t2 = vt;         // identity conversion
-
-```
-
-Because of the dual nature of tuples, it is not allowed to assign field names that overlap with preexisting member names of the underlying type. The only exception is the use of predefined `Item1`, `Item2`,...`ItemN` at corresponding position N, since that would not be ambiguous.
-
-```csharp
-var t =  (ToString: 0, ObjectEquals: 1);  // error: names match underlying member names
-var t1 = (Item1: 0, Item2: 1);            // valid
-var t2 = (misc: 0, Item1: 1);             // error: "Item1" was used in a wrong position
-```
-
-If the tuple is bigger than the limit of 7, the implementation will nest the "tail" as a tuple into the eighth element recursively. This nesting is visible by accessing the `Rest` field of a tuple, but that field my be hidden from e.g. auto-completion, just as the `ItemX` field names may be hidden but allowed when a tuple has named elements.
-
-A well formed "big tuple" will have names `Item1` etc. all the way up to the number of tuple elements, even though the underlying type doesn't physically have those fields directly defined. The same goes for the tuple returned from the `Rest` field, only with the numbers "shifted" appropriately.
-
-For tuple element names occurring in partial type declarations, the names must be the same.
-
+\[Example:
 ```csharp
 partial class C : IEnumerable<(string name, int age)> { ... }
-partial class C : IEnumerable<(string fullname, int)> { ... } // error: names must be specified and the same
+partial class C : IEnumerable<(string fullname, int)> { ... } // Error: names must be specified and the same
 ```
+end example\]
 
-When tuple elements names are used in overridden signatures, or implementations of interface methods, tuple element names in parameter and return types must be preserved. It is an error for the same generic interface to be inherited or implemented twice with identity convertible type arguments that have conflicting tuple element names
+A tuple cannot be created with the `new` operator. However, the `new` operator can be used to create and initialize an array of tuple or a nullable tuple.
 
-> note:
->
-> For the purpose of overloading, overriding and hiding, tuples of the same types and lengths as well as their underlying ValueTuple types are considered equivalent. All other differences are immaterial. When overriding a member it is permitted to use tuple types with same or different field names than in the base member.
->
-> A situation where same field names are used for non-matching fields between base and derived member signatures, a warning is reported by the compiler.  
+#### A tuple's underlying type
+
+Each tuple type maps to an underlying type. Specifically, a tuple having two elements maps to `System.ValueTuple<T1, T2>`, one with three elements maps to `System.ValueTuple<T1, T2, T3>`, and so on, up to seven elements. Tuple types having eight or more elements map to `System.ValueTuple<T1, T2, T3,..., T7, TRest>`. The first element in an underlying type has the public name `Item1`, the second `Item2`, and so on through `Item7`. Any elements beyond seven can be accesed as a group by the public name `Rest`, whose type is "tuple of the remaining elements". Alternatively, those elements can be accessed individually using the names `Item8` through `Item`*N*, where *N* is the total number of elements, even though the underlying type has no such names defined.
+
+A tuple type shall behave exactly like its underlying type. The only additional enhancement in the tuple type case is the ability to provide a more expressive name for each element.
+
+\[Example:
+```csharp
+var t1 = (sum: 0, 1);
+t1.sum   = 1;		// access the first element by its declared name
+t1.Item1 = 1;		// access the first element by its underlying name
+t1.Item2 = 3;		// access the second element by its underlying name
+
+System.ValueTuple<int, int> vt = t1;	// identity conversion
+
+ var t2 = (1, 2, 3, 4, 5, 6, 7, 8, 9);	// t2 is a System.ValueTuple<T1, T2, T3,..., T7, TRest>
+ var t3 = t4.Rest;						// t3 is a (int, int); that is, a System.ValueTuple<T1, T2>
+ System.Console.WriteLine("Item9 = {0}", t1.Item9);	// outputs 9 even though no such name Item9 exists!
+```
+end example\]
+
+**ISSUE:** Where will the type System.Value be defined (and its public names listed)?
+
+The name given explicitly to any element shall not be the same as any name in the underlying type, except that an explicit name may have the form `Item`*N* provided it corresponds position-wise with an element of the same name in the underlying type.
+
+\[Example:
+```csharp
+var t =  (ToString: 0, GetHashCode: 1);	// Error: names match underlying member names
+var t1 = (Item1: 0, Item2: 1);			// OK
+var t2 = (misc: 0, Item1: 1);			// Error: "Item1" used in a wrong position
+```
+end example\]
+
+#### Element names and overloading, overriding, and hiding
+
+When tuple element names are used in overridden signatures or implementations of interface methods, tuple element names in parameter and return types shall be preserved. It is an error for the same generic interface to be inherited or implemented twice with identity-convertible type arguments that have conflicting tuple element names.
+
+For the purpose of overloading, overriding and hiding, tuples of the same types and lengths as well as their underlying ValueTuple types shall be considered equivalent. All other differences are immaterial. When overriding a member it is permitted to use tuple types with the same names or names different than in the base member.
+
+**ISSUE:** Re the mention of "tuple length", which is not mentioned elsewhere; it seems to me that a tuple type includes an implied length based on the number of elements.
+
+If the same element name is used for non-matching elements in base and derived member signatures, the implementation shall issue a warning.  
 
 ```csharp
 class Base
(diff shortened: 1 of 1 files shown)


BillWagner:
I would like at least one approval before this gets merged. It's larger, and has more semantic changes than I feel comfortable merging without thorough review.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+    ;
+```
+
+Note that because it is not a constant expression, a tuple literal cannot be used as default value for an optional parameter.
+
+## Changes to [Types](../../spec/types.md)
+
+> The first paragraph of [Value types](../../spec/types.md#value-types) is replaced with the following text:
+
+A value type is either a `struct` type, an enumeration type, or a tuple type. C# provides a set of predefined struct types called the ***simple types***. The simple types are identified through reserved words.
```
gafter: >or a tuple type [](start = 61, length = 15)

This is not correct.  A tuple type is defined as an instantiation of a struct type, so it is just a struct type.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+> The `value_type` grammar is updated to include `tuple_type`:
```
gafter: >The [](start = 2, length = 3)

This should be deleted.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+    ;
+```
+
+## Additions to [Types](../../spec/types.md)
+
+> The following sections should be added after [Nullable types](../../spec/types.md#nullable-types) (at the end of the current *Value types* section.)
+
+### Tuple types
+
+Tuple types are declared with the following syntax:
```
gafter: >declared [](start = 16, length = 8)

Tuple types are not "declared".  They are used.  Moreover, your example is not a declaration of a type.  It is an *example* of a method declaration that uses a particular tuple type.

If you are intending to write spec text, you need to clearly distinguish normative (spec) language from examples.  Text such as this is an example, not a normative specification.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+The syntax `(int sum, int count)` indicates an anonymous data structure with public fields of the given names and types, referred to as *tuple*.
+
+Tuple values can be created using *tuple literals*:
+
+```csharp
+var t1 = (sum: 0, count: 1);
+var t2 = (0, 1);     // field names are optional
+```
+
+Tuples cannot be created with the `new` operator. The `new` operator can be used to initialize arrays of tuples, or nullable tuples:
```
gafter: >Tuples cannot be created with the `new` operator. [](start = 0, length = 49)

Is this a language rule or a comment about what constructors exist?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+1. Convert each of the RHS expressions to the LHS types expected, one by one, left to right.
+1. Assign each of the conversion results from 3 to the storage locations found in 
+
+> **Note to reviewers**: I found this in the LDM notes for July 13-16, 2016. I don't think it is still accurate:
+
+A deconstructing assignment is a *statement-expression* whose type could be `void`.
+
+### Duality with underlying type
+
+Tuples map to underlying types of particular names.
```
gafter: >map to underlying types of particular names [](start = 7, length = 43)

This section does not specify the mapping.  For example, VT`8 is not itself a tuple type unless the rest field is substituted with a tuple type.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+```csharp
+System.ValueTuple<T1, T2>
+System.ValueTuple<T1, T2, T3>
+...
+System.ValueTuple<T1, T2, T3,..., T7, TRest>
+```
+
+Tuple types behave exactly like underlying types. The only additional enhancement is the more expressive field names given by the programmer.
```
gafter: >The only additional enhancement [](start = 50, length = 31)

How is this enhancement affected in the language?  I would expect it to change name lookup rules.  Are those rules given here?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+t.Item1 = 1;     // underlying field name is still available
+
+t.ToString();     // ToString on the underlying tuple type is called.
+
+System.ValueTuple<int, int> vt = t; // identity conversion
+(int moo, int boo) t2 = vt;         // identity conversion
+
+```
+
+Because of the dual nature of tuples, it is not allowed to assign field names that overlap with preexisting member names of the underlying type. The only exception is the use of predefined `Item1`, `Item2`,...`ItemN` at corresponding position N, since that would not be ambiguous.
```
gafter: >with preexisting member names of the underlying type [](start = 91, length = 52)

Not quite true.  We permit, for example, ObjectEquals and GetType.  [https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZQAs6AsgBQCU+hxTA+ugLzrUDyARgFYBTAMYAXAKIBHAK4BDADYBnEOjgAadAHEBIgCoBPAA4DlMWgG5GRAL7IrQA=](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZQAs6AsgBQCU+hxTA+ugLzrUDyARgFYBTAMYAXAKIBHAK4BDADYBnEOjgAadAHEBIgCoBPAA4DlMWgG5GRAL7IrQA=)

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+```csharp
+var t =  (ToString: 0, ObjectEquals: 1);  // error: names match underlying member names
+var t1 = (Item1: 0, Item2: 1);            // valid
+var t2 = (misc: 0, Item1: 1);             // error: "Item1" was used in a wrong position
+```
+
+If the tuple is bigger than the limit of 7, the implementation will nest the "tail" as a tuple into the eighth element recursively. This nesting is visible by accessing the `Rest` field of a tuple, but that field my be hidden from e.g. auto-completion, just as the `ItemX` field names may be hidden but allowed when a tuple has named elements.
```
gafter: >auto-completion [](start = 236, length = 15)

IDE concerns have no place in a language spec.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+A well formed "big tuple" will have names `Item1` etc. all the way up to the number of tuple elements, even though the underlying type doesn't physically have those fields directly defined. The same goes for the tuple returned from the `Rest` field, only with the numbers "shifted" appropriately.
```
gafter: >"big tuple" [](start = 14, length = 11)

define

gafter: >have names [](start = 31, length = 10)

"have names" meaning affecting name lookup?  Please give the name lookup rules.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+For tuple element names occurring in partial type declarations, the names must be the same.
+
+```csharp
+partial class C : IEnumerable<(string name, int age)> { ... }
+partial class C : IEnumerable<(string fullname, int)> { ... } // error: names must be specified and the same
+```
+
+When tuple elements names are used in overridden signatures, or implementations of interface methods, tuple element names in parameter and return types must be preserved. It is an error for the same generic interface to be inherited or implemented twice with identity convertible type arguments that have conflicting tuple element names
```
gafter: >that have conflicting tuple element names [](start = 295, length = 41)

Not precisely true.  See [https://sharplab.io/#v2:CYLg1APglgdgLgUwE4DMCGBjBACAkgRgB4AVAPmwG8BfAWACh6ABAJmwGF9sQ8iAKWONgAeAGmwDsATwCU5avSas2rbhzEFC/eNjRiJAI1mUqQA=](https://sharplab.io/#v2:CYLg1APglgdgLgUwE4DMCGBjBACAkgRgB4AVAPmwG8BfAWACh6ABAJmwGF9sQ8iAKWONgAeAGmwDsATwCU5avSas2rbhzEFC/eNjRiJAI1mUqQA=)

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+> note:
+>
+> For the purpose of overloading, overriding and hiding, tuples of the same types and lengths as well as their underlying ValueTuple types are considered equivalent. All other differences are immaterial. When overriding a member it is permitted to use tuple types with same or different field names than in the base member.
+>
+> A situation where same field names are used for non-matching fields between base and derived member signatures, a warning is reported by the compiler.  
+
+```csharp
+class Base
+{
+    virtual void M1(ValueTuple<int, int> arg){...}
```
gafter: >virtual [](start = 4, length = 7)

virtual members may not be private

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+```csharp
+partial class C : IEnumerable<(string name, int age)> { ... }
+partial class C : IEnumerable<(string fullname, int)> { ... } // error: names must be specified and the same
+```
+
+When tuple elements names are used in overridden signatures, or implementations of interface methods, tuple element names in parameter and return types must be preserved. It is an error for the same generic interface to be inherited or implemented twice with identity convertible type arguments that have conflicting tuple element names
+
+> note:
+>
+> For the purpose of overloading, overriding and hiding, tuples of the same types and lengths as well as their underlying ValueTuple types are considered equivalent. All other differences are immaterial. When overriding a member it is permitted to use tuple types with same or different field names than in the base member.
```
gafter: >purpose [](start = 10, length = 7)

this sentence contradicts the previous sentence.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+```csharp
+class Base
+{
+    virtual void M1(ValueTuple<int, int> arg){...}
+}
+class Derived : Base
+{
+    override void M1((int c, int d) arg){...} // valid override, signatures are equivalent
+}
```
gafter: Incorrect.  See [https://sharplab.io/#v2:CYLg1APgAgTAjAWAFDNgAgEIEMDOBTZAb2TVLSgGY0A3ASwCcAXAVywBtyAWNAWTgAoAau2Z4AKswAObPAB5aAO0YAaNIsYA+NFnoBzAJSEAvshNJ0AETz1a1PMDQhMuAkmJIy5KgHs79G8B4XLwC/OpoAMaq4cD62nqGRmgA9Mk07LQOvtYBeKo4tLoKWCz0eDjxQXgAjsy27HhKpqgwaFY2dsCtTu229mhEJGSUaNn+mUFQ3Hz8YUqRcNHzEXE6BsYpaew43ulsmaoA7joKirqjCmiS3jgFAEYyaAC2tDjM+KMAZmjFT0EA5BF/gMkGYgA](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFDNgAgEIEMDOBTZAb2TVLSgGY0A3ASwCcAXAVywBtyAWNAWTgAoAau2Z4AKswAObPAB5aAO0YAaNIsYA+NFnoBzAJSEAvshNJ0AETz1a1PMDQhMuAkmJIy5KgHs79G8B4XLwC/OpoAMaq4cD62nqGRmgA9Mk07LQOvtYBeKo4tLoKWCz0eDjxQXgAjsy27HhKpqgwaFY2dsCtTu229mhEJGSUaNn+mUFQ3Hz8YUqRcNHzEXE6BsYpaew43ulsmaoA7joKirqjCmiS3jgFAEYyaAC2tDjM+KMAZmjFT0EA5BF/gMkGYgA)

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+> note:
+>
+> An element name at one position on one side of a conversion, and the same name at another position on the other side almost certainly have bug in the code:
+
+```csharp
+(string first, string last) GetNames() { ... }
+(string last, string first) names = GetNames(); // Oops!
+```
+
+> Compilers should issue a warning for the preceding code.
```
gafter: >should [](start = 12, length = 6)

Yes, they should.  They do not.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+> The following text should be added to [Boxing conversions](../../spec/conversions.md#boxing-conversions) after the first paragraph:
+
+Tuples, like all value types, have a boxing conversion. Importantly, the names aren't part of the runtime representation of tuples, but are tracked only by the compiler. Thus, once you've "cast away" the names, you cannot recover them. In alignment with the identity conversions, a boxed tuple will unbox to any tuple type that has the same element types in the same order.
+
+### Tuple conversions
+
+> This section should be added after [Implicit enumeration conversions](../../spec/conversions.md#implicit-enumeration-conversions)
+
+Tuple types and expressions support a variety of conversions by "lifting" conversions of the elements into overall *tuple conversion*.
+For the classification purpose, all element conversions are considered recursively. For example: To have an implicit conversion, all element expressions/types must have implicit conversions to the corresponding element types.
```
gafter: >considered recursively [](start = 60, length = 22)

Then they need to be specified recursively.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+Tuple conversions are *Standard Conversions* and therefore can stack with user-defined operators to form user-defined conversions.
+
+An implicit tuple conversion is a standard conversion. It applies between two tuple types of equal arity when there is any implicit conversion between each corresponding pair of types.
+
+An explicit tuple conversion is a standard conversion. It applies between two tuple types of equal arity when there is any explicit conversion between each corresponding pair of types.
+
+A tuple conversion can be classified as a valid instance conversion or an extension method invocation as long as all element conversions are applicable as instance conversions.
```
gafter: >instance conversion [](start = 48, length = 19)

I don't know what this means.

gafter: >extension method invocation [](start = 74, length = 27)

Ditto.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+On top of the member-wise conversions implied by target typing, implicit conversions between tuple types themselves are allowed.
```
gafter: >implicit conversions between tuple types themselves are allowed [](start = 64, length = 63)

I don't know what this means.  Is this asserting that there is something that isn't forbidden?  Normally things described by the spec that aren't forbidden are allowed.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+var t = ((int moo, int boo))o;     // unboxing conversion
+```
+
+## Additions to [Variables](../../spec/variables.md)
+
+> The following text should be added at the end of the [Variables](../../spec/variables.md) section.
+
+### "Discards"
+
+The `_` can be used as a *discard* under these rules:
```
gafter: >used as a *discard* [](start = 15, length = 19)

I don't know what "used as a *discard*" means.

gafter: >The `_` [](start = 0, length = 7)

I suspect this is intended to imply something about name lookup.  Do you have proposed changes for name lookup?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+M(out _, out var _, out int _); // three out variable discards
+(_, var _, int _) = GetCoordinates(); // deconstruction into discards
+if (x is var _ && y is int _) { ... } // discards in patterns
+```
+
+## Changes to [Conversions](../spec/conversions.md)
+
+> The following text should be added to [Identity conversions](../../spec/conversions.md#identity-conversions), after the bullet point on `object` and `dynamic`:
+
+* Element names are immaterial to tuple conversions. Tuples with the same types in the same order are identity convertible to each other or to and from corresponding underlying `ValueTuple` types, regardless of the names.
```
gafter: >to and from corresponding underlying `ValueTuple` types [](start = 140, length = 55)

I think it needs to be said somewhere what "underlying type" means.  For example, the type `(int, int)` and `ValueTuple<int, int>` are defined to designate the *same* type.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+Tuples, like all value types, have a boxing conversion. Importantly, the names aren't part of the runtime representation of tuples, but are tracked only by the compiler. Thus, once you've "cast away" the names, you cannot recover them. In alignment with the identity conversions, a boxed tuple will unbox to any tuple type that has the same element types in the same order.
+
+### Tuple conversions
+
+> This section should be added after [Implicit enumeration conversions](../../spec/conversions.md#implicit-enumeration-conversions)
+
+Tuple types and expressions support a variety of conversions by "lifting" conversions of the elements into overall *tuple conversion*.
+For the classification purpose, all element conversions are considered recursively. For example: To have an implicit conversion, all element expressions/types must have implicit conversions to the corresponding element types.
+
+Tuple conversions are *Standard Conversions* and therefore can stack with user-defined operators to form user-defined conversions.
```
gafter: >stack [](start = 63, length = 5)

We don't have a definition of "stack", and in any case we don't repeat the implications of standard conversion on every conversion that is a standard conversion.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+```csharp
+void M1((int x, int y) arg){...};
+void M1((object x, object y) arg){...};
+
+M1((1, 2));            // first overload is used. Identity conversion is better than implicit conversion.
+M1(("hi", "hello"));   // second overload is used. Implicit tuple conversion is better than no conversion.
+```
+
+Target typing will "see through" nullable target types. A successful conversion from tuple expression to a nullable tuple type is classified as *ImplicitNullable* conversion.
```
gafter: >Target typing will "see through" nullable target types. [](start = 0, length = 55)

Isn't this already the way nullable conversions work?  What is changing here?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+}
+```
+
+## Additions to [Expressions](../../spec/expressions.md)
+
+### Overload resolution and tuples with no natural types
+
+> The following text is added after after the bullet list in [Exactly matching expressions](../../spec/expressions.md#Exactly-matching-expressions):
+
+The exact match rule for tuple expressions is based on the natural types of the constituent tuple arguments. The rule is mutually recursive with respect to other containing or contained expressions not in a possession of a natural type.
```
gafter: >exact match rule for tuple expressions is based on [](start = 4, length = 50)

What is the rule?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+### Deconstruction expressions
+
+> This section should be added at the end of the [Expressions](../../spec/expressions.md) chapter.
+
+Tuple deconstruction expressions receive and "split out" a tuple:
+
+```csharp
+(var sum, var count) = Tally(myValues); // deconstruct result
```
gafter: >var sum [](start = 1, length = 7)

What grammar rules permit this syntax?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+The `_` wildcard indicates that the one or more of the tuple fields are discarded:
+
+```csharp
+(var sum, _) = Tally(myValues); // deconstruct result
+Console.WriteLine($"Sum: {sum}, count was ignored");  
+```
+
+Any object may be deconstructed by providing an accessible `Deconstruct` method, either as a member or as an extension method. A `Deconstruct` method converts an object to a set of discrete values. The Deconstruct method "returns" the component values by use of individual `out` parameters. Deconstruct is overloadable.
```
gafter: >a [](start = 91, length = 1)

an instance

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+The deconstructor pattern could be implemented as a member method, or an extension method:
```
gafter: >deconstructor pattern [](start = 4, length = 21)

This is not defined.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+(int x, int y) = (__x, __y);
+```
+
+### Additions to [Classes](../../spec/classes.md)
+
+> The following note should be added to the end of the section on [extension methods](../../spec/classes.md#extension-methods):
+
+> note: 
+>Extension methods on a tuple type apply to tuples with different element names:
+
```
gafter: This is not normative.  It is a consequence of the identity conversions.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+Tuples cannot be created with the `new` operator. The `new` operator can be used to initialize arrays of tuples, or nullable tuples:
+
+```csharp
+var array = new (int x, int y)[10];
+var nullable = new (int x, int y)?();
+```
+
+Tuples can be created for a known target type:
```
gafter: >Tuples can be created for a known target type [](start = 0, length = 45)

Is this part of the *Tuple Literals* section?  That is what it appears to be describing.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+### Tuple literals
+
+```csharp
+var t1 = (0, 2);              // infer tuple type from values
+var t2 = (sum: 0, count: 1);  // infer tuple type from names and values
+```
+
+A tuple literal is "target typed" whenever possible. The tuple literal has a "conversion from expression" to any tuple type, as long as the element expressions of the tuple literal have an implicit conversion to the element types of the tuple type.
```
gafter: >type [](start = 119, length = 4)

of the same arity

gafter: >element [](start = 216, length = 7)

corresponding element

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+Examples:
+
+```csharp
+M(out _, out var _, out int _); // three out variable discards
+(_, var _, int _) = GetCoordinates(); // deconstruction into discards
+if (x is var _ && y is int _) { ... } // discards in patterns
+```
+
+## Changes to [Conversions](../spec/conversions.md)
```
gafter: >Changes [](start = 3, length = 7)

You do not have a spec for the tuple conversion.  See [https://sharplab.io/#v2:CYLg1APgAgTABAQQLACgDeqC+rWzgIVQxTlLigGZyBGANjgEsBbABwBsGBjBgFzgHsWAUwBOAQx78RBABQI4YgJRwAvAD44AOwCubNgG4sOFHijUA7EVRlyMGNbLEbN+WNVbdBh87hyANIjKPNTuMmIBSoYkPr74AfhB8CpwwfpwAPTpKQAWDADOcJz8mgBuonkMxd5w2Ci1QA==](https://sharplab.io/#v2:CYLg1APgAgTABAQQLACgDeqC+rWzgIVQxTlLigGZyBGANjgEsBbABwBsGBjBgFzgHsWAUwBOAQx78RBABQI4YgJRwAvAD44AOwCubNgG4sOFHijUA7EVRlyMGNbLEbN+WNVbdBh87hyANIjKPNTuMmIBSoYkPr74AfhB8CpwwfpwAPTpKQAWDADOcJz8mgBuonkMxd5w2Ci1QA==)

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+### Tuple conversions
+
+> This section should be added after [Implicit enumeration conversions](../../spec/conversions.md#implicit-enumeration-conversions)
+
+Tuple types and expressions support a variety of conversions by "lifting" conversions of the elements into overall *tuple conversion*.
+For the classification purpose, all element conversions are considered recursively. For example: To have an implicit conversion, all element expressions/types must have implicit conversions to the corresponding element types.
+
+Tuple conversions are *Standard Conversions* and therefore can stack with user-defined operators to form user-defined conversions.
+
+An implicit tuple conversion is a standard conversion. It applies between two tuple types of equal arity when there is any implicit conversion between each corresponding pair of types.
```
gafter: >implicit conversion between [](start = 123, length = 27)

In either direction?

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+An explicit tuple conversion is a standard conversion. It applies between two tuple types of equal arity when there is any explicit conversion between each corresponding pair of types.
+
+A tuple conversion can be classified as a valid instance conversion or an extension method invocation as long as all element conversions are applicable as instance conversions.
+
+On top of the member-wise conversions implied by target typing, implicit conversions between tuple types themselves are allowed.
+
+### Target typing
```
gafter: >Target typing [](start = 4, length = 13)

Tuple Literal Conversion

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+> This section should be added after [Anonymous function conversions and method group conversions](../../spec.md#Anonymous-function-conversions-and-method-group-conversions)
+
+A tuple literal is "target typed" when used in a context specifying a tuple type. The tuple literal has a "conversion from expression" to any tuple type, as long as the element expressions of the tuple literal have an implicit conversion to the element types of the tuple type.
```
gafter: >element [](start = 245, length = 7)

corresponding

gafter: >type [](start = 148, length = 4)

of the same length

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+    ;
+
+tuple_literal
+    : '(' tuple_literal_element_list ')'
+    ;
+
+tuple_literal_element_list
+    : tuple_literal_element ',' tuple_literal_element
+    | tuple_literal_element_list ',' tuple_literal_element
+    ;
```
gafter:
```antlr
tuple_literal
    : '('  ( tuple_literal_element ',' )+ tuple_literal_element ')'
    ;
```

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
+# Tuples 
+
+This proposal specifies the changes required to the [C# 6.0 (draft) Language specification](../../spec/introduction.md) to support *Tuples* as a new [value type](../../spec/types.md#value-types).
+
+## Changes to [Lexical structure](../../spec/lexical-structure.md)
+
+> The grammar for [Literals](../../spec/lexical-structure.md#literals) is extended to include `tuple_literal`:
```
gafter: >lexical-structure.md [](start = 40, length = 20)

Tuple literals are not a new lexical structure.  They belong in a new section in expressions, and require semantics to be defined there.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+
+```csharp
+var t = (name: "John", age: 5); // The type of t is (string name, int age)
+```
+
+Tuple literals may be deconstructed directly:
+
+```csharp
+(string x, byte y, var z) = (null, 1, 2);
+(string x, byte y) t = (null, 1);
```
gafter: No grammar rule in this specification permits this statement form.

proposals/csharp-7.0/tuples.md
```diff
@@ -0,0 +1,437 @@
...
+```
+
+Or for deconstructing assignment:
+
+```csharp
+string x;
+byte y;
+
+(x, y) = (null, 1);
+(x, y) = (y, x); // swap!
```
gafter: This appears to be an assignment in which the left-hand-side isn't an lvalue, which would be an error.  How is it permitted?

gafter:
Generally, this relies too much on examples to convey what need to be normative rules.  I suggest separating ALL examples into non-normative "[Example: ..." sections as in the language spec, and then making sure that each language construct has sufficient specification to be understood fully without reference to the examples.  Examples should be present just to give the reader motivating examples and some intuition.

Every construct requires (where applicable)
- Syntax (through a set of grammar rules)
- Semantic meaning for the constructs, including type
- Semantic constraints including required errors ("it is an error if...")
- Runtime/dynamic semantics of what it means at runtime.

Every conversion needs a name, it needs to be clear if it is a conversion from expression or from type, whether implicit or explicit, whether a standard conversion, etc.  Where lists of conversions appear elsewhere, they need to be reviewed to see if they need to be amended with the new conversion.

BillWagner:
I changed the base branch to a feature branch that @RexJaeschke and I will use to keep working on this project.

I'll merge this PR into that branch. Then, Rex and I will keep iterating on this (and feature specs).

gafter:
I am disappointed that this was merged without addressing the many comments.  The comments are not even retained as issues somewhere, so I'm not sure how you intend to track them.

BillWagner:
@gafter  I'll be adding issues for all those comments by tomorrow morning. They'll be in this project: https://github.com/dotnet/csharplang/projects/5

As I add the issues, I'll add links to all the original comments so we don't lose any of the review work.

proposals/csharp-7.0/tuples.md
BillWagner: Addressed in #3150 (targeting the feature branch)

proposals/csharp-7.0/tuples.md
BillWagner: Addressed in #3150 (targeting the feature branch)

proposals/csharp-7.0/tuples.md
BillWagner: The first comment is address in #3150 (targeting the feature branch).

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3151

proposals/csharp-7.0/tuples.md
BillWagner: logged as #3152

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3153

proposals/csharp-7.0/tuples.md
BillWagner: Rex has updated this description. I've noted (in our working draft) that the language used may still need refinement. We need to define the term "element" to mean "ItemN" and "Rest" elements, excluding other members.

proposals/csharp-7.0/tuples.md
BillWagner: Rex has addressed this as a note in the first paragraph in [this clause](https://github.com/dotnet/csharplang/blob/standard-proposals/proposals/csharp-7.0/tuples.md#tuple-element-name-erasure-at-runtime)

proposals/csharp-7.0/tuples.md
BillWagner: Rex has re-written this paragraph. The "big tuple" term no longer appears.

BillWagner: also logged as #3153

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3154

proposals/csharp-7.0/tuples.md
BillWagner: added to #3154

proposals/csharp-7.0/tuples.md
BillWagner: addressed in #3150.  All members and types in this example are public.

proposals/csharp-7.0/tuples.md
BillWagner: added to #3154

proposals/csharp-7.0/tuples.md
BillWagner: Addressed in #3150. I removed this paragraph, so if a warning is added later it would be compliant.

proposals/csharp-7.0/tuples.md
BillWagner: Rex has addressed this in the feature branch.

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3155

BillWagner: logged in #3155

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3156

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3153

BillWagner: logged in #3181

proposals/csharp-7.0/tuples.md
BillWagner: Addressed in #3150

proposals/csharp-7.0/tuples.md
BillWagner: addressed in #3150

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3183

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3184

proposals/csharp-7.0/tuples.md
BillWagner: instance added in #3146 

Other comments logged in #3185

proposals/csharp-7.0/tuples.md
BillWagner: also logged in #3185

proposals/csharp-7.0/tuples.md
BillWagner: Addressed. This note is not marked as informative.

proposals/csharp-7.0/tuples.md
BillWagner: This section has been re-written. @RexJaeschke is writing an issue to discuss and resolve "tuple literals" and "tuple expressions". I'll update this comment to tag that issue.

proposals/csharp-7.0/tuples.md
BillWagner: resolved in #3150

BillWagner: resolved.

proposals/csharp-7.0/tuples.md
BillWagner: addressed in #3150

proposals/csharp-7.0/tuples.md
BillWagner: addressed in #3150

proposals/csharp-7.0/tuples.md
BillWagner: fixed in #3150

BillWagner: Fixed in #3150. Used "arity" for consistency.

proposals/csharp-7.0/tuples.md
BillWagner: fixed in #3150

proposals/csharp-7.0/tuples.md
BillWagner: logged in #3186

proposals/csharp-7.0/tuples.md
BillWagner: Logged in #3187

PR #3035, opened 2019-12-17
Unconstrained type parameter annotation proposal

cston:


proposals/unconstrained-type-parameter-annotation.md
```diff
@@ -0,0 +1,114 @@
...
+_There is a potential for cycles decoding a `[Nullable]T` type reference, particularly when the type parameter
+reference appears in the constraints of `T` which are used in determining if the type parameter is a
+value type or reference type._
+
+### Public API
+There are no changes to the compiler API.
+
+`ITypeParameterSymbol.NullableAnnotation` will return `NullableAnnotation.Annotated` for `T??`, indistinquisable from `T?`.
+To determine if that type parameter annotation would be represented in source as `??` would require checking the
+constraints of the `ITypeParameterSymbol` which might involve checking constraints inherited from base types.
```
CyrusNajmabadi: seems painful fwiw.

proposals/unconstrained-type-parameter-annotation.md
```diff
@@ -0,0 +1,114 @@
...
+{
+    internal override U?? F<U>() => default; // Is ?? allowed or required?
+}
+class B3 : A<int?>
+{
+    internal override U?? F<U>() => default; // Is ?? allowed or required?
+}
+```
+
+`T??` may be used interchangably with `[MaybeNull]T` in cases where `T` is not constrained to a value type or reference type.
```
alrz: I really hope this lands as `T?` at the end - for the same reasons that the team were avoiding introducing `T!`.

Furthermore, the fact that you need to change annotations `T??` -> `T?` if you add a constraint is a pain point.

orthoxerox: I really hope unconstrained `T?` will sooner or later end up as an unconstrained `T?` that is `T?` for reference types or `T?` for value types.

alrz: >I really hope unconstrained T? will sooner or later end up as an unconstrained T? that is T? for reference types or T? for value types.

That would probably require runtime changes - personally I'm ok with `[MaybeNull]` for the time being. `T??` would be just an oddball that I don't want to have to deal with.

mikernet:
Sorry for posting here but I can't for the life of me find the discussion where @gafter and others are discussing unconstrained `T` nullable syntax. 

Would `T??` also be possible when `T : notnull` or only when it is strictly unconstrained? Please tell me it's the former.

cston:
@mikernet, `T??` would be allowed `where T : notnull` and in other cases where `T` is not constrained to a value type or reference type.

alrz:
Looks like this finally implemented as `T?`?


BTW it's not documented in language feature status or https://github.com/dotnet/csharplang/tree/master/proposals/csharp-9.0

cston:
Replaced by support for `T?` (see [LDM-2020-06-17](https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-06-17.md#t)).

PR #3076, opened 2020-01-06
Add spec work for record ctor

agocke:


proposals/records-wip.md
```diff
@@ -28,3 +28,17 @@ struct_body
 ```
 
 The `attributes` non-terminal will also permit a new contextual attribute, `data`.
+
+A class (struct) declared with the a parameter list is called a record class (record struct), either of which is a record type.
```
YairHalberstadt: This looks like it contains some typos.

proposals/records-wip.md
```diff
@@ -28,3 +28,17 @@ struct_body
...
+
+## Members of a record type
+
+In addition to the members declared in the class-body, a record type has the following additional members:
+
+**Primary Constructor**
```
gafter: OK, but what does the primary constructor *do*?

proposals/records-wip.md
```diff
@@ -28,3 +28,17 @@ struct_body
...
+
+A record type has a public constructor whose signature corresponds to the value parameters of the
+type declaration. This is called the primary constructor for the type, and causes the implicitly
+declared default constructor to be suppressed. If a constructor with the same signature is
+already present in the class, either through inheritance or user declaration, the synthesized
```
gafter: Constructors are not inherited.

proposals/records-wip.md
```diff
@@ -28,3 +28,17 @@ struct_body
...
+primary constructor is suppressed.
```
gafter: If it is suppressed, where is the code to do the things the implicitly generated constructor would do?

proposals/records-wip.md
```diff
@@ -13,7 +13,7 @@ class_declaration
 
 struct_declaration
```
gafter: This spec is very hard to review because it describes only a fragment of the feature.  Please provide a more complete specification, even if you are implementing only part of it at this time.  That is the only way global issues with the specification can be raised and addressed.

proposals/records-wip.md
gafter: > ... with the a parameter ...

agocke:
@gafter Fair point. To avoid writing out the whole record proposal I've tried to make a decent spec for a subset of the record feature, including generated constructors and properties. I'll add generated equality next.

proposals/records-wip.md
```diff
@@ -28,3 +28,35 @@ struct_body
...
+## Members of a record type
+
+In addition to the members declared in the class-body, a record type has the following additional members:
+
+### Primary Constructor
+
+A record type has a public constructor whose signature corresponds to the value parameters of the
+type declaration. This is called the primary constructor for the type, and causes the implicitly
+declared default constructor to be suppressed. If a constructor with the same signature is
+already present in the class, the synthesized primary constructor is suppressed.
```
gafter: >already present [](start = 0, length = 15)

explicitly declared?

proposals/records-wip.md
```diff
@@ -28,3 +28,35 @@ struct_body
...
+1. initializes compiler-generated backing fields for the properties corresponding to the value parameters (if these properties are compiler-provided; see [Synthesized properties](#Synthesized Properties)); then
+
+2. executes the instance field initializers appearing in the class-body; and then
+    invokes the base class constructor with no arguments.
+
+[ ] TODO: add base call syntax and specification about choosing base constructor through overload resolution
+
+### Properties
+
+For each record parameter of a record type declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) public property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:
```
gafter: >public [](start = 206, length = 6)

probably delete `public`.  I think you want to permit an explicit declaration to change the protection.

proposals/records-wip.md
```diff
@@ -28,3 +28,35 @@ struct_body
...
+
+For a record struct or a sealed record class:
```
gafter: >sealed [](start = 25, length = 6)

And what if it isn't sealed?

proposals/records-wip.md
```diff
@@ -13,7 +13,7 @@ class_declaration
 
 struct_declaration
     : attributes? struct_modifiers? 'partial'? 'struct' identifier type_parameter_list?
-      record_parameters? struct_interfaces? type_parameter_constraints_clauses? struct_body
+      parameter_list? struct_interfaces? type_parameter_constraints_clauses? struct_body
```
gafter: >parameter_list [](start = 6, length = 14)

Are these in scope anywhere?  If so, where?

proposals/records-wip.md
```diff
@@ -28,3 +28,35 @@ struct_body
...
+
+[ ] TODO: add base call syntax and specification about choosing base constructor through overload resolution
+
+### Properties
+
+For each record parameter of a record type declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) public property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:
+
+For a record struct or a sealed record class:
+
+* A public get-only auto-property is created. Its value is initialized during construction with the value of the corresponding primary constructor parameter. Each "matching" inherited virtual property's get accessor is overridden.
```
gafter: >inherited virtual [](start = 174, length = 17)

But the previous paragraph says that inherited virtual (non-abstract) properties suppress it.

proposals/records-wip.md
YairHalberstadt: The naming convention for parameters is lower case, and for properties is uppercase. It's there any intention of fixing up the casing?

proposals/records-wip.md
```diff
@@ -28,3 +28,35 @@ struct_body
 ```
 
 The `attributes` non-terminal will also permit a new contextual attribute, `data`.
+
+A class (struct) declared with a parameter list or `data` modifier is called a record class (record struct), either of which is a record type.
+
+It is an error to declare a record type without both a parameter list and the `data` modifier.
```
YairHalberstadt: Is there any consideration of splitting these two? So a parameter lost on its own is a primary constructor, and data causes it to generate properties and equality.

proposals/records-wip.md
gafter: Our current thinking is "no".  Just use uppercase.

---
In reply to: [364061234](https://github.com/dotnet/csharplang/pull/3076#discussion_r364061234) [](ancestors = 364061234)

proposals/records-wip.md
gafter: The reason the `data` keyword is being introduced is to keep the door open to doing that.  We'll be discussing next week whether or not we want to stick with that (or, alternatively, drop the `data` keyword).

---
In reply to: [364061435](https://github.com/dotnet/csharplang/pull/3076#discussion_r364061435) [](ancestors = 364061435)

agocke:
I'm gonna merge this then follow up with more detail for the next PR

PR #3122, opened 2020-01-23
Added case to clarify code change (Fixes #16279)

IntegerMan:
## Summary

Added the case statement described in the issue. Technically it was already covered by the default case's fall through, but when illustrating a before and after in a language specification, it's probably best to be explicit and avoid confusion.

Fixes dotnet/docs#16279


Thraka:
I don't have permissions here. Tagging @BillWagner

PR #3123, opened 2020-01-23
Improved syntax of switch condition (#16280)

IntegerMan:
## Summary

Addressed a concern on the switch condition potentially containing null values and being more of a correct example using `ToLowerInvariant` instead of `ToLower`.

I also agreed with @Thraka that the `InvalidCommand` case should not be changed and so I left out the null / empty case handling as well since it didn't make sense without the default case change.

Fixes dotnet/docs#16280

### DIFF
--- a/spec/statements.md
+++ b/spec/statements.md
@@ -539,7 +539,7 @@ case 2:
 The governing type of a `switch` statement may be the type `string`. For example:
 ```csharp
 void DoCommand(string command) {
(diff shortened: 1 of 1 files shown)


Thraka:
@BillWagner

BillWagner:
I'm also closing this, as it matches the current ECMA standard.

We'll take up this issue in the standardization committee meetings.

PR #3127, opened 2020-01-27
Add proposal for 'key' modifier

agocke:
Implements structural equality.

### DIFF
--- a/proposals/key-equals.md
+++ b/proposals/key-equals.md
@@ -1,0 +1,133 @@
+
+# Ad hoc structural equality
+
+One of the component features proposed as a part of records is structural equality defined on certain members. This seems like a feature which could be orthogonal and useful for regular classes. The following proposal details how we could mark specific members as "keys" to a generated structural equality.
+
+Much like the `Key` modifier on anonymous properties in VB, a `key` modifier will be permitted on C# fields or properties.
+
+If there are any `key` modifiers present on any members in a class or struct declaration,
+equality will be generated for the type based on the annotated members as follows.
+
+If no `key` members are present, the type uses inherited equality, like today.
+
+## Synthesized members
+
+An example of a complex hierarchy with multiple types with `key` modifiers:
+
+```C#
+class A
+{
+    public key int P1 { get; }
+    public int P2 { get; }
+}
+class B : A
+{
+    public key int P3 { get; }
+    public int P4 { get; }
+}
+```
+
+produces
+
+```C#
+class A : IEquatable<A>
+{
+    public key int P1 { get; }
+    public int P2 { get; }
+
+    public override bool Equals(object o)
+        => o is A a && Equals(a);
+
+    public override int GetHashCode()
+        => P1;
+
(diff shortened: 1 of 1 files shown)


proposals/key-equals.md
```diff
@@ -0,0 +1,133 @@
...
+        => EqualityContractOrigin == a.EqualityContractOrigin
+        && KeyEquals(a);
+
+    protected virtual Type EqualityContractOrigin
+        => typeof(A);
+
+    protected bool KeyEquals(A a)
+        => P1 == a.P1;
+
+    public static bool operator==(A left, A right) => left.Equals(right);
```
YairHalberstadt: This doesn't check for null?

proposals/key-equals.md
```diff
@@ -0,0 +1,133 @@
...
+    protected bool KeyEquals(B b)
+        => base.KeyEquals(b) && P3 == a.P3;
+
+    public static bool operator==(B left, B right) => left.Equals(right);
+    public static bool operator!=(B left, B right) => !(left == right);
+}
+```
+
+
+Three synthesized members will be added to the type, `T`:
```
333fred: This is well more than 3 synthesized members.

proposals/key-equals.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. override of `object.Equals(object)`. It is an error if a base class has sealed this member.
```
333fred: Also consider commenting on what happens if the class has written out implementations of any of these members.

proposals/key-equals.md
agocke: Good point! Yeah, it seems good to make this null-aware.

proposals/key-equals.md
agocke: heh, just made up a number before writing this list and forgot to go back

omariom:
`EqualityContractOrigin` is basically `GetType()`, right?

Then B.Equals could be:
```C#
public bool Equals(B other)  => other.GetType() == typeof(B) && KeyEquals(b);
```
```other.GetType() == typeof(SomeType)``` is optimised by JIT since probably .NET 2.0 and is extremely efficient. Using `EqualityContractOrigin`  means Type objects will have to be instantiated.

CyrusNajmabadi:
> EqualityContractOrigin is basically GetType(), right?

No.  It *can* be implemented that way, but it's not required to be implemented that way.  The point is to allow the impl (including user overrides) to make the call on what determines equality.

gafter:
> `EqualityContractOrigin` is basically `GetType()`, right?

No.  Having it be a virtual method means that you can extend a class (implementation inheritance) without changing its equality contract.

omariom:
Can `Equals`  be implemented in a more efficient manner if **B** is sealed?
What if **B** already has its own implementation of `IEquatable<B>`.

bugproof:
Why closed?

agocke:
We ended up implementing records without the notion and at the moment I don't think it's worth the additional complexity.

PR #3147, opened 2020-02-04
csharplang-attributes-1

chamukallem-msft:


spec/attributes.md
```diff
@@ -162,84 +162,67 @@ The types of positional and named parameters for an attribute class are limited
...
-
-attribute_argument_expression
-    : expression
-    ;
+global_attributes: 
+    global_attribute_section+;
+
+global_attribute_section:
+     '[' global_attribute_target_specifier attribute_list ']' |
+     '[' global_attribute_target_specifier attribute_list ',' ']';
```
CyrusNajmabadi: why was this changed?  The previous way is much easier to read.

CyrusNajmabadi: specifically, having each production start with either `:` or `|` scans really well.  It's also how we generate our grammar files here: https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Generated/CSharp.Generated.g4

chamukallem-msft:
I should I have branched out of csharp5-standard

PR #3159, opened 2020-02-06
Add proposal for "Simple programs" feature

AlekseyTs:


proposals/Simple-programs.md
```diff
@@ -0,0 +1,169 @@
...
+[summary]: #summary
+
+Allow a sequence of *statements* to occur right before the *namespace_member_declaration*s of a *compilation_unit* (i.e. source file).
+
+The semantics are that if such a sequence of *statements* is present, the following type declaration, modulo the actual type name and the method name, would be emitted:
+
+``` c#
+static class Program
+{
+    static async Task Main()
```
YairHalberstadt: Can you return a Task<int>?
What about accessing args?

proposals/Simple-programs.md
```diff
@@ -0,0 +1,169 @@
...
+        await System.Threading.Tasks.Task.Delay(1000);
+        System.Console.WriteLine("Hi!");
+    }
+}
+```
+
+### Scope of top-level local variables and local functions
+
+Even though top-level local variables and functions are "wrapped" 
+into the generated entry point method, they should still be in scope throughout the program.
```
orthoxerox: This line without the rest of the explanation suggests the wrong interpretation of the feature. They are not "in scope" as most users understand it (reachable and usable), they just "poison" it (reachable and unusable) to provide more useful diagnostics.

proposals/Simple-programs.md
```diff
@@ -0,0 +1,169 @@
...
+[design]: #detailed-design
+
+### Syntax
+
+The only additional syntax is allowing a sequence of *statement*s in a compilation unit,
+just before the *namespace_member_declaration*s:
+
+``` antlr
+compilation_unit
+    : extern_alias_directive* using_directive* global_attributes? statement* namespace_member_declaration*
```
orthoxerox: Should statements come before declarations or *instead of* them at this stage? What is the presumed process of migrating learners to "complex programs"?

proposals/Simple-programs.md
```diff
@@ -0,0 +1,169 @@
...
+    ;
+```
+
+In all but one *compilation_unit* the *statement*s must all be local function declarations.
```
orthoxerox: Should there even be multiple compilation units with statements to support local functions? Or can this be covered by scenario 2 (top level functions) later?

proposals/Simple-programs.md
AlekseyTs: > Can you return a Task<int>?

Not at the moment.

> What about accessing args?

At the moment the method has no args.


---
In reply to: [375636464](https://github.com/dotnet/csharplang/pull/3159#discussion_r375636464) [](ancestors = 375636464)

proposals/Simple-programs.md
AlekseyTs: > Should there even be multiple compilation units with statements to support local functions?

There is no requirement to place Local Functions into a separate compilation unit.

---
In reply to: [375677410](https://github.com/dotnet/csharplang/pull/3159#discussion_r375677410) [](ancestors = 375677410)

proposals/Simple-programs.md
AlekseyTs: > Should statements come before declarations or instead of them at this stage? 

As the grammar indicates, top level statements should come before type/namespace declarations

---
In reply to: [375676793](https://github.com/dotnet/csharplang/pull/3159#discussion_r375676793) [](ancestors = 375676793)

proposals/Simple-programs.md
```diff
@@ -0,0 +1,169 @@
...
+If `await` expressions and other async operations are omitted, no warning is produced. Instead the
+signature of the generated entry point method is equivalent to 
+``` c#
+    static void Main()
+```
+
+The example above would yield the following `$Main` method declaration:
+
+``` c#
+static class $Program
```
jkotas, line 110: Do you plan to actually emit a hidden type for this, or is the Main going to emitted as a global function in this case?

YairHalberstadt: The current implementation emits a hidden type: https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQGUCWBbADgGwgAUEB7AcwSm2QBoYRN8AfAAQCYBGAWAChXOATgAUAIgAWEfPlKiAlAG4+AAlXKgA

AlekseyTs:
> Do you plan to actually emit a hidden type for this, or is the Main going to emitted as a global function in this case?

The current plan is to synthesize a type with a method inside it.

proposals/Simple-programs.md
AlekseyTs: > Do you plan to actually emit a hidden type for this, or is the Main going to emitted as a global function in this case?

The current plan is to synthesize a type with a method inside it.

jkotas: Is there a reason why not emit it as global function? The global functions were kind of meant for this. Just wondering.

AlekseyTs: > Is there a reason why not emit it as global function? The global functions were kind of meant for this. Just wondering.

I do not believe there is a good reason to complicate things. As it stands right now, compiler never emits global functions, not to mention global functions with user code. Our pipeline from beginning to the end assumes there is a type and that works just fine for this scenario.

PR #3171, opened 2020-02-07
Csharp5 standard

chamukallem-msft:
PR for basic-concepts.md file

### DIFF
--- a/spec/arrays.md
+++ b/spec/arrays.md
@@ -34,7 +34,7 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 ### The `System.Array` type
 
-The type ``System.Array`` is the abstract base type of all array types. An implicit reference conversion (§11.2.7) exists from any array type to `System.Array` and to any interface type implemented by `System.Array`. An explicit reference conversion (§11.3.5) exists from `System.Array` and any interface type implemented by `System.Array` to any array type. `System.Array` is not itself an *array-type*. Rather, it is a *class-type* from which all *array-type*s are derived.
+The type `System.Array` is the abstract base type of all array types. An implicit reference conversion (§11.2.7) exists from any array type to `System.Array` and to any interface type implemented by `System.Array`. An explicit reference conversion (§11.3.5) exists from `System.Array` and any interface type implemented by `System.Array` to any array type. `System.Array` is not itself an *array-type*. Rather, it is a *class-type* from which all *array-type*s are derived.
 
 At run-time, a value of type `System.Array` can be `null` or a reference to an instance of any array type.
 
@@ -118,7 +118,7 @@ class Test
     }
 }
 ```
-The assignment to `array[i]` in the `Fill` method implicitly includes a run-time check, which ensures that `value` is either a `null` reference or a reference to an object of a type that is compatible with the actual element type of `array`. In `Main`, the first two invocations of `Fill` succeed, but the third invocation causes a `System.Array`TypeMismatchException to be thrown upon executing the first assignment to `array[i]`. The exception occurs because a boxed `int` cannot be stored in a `string` array. *end example*]
+The assignment to `array[i]` in the `Fill` method implicitly includes a run-time check, which ensures that `value` is either a `null` reference or a reference to an object of a type that is compatible with the actual element type of `array`. In `Main`, the first two invocations of `Fill` succeed, but the third invocation causes a `System.Array` TypeMismatchException to be thrown upon executing the first assignment to `array[i]`. The exception occurs because a boxed `int` cannot be stored in a `string` array. *end example*]
 
 Array covariance specifically does not extend to arrays of *value-type*s. For example, no conversion exists that permits an `int[]` to be treated as an `object[]`.
 
--- a/spec/arrays.md
+++ b/spec/arrays.md
@@ -40,7 +40,7 @@ At run-time, a value of type `System.Array` can be `null` or a reference to an i
 
 ### Arrays and the generic collection interfaces
 
-A single-dimensional array `T[]` implements the interface System.Collections.Generic.`IList<T>` (`IList<T>` for short) and its base interfaces. Accordingly, there is an implicit conversion from `T[]` to `IList<T>` and its base interfaces. In addition, if there is an implicit reference conversion from S to T then `S[]` implements `IList<T>` and there is an implicit reference conversion from `S[]` to `IList<T>` and its base interfaces (§11.2.7). If there is an explicit reference conversion from `S` to `T` then there is an explicit reference conversion from `S[]` to `IList<T>` and its base interfaces (§11.3.5).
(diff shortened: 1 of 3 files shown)


spec/arrays.md
```diff
@@ -18,196 +18,181 @@ An array type is written as a *non-array-type* followed by one or more *rank-spe
...
+A single-dimensional array `T[]` implements the interface System.Collections.Generic.`IList<T>` (`IList<T>` for short) and its base interfaces. Accordingly, there is an implicit conversion from `T[]` to `IList<T>` and its base interfaces. In addition, if there is an implicit reference conversion from S to T then `S[]` implements `IList<T>` and there is an implicit reference conversion from `S[]` to `IList<T>` and its base interfaces (§11.2.7). If there is an explicit reference conversion from `S` to `T` then there is an explicit reference conversion from `S[]` to `IList<T>` and its base interfaces (§11.3.5).
```
BillWagner: a couple extra code-fences:

```suggestion
A single-dimensional array `T[]` implements the interface System.Collections.Generic.`IList<T>` (`IList<T>` for short) and its base interfaces. Accordingly, there is an implicit conversion from `T[]` to `IList<T>` and its base interfaces. In addition, if there is an implicit reference conversion from `S` to `T` then `S[]` implements `IList<T>` and there is an implicit reference conversion from `S[]` to `IList<T>` and its base interfaces (§11.2.7). If there is an explicit reference conversion from `S` to `T` then there is an explicit reference conversion from `S[]` to `IList<T>` and its base interfaces (§11.3.5).
```

spec/arrays.md
```diff
@@ -18,196 +18,181 @@ An array type is written as a *non-array-type* followed by one or more *rank-spe
...
-
-a\[0\] = 0; a\[1\] = 2; a\[2\] = 4; a\[3\] = 6; a\[4\] = 8;
-
-*end example*\]
-
-For a multi-dimensional array, the array initializer shall have as many levels of nesting as there are dimensions in the array. The outermost nesting level corresponds to the leftmost dimension and the innermost nesting level corresponds to the rightmost dimension. The length of each dimension of the array is determined by the number of elements at the corresponding nesting level in the array initializer. For each nested array initializer, the number of elements shall be the same as the other array initializers at the same level. \[*Example*: The example:
-
-int\[,\] b = {{0, 1}, {2, 3}, {4, 5}, {6, 7}, {8, 9}};
-
+```csharp
```
BillWagner: This block should be ANTLR:

```suggestion
```ANTLR
```

spec/attributes.md
```diff
@@ -2,496 +2,509 @@
...
 
-\[@ExampleAttribute\] // Refers to ExampleAttribute\
+[@ExampleAttribute] // Refers to ExampleAttribute
 class Class4 {}
+```
 
-shows two attribute classes named Example and ExampleAttribute. The attribute \[Example\] is ambiguous, since it could refer to either Example or ExampleAttribute. Using a verbatim identifier allows the exact intent to be specified in such rare cases. The attribute \[ExampleAttribute\] is not ambiguous (although it would be if there was an attribute class named ExampleAttributeAttribute!). If the declaration for class Example is removed, then both attributes refer to the attribute class named ExampleAttribute, as follows:
+shows two attribute classes named `Example` and `ExampleAttribute`. The attribute [Example] is ambiguous, since it could refer to either `Example` or `ExampleAttribute`. Using a verbatim identifier allows the exact intent to be specified in such rare cases. The attribute [ExampleAttribute] is not ambiguous (although it would be if there was an attribute class named `ExampleAttributeAttribute`!). If the declaration for class `Example` is removed, then both attributes refer to the attribute class named `ExampleAttribute`, as follows:
```
BillWagner: Missing code-fence:

```suggestion
shows two attribute classes named `Example` and `ExampleAttribute`. The attribute `[Example]` is ambiguous, since it could refer to either `Example` or `ExampleAttribute`. Using a verbatim identifier allows the exact intent to be specified in such rare cases. The attribute `[ExampleAttribute]` is not ambiguous (although it would be if there was an attribute class named `ExampleAttributeAttribute`!). If the declaration for class `Example` is removed, then both attributes refer to the attribute class named `ExampleAttribute`, as follows:
```

spec/attributes.md
```diff
@@ -505,310 +518,329 @@ Retrieval of an attribute instance involves both compile-time and run-time proce
...
 
-The compilation of an *attribute* with attribute class T, *positional-argument-list *P, *named-argument-list* N, and specified on a program entity E is compiled into an assembly A viathe following steps:
-
--   Follow the compile-time processing steps for compiling an *object-creation-expression* of the form new T(P). These steps either result in a compile-time error, or determine an instance constructor C on T that can be invoked at run-time.
-
--   If C does not have public accessibility, then a compile-time error occurs.
+The compilation of an *attribute* with attribute class `T`, *positional-argument-list * `P`, *named-argument-list* `N`, and specified on a program entity `E` is compiled into an assembly `A` via the following steps:
 
--   For each *named-argument* Arg in N:
+*   Follow the compile-time processing steps for compiling an *object-creation-expression* of the form new `T(P)`. These steps either result in a compile-time error, or determine an instance constructor C on T that can be invoked at run-time.
```
BillWagner: a couple missing code fences:

```suggestion
*   Follow the compile-time processing steps for compiling an *object-creation-expression* of the form new `T(P)`. These steps either result in a compile-time error, or determine an instance constructor `C` on `T` that can be invoked at run-time.
```

spec/basic-concepts.md
```diff
@@ -781,93 +799,88 @@ A *type-name* is a *namespace-or-type-name* that refers to a type. Following res
...
 
--   Otherwise, if the *namespace-or-type-name* appears within the body of the type declaration, and T or any of its base types contain a nested accessible type having name I and K type parameters, then the *namespace-or-type-name* refers to that type constructed with the given type arguments. If there is more than one such type, the type declared within the more derived type is selected. \[*Note*: Non-type members (constants, fields, methods, properties, indexers, operators, instance constructors, finalizers, and static constructors) and type members with a different number of type parameters are ignored when determining the meaning of the *namespace-or-type-name*. *end note*\]
+-   Otherwise, if the *namespace-or-type-name* appears within the body of the type declaration, and `T` or any of its base types contain a nested accessible type having name I and `K` type parameters, then the *namespace-or-type-name* refers to that type constructed with the given type arguments. If there is more than one such type, the type declared within the more derived type is selected.
```
BillWagner: a missing code-fence:

```suggestion
-   Otherwise, if the *namespace-or-type-name* appears within the body of the type declaration, and `T` or any of its base types contain a nested accessible type having name `I` and `K` type parameters, then the *namespace-or-type-name* refers to that type constructed with the given type arguments. If there is more than one such type, the type declared within the more derived type is selected. 
```

BillWagner:
Thanks again @chamukallem-msft 

I'll :shipit: now.

Issue #3008, opened 2019-12-06
Draft spec for local function attributes


RikkiGibson:
Related to #1888

## Attributes

Local function declarations are now permitted to have [attributes](https://github.com/dotnet/csharplang/blob/bb997294540bd02f939d60514634ccaa7abc659c/spec/attributes.md). Parameters and type parameters on local functions are also allowed to have attributes.

Attributes with a specified meaning when applied to a method, its parameters, or its type parameters will have the same meaning when applied to a local function, its parameters, or its type parameters, respectively.

A local function can be made conditional in the same sense as a [conditional method](https://github.com/dotnet/csharplang/blob/98043cdc889303d956d540d7ab3bc4f5044a9d3b/spec/attributes.md#the-conditional-attribute) by decorating it with a `[ConditionalAttribute]`. A conditional local function must also be `static`.

## Extern

The `extern` modifier is now permitted on local functions. This makes the local function external in the same sense as an [external method](https://github.com/dotnet/csharplang/blob/892af9016b3317a8fae12d195014dc38ba51cf16/spec/classes.md#external-methods).

Similarly to an external method, the *local-function-body* of an external local function must be a semicolon. A semicolon *local-function-body* is only permitted on an external local function. 

An external local function must also be `static`.

## Syntax

The [local functions grammar](https://github.com/dotnet/csharplang/blob/bb997294540bd02f939d60514634ccaa7abc659c/proposals/csharp-7.0/local-functions.md#syntax-grammar) is modified as follows:
```
local-function-header
    : attributes? local-function-modifiers? return-type identifier type-parameter-list?
        ( formal-parameter-list? ) type-parameter-constraints-clauses
    ;

local-function-modifiers
    : (async | unsafe | static | extern)*
    ;

local-function-body
    : block
    | arrow-expression-body
    | ';'
    ;
```

svick:
For `extern` methods that use `DllImport`, the method name is usually used by the runtime to figure out which function to import. Since the compiler mangles names for local functions, does this mean that `extern` local functions would be required to explicitly specify `DllImportAttribute.EntryPoint`?

In other words, would the code in https://github.com/dotnet/csharplang/issues/2740 work as is at runtime?

HaloFour:
@svick

`DllImport` is a pseudo-attribute, no?  I'd think that the compiler would write out the metadata with the original name and not the mangled name.  That wouldn't work for custom attributes, though.

RikkiGibson:
[...]

PR #3198, opened 2020-02-13
Add draft spec for local function attributes

RikkiGibson:
Related to dotnet/roslyn#38801
This adds the draft spec from #3008 into the repository.

@gafter please let me know if this should go somewhere else, or if any other action is needed.

### DIFF
--- a/proposals/local-function-attributes.md
+++ b/proposals/local-function-attributes.md
@@ -2,15 +2,15 @@
 
 ## Attributes
 
-Local function declarations are now permitted to have [attributes](https://github.com/dotnet/csharplang/blob/bb997294540bd02f939d60514634ccaa7abc659c/spec/attributes.md). Parameters and type parameters on local functions are also allowed to have attributes.
+Local function declarations are now permitted to have [attributes](../spec/attributes.md). Parameters and type parameters on local functions are also allowed to have attributes.
 
 Attributes with a specified meaning when applied to a method, its parameters, or its type parameters will have the same meaning when applied to a local function, its parameters, or its type parameters, respectively.
 
-A local function can be made conditional in the same sense as a [conditional method](https://github.com/dotnet/csharplang/blob/98043cdc889303d956d540d7ab3bc4f5044a9d3b/spec/attributes.md#the-conditional-attribute) by decorating it with a `[ConditionalAttribute]`. A conditional local function must also be `static`.
+A local function can be made conditional in the same sense as a [conditional method](../spec/attributes.md#the-conditional-attribute) by decorating it with a `[ConditionalAttribute]`. A conditional local function must also be `static`.
 
 ## Extern
 
-The `extern` modifier is now permitted on local functions. This makes the local function external in the same sense as an [external method](https://github.com/dotnet/csharplang/blob/892af9016b3317a8fae12d195014dc38ba51cf16/spec/classes.md#external-methods).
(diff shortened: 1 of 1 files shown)


proposals/local-function-attributes.md
```diff
@@ -0,0 +1,37 @@
+# Attributes on local functions
+
+## Attributes
+
+Local function declarations are now permitted to have [attributes](https://github.com/dotnet/csharplang/blob/bb997294540bd02f939d60514634ccaa7abc659c/spec/attributes.md). Parameters and type parameters on local functions are also allowed to have attributes.
```
RikkiGibson: These links should probably use relative paths instead of linking to the web at a particular commit.

proposals/local-function-attributes.md
```diff
@@ -0,0 +1,37 @@
+# Attributes on local functions
+
+## Attributes
+
+Local function declarations are now permitted to have [attributes](../spec/attributes.md). Parameters and type parameters on local functions are also allowed to have attributes.
+
+Attributes with a specified meaning when applied to a method, its parameters, or its type parameters will have the same meaning when applied to a local function, its parameters, or its type parameters, respectively.
+
+A local function can be made conditional in the same sense as a [conditional method](../spec/attributes.md#the-conditional-attribute) by decorating it with a `[ConditionalAttribute]`. A conditional local function must also be `static`.
```
gafter: >A conditional local function must also be `static` [](start = 184, length = 50)

Not sure the reason for this restriction. Is it to make lowering easier (we just don't have to make frame objects, instead of having to analyze whether frame objects are needed)?

proposals/local-function-attributes.md
gafter: The relative path will be broken when this moves into the `csharp-9.0` directory :(

---
In reply to: [393936460](https://github.com/dotnet/csharplang/pull/3198#discussion_r393936460) [](ancestors = 393936460)

proposals/local-function-attributes.md
RikkiGibson: > Not sure the reason for this restriction. Is it to make lowering easier (we just don't have to make frame objects, instead of having to analyze whether frame objects are needed)?

It came up in design review that allowing unrestricted use of `ConditionalAttribute` could affect definite assignment safety. e.g.

```cs
int x;
local1();
Console.WriteLine(x);

[Conditional("DEBUG")]
void local1()
{
    x = 42;
}
```

We decided to get around it by requiring local functions to be 'static' in order to use the attribute.

> You should mention that its return type must be void.

I'll add language to clarify that all restrictions from the linked spec apply, such as requiring a 'void' return type, the conditional method may not take 'ref' or 'out' parameters, etc.

gafter: Definite assignment already has to deal with calling conditional methods (e.g. to know that any side-effects of the arguments and the call might not happen) and handles that scenario just fine. See AbstractFlowPass.VisitCall, which begins

``` csharp
            // If the method being called is a partial method without a definition, or is a conditional method
            // whose condition is not true, then the call has no effect and it is ignored for the purposes of
            // definite assignment analysis.
            bool callsAreOmitted = node.Method.CallsAreOmitted(node.SyntaxTree);
            TLocalState savedState = default(TLocalState);

            if (callsAreOmitted)
            {
                savedState = this.State.Clone();
                SetUnreachable();
            }
```

Having said that, I don't see any serious problem with the restriction.

Issue #1510, opened 2018-05-11
Discussion: slice patterns


alrz:
This would be an increment on the "list pattern" proposal (https://github.com/dotnet/csharplang/issues/1039)
```cs
static bool IsSymmetric(Span<int> list) {
  switch (list) {
    case []:
    case [_]:
      return true;
    case [var first, ..var others, var last] when first == last:
      return IsSymmetric(others);
    default:
      return false;
  }
}
```
->
```cs
static bool IsSymmetric(Span<int> list) {
  int length = list.Length;
  if (length == 0 || length == 1) {
    return true;     
  }

  int first = list[0];
  int last = list[^1];
  Span<int> others = list[1..^1]; 
  if (first == last) {
    return IsSymmetric(others);
  }

  return false;
}

```

- The `..` expansion should work with any type with a range indexer.
- The resulting type of `..` expansion would be the return type of the indexer, e.g. `Span<int>` here.
- Only a single `..` expansion would be allowed in the pattern.
- As mentioned in #1039,  we may use `..` as an standalone pattern to skip elements.

> ```cs
> if (collection is [.., 99, 100]) 
>     // matches on the last two elements only
> 
> if (collection is [1, 2, ..]) 
>     // matches on the first two elements
> 
> if (collection is [1, .., 100]) 
>     // matches the first and last elements
> ```

ufcpp:
Can the following pattern substitute this proposal to some extent?

```cs
    public static bool Deconstruct<T>(this Span<T> x, out T first, out Range range, out T last)
    {
        first = x[0];
        range = ..;
        last = x[^2];
    }
```

```cs
if (collection is (1, .., 100))
```

alrz:
@ufcpp I'd say that would be a very limited helper, to only match first and last element. Note that the slice pattern could appear anywhere in the list pattern (first, last, or anywhere in between). Consider a list pattern with more that one element to match. You'd need N Deconstruct overloads to make that work, and still since all Deconstruct overloads must differ in parameter counts, you'd be limited to a particular set of overloads.

Issue #1039, opened 2017-10-24
[Discussion] Revisiting pattern matching on collections


DavidArno:
My aim with this discussion is to explore how pattern matching could be supported over any `IEnumerable<T>`. Of course, this immediately raises a problem: how do we avoid re-enumerating the collection as we pattern match? Further, how do we avoid the need for an enumerator, for [as @gafter says, patterns shouldn't generally be executing methods like GetEnumerator/MoveNext](https://github.com/dotnet/roslyn/issues/10631)?

Taking the second concern first, typical applications of pattern matching over a collection involve some sort of matching against the head and then recursively re-applying the patterns to the tail. Leaving this as a recusive implementation leads too easily to a stack overflow. So tail-call optimisation is used to turn it into an iterative implementation. Such an implementation naturally lends itself to using an enumerator. So my gut feeling here is that trying to avoid `GetEnumerator` and `MoveNext` is not practical. Whilst it's highly likely that folk way more clever than me have found a way around this, eg for F#'s pattern matching over linked lists, the rest of this discussion works on the assumption that this isn't really a problem.

Regarding the problem of re-enumerating an `IEnumerable<T>`, the two solutions I have come up with for this is (1) to cache the enumeration as it is read, using a linked list and (2) copy it to an array (or other indexed collection) first. 

The advantages I see to the linked list approach are:
1. The collection only needs to be enumerated as deeply as needed (eg to determine no pattern matches the start of the collection),
2. Each node effectively forms the head of a sub collection, allowing easy support of recursive matching,
3. The original enumeration, and its enumerator, can be stored in the last node. This allows the linked list to be "self contained".

However, the linked list approach has an obvious disadvantage over copying to an array, namely that elements cannot be accessed via an index and must be found be starting at the beginning of the list each time. This would likely add huge overheads to all but the most trivial of patterns.

For the rest of this discussion, I'll ignore which method is used as the correct way to determine the best solution would be through performance testing, rather than me guessing.

### Basic pattern matching ###
Taking ideas from [Roslyn #10631 (ie, @gafter's link above)](https://github.com/dotnet/roslyn/issues/10631), the simplest form of pattern matching would match the whole collection. I've changed the syntax here as @alrz points out that the proposed property pattern syntax would likely clash with the ideas of using `{}` for those collection patterns. So I've used `[]` here (until someone points out how that will clash too!)

```cs
if (collection is [1, 2, 3]) 
    // true if collection contains exactly the three elements, [1,2,3]

[...]

Issue #3107, opened 2020-01-16
Notes on the theme "Programming With Data"


gafter:
### Programming languages are opinionated

Programming languages are opinionated, whether we like it or not.
Any given programming language makes some things easy, and other things not as easy.
If someone is programming and has to choose between the two, the language effectively encourages them to do things the easy way.
Sometimes the opinions of a language are intentional, but sometimes the opinions are incidental and do not align with the language author's own opinions.
When that happens, that is a language design smell.
When we design and modify a language, we should aim to have its opinions align with our own.

### C# grew up opinionated

C# came preloaded with a set of opinions inherited from its predecessors.
- It is *object-oriented*: operations on a data type should be placed on the data type on which they operate. If an operation varies in behavior from one "kind" of instance to another, it should be `virtual` (*internal dispatch*). The exceptions to this, inherited from non-object-oriented C, are the integral and enum types, for which `switch` permits *external dispatch*.
- C# is *imperative*: algorithms are expressed as a sequence of steps described by *statements*.
- C# is *mutation-based*: objects are mutable and the purpose of computation is to mutate data.
- *Concurrency* is expressed by protecting mutable data with monitors (`lock`) to provide mutual exclusion during mutation.

### Opinions have changed over time

As we've learned more about good software engineering practices, we've reconsidered these opinions and modified the language and its libraries accordingly.
We added *extension methods* to permit a non-`virtual` operation to be declared outside the type on which it operates while being used by the client as if it were a member.
We added a number of enhancements to make more declarative (and less imperative) code possible, such as expression-bodied methods, Linq, the "elvis" `?.` operator, and local functions.
We added libraries that support fork-join concurrency and asynchronous computation pipelines.
We added auto-properties to make it easier to declare APIs with immutable members.
In all of these, we moved the language's opinion not by making the existing things harder, but by adding yet easier ways of doing things that align with our opinions of what is "right".

### The opinionated purpose of Programming With Data features

This cluster of features aims to move C#'s opinion:
- From *internal dispatch* to *external dispatch*
- From *mutation based* to patterns that work well with immutable data.
- From *imperative* (statement oriented) to *declarative* (e.g. expression oriented)

One shortcoming of C#'s opinion is that *internal dispatch* is frequently inappropriate for programming-in-the-large (i.e. for data types that span component boundaries).
This is particularly true when the data type cannot anticipate all of the operations that clients will want to perform on it, or when the data type is described by a set of related types.
[...]

PR #3245, opened 2020-03-02
Add list patterns proposal

alrz:
Tracking issue: https://github.com/dotnet/csharplang/issues/3435
Discussion: https://github.com/dotnet/csharplang/issues/1510, https://github.com/dotnet/csharplang/issues/1039, https://github.com/dotnet/csharplang/issues/898
Prototype: https://github.com/dotnet/roslyn/compare/main...alrz:list-patterns
Proposal: [Rendered](https://github.com/alrz/csharplang/blob/list-patterns/proposals/list-patterns.md)

### DIFF
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -1,3 +1,4 @@
+
 # List patterns
 
 ## Summary
@@ -44,25 +45,37 @@ There are two new additions to the *recursive_pattern* syntax as well as a *slic
 - The *list_pattern_clause* is used to match elements and the *length_pattern_clause* is used to match the length.
 - A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
 
-Due to the ambiguity with *property_pattern_clause*, the *list_pattern_clause* cannot be empty and a *length_pattern_clause* should be used instead.
+Notes:
+
+- Due to the ambiguity with *property_pattern_clause*, the *list_pattern_clause* cannot be empty and a *length_pattern_clause* should be used instead to match a list with the length of zero, e.g. `[0]`. 
+- The *length_pattern_clause* must be in agreement with the inferred length from the pattern (if any), e.g. `[0] {1}` is an error.
+	- However, `[1] {}` is **not** an error due to the length mismatch, rather, `{}` would be always parsed as an empty *property_pattern_clause*. We may want to add a warning for it so it would not be confused that way.
+- If the *type* is an *array_type*, the *length_pattern_clause* is disambiguated so that `int[] [0]` would match an empty integer array.
+- All other combinations are valid, for instance `T (p0, p1) [p2] { name: p3 } v` or `T (p0, p1) [p2] { p3 } v` where each clause can be omitted.
+
+> **Open question**: Should we support all these combinations?
+
+
 
 #### Pattern compatibility
 
-A *length_pattern_clause* is compatible with any type that conforms to the following rules:
-
-1. Has an accessible property getter that returns an `int` and has the name `Length` or `Count`
+A *length_pattern_clause* is compatible with any type that is *countable*, i.e. has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.
 
 A *list_pattern_clause* is compatible with any type that conforms to the following rules:
 
-1. Is compatible with the length pattern
+1. Is compatible with the *length_pattern_clause*
 2. Has an accessible indexer with a single `int` parameter
+
+ > **Open question**: Should we support `this[Index]` indexers? If so, which one is preferred if `this[int]` is also present?
 
 A *slice_pattern* is compatible with any type that conforms to the following rules:
 
-1. Is compatible with the list pattern
+1. Is compatible with the *length_pattern_clause*
 2. Has an accessible `Slice` method that takes two `int` parameters (required only if a subpattern is specified)
 
-This set of rules is already specified as the ***range indexer pattern*** and required for range indexers.
+ > **Open question**: Should we support `this[Range]` indexers? If so, which one is preferred if `this[int]` is also present?
+
+This set of rules is already specified as the [***range indexer pattern***](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md#implicit-index-support) and required for range indexers.
 
 #### Subsumption checking
 
@@ -84,6 +97,8 @@ expr.Length    is >= 2
 && expr[1..^1] is var s
 && expr[^1]    is 3
 ```
+The *input type* for the *slice_pattern* is the return type of the underlying `Slice` method with two exceptions: For `string` and arrays, `string.Substring` and `RuntimeHelpers.GetSubArray` will be used, respectively.
+
 Note: the lowering is presented in the pattern form here to show how subsumption checking works, for example, the following code produces an error because both patterns yield the same DAG:
 
 ```cs
@@ -96,7 +111,7 @@ case {_, 1, ..}: // expr.Length is >= 2 && expr[1] is 1
 case {.., 1, _}: // expr.Length is >= 2 && expr[^2] is 1
 ```
 
-Note: The pattern `{..}` lowers to `expr.Length >= 0` so it would not be considered as a catch-all.
+> **Open question**: The pattern `{..}` lowers to `expr.Length >= 0` so it would not be considered as a catch-all. Should we omit such test (assuming `Length` is always non-negative)?
 
 ### Additional types
 
@@ -105,7 +120,8 @@ Beyond the pattern-based mechanism outlined above, there are an additional two s
 - **Multi-dimensional arrays**: All nested list patterns must agree to a length range.
 - **Foreach-able types**: This includes pattern-based and extension `GetEnumerator`.
 
-A slice subpattern (i.e. the pattern followed by `..` in a *slice_pattern*) is disallowed for either of the above.
+A slice subpattern (i.e. the pattern following `..` in a *slice_pattern*) is disallowed for either of the above.
+
 
 ## Unresolved questions
 
@@ -114,9 +130,3 @@ All multi-dimensional arrays can be non-zero-based. We can either:
 1. Add a runtime helper to check if the array is zero-based across all dimensions.
 2. Call `GetLowerBound` and add it to each indexer access to pass the *correct* index.
 3. Assume all arrays are zero-based since that's the default for arrays created by `new` expressions.
-
-## Design meetings
-
-https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-12-14.md
-https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-12-16.md
-https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-02-03.md
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -1,3 +1,4 @@
+
 # List patterns
 
 ## Summary
@@ -9,8 +10,20 @@ Lets you to match an array or a list with a sequence of patterns e.g. `array is 
 The pattern syntax is modified as follow:
 
 ```antlr
-recursive_pattern
-  : type? positional_pattern_clause? length_pattern_clause? property_or_list_pattern_clause? simple_designation?
+positional_pattern
+  : type? positional_pattern_clause length_pattern_clause? property_or_list_pattern_clause? simple_designation?
+  ;
+
+length_pattern
+  : type? length_pattern_clause property_or_list_pattern_clause? simple_designation?
+  ;
+
+list_pattern
+  : type? list_pattern_clause simple_designation?
+  ;
+
+property_pattern
+  : type? property_pattern_clause simple_designation?
   ;
 
 length_pattern_clause
@@ -35,20 +48,21 @@ slice_pattern
   ;
 
 primary_pattern
-  : recursive_pattern
+  : list_pattern
+  | length_pattern
   | slice_pattern
   | // all of the pattern forms previously defined
   ;
 ```
-There are two new additions to the *recursive_pattern* syntax as well as a *slice_pattern*:
+There are three new patterns:
 
-- The *list_pattern_clause* is used to match elements and the *length_pattern_clause* is used to match the length.
+- The *list_pattern* is used to match elements and the *length_pattern* is used to match the length.
 - A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
 
 Notes:
 
-- Due to the ambiguity with *property_pattern_clause*, the *list_pattern_clause* cannot be empty and a *length_pattern_clause* should be used instead to match a list with the length of zero, e.g. `[0]`. 
-- The *length_pattern_clause* must be in agreement with the inferred length from the pattern (if any), e.g. `[0] {1}` is an error.
+- Due to the ambiguity with *property_pattern*, a *list_pattern* cannot be empty and a *length_pattern* should be used instead to match a list with the length of zero, e.g. `[0]`. 
+- The *length_pattern_clause* must be in agreement with the inferred length from the *list_pattern_clause* (if any), e.g. `[0] {1}` is an error.
 	- However, `[1] {}` is **not** an error due to the length mismatch, rather, `{}` would be always parsed as an empty *property_pattern_clause*. We may want to add a warning for it so it would not be confused that way.
 - If the *type* is an *array_type*, the *length_pattern_clause* is disambiguated so that `int[] [0]` would match an empty integer array.
 - All other combinations are valid, for instance `T (p0, p1) [p2] { name: p3 } v` or `T (p0, p1) [p2] { p3 } v` where each clause can be omitted.
@@ -57,18 +71,18 @@ Notes:
 
 #### Pattern compatibility
 
-A *length_pattern_clause* is compatible with any type that is *countable*, i.e. has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.
+A *length_pattern* is compatible with any type that is *countable*, i.e. has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.
 
-A *list_pattern_clause* is compatible with any type that conforms to the following rules:
+A *list_pattern* is compatible with any type that conforms to the following rules:
 
-1. Is compatible with the *length_pattern_clause*
+1. Is compatible with the *length_pattern*
 2. Has an accessible indexer with a single `int` parameter
 
  > **Open question**: Should we support `this[Index]` indexers? If so, which one is preferred if `this[int]` is also present?
 
 A *slice_pattern* is compatible with any type that conforms to the following rules:
 
-1. Is compatible with the *length_pattern_clause*
+1. Is compatible with the *length_pattern*
 2. Has an accessible `Slice` method that takes two `int` parameters (required only if a subpattern is specified)
 
  > **Open question**: Should we support `this[Range]` indexers? If so, which one is preferred if `Slice(int, int)` is also present?
@@ -77,7 +91,7 @@ This set of rules is already specified as the [***range indexer pattern***](http
 
 #### Subsumption checking
 
-Subsumption checking works just like recursive patterns with `ITuple` - corresponding subpatterns are matched by position plus an additional node for testing length.
+Subsumption checking works just like positional patterns with `ITuple` - corresponding subpatterns are matched by position plus an additional node for testing length.
 
 #### Lowering
 
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -13,22 +13,6 @@ positional_pattern
   : type? positional_pattern_clause length_pattern_clause? property_or_list_pattern_clause? simple_designation?
   ;
 
-length_pattern
-  : type? length_pattern_clause property_or_list_pattern_clause? simple_designation?
-  ;
-
-list_pattern
-  : type? list_pattern_clause simple_designation?
-  ;
-
-property_pattern
-  : type? property_pattern_clause simple_designation?
-  ;
-
-length_pattern_clause
-  : '[' pattern ']'
-  ;
-
 property_or_list_pattern_clause
   : list_pattern_clause
   | property_pattern_clause
@@ -40,6 +24,22 @@ property_pattern_clause
 
 list_pattern_clause
   : '{' pattern (',' pattern)* ','? '}'
+  ;
+
+length_pattern_clause
+  : '[' pattern ']'
+  ;
+
+length_pattern
+  : type? length_pattern_clause property_or_list_pattern_clause? simple_designation?
+  ;
+
+list_pattern
+  : type? list_pattern_clause simple_designation?
+  ;
+
+property_pattern
+  : type? property_pattern_clause simple_designation?
   ;
 
 slice_pattern
@@ -55,7 +55,8 @@ primary_pattern
 ```
 There are three new patterns:
 
-- The *list_pattern* is used to match elements and the *length_pattern* is used to match the length.
+- The *list_pattern* is used to match elements.
+- The *length_pattern* is used to match the length.
 - A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
 
 > **Open question**: Should we accept a general *pattern* following `..` in a *slice_pattern*?
@@ -101,7 +102,7 @@ case {.., 1, _}: // expr.Length is >= 2 && expr[^2] is 1
 The order in which subpatterns are matched at runtime is unspecified, and a failed match may not attempt to match all subpatterns.
 
 > **Open question**: The pattern `{..}` tests for  `expr.Length >= 0`. Should we omit such test (assuming `Length` is always non-negative)?
-> 
+ 
 #### Lowering
 
 A pattern of the form `expr is {1, 2, 3}` is equivalent to the following code (if compatible via implicit `Index` support):
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -39,7 +39,7 @@ primary_pattern
   | slice_pattern
   ;
 ```
-There are two new additions to the *recursive_pattern* syntax as well as a *slice pattern*:
+There are two new additions to the *recursive_pattern* syntax as well as a *slice_pattern*:
 
 - The *list_pattern_clause* is used to match elements and the *length_pattern_clause* is used to match the length.
 - A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -11,53 +11,53 @@ The pattern syntax is modified as follow:
 ```antlr
 recursive_pattern
   : type? positional_pattern_clause? length_pattern_clause? property_or_list_pattern_clause? simple_designation?
+  ;
 
 length_pattern_clause
   : '[' pattern ']'
+  ;
 
 property_or_list_pattern_clause
   : list_pattern_clause
   | property_pattern_clause
+  ;
 
 property_pattern_clause
   : '{' (subpattern (',' subpattern)* ','?)? '}'
-  
+  ;
+
 list_pattern_clause
   : '{' pattern (',' pattern)* ','? '}'
+  ;
 
 slice_pattern
   : '..' negated_pattern?
-	
+  ;
+
 primary_pattern
   : recursive_pattern
   | slice_pattern
-
+  ;
 ```
-There are two new additions to the *recursive pattern* syntax as well as a *slice pattern*: 
+There are two new additions to the *recursive_pattern* syntax as well as a *slice pattern*:
 
-- The *list pattern clause* is used to match elements and the *length pattern clause* is used to match the length.
-- A _slice pattern_ is only permitted once and only directly in a *list pattern* and discards _**zero or more**_ elements.
+- The *list_pattern_clause* is used to match elements and the *length_pattern_clause* is used to match the length.
+- A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
 
-
-Due to the ambiguity with *property pattern clause*, the *list pattern clause* cannot be empty and a *length pattern* should be used. 
-
-
-
-
-
+Due to the ambiguity with *property_pattern_clause*, the *list_pattern_clause* cannot be empty and a *length_pattern_clause* should be used instead.
 
 #### Pattern compatibility
 
-A *length pattern* is compatible with any type that conforms to the following rules:
+A *length_pattern_clause* is compatible with any type that conforms to the following rules:
 
 1. Has an accessible property getter that returns an `int` and has the name `Length` or `Count`
 
-A *list pattern* is compatible with any type that conforms to the following rules:
+A *list_pattern_clause* is compatible with any type that conforms to the following rules:
 
 1. Is compatible with the length pattern
 2. Has an accessible indexer with a single `int` parameter
 
-A *slice pattern* is compatible with any type that conforms to the following rules:
+A *slice_pattern* is compatible with any type that conforms to the following rules:
 
 1. Is compatible with the list pattern
 2. Has an accessible `Slice` method that takes two `int` parameters (required only if a subpattern is specified)
@@ -77,7 +77,7 @@ expr.Length is 3
 && expr[1] is 2
 && expr[2] is 3
 ```
-A *slice pattern* acts like a proper discard i.e. no tests will be emitted for such pattern, rather it only affects other nodes, namely the length and indexer. For instance, a pattern of the form `expr is {1, .. var s, 3}`  is equivalent to the following code: 
+A *slice_pattern* acts like a proper discard i.e. no tests will be emitted for such pattern, rather it only affects other nodes, namely the length and indexer. For instance, a pattern of the form `expr is {1, .. var s, 3}`  is equivalent to the following code:
 ```cs
 expr.Length    is >= 2
 && expr[0]     is 1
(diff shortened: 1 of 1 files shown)


gafter:
Really nice work!

List initializers (e.g. in an object creation expression) look like `{ 1, 2, 3 }`, so by *A design principle for Programming With Data* in https://github.com/dotnet/csharplang/issues/3107 the pattern form "should" use curly braces instead of square brackets.  Do you see issues with that?

proposals/list-patterns.md
```diff
@@ -0,0 +1,93 @@
...
+	: list_pattern
+list_pattern
+	: '[' (pattern (',' pattern)* ','?)? ']'
+```
+
+**Pattern compatibility:** A *list_pattern* is compatible with any type that conforms to the ***range indexer pattern***:
+
+1. Has an accessible property getter that returns an `int` and has the name `Length` or `Count`
+2. Has an accessible indexer with a single `int` parameter
+3. Has an accessible `Slice` method that takes two `int` parameters (for slice patterns)
```
yaakov-h: No love for System.Index or System.Range?

proposals/list-patterns.md
```diff
@@ -0,0 +1,93 @@
...
+	: `..`
+```
+
+A *slice_pattern* is only permitted once and only directly in a *list_pattern* and discards ***zero or more*** elements. Note that it's possible to use a *slice_pattern* in a nested *list_pattern* e.g. `[.., [.., 1]]` will match `new int[][]{new[]{1}}`.
+
+A *slice_pattern* acts like a proper discard i.e. no tests will be emitted for such pattern, rather it only affects other nodes, namely the length and indexer. For instance, a pattern of the form `expr is [1, .., 3]`  is equivalent to the following code: 
+```cs
+expr.Length is >= 2
+&& expr[0]  is 1
+&& expr[^1] is 3
```
yaakov-h: Does this represent System.Index, or are you just using this to express `&& expr[expr.Length - 1] is 3`?

alrz:
>the pattern form "should" use curly braces instead of square brackets. Do you see issues with that?

Unfortunately, with braces, empty list pattern would be ambigious with property pattern clause.

proposals/list-patterns.md
alrz: No Index nor Range would be created except for slice pattern usage on arrays.

proposals/list-patterns.md
```diff
@@ -0,0 +1,93 @@
...
+expr.Length    is >= 2
+&& expr[0]     is 1
+&& expr[1..^1] is var s
+&& expr[^1]    is 3
+```
+
+
+## Questions
+
+- Should we support multi-dimensional (including non-zero-based) arrays?
```
alrz: IMO this would not worth if it includes non-zero-based arrays (it wouldn't be efficient). Although, given a runtime helper to check if the array is zero-based over all ranks (which is the default for C# multi-dimensional arrays), we can accept it as a compatible type.

proposals/list-patterns.md
```diff
@@ -0,0 +1,95 @@
...
+2. Has an accessible indexer with a single `int` parameter
+3. Has an accessible `Slice` method that takes two `int` parameters (for slice patterns)
+
+This rule includes `T[]`,  `string`,  `Span<T>`, `ImmutableArray<T>` and more.
+
+**Subsumption checking:**  This construct will be built entirely on top of the existing DAG nodes. The subsumption is checked just like recursive patterns with `ITuple` - corresponding subpatterns are matched by position plus an additional node for testing length.
+
+**Lowering:** A pattern of the form `expr is [1, 2, 3]` is equivalent to the following code:
+```cs
+expr.Length is 3
```
ILMTitan, line 33: Does this need to include a null check?

```cs
expr is not null
&& expr.Lenght is 3
&& expr[0] is 1
&& expr[1] is 2
&& expr[2] is 3
```

Alternatively, could it use a property pattern

```cs
expr is { Lenght: 3 }
&& expr[0] is 1
&& expr[1] is 2
&& expr[2] is 3
```

CyrusNajmabadi: yes.  :)

alrz:
Talking with folks at gitter, IEnumerable support can def be added as a special case (alongside multidimensional arrays), but we might want to limit possible pattern forms to keep things simple and predictable.

YairHalberstadt:
> IEnumerable support can def be added

I would suggest allowing pattern based enumeration as well, same as foreach.

alrz:
Sorry I removed my comment to add further info, here's how I imagine it could work:
```cs

switch (expr)
{
    // MoveNext() would be an evaluation node followed by a value test

    case [0]:    // DAG: e.MoveNext() && e.Current is 0 && !e.MoveNext()
    case [0, 1]: // DAG: e.MoveNext() && e.Current is 0 && e.MoveNext() && e.Current is 1 && !e.MoveNext()
}

using (var e = expr.GetEnumeable())
{
    if (e.MoveNext())
    {
        var t0 = e.Current;
        if (t0 is 0)
        {
            if (!e.MoveNext())
            {
                goto case0;
            }
            else
            {
                var t1 = e.Current;
                if (t1 is 1)
                {
                    if (!e.MoveNext())
                    {
                        goto case1;
                    }
                }
            }
        }
    }

    goto @default;
}

```
We probably can use `foreach` lowering for some pieces so struct enumerators should be possible. 

Lowering of that `using` statement gets a little bit tricky but I think it's doable.

Note this strategy does cover `..` as the last element, beyond that it gets easily complicated.

alrz:
Preliminary support for multi-dimensional arrays and IEnumerable is added to the darf impl.

(Note for IEnumerable support a big chunk of loop binder members needed to move around.)

proposals/list-patterns.md
```diff
@@ -87,11 +87,84 @@ expr.Length    is >= 2
...
+- For an array of rank N, only N level of nested list-patterns are accepted.
+- Except for that last level, all subpatterns must be either a slice pattern without a subpattern (`..`) or a list-pattern.
+- For slice patterns, the usual rules apply - only permitted once and only directly inside the pattern.
+- All nested list-patterns must be of an exact or minimum size. Given X = the minimum or exact required length so far and Y = the new length from current nested list pattern, the expected size is calculated as follow:
+  ```
+  AtLeast(X) + AtLeast(Y) = AtLeast(Max(X, Y))
+  Exactly(X) + Exactly(Y) = Exactly(X) only if X==Y
+  Exactly(X) + AtLeast(Y) = Exactly(X) only if X>=Y
+  ```
+  Note: The presense of a slice pattern implies a minimum required length.
```
jnm2, line 121:
```suggestion
  Note: The presence of a slice pattern implies a minimum required length.
```

alrz: Blunder!

alrz:
Thia proposal is fairly outdated and needs a complete rewrite. open issues are tracked in the prototype PR.

alrz:
Updated based on the latest LDM decisions. @333fred @jcouv please review.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+Note: The pattern `{..}` lowers to `expr.Length >= 0` so it would not be considered as a catch-all.
+
+### Additional types
+
+Beyond the pattern-based mechanism outlined above, there are an additional two set of types that can be covered as a special case.
+
+- **Multi-dimensional arrays**: All nested list patterns must agree to a length range. 
+- **Foreach-able types**: This includes pattern-based and extension `GetEnumerator`.
```
alrz, line 106: I did not go through the details on these. I can update later when I work out the specifics.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+Lets you to match an array or a list with a sequence of patterns e.g. `array is {1, 2, 3}` will match an integer array of the length three with 1, 2, 3 as its elements, respectively.
+
+## Detailed Design
+
+The pattern syntax is modified as follow:
+
+```antlr
+recursive_pattern
+  : type? positional_pattern_clause? length_pattern_clause? property_or_list_pattern_clause? simple_designation?
```
333fred, line 13: Missing trailing `;`'s in all the sections here.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+```
+There are two new additions to the *recursive pattern* syntax as well as a *slice pattern*: 
+
+- The *list pattern clause* is used to match elements and the *length pattern clause* is used to match the length.
+- A _slice pattern_ is only permitted once and only directly in a *list pattern* and discards _**zero or more**_ elements.
+
+
+Due to the ambiguity with *property pattern clause*, the *list pattern clause* cannot be empty and a *length pattern* should be used. 
+
```
333fred, line 43: Lots of new lines here?

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+  : '..' negated_pattern?
+	
+primary_pattern
+  : recursive_pattern
+  | slice_pattern
+
+```
+There are two new additions to the *recursive pattern* syntax as well as a *slice pattern*: 
+
+- The *list pattern clause* is used to match elements and the *length pattern clause* is used to match the length.
```
333fred, line 38:
```suggestion
- The *list\_pattern\_clause* is used to match elements and the *length\_pattern\_clause* is used to match the length.
```

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+- A _slice pattern_ is only permitted once and only directly in a *list pattern* and discards _**zero or more**_ elements.
```
333fred, line 39:
```suggestion
- A _slice\_pattern_ is only permitted once and only directly in a *list\_pattern* and discards _**zero or more**_ elements.
```

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+Due to the ambiguity with *property pattern clause*, the *list pattern clause* cannot be empty and a *length pattern* should be used.
```
333fred, line 42:
```suggestion
Due to the ambiguity with *property pattern clause*, the *list\_pattern\_clause* cannot be empty and a *length\_pattern* should be used. 
```

333fred: Similar changes are needed throughout the rest. Not going to go through and suggest for them all.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+Note: The pattern `{..}` lowers to `expr.Length >= 0` so it would not be considered as a catch-all.
+
+### Additional types
+
+Beyond the pattern-based mechanism outlined above, there are an additional two set of types that can be covered as a special case.
+
+- **Multi-dimensional arrays**: All nested list patterns must agree to a length range. 
+- **Foreach-able types**: This includes pattern-based and extension `GetEnumerator`.
+
+A slice subpattern is disallowed for either of the above.
```
333fred, line 108: I'm not sure this is true for foreachable types. A slice pattern with a designator is certainly disallowed, but we talked about allowing regular slice patterns as discards.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+list_pattern_clause
+  : '{' pattern (',' pattern)* ','? '}'
+  ;
+
+slice_pattern
+  : '..' negated_pattern?
+  ;
+
+primary_pattern
```
jcouv, line 37: What's the benefit of overloading `recursive_pattern`?
What if we instead `primary_pattern` to have a choice of `list_pattern` defined as  `length_pattern_clause? list_pattern_clause`?

The `positional_pattern_clause` and `simple_designation` portions would not be allowed in a list pattern anyways, right?

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+- The *list_pattern_clause* is used to match elements and the *length_pattern_clause* is used to match the length.
+- A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
+
+Due to the ambiguity with *property_pattern_clause*, the *list_pattern_clause* cannot be empty and a *length_pattern_clause* should be used instead.
+
+#### Pattern compatibility
+
+A *length_pattern_clause* is compatible with any type that conforms to the following rules:
+
+1. Has an accessible property getter that returns an `int` and has the name `Length` or `Count`
```
jcouv, line 53: nit: we'll want to define which property getter is picked first.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+A *list_pattern_clause* is compatible with any type that conforms to the following rules:
+
+1. Is compatible with the length pattern
+2. Has an accessible indexer with a single `int` parameter
```
jcouv, line 58: For other recent pattern-based features, we've used a different formulation, instead on relying on equivalent invocation code. This means that an indexer with an optional or `params` parameter would be found acceptable.

Tagging @chsienki to confirm whether I'm remembering right. I wonder if we have any document or example for this guideline.

jcouv: It looks like we may have taken the old approach for range...
Let's add a link to the ranges spec here.
https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md#implicit-index-support

chsienki: We have a mixed record on this, as earlier language features deviate in different ways. 

I think in general though there is a desire for new language features to match the same definition of 'pattern' and we have machinery in the compiler now to do so: essentially if it's invocable as a method at the call site then it should satisfy the pattern.

I actually wrote a doc on this, but it never got out of PR. I should probably revive it so we have something to point to going forward.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+A *slice_pattern* is compatible with any type that conforms to the following rules:
+
+1. Is compatible with the list pattern
```
jcouv, line 62: Do you mean "length" pattern here?
Let's try to re-use the terminology from the ranges spec: "The type is Countable".

What if the type has an `this[Range]` indexer?

proposals/list-patterns.md
jcouv: What if the type has an `this[Index]` indexer?

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+2. Has an accessible `Slice` method that takes two `int` parameters (required only if a subpattern is specified)
+
+This set of rules is already specified as the ***range indexer pattern*** and required for range indexers.
+
+#### Subsumption checking
+
+Subsumption checking works just like recursive patterns with `ITuple` - corresponding subpatterns are matched by position plus an additional node for testing length.
+
+#### Lowering
```
jcouv, line 71: The ranges spec called out special cases for `string` and array types. Are we going to do something special of list and slice patterns on those types as well?

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+A slice subpattern (i.e. the pattern followed by `..` in a *slice_pattern*) is disallowed for either of the above.
+
+## Unresolved questions
+
+All multi-dimensional arrays can be non-zero-based. We can either:
+
+1. Add a runtime helper to check if the array is zero-based across all dimensions.
+2. Call `GetLowerBound` and add it to each indexer access to pass the *correct* index.
+3. Assume all arrays are zero-based since that's the default for arrays created by `new` expressions.
```
jcouv, line 116: We should investigate and spec how list and slice patterns would work on `IEnumerable`. Let's leave that as an open issue for now.
https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-02-03.md#list-patterns-on-ienumerable

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Design meetings
```
jcouv, line 118: I think it's okay to list design notes here if we'll update the list. Otherwise, I'd suggest leaving those in the "proposal" issue.

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+property_pattern_clause
+  : '{' (subpattern (',' subpattern)* ','?)? '}'
+  ;
+
+list_pattern_clause
+  : '{' pattern (',' pattern)* ','? '}'
+  ;
+
+slice_pattern
+  : '..' negated_pattern?
```
jcouv, line 34: Negated_pattern allows other patterns. Is that what we want?
What does `..Type (..., ...) { Property: ...}` mean? Would it be allowed?

```
negated_pattern
    : 'not' negated_pattern
    | primary_pattern
```

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+A *length_pattern_clause* is compatible with any type that conforms to the following rules:
+
+1. Has an accessible property getter that returns an `int` and has the name `Length` or `Count`
+
+A *list_pattern_clause* is compatible with any type that conforms to the following rules:
+
+1. Is compatible with the length pattern
+2. Has an accessible indexer with a single `int` parameter
+
+A *slice_pattern* is compatible with any type that conforms to the following rules:
```
jcouv, line 60: What is the type of a slice pattern? (what's the `var` in `..var x`?)

proposals/list-patterns.md
```diff
@@ -0,0 +1,122 @@
...
+  : '{' pattern (',' pattern)* ','? '}'
+  ;
+
+slice_pattern
+  : '..' negated_pattern?
+  ;
+
+primary_pattern
+  : recursive_pattern
+  | slice_pattern
```
jcouv, line 39: Let's put an ellipsis here (`: // all of the patterns forms previously defined`).

This implies that slice_pattern can be used like this `x is ..var y`. Would that be allowed? What does it mean?

proposals/list-patterns.md
alrz: >The positional_pattern_clause and simple_designation portions would not be allowed in a list pattern anyways, right?

Actually that's the case. LDT wants this to be a part of recursive patterns mainly to allow a type e.g. `IEnumerable {4}`, but positionals and designations would be allowed as well. Same for length patterns. The only limitation is that list and property clauses cannot appear side by side, otherwise all combinations are valid. 

Note that `[1] {}` would NOT be an error because the length doesn't match, rather `{}` is still (and always) a property pattern clause. Fred suggested we warn here to be not confused that way.

proposals/list-patterns.md
alrz: Length, same as for ranges. I'll add a note.

proposals/list-patterns.md
alrz: >It looks like we may have taken the old approach for range

That's right. I did mention that this is exactly what specified for ranges a few lines below. This was part of the original proposal but afaik never called out in the LDM (presumably ok'd). I'll add a link.

proposals/list-patterns.md
alrz: `this[Range]` is not part of range spec. It just "accidentally" supported via overload resolution and `Range` natural type for `..` expressions.

If we want to support it, we should explicitly add a section.

proposals/list-patterns.md
alrz: Lowering for arrays and string would be different just like for ranges (using GetSubArray and Substring). Should we call that out?

proposals/list-patterns.md
jcouv: Ah, thanks. Could we include an example with the type and other parts?
I must say that re-using the recursive pattern still seems strange to me. Why allow combining a positional check with a list check, but not a pattern check with a list check.

By the way, I think the patterns spec was re-writing without `recursive_pattern` as it had all its parts optional. See [here](https://github.com/dotnet/csharplang/blob/a3d67d141501101af9dbb9e8192ca8b4c717e316/proposals/csharp-8.0/patterns.md#patterns) for way the patterns were divided.

I'd propose we do something similar for length and list patterns, which disallows their combination with positional patterns.

proposals/list-patterns.md
alrz: That is an open issue for now. I intended to expand that part when I have more specifics. I'll add a note.

proposals/list-patterns.md
jcouv: Let's confirm, but my assumption is that we'd want to support `this[Range]`. 
The implicit Range indexing was added to make the transition easier (legacy types didn't have Range indexer). That was not called out in the range spec, because it already worked (normal indexing).

proposals/list-patterns.md
jcouv: Yes, please. Let's call out the special lowering cases for `string` and array.

proposals/list-patterns.md
alrz: That's just a var pattern. Currently that's the only sensible pattern here, but I think user-defined positional patterns (https://github.com/dotnet/csharplang/issues/4131) should be useful here in the future.

proposals/list-patterns.md
alrz: Slice pattern is only allowed directly in the list pattern clause. I mentioned that a few lines below.

proposals/list-patterns.md
jcouv: Sorry I didn't follow. How is `..var x` just a var pattern? It has `..` for slicing.

proposals/list-patterns.md
jcouv: Thanks! I'd missed that.

proposals/list-patterns.md
alrz: Given:
```antlr
slice_pattern
  : '..' negated_pattern?
  ;
```
`.. var x` is `..` followed by a var pattern which is matched against the slice (just capturing it in this case)

You could imagine a user-defined pattern against `Span` so that you could write:

```cs
expr is {1, .. SomePattern(), 3}
```
```cs
expr.Length    is >= 2
&& expr[0]     is 1
&& expr[1..^1] is SomePattern()
&& expr[^1]    is 3
```

jcouv: Let me illustrate the question with an example:
`x is { 1, .. var { Property: 0 } y, 2 }` // if this is allowed, what is the type of `var` and `y`, and what is the receiver type for `Property`?

alrz: `var { Property: 0 } y` is not a valid pattern anywhere. 

Note that `var` is not part of the slice pattern, rather it's a "pattern combinator" so that it could contain other patterns in it.

proposals/list-patterns.md
alrz: As long as it's compatible with the return type of `Slice` method, it is allowed.

proposals/list-patterns.md
jcouv: Sorry my example was invalid. We can break it into two separate patterns `..var y` and  `..{ Property: 0 } y`, and my question stands.

alrz: >We'll want to say what is the input type for patterns following `..`. It's whatever type the range indexer returned.

Correct. I'll add a note to clarify that.

proposals/list-patterns.md
```diff
@@ -0,0 +1,129 @@
...
+## Summary
+
+Lets you to match an array or a list with a sequence of patterns e.g. `array is {1, 2, 3}` will match an integer array of the length three with 1, 2, 3 as its elements, respectively.
+
+## Detailed Design
+
+The pattern syntax is modified as follow:
+
+```antlr
+recursive_pattern
```
jcouv, line 12: Previous patterns specs did away with this single syntax form. There is no `recursive_pattern` production.
See https://github.com/dotnet/csharplang/blob/a3d67d141501101af9dbb9e8192ca8b4c717e316/proposals/csharp-8.0/patterns.md#patterns

alrz: Sorry I missed your comment last time. I was wrongly reflecting the AST so the grammar was all wrong. =(

proposals/list-patterns.md
```diff
@@ -0,0 +1,142 @@
...
+Lets you to match an array or a list with a sequence of patterns e.g. `array is {1, 2, 3}` will match an integer array of the length three with 1, 2, 3 as its elements, respectively.
+
+## Detailed Design
+
+The pattern syntax is modified as follow:
+
+```antlr
+positional_pattern
+  : type? positional_pattern_clause length_pattern_clause? property_or_list_pattern_clause? simple_designation?
+  ;
```
alrz, line 14: This is what's been discussed so far, I believe. cc @333fred

proposals/list-patterns.md
```diff
@@ -0,0 +1,142 @@
...
+property_pattern_clause
+  : '{' (subpattern (',' subpattern)* ','?)? '}'
+  ;
+
+list_pattern_clause
+  : '{' pattern (',' pattern)* ','? '}'
+  ;
+
+slice_pattern
+  : '..' negated_pattern?
```
jcouv, line 46: I'm not sure whether this should be `negated_pattern` or `pattern`. For example, should `..{ 1, 2 } or { 1, 2, 3 }` be allowed, or would that require parens `..({ 1, 2 } or { 1, 2, 3 })`?

proposals/list-patterns.md
jcouv: Thanks. Confirmed with thread with LDM.

proposals/list-patterns.md
```diff
@@ -0,0 +1,142 @@
...
+#### Pattern compatibility
+
+A *length_pattern* is compatible with any type that is *countable*, i.e. has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.
+
+A *list_pattern* is compatible with any type that conforms to the following rules:
+
+1. Is compatible with the *length_pattern*
+2. Has an accessible indexer with a single `int` parameter
+
+ > **Open question**: Should we support `this[Index]` indexers? If so, which one is preferred if `this[int]` is also present?
```
jcouv, line 80: Started thread with LDM and @alrz. 
I think we should just say that:
- if the type is "indexable" (explicitly or implicitly, ie. can you write `expr[^1]`?) then the list pattern can be used
- if the type is "sliceable" (explicitly or implicitly, ie. can you write `expr[1..^1]`?) then the slice pattern can be used
- if the type is countable (defined in ranges spec) then the length pattern can be used

proposals/list-patterns.md
```diff
@@ -0,0 +1,142 @@
...
+1. Is compatible with the *length_pattern*
+2. Has an accessible `Slice` method that takes two `int` parameters (required only if a subpattern is specified)
+
+ > **Open question**: Should we support `this[Range]` indexers? If so, which one is preferred if `Slice(int, int)` is also present?
+
+This set of rules is already specified as the [***range indexer pattern***](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md#implicit-index-support) and required for range indexers.
+
+#### Subsumption checking
+
+Subsumption checking works just like positional patterns with `ITuple` - corresponding subpatterns are matched by position plus an additional node for testing length.
```
jcouv, line 93: nit: Could you add a link and possibly an example?

proposals/list-patterns.md
```diff
@@ -0,0 +1,142 @@
...
+Note: the lowering is presented in the pattern form here to show how subsumption checking works, for example, the following code produces an error because both patterns yield the same DAG:
+
+```cs
+case {_, .., 1}: // expr.Length is >= 2 && expr[^1] is 1
+case {.., _, 1}: // expr.Length is >= 2 && expr[^1] is 1
+```
+Unlike:
+```cs
+case {_, 1, ..}: // expr.Length is >= 2 && expr[1] is 1
+case {.., 1, _}: // expr.Length is >= 2 && expr[^2] is 1
```
jcouv, line 122: Just curious, would we detect the redundancy if we know the length? (my assumption is _no_)
```cs
case [3] {_, 1, ..}: // expr.Length is >= 2 && expr[1] is 1
case [3] {.., 1, _}: // expr.Length is >= 2 && expr[^2] is 1
```

jcouv (commented):
Done with review pass (iteration 11)

proposals/list-patterns.md
alrz: I'm open to both. I started this as an unary operator hence the current limitation. If you look at it as a kind of *prefix* we could possibly allow a general pattern instead. I do believe parens make that more clear but we probably shouldn't *force* it. Either way, something like `.. p or .. q` would be an error just for different reasons.

proposals/list-patterns.md
alrz: That makes perfect sense (I assume this is the terminology we used for ranges). Following range semantics that would imply we prefer this[Index] and this[Range] if both are present.

proposals/list-patterns.md
alrz: The correct codegen for that example checks `Length=3` but we wouldn't warn or anything about those redundancies.

(The effective length range is defined as the intersection of the length pattern value set and the inferred length value set from the list, if this results in an empty value set we produce an error.)

proposals/list-patterns.md
alrz: (pending)

proposals/list-patterns.md
alrz: Quick question: Does this mean we may use `this[int]` for leading patterns and `this[Index]` for trailing patterns?

proposals/list-patterns.md
jcouv: Let's keep `negated_pattern` for now and add an open question. Thanks

proposals/list-patterns.md
jcouv: Confirmed with LDM that we should use this approach.
The ranges spec only defined "countable". But it unfortunately doesn't define "indexable" or "sliceable" as concepts. It uses terms like "Index support" and "Range support". I think that's okay, we can introduce the terminology now.

> However, if we were to completely match indexer access, does that mean we may use this[int] for leading patterns and this[Index] for trailing patterns?

Good point. Both the leading and trailing patterns should bind the same way.
We can solve this by saying that we bind like `expr[new Index(index)]`.

proposals/list-patterns.md
```diff
@@ -0,0 +1,136 @@
...
+expr.Length    is >= 2
+&& expr[0]     is 1
+&& expr[1..^1] is var s
+&& expr[^1]    is 3
+```
+The *input type* for the *slice_pattern* is the return type of the underlying `this[Range]` or `Slice` method with two exceptions: For `string` and arrays, `string.Substring` and `RuntimeHelpers.GetSubArray` will be used, respectively.
+
+### Additional types
+
+Beyond the pattern-based mechanism outlined above, there are an additional two set of types that can be covered as a special case.
```
alrz, line 123: Too many puns. #WontFix

jcouv (approved):
LGTM Thanks (iteration 12) with minor follow-up comments.

proposals/list-patterns.md
```diff
@@ -0,0 +1,136 @@
...
+The order in which subpatterns are matched at runtime is unspecified, and a failed match may not attempt to match all subpatterns.
+
+> **Open question**: The pattern `{..}` tests for  `expr.Length >= 0`. Should we omit such test (assuming `Length` is always non-negative)?
+> 
+#### Lowering
+
+A pattern of the form `expr is {1, 2, 3}` is equivalent to the following code:
+```cs
+expr.Length is 3
+&& expr[0] is 1
```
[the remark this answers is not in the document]
alrz: This is still correct assuming `expr` is compatible via "implicit indexer support". This is just an example..

proposals/list-patterns.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### Pattern compatibility
+
+A *length_pattern* is compatible with any type that is *countable* - it has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.
+
+A *list_pattern* is compatible with any type that is *countable* as well as *indexable* - it has an accessible indexer that takes an `Index` or `int` argument. If both indexers are present, the former is preferred.
+
+A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* - it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.
+
+This set of rules is derived from the [***range indexer pattern***](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md#implicit-index-support) but relaxed to ignore optional or `params` parameters, if any.
```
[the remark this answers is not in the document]
alrz: I reflected on https://github.com/dotnet/csharplang/pull/3245#discussion_r594626501

If you can write `expr[^1]` you might have some optional and `params` parameters in the actual signature.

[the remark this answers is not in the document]
alrz: I added a note to confirm binding rules.

jcouv:
Tagging @333fred for a second look. Thanks

proposals/list-patterns.md
```diff
@@ -0,0 +1,138 @@
...
+```antlr
+positional_pattern
+  : type? positional_pattern_clause length_pattern_clause? property_or_list_pattern_clause? simple_designation?
+  ;
+
+length_pattern
+  : type? length_pattern_clause property_or_list_pattern_clause? simple_designation?
+  ;
+
+list_pattern
```
333fred, line 20: Some reordering here to put the clause definitions after first use would be very helpful as a reader.

proposals/list-patterns.md
```diff
@@ -0,0 +1,138 @@
...
+primary_pattern
+  : list_pattern
+  | length_pattern
+  | slice_pattern
+  | // all of the pattern forms previously defined
+  ;
+```
+There are three new patterns:
+
+- The *list_pattern* is used to match elements and the *length_pattern* is used to match the length.
```
333fred, line 58: Please put *length_pattern* as it's own bullet.

proposals/list-patterns.md
```diff
@@ -0,0 +1,138 @@
...
+Unlike:
+```cs
+case {_, 1, ..}: // expr.Length is >= 2 && expr[1] is 1
+case {.., 1, _}: // expr.Length is >= 2 && expr[^2] is 1
+```
+
+The order in which subpatterns are matched at runtime is unspecified, and a failed match may not attempt to match all subpatterns.
+
+> **Open question**: The pattern `{..}` tests for  `expr.Length >= 0`. Should we omit such test (assuming `Length` is always non-negative)?
+>
```
333fred, line 104: Nit: extra >

333fred (commented):
A couple of clarity comments.

jcouv:
Cool. Thanks for writing this up and sorting many questions. We'll start looking at your roslyn-side change now :-)

PR #3263, opened 2020-03-10
Add conversions spec for function pointers

333fred:
The summary for these specs are:

- Value parameters are treated as contravariant.
- Value return types are treated as covariant.
- Reference parameter/return types and modifiers must match exactly.
- Calling convention must match exactly.
- An address-of expression that contains method group with exactly 1 static method is convertible to `void*`.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -129,6 +129,18 @@ delegate* managed<string, int>;
 delegate*<delegate* managed<string, int>, delegate*<string, int>>;
 ```
 
+### Function pointer conversions
+
+In an unsafe context, the set of available implicit conversions (Implicit conversions) is extended to include the following implicit pointer conversions:
+- [_Existing conversions_](https://github.com/dotnet/csharplang/blob/master/spec/unsafe-code.md#pointer-conversions)
+- From _funcptr\_type_ `F0` to another _funcptr\_type_ `F1`, provided all of the following are true:
+    - `F0` and `F1` have the same number of parameters, and each parameter `D0n` in `F0` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter `D1n` in `F1`.
+    - For each value parameter (a parameter with no `ref`, `out`, or `in` modifier), an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `F0` to the corresponding parameter type in `F1`.
+    - For each `ref`, `out`, or `in` parameter, the parameter type in `F0` is the same as the corresponding parameter type in `F1`.
+    - If the return type is by value (no `ref` or `ref readonly`), an identity, implicit reference, or implicit pointer conversion exists from the return type of `F1` to the return type of `F0`.
+    - If the return type is by reference (`ref` or `ref readonly`), the return type and `ref` modifiers of `F1` are the same as the return type and `ref` modifiers of `F0`.
+    - The calling convention of `F0` is the same as the calling convention of `F1`.
+
 ### Allow address-of to target methods
 
 Method groups will now be allowed as arguments to an address-of expression. The type of such an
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -151,11 +163,26 @@ unsafe class Util {
...
-- Only a `delegate*` with a managed calling convention can be the target of such a conversion.
+In an unsafe context, a method `M` is compatible with a function pointer type `F` if all of the following are true:
+- `M` and `F` have the same number of parameters, and each parameter in `D` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter in `F`.
+- For each value parameter (a parameter with no `ref`, `out`, or `in` modifier), an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `M` to the corresponding parameter type in `F`.
+- For each `ref`, `out`, or `in` parameter, the parameter type in `M` is the same as the corresponding parameter type in `F`.
+- If the return type is by value (no `ref` or `ref readonly`), an identity, implicit reference, or implicit pointer conversion exists from the return type of `F` to the return type of `M`.
+- If the return type is by reference (`ref` or `ref readonly`), the return type and `ref` modifiers of `F` are the same as the return type and `ref` modifiers of `M`.
+- The calling convention of `M` is the same as the calling convention of `F`.
```
jkotas: How you determine calling convention of `M` ? Is this going to take into account the NativeCallable attribute?

proposals/function-pointers.md
```diff
@@ -151,11 +163,26 @@ unsafe class Util {
...
+   - The arguments list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref`, `out`, or `in`) of the corresponding _formal\_parameter\_list_ of `D`.
+   - The candidate methods are only those methods that are only those methods that are applicable in their normal form, not those applicable in their expanded form.
+- If the algorithm of Method invocations produces an error, then a compile-time error occurs. Otherwise, the algorithm produces a single best method `M` having the same number of parameters as `F` and the conversion is considered to exist.
+- The selected method `M` must be compatible (as defined above) with the function pointer type `F`. Otherwise, a compile-time error occurs.
+- The result of the conversion is a function pointer of type `F`.
+
+An implicit conversion exists from an address-of expression whose target is a method group `E` to `void*` if there is only one static method `M` in `E`.
+If there is one static method, then the single best method from `E` is `M`.
+Otherwise, a compile-tiem error occurs.
```
jkotas:
```suggestion
Otherwise, a compile-time error occurs.
```

proposals/function-pointers.md
333fred: > Is this going to take into account the NativeCallable attribute?

I was planning on it doing so, yes, but I'm not sure if that's something that should be explicitly called out in our spec or if should be an implementation detail. @gafter for thoughts on that?

proposals/function-pointers.md
```diff
@@ -129,6 +129,18 @@ delegate* managed<string, int>;
 delegate*<delegate* managed<string, int>, delegate*<string, int>>;
 ```
 
+### Function pointer conversions
+
+In an unsafe context, the set of available implicit conversions (Implicit conversions) is extended to include the following implicit pointer conversions:
+- [_Existing conversions_](https://github.com/dotnet/csharplang/blob/master/spec/unsafe-code.md#pointer-conversions)
+- From _funcptr\_type_ `F0` to another _funcptr\_type_ `F1`, provided all of the following are true:
+    - `F0` and `F1` have the same number of parameters, and each parameter `D0n` in `F0` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter `D1n` in `F1`.
+    - For each value parameter (a parameter with no `ref`, `out`, or `in` modifier), an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `F0` to the corresponding parameter type in `F1`.
```
AlekseyTs: >implicit reference conversion, or implicit pointer conversion exists from the parameter type in `F0` to the corresponding parameter type in `F1`. [](start = 109, length = 145)

Does runtime support this kind of mismatch natively, or will compiler have to generate a stub to make this work?

proposals/function-pointers.md
```diff
@@ -151,11 +163,26 @@ unsafe class Util {
...
 ```
 
-The conversion of an address-of method group to `delegate*` has roughly the same process as method group to `delegate`
-conversion. There are two additional restrictions to the existing process:
-
-- Only members of the method group that are marked as `static` will be considered.
-- Only a `delegate*` with a managed calling convention can be the target of such a conversion.
+In an unsafe context, a method `M` is compatible with a function pointer type `F` if all of the following are true:
+- `M` and `F` have the same number of parameters, and each parameter in `D` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter in `F`.
+- For each value parameter (a parameter with no `ref`, `out`, or `in` modifier), an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `M` to the corresponding parameter type in `F`.
```
AlekseyTs: >implicit reference conversion, or implicit pointer conversion exists from the parameter type in `M` to the corresponding parameter type in `F`. [](start = 105, length = 143)

Same question here.

proposals/function-pointers.md
333fred: > Does runtime support this kind of mismatch natively, or will compiler have to generate a stub to make this work?

Natively. We don't even emit a cast for explicit conversions.

AlekseyTs:
Do we want to/is it possible to support a conversion from a function pointer to a delegate?

333fred:
@AlekseyTs the runtime has an API for this today, `Marshal.GetFunctionPointerForDelegate<TDelegate>(TDelegate)`. I don't think we need to try and do anything special in the language for that.

AlekseyTs:
> the runtime has an API for this today, Marshal.GetFunctionPointerForDelegate<TDelegate>(TDelegate). I don't think we need to try and do anything special in the language for that.

I think I am talking about conversion in the opposite direction. I.e. passing  a pointer to delegate's constructor.

333fred:
> I think I am talking about conversion in the opposite direction.

The runtime has a method going in the other direction as well, `Marshal.GetDelegateForFunctionPointer<TDelegate>(IntPtr)`. While it's not ideal as it's not a strongly-typed guarantee that the function pointer is compatible with the type of the delegate, unless I hear strong desire for these to be compatible I don't think it merits special language treatment either.

AlekseyTs:
> The runtime has a method going in the other direction as well, `Marshal.GetDelegateForFunctionPointer<TDelegate>(IntPtr)`.

I am not sure why would we want to force a user to use this API and not have a benefit of type safety checks performed by a compiler.  A delegate constructor takes a pointer (which we obtain for a method today) and, in function pointer case, we would have a pointer available. We also already have rules for method signature to delegate compatibility, which we can simply reuse.

jkotas:
- `Marshal.GetDelegateForFunctionPointer` is an expensive API. I do not think you want compiler to add calls to it.
- `Marshal.GetDelegateForFunctionPointer` works for unmanaged methods only
- Users who are using function pointers are unlikely to use `Marshal.GetDelegateForFunctionPointer`. Function pointers are better.

333fred:
I added a note to the open issues to consider it.

333fred:
@AlekseyTs @jkotas did you have any more feedback? I'll address the typo when I'm confident that all the rest of the minor changes are accounted for.

jkotas:
Looks fine to me otherwise.

proposals/function-pointers.md
jkotas: It does not feel right to me to talk about calling convention in the spec, but not explain where it comes from.

gafter: I agree that if the spec depends on a concept, it should define it. However, that can be left for a future refinement of this spec. Put it in an open issue?

333fred: Added to the open issues list.

PR #3280, opened 2020-03-19
'conceit' → 'concept'

jnm2:


333fred:
Conceit does make sense in context, I'll let @agocke decide if concept is a better fit.

jcouv:
Merged. Thanks

Issue #3277, opened 2020-03-19
[Proposal] Improve overload resolution for delegate compatibility


333fred:
# Improve overload resolution for delegate compatibility

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

## Summary
[summary]: #summary

Improve overload resolution when looking for applicable methods by removing methods that cannot be compatible. This will allow the following code to compile (it does not today):
```cs
public class Program1
{
    delegate void MyAction<T>(T x);

    void Y(long x) { }

    void D(MyAction<int> o) { }
    void D(MyAction<long> o) { }

    void T()
    {
        D(Y); // Ambiguous between both D calls, despite the fact that `void D(MyAction<int>)` is not a valid target.
    }
}
```

## Motivation
[motivation]: #motivation

In the spirit of the other overload resolution changes we've made recently, removing things that already don't work, this would allow currently ambiguous code to compile and improve user experience.

## Detailed design
[design]: #detailed-design

Given the example in the Summary section, the reason why it does not compile today is straightforward, if requiring an intimate knowledge of the spec to figure out:

1. We attempt to perform overload resolution on the call to `D`. To do that, we start by computing the applicable function members in the method group `D`.
2. Overload resolution for applicable function members states that, for value parameters, there must be an implicit conversion from every parameter in the call to the given method in `D` if the method in `D` is applicable.
3. In attempting to determine if that is true for `D(MyAction<int>)`, we need to determine if the method group `Y` is applicable to `MyAction<T>` where `T` is `int`.
4. We again perform overload resolution, this time on `Y`.
5. The applicable function member definition looks that the first parameter of `Y`, which is of type `long`. There is an implicit numeric conversion from `int` to `long`. There are no more parameters, so `Y` is applicable to `MyAction<int>` and a method group conversion exists.
    * Delegate compatibility has _not_ come into play here yet! Even though `Y` is not compatible with `MyAction<int>`, it is _applicable_.
6. The first overload resolution determines that `D(MyAction<int>)` is an applicable member.
7. Overload resolution looks at the next member, `D(MyAction<long>)`. The same sub-steps run again, and `D(MyAction<long>)` is determined to be _applicable_.
8. Neither method is more specific, so overload resolution fails.

[...]

PR #3281, opened 2020-03-19
Only allow static methods for applicable members

333fred:
This fixes a very similar problem to https://github.com/dotnet/csharplang/issues/3277, where this code is unable to be resolved:
```cs
interface I1{}
interface I2{}

public unsafe class C : I1, I2 {
    void M(I1 i) {}
    static void M(I2 i) {}
    public void M1() {
        delegate*<C, void> a = M; // Ambiguous because both M's are applicable
    }
}
```
With this change, the instance method M is not applicable, so there is no ambiguity.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -176,6 +176,7 @@ In an unsafe context, an implicit conversion exists from an address-of expressio
 - A single method `M` is selected corresponding to a method invocation of the form `E(A)` with the following modifications:
    - The arguments list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref`, `out`, or `in`) of the corresponding _formal\_parameter\_list_ of `D`.
    - The candidate methods are only those methods that are only those methods that are applicable in their normal form, not those applicable in their expanded form.
+   - The candidate methods are only those methods that are static.
 - If the algorithm of Method invocations produces an error, then a compile-time error occurs. Otherwise, the algorithm produces a single best method `M` having the same number of parameters as `F` and the conversion is considered to exist.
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -176,6 +176,7 @@ In an unsafe context, an implicit conversion exists from an address-of expressio
 - A single method `M` is selected corresponding to a method invocation of the form `E(A)` with the following modifications:
    - The arguments list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref`, `out`, or `in`) of the corresponding _formal\_parameter\_list_ of `D`.
    - The candidate methods are only those methods that are only those methods that are applicable in their normal form, not those applicable in their expanded form.
+   - The candidate methods are only those methods that are static.
```
AlekseyTs: It looks like the previous bullet should be adjusted: "those methods that are only those methods that are applicable"

tannergooding:
Doesn't this preclude allowing binding to instance methods in the future or would that likely involve some special syntax (hypothetically `delegate*<this C, void>`)?

333fred:
It does preclude that, yes. If we really need to we can introduce a `this` modifier, but I'm not of the opinion that we need to allow this. It's easy to write a small static wrapper that takes a first parameter and invokes a method on it.

tannergooding:
> It's easy to write a small static wrapper that takes a first parameter and invokes a method on it.

Easy yes, but possibly not performant, as it will force an additional indirection.

(I don't think its needed for v1, just wanted to ensure I understood the state we'd be in and what might be needed in the future to enable it).

PR #3282, opened 2020-03-19
Add 'with' expression to the spec

agocke:


proposals/records-wip.md
```diff
@@ -76,3 +76,45 @@ each primary constructor parameter to the corresponding property of the other ty
...
+A `with` expression allows for "non-destructive mutation", designed to
+produce a copy of the receiver expression with modifications to properties
+listed in the `anonymous_object_initializer`.
+
+A valid `with` expression has a receiver with a non-target-typed, non-null type. The receiver type must contain an instance method called `With` with
+the appropriate parameters and return type. It is an error if there are multiple non-override `With` methods. If there are multiple `With` overrides,
+there must be a non-override `With` method, which is the target method. Otherwise, there must be exactly one `With` method.
+
+On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
+sequence of assignments with a field or property of the receiver on the left hand side of the
```
gafter: >a [](start = 29, length = 1)

accessible?

proposals/records-wip.md
```diff
@@ -76,3 +76,45 @@ each primary constructor parameter to the corresponding property of the other ty
...
+assignment, and an arbitrary expression on the right hand side.
```
gafter: >arbitrary expression [](start = 19, length = 20)

Does the expression have to be implicitly convertible to the type of the field or property?

proposals/records-wip.md
```diff
@@ -76,3 +76,45 @@ each primary constructor parameter to the corresponding property of the other ty
...
+receiver type with the same name and the same type.
+
+Given a valid `With` method, the evaluation of a `with` expression is equivalent to calling the `With` method with the expressions in the
+`anonymous_object_initializer` substituted for the parameter of the same
+name as the property on the left hand side. If there is no matching property
+for a given parameter in the `anonymous_object_initializer`, the argument
+is the evaluation of the field or property of the same name on the receiver.
+
+The order of evaluation of side effects is as follows, with each expression
+evaluated exactly once:
```
gafter: >: [](start = 22, length = 1)

What happens to the initializers that do not correspond to parameters of the `With` method?

proposals/records-wip.md
```diff
@@ -76,3 +76,45 @@ each primary constructor parameter to the corresponding property of the other ty
...
+the appropriate parameters and return type. It is an error if there are multiple non-override `With` methods. If there are multiple `With` overrides,
+there must be a non-override `With` method, which is the target method. Otherwise, there must be exactly one `With` method.
+
+On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
+sequence of assignments with a field or property of the receiver on the left hand side of the
+assignment, and an arbitrary expression on the right hand side.
+
+Given a target `With` method, the return type must be the type of the receiver expression type, or a base type thereof. For each parameter of
+the `With` method, there must be a corresponding field or property on the
+receiver type with the same name and the same type.
```
gafter: >same name and the same type [](start = 23, length = 27)

I'm confused as to why we're going with name matching. Mads was strongly opposed to that.

proposals/records-wip.md
```diff
@@ -76,3 +76,45 @@ each primary constructor parameter to the corresponding property of the other ty
...
+with_expression
+    : switch_expression
+    | switch_expression 'with' anonymous_object_initializer
+```
+
+A `with` expression allows for "non-destructive mutation", designed to
+produce a copy of the receiver expression with modifications to properties
+listed in the `anonymous_object_initializer`.
+
+A valid `with` expression has a receiver with a non-target-typed, non-null type. The receiver type must contain an instance method called `With` with
```
gafter: >non-target-typed, non-null type [](start = 48, length = 31)

These are not concepts in the language.  If you don't define a target for the conversion, then any expression in the structure of a program isn't target-typed.

proposals/records-wip.md
```diff
@@ -76,3 +76,45 @@ each primary constructor parameter to the corresponding property of the other ty
...
+the appropriate parameters and return type. It is an error if there are multiple non-override `With` methods. If there are multiple `With` overrides,
+there must be a non-override `With` method, which is the target method. Otherwise, there must be exactly one `With` method.
+
+On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
+sequence of assignments with a field or property of the receiver on the left hand side of the
+assignment, and an arbitrary expression on the right hand side.
+
+Given a target `With` method, the return type must be the type of the receiver expression type, or a base type thereof. For each parameter of
+the `With` method, there must be a corresponding field or property on the
```
gafter: >corresponding [](start = 35, length = 13)

instance?

proposals/records-wip.md
agocke: Talked to him about this, he said it was fine for the prototype. I suspect this will not be the final design.

proposals/records-wip.md
```diff
@@ -76,3 +76,47 @@ each primary constructor parameter to the corresponding property of the other ty
...
+A `with` expression allows for "non-destructive mutation", designed to
+produce a copy of the receiver expression with modifications to properties
+listed in the `anonymous_object_initializer`.
+
+A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible instance method called `With` with
+the appropriate parameters and return type. It is an error if there are multiple non-override `With` methods. If there are multiple `With` overrides,
+there must be a non-override `With` method, which is the target method. Otherwise, there must be exactly one `With` method.
+
+On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
+sequence of assignments with a field or property of the receiver on the left-hand side of the
```
gafter: >a [](start = 29, length = 1)

... with the name of a ...

gafter:
Finished reviewing iteration 3.

proposals/records-wip.md
```diff
@@ -76,3 +76,47 @@ each primary constructor parameter to the corresponding property of the other ty
...
+assignment, and an arbitrary expression on the right-hand side which is implicitly convertible to the type
+of the left-hand side.
+
+Given a target `With` method, the return type must be the type of the receiver expression type, or a base type thereof. For each parameter of
```
333fred: > or a base type thereof

Did you mean derived type?

agocke: I don't think so... given the example

```C#
class C { }
class B : C {}
```

if `B` has a `With` method, it could return `C` instead of `B`, in that it would be an override of the `With` in `C`, meaning the base type.

proposals/records-wip.md
agocke: I think that's covered below in

> Each property or field in the right-hand side of the With
expression must also correspond to a parameter of the same name in the `With` method.

proposals/records-wip.md
333fred: Should we then spec that even if the return type is `C`, you must return at least `B`, if not a further derived class? Since we have to be able to assign back to a field of `B`? Sure, it's trivially possible to define a `With` method that violates such constraints, but those programs are not in compliance with the spec.

agocke: This seems fine for now. We can always strengthen it later

PR #3292, opened 2020-03-20
Limit top-level statements to a single compilation unit within a program

AlekseyTs:
Reflect LDM decision made on 2020-03-09 to limit top-level statements to a single compilation unit within a program.


proposals/Simple-programs.md
```diff
@@ -93,11 +89,6 @@ The method is designated as the entry point of the program. Explicitly declared
 could be considered as an entry point candidates are ignored. A warning is reported when that happens. It is
 an error to specify `-main:<type>` compiler switch.
 
-If any one compilation unit has statements other than local function declarations, statements from that
```
gafter: >If [](start = 0, length = 2)

Deleting this paragraph (and related examples) doesn't make it illegal for a local function in one compilation unit to reference a local declared in another.  That is controlled by the scope, and your section "Scope of top-level local variables and local functions" below indeed makes that legal.  If you want it to be illegal, you should do so explicitly.  Otherwise it would be better to restore this text.

AlekseyTs: > If you want it to be illegal, you should do so explicitly.

The following is added above: "Only one *compilation_unit* is allowed to have the *statement*s." I think that is explicit.

AlekseyTs: It replaces the old rule in place: "In all but one *compilation_unit* the *statement*s must all be local function declarations."

proposals/Simple-programs.md
```diff
@@ -47,18 +47,16 @@ compilation_unit
     ;
 ```
 
-In all but one *compilation_unit* the *statement*s must all be local function declarations. 
+Only one *compilation_unit* is allowed to have the *statement*s.
```
gafter: >the [](start = 47, length = 3)

Delete "the".

proposals/Simple-programs.md
```diff
@@ -93,11 +89,6 @@ The method is designated as the entry point of the program. Explicitly declared
 could be considered as an entry point candidates are ignored. A warning is reported when that happens. It is
 an error to specify `-main:<type>` compiler switch.
```
gafter: >switch [](start = 44, length = 6)

... when there are top-level statements.

proposals/Simple-programs.md
```diff
@@ -113,13 +104,11 @@ static class $Program
 {
     static void $Main()
     {
-        // Statements from File 1
         if (args.Length == 0
```
gafter: >args [](start = 12, length = 4)

You probably want to declare a parameter named `args`?

proposals/Simple-programs.md
```diff
@@ -47,18 +47,16 @@ compilation_unit
...
 ```
 
-In all but one *compilation_unit* the *statement*s must all be local function declarations. 
+Only one *compilation_unit* is allowed to have the *statement*s. 
 
 Example:
 
 ``` c#
-// File 1 - any statements
 if (args.Length == 0
```
gafter: >args [](start = 4, length = 4)

By this spec `args` is not declared and this is an error.

AlekseyTs: > By this spec args is not declared and this is an error.

Thanks, I will adjust the example.

gafter:
> into the generated entry point method, they should still be in scope throughout the program.

Presumably this means in every compilation unit.  Please be explicit about that.

---
Refers to: proposals/Simple-programs.md:143 in d6124ad. [](commit_id = d6124ad0a8334ce9357ee93f6d57d702c8fd3fd2, deletion_comment = False)

PR #3299, opened 2020-03-23
specify using accessibility requirements

YairHalberstadt:
using alias and static specifies that types should be accessible

See discussion in https://github.com/dotnet/roslyn/pull/42692


AlekseyTs:
Don't regular name lookup rules outlined in section **3.8 Namespace and type names** involve type accessibility? I am not sure why sections about using directives are expected to reiterate them.

PR #3310, opened 2020-03-25
Clean up the specification portion of target-typed new

gafter:
@jcouv Please have a look

### DIFF
--- a/proposals/target-typed-new.md
+++ b/proposals/target-typed-new.md
@@ -47,7 +47,7 @@ target_typed_new
     ;
 ```
 
-There is a new *object creation conversion* that is an implicit conversion from expression, that exists from a *target_typed_new* to every type. It is an error for this conversion to be applied unless the type being converted to is
+A *target_typed_new* expression does not have a type. However, there is a new *object creation conversion* that is an implicit conversion from expression, that exists from a *target_typed_new* to every type. It is an error for this conversion to be applied unless the type being converted to is
 - a `struct` type; or
 - a `class` type; or
 - a `delegate` type; or
--- a/proposals/target-typed-new.md
+++ b/proposals/target-typed-new.md
@@ -55,6 +55,8 @@ A *target_typed_new* expression does not have a type. However, there is a new *o
 
 Given a target type `T`, the type `T0` is `T`'s underlying type if `T` is an instance of `System.Nullable`. Otherwise `T0` is `T`. The meaning of a *target_typed_new* expression that is converted to the type `T` is the same as the meaning of a corresponding *object_creation_expression* that specifies `T0` as the type.
 
+It is a compile-time error if a *target_typed_new* is used as an operand of a unary or binary operator.
+
 > **Open Issue:** should we allow delegates and tuples as the target-type?
 
 The above rules include delegates (a reference type) and tuples (a struct type). Although both types are constructible, if the type is inferable, an anonymous function or a tuple literal can already be used.
--- a/proposals/target-typed-new.md
+++ b/proposals/target-typed-new.md
@@ -31,37 +31,29 @@ Instantiate an object without spelling out the type.
 private readonly static object s_syncObj = new();
 ```
 
-## Detailed design
+## Specification
 [design]: #detailed-design
 
-The *object_creation_expression* syntax would be modified to make the *type* optional when parentheses are present. This is required to address the ambiguity with *anonymous_object_creation_expression*.
+A new syntactic form, *target_typed_new* of the *object_creation_expression* is accepted in which the *type* is optional.
+
 ```antlr
 object_creation_expression
-    : 'new' type? '(' argument_list? ')' object_or_collection_initializer?
+    : 'new' type '(' argument_list? ')' object_or_collection_initializer?
     | 'new' type object_or_collection_initializer
+    | target_typed_new
+    ;
+target_typed_new
+    : 'new' '(' argument_list? ')' object_or_collection_initializer?
     ;
 ```
 
-A target-typed `new` is convertible to any type. As a result, it does not contribute to overload resolution. This is mainly to avoid unpredictable breaking changes.
+There is a new *object creation conversion* that is an implicit conversion from expression, that exists from a *target_typed_new* to every type. It is an error for this conversion to be applied unless the type being converted to is
+- a `struct` type; or
+- a `class` type; or
+- a `delegate` type; or
+- a type parameter.
 
-The argument list and the initializer expressions will be bound after the type is determined.
-
-The type of the expression would be inferred from the target-type which would be required to be one of the following:
-
-- **Any struct type** (including tuple types)
-- **Any reference type** (including delegate types)
-- **Any type parameter** with a constructor or a `struct` constraint
-
-with the following exceptions:
-
-- **Enum types:** not all enum types contain the constant zero, so it should be desirable to use the explicit enum member.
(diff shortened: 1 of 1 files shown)


proposals/target-typed-new.md
```diff
@@ -31,37 +31,29 @@ Instantiate an object without spelling out the type.
...
-The type of the expression would be inferred from the target-type which would be required to be one of the following:
-
-- **Any struct type** (including tuple types)
-- **Any reference type** (including delegate types)
-- **Any type parameter** with a constructor or a `struct` constraint
+A *target_typed_new* expression does not have a type. However, there is a new *object creation conversion* that is an implicit conversion from expression, that exists from a *target_typed_new* to every type. It is an error for this conversion to be applied unless the type being converted to is
+- a `struct` type; or
+- a `class` type; or
+- a `delegate` type; or
+- a type parameter.
```
333fred: I think this is missing interfaces, due to COM.

333fred: Ah, I see the deleted diff has an explicit exception. I'm not sure that I agree with the exception, it feels inconsistent for no reason.

proposals/target-typed-new.md
```diff
@@ -76,13 +68,12 @@ Action a = new(); // no constructor found
...
-
-It is disallowed when there is no type to target: unary operators, collection of a `foreach`, in a `using`, in a deconstruction, in an `await` expression, as an anonymous type property (`new { Prop = new() }`), in a `lock` statement, in a `sizeof`, in a `fixed` statement, in a member access (`new().field`), in a dynamically dispatched operation (`someDynamic.Method(new())`), in a LINQ query, as the operand of the `is` operator, as the left operand of the `??` operator,  ...
+The following are consequences of the specification:
 
-It is also disallowed as a `ref`.
+- `throw new()` is allowed (the target type is `System.Exception`)
+- Target-typed `new` is not allowed with binary operators.
+- It is disallowed when there is no type to target: unary operators, collection of a `foreach`, in a `using`, in a deconstruction, in an `await` expression, as an anonymous type property (`new { Prop = new() }`), in a `lock` statement, in a `sizeof`, in a `fixed` statement, in a member access (`new().field`), in a dynamically dispatched operation (`someDynamic.Method(new())`), in a LINQ query, as the operand of the `is` operator, as the left operand of the `??` operator,  ...
```
alrz: >as the operand of the `is` operator

:memo: This is only the left-hand-side operand of the `is` operator.

proposals/target-typed-new.md
```diff
@@ -76,13 +68,12 @@ Action a = new(); // no constructor found
 
 ### Miscellaneous
 
-`throw new()` is allowed.
-
-Target-typed `new` is not allowed with binary operators.
-
-It is disallowed when there is no type to target: unary operators, collection of a `foreach`, in a `using`, in a deconstruction, in an `await` expression, as an anonymous type property (`new { Prop = new() }`), in a `lock` statement, in a `sizeof`, in a `fixed` statement, in a member access (`new().field`), in a dynamically dispatched operation (`someDynamic.Method(new())`), in a LINQ query, as the operand of the `is` operator, as the left operand of the `??` operator,  ...
+The following are consequences of the specification:
```
alrz: We specifically do not permit to target unspeakable types i.e. anonymous types, I think that needs to be specified as it's not a consequence of the above.

proposals/target-typed-new.md
alrz: Interface were excluded because we only target *constructible types* as per LDM. Interfaces were not deemed to be constructible.

alrz: Interface were excluded because we only target constructible types as per LDM. Interfaces were not deemed to be constructible.

333fred: Will bring this up today. Personally, I think that's true everywhere except the COM case, and limiting it there just feels inconsistent.

jcouv: The previous revision had some additional details on type parameter case, which I think should be kept:

> **Any type parameter** with a constructor or a `struct` constraint

proposals/target-typed-new.md
```diff
@@ -31,37 +31,29 @@ Instantiate an object without spelling out the type.
...
-- **Any struct type** (including tuple types)
-- **Any reference type** (including delegate types)
-- **Any type parameter** with a constructor or a `struct` constraint
+A *target_typed_new* expression does not have a type. However, there is a new *object creation conversion* that is an implicit conversion from expression, that exists from a *target_typed_new* to every type. It is an error for this conversion to be applied unless the type being converted to is
+- a `struct` type; or
+- a `class` type; or
+- a `delegate` type; or
+- a type parameter.
 
-with the following exceptions:
```
jcouv: Could we keep the description/rationales for enum/interface/arrays/dynamic, if only in the "Miscellaneous" section?

proposals/target-typed-new.md
```diff
@@ -76,13 +68,12 @@ Action a = new(); // no constructor found
...
-`throw new()` is allowed.
-
-Target-typed `new` is not allowed with binary operators.
-
-It is disallowed when there is no type to target: unary operators, collection of a `foreach`, in a `using`, in a deconstruction, in an `await` expression, as an anonymous type property (`new { Prop = new() }`), in a `lock` statement, in a `sizeof`, in a `fixed` statement, in a member access (`new().field`), in a dynamically dispatched operation (`someDynamic.Method(new())`), in a LINQ query, as the operand of the `is` operator, as the left operand of the `??` operator,  ...
+The following are consequences of the specification:
 
-It is also disallowed as a `ref`.
+- `throw new()` is allowed (the target type is `System.Exception`)
+- Target-typed `new` is not allowed with binary operators.
```
jcouv: The restriction on binary operators is not really a consequence of above spec. It is a separate spec rule, no?

proposals/target-typed-new.md
alrz: Thinking more about this, we disallow `default` in patterns, particularly to be not confused with `default:`, but perhaps it would be more consistent to disallow all typeless expressions.

proposals/target-typed-new.md
gafter: Why?  It will be a semantic error to apply the conversion.  I don't think we should condition the existence of the conversion on constraints.

---
In reply to: [397972206](https://github.com/dotnet/csharplang/pull/3310#discussion_r397972206) [](ancestors = 397972206)

gafter: The existing language spec does not permit an object creation expression to specify an interface type. See [https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#object-creation-expressions](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#object-creation-expressions)
This is not the place to start doing that.

---
In reply to: [397967636](https://github.com/dotnet/csharplang/pull/3310#discussion_r397967636) [](ancestors = 397967636)

proposals/target-typed-new.md
gafter: There is no need.  The conversion exists but the creation expression is an error.

---
In reply to: [397953799](https://github.com/dotnet/csharplang/pull/3310#discussion_r397953799) [](ancestors = 397953799)

proposals/target-typed-new.md
gafter: Yes, that is in iteration 3.

gafter:
Will integrate this now to permit LDM review, but address additional comments in a separate PR.

PR #3311, opened 2020-03-25
target-typed new: reflect LDM decisions 2020-03-25

gafter:


proposals/target-typed-new.md
```diff
@@ -78,7 +74,7 @@ The following are consequences of the specification:
 - It is also disallowed as a `ref`.
 - The following kinds of types are not permitted as targets of the conversion
   - **Enum types:** not all enum types contain the constant zero, so it should be desirable to use the explicit enum member.	It is a compile-time error if a *target_typed_new* is used as an operand of a unary or binary operator.
```
jcouv: enum types should be updated (ie. it will work)

proposals/target-typed-new.md
```diff
@@ -64,7 +60,7 @@ The above rules include delegates (a reference type) and tuples (a struct type).
 (int a, int b) t = new(1, 2); // "new" is redundant
 Action a = new(() => {}); // "new" is redundant
 
-(int a, int b) t = new(); // ruled out by "use of struct default constructor"
+(int a, int b) t = new(); // OK; same as (0, 0)
```
jcouv: I thought this would still be an error: `error CS8181: 'new' cannot be used with tuple type. Use a tuple literal expression instead.`

alrz: That is a syntactical limitation only, we still can write `new ValueTuple(..)` and that's what we emit for `new()` here.

>`(int a, int b) t = new(); // OK; same as (0, 0)
`

It would be the same as `new ValueTuple<int, int>()`

333fred: No, this one is fine because we have a real target-type of `ValueTuple`. The one that's not allowed is deconstruction assignment.

alrz:
To summarize, all type restrictions on `new()` are lifted (permitting enum and interface as target-type) and no changes to the conversion. Is that correct?

proposals/target-typed-new.md
jcouv: Right, I think `ValueTuple<int, int> t = new();` is okay, but `(int, int) t = new();` is not (one should write a tuple literal instead).

alrz: >but (int, int) t = new(); is not (one should write a tuple literal instead).

I think we can't even distinguish between the two per "tuple compatible types" rule. 

Basically, [compiler sees both as if they are the same type](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQygg5hG6CeADhACaoB2EA7gDQwgBuECAlmPgD4ACATAIwBYAFC8AzAAIA3sPGzxvHlJlyVAelXiqmci3JZxAYQDKADj49RIcQBUAFhHHxCAGwcRXAWwjkY48lC9xAHIWIPEWZHCscgB7JBJxYAgAYyg4ZAcocRI2SCQfPwCIZRVZdXE4vxjCwIjxZGJkthZSRPxHe0dMHF8YIgcggApdGBpwnwBKIIA6EtKANShnOAhrOBcIAB4RmhGAPkdxAF5hkD4xngmAbjmAX2FboA==).

(btw this was brought up in an LDM before and decided to be allowed)

gafter: Those two examples have precisely the same target type, so both would work.

gafter:
That is correct.

proposals/target-typed-new.md
jcouv: Hum, I guess that's okay and possibly unavoidable.
I added a note to issue tracking [ExplicitType refactoring ](https://github.com/dotnet/roslyn/issues/42554) work for `new()` as it should specially handle tuples

PR #3334, opened 2020-04-04
Add the missing letter

andrewandzz:
"The type parameters can the be used" -> "The type parameters can then be used"


jcouv:
Thanks ;-)

PR #3348, opened 2020-04-09
Update function pointers proposal for binary operators.

333fred:
This forbids all binary operators except comparison operators.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -212,6 +212,25 @@ Restrictions of this feature:
   deliberately not specified by the language. This includes whether they are static vs. instance or
   exactly what signature they are emitted with.
 
+
+### Operators on Function Pointer Types
+
+The section in unsafe code on operators is modified as such:
+
+> In an unsafe context, several constructs are available for operating on all _pointer\_type_s but _funcptr\_type_s (except where noted):
+>
+> *  The `*` operator may be used to perform pointer indirection ([Pointer indirection](unsafe-code.md#pointer-indirection)).
+> *  The `->` operator may be used to access a member of a struct through a pointer ([Pointer member access](unsafe-code.md#pointer-member-access)).
+> *  The `[]` operator may be used to index a pointer ([Pointer element access](unsafe-code.md#pointer-element-access)).
+> *  The `&` operator may be used to obtain the address of a variable ([The address-of operator](unsafe-code.md#the-address-of-operator)).
+> *  The `++` and `--` operators may be used to increment and decrement pointers ([Pointer increment and decrement](unsafe-code.md#pointer-increment-and-decrement)).
+> *  The `+` and `-` operators may be used to perform pointer arithmetic ([Pointer arithmetic](unsafe-code.md#pointer-arithmetic)).
+> *  The `==`, `!=`, `<`, `>`, `<=`, and `=>` operators may be used to compare pointers (including function pointers) ([Pointer comparison](unsafe-code.md#pointer-comparison)).
+> *  The `stackalloc` operator may be used to allocate memory from the call stack ([Fixed size buffers](unsafe-code.md#fixed-size-buffers)).
+> *  The `fixed` statement may be used to temporarily fix a variable so its address can be obtained ([The fixed statement](unsafe-code.md#the-fixed-statement)).
+
+Additionally, we modify all the sections in `Pointers in expressions` to forbid function pointer types, except `Pointer comparison` and `The sizeof operator`, to forbid usage of function pointer types.
+
 ### Better function member
 
 The better function member specification will be changed to include the following line:
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -217,7 +217,7 @@ Restrictions of this feature:
 
 The section in unsafe code on operators is modified as such:
 
-> In an unsafe context, several constructs are available for operating on all _pointer\_type_s but _funcptr\_type_s (except where noted):
+> In an unsafe context, several constructs are available for operating on all _pointer\_type_s that are not _funcptr\_type_s:
 >
 > *  The `*` operator may be used to perform pointer indirection ([Pointer indirection](unsafe-code.md#pointer-indirection)).
 > *  The `->` operator may be used to access a member of a struct through a pointer ([Pointer member access](unsafe-code.md#pointer-member-access)).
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -212,6 +212,25 @@ Restrictions of this feature:
   deliberately not specified by the language. This includes whether they are static vs. instance or
   exactly what signature they are emitted with.
 
+
+### Operators on Function Pointer Types
+
+The section in unsafe code on operators is modified as such:
+
+> In an unsafe context, several constructs are available for operating on all _pointer\_type_s but _funcptr\_type_s (except where noted):
```
gafter, line 220: Suggest changing "but" to "that are not".  Better yet, because an exception to an exception is confusing, make the one bullet that permits function pointers a separate paragraph not included in this bullet list.

proposals/function-pointers.md
```diff
@@ -212,6 +212,25 @@ Restrictions of this feature:
...
+> *  The `&` operator may be used to obtain the address of a variable ([The address-of operator](unsafe-code.md#the-address-of-operator)).
+> *  The `++` and `--` operators may be used to increment and decrement pointers ([Pointer increment and decrement](unsafe-code.md#pointer-increment-and-decrement)).
+> *  The `+` and `-` operators may be used to perform pointer arithmetic ([Pointer arithmetic](unsafe-code.md#pointer-arithmetic)).
+> *  The `==`, `!=`, `<`, `>`, `<=`, and `=>` operators may be used to compare pointers (including function pointers) ([Pointer comparison](unsafe-code.md#pointer-comparison)).
+> *  The `stackalloc` operator may be used to allocate memory from the call stack ([Fixed size buffers](unsafe-code.md#fixed-size-buffers)).
+> *  The `fixed` statement may be used to temporarily fix a variable so its address can be obtained ([The fixed statement](unsafe-code.md#the-fixed-statement)).
+
+Additionally, we modify all the sections in `Pointers in expressions` to forbid function pointer types, except `Pointer comparison` and `The sizeof operator`, to forbid usage of function pointer types.
```
gafter, line 232: The phrase "to forbid usage of function pointer types" appears twice in this sentence.

333fred:
@gafter addressed feedback.

proposals/function-pointers.md
```diff
@@ -212,6 +212,29 @@ Restrictions of this feature:
...
+> *  The `++` and `--` operators may be used to increment and decrement pointers ([Pointer increment and decrement](unsafe-code.md#pointer-increment-and-decrement)).
+> *  The `+` and `-` operators may be used to perform pointer arithmetic ([Pointer arithmetic](unsafe-code.md#pointer-arithmetic)).
+> *  The `==`, `!=`, `<`, `>`, `<=`, and `=>` operators may be used to compare pointers ([Pointer comparison](unsafe-code.md#pointer-comparison)).
+> *  The `stackalloc` operator may be used to allocate memory from the call stack ([Fixed size buffers](unsafe-code.md#fixed-size-buffers)).
+> *  The `fixed` statement may be used to temporarily fix a variable so its address can be obtained ([The fixed statement](unsafe-code.md#the-fixed-statement)).
+> 
+> In an unsafe context, several constructs are available for operating on all _funcptr\_type_s:
+> *  The `&` operator may be used to obtain the address of static methods ([Allow address-of to target methods](function-pointers.md#allow-address-of-to-target-methods))
+> *  The `==`, `!=`, `<`, `>`, `<=`, and `=>` operators may be used to compare pointers ([Pointer comparison](unsafe-code.md#pointer-comparison)).
```
jkotas, line 234: `<`, `>`, `<=`, and `=>` do not make sense for function pointers. I know that C/C++ allows it, but it is not necessarily a good prior art to follow.

333fred: I wish we could block them, but they're defined on `void*`, and function pointers are implicitly convertible.

jkotas: `==` and `!=` are questionable too. Function pointers in .NET are not stable. For example, `ldftn` on same function can return different pointer values in the two places in the program. It does actually happen in practice, e.g. a typical example is `ldftn M` before M was executed for the first time and after M was executed for the first time.

jkotas: Always emit a warning when any of these are used on function pointers?

proposals/function-pointers.md
```diff
@@ -212,6 +212,29 @@ Restrictions of this feature:
...
+> *  The `&` operator may be used to obtain the address of a variable ([The address-of operator](unsafe-code.md#the-address-of-operator)).
+> *  The `++` and `--` operators may be used to increment and decrement pointers ([Pointer increment and decrement](unsafe-code.md#pointer-increment-and-decrement)).
+> *  The `+` and `-` operators may be used to perform pointer arithmetic ([Pointer arithmetic](unsafe-code.md#pointer-arithmetic)).
+> *  The `==`, `!=`, `<`, `>`, `<=`, and `=>` operators may be used to compare pointers ([Pointer comparison](unsafe-code.md#pointer-comparison)).
+> *  The `stackalloc` operator may be used to allocate memory from the call stack ([Fixed size buffers](unsafe-code.md#fixed-size-buffers)).
+> *  The `fixed` statement may be used to temporarily fix a variable so its address can be obtained ([The fixed statement](unsafe-code.md#the-fixed-statement)).
+> 
+> In an unsafe context, several constructs are available for operating on all _funcptr\_type_s:
+> *  The `&` operator may be used to obtain the address of static methods ([Allow address-of to target methods](function-pointers.md#allow-address-of-to-target-methods))
```
jkotas, line 233: This not operating on funcptr type (contrary to what the previous line says).

Should this rather say that ` The `&` operator may be used to obtain the address of a variable` ?

ie. should the following be possible?

```
delegate*<int> p1 = ...;
var x = &p1;
```

proposals/function-pointers.md
333fred: Hmm, that's interesting. Seems like a good candidate for a warning then. I'll add a note to the tracking issue.

proposals/function-pointers.md
333fred: You can certainly do that, just like you can do that for any local. The operating on is taken from existing examples for pointer types.

jkotas: > You can certainly do that, just like you can do that for any local. 

Good, it is what I have expected.

The formatting of this section makes it sound like that it is not possible. Taking address of a local is explicitly mentioned in the pointers that are not function pointers section above; but it is not mentioned in this section.

PR #3353, opened 2020-04-10
Update records-wip.md

agocke:
Relates to https://github.com/dotnet/roslyn/issues/40726 (test plan for records)

### DIFF
--- a/proposals/records-wip.md
+++ b/proposals/records-wip.md
@@ -53,11 +53,13 @@ At runtime the primary constructor
 
 ### Properties
 
-For each record parameter of a record type declaration there is a corresponding public property member (record property) whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:
+For each record parameter of a record type declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:
 
 For a record struct or a record class:
 
 * A public get-only auto-property is created. Its value is initialized during construction with the value of the corresponding primary constructor parameter. Each "matching" inherited abstract property's get accessor is overridden.
+
+  * This property is also `initonly`, meaning the backing field can be modified in the [with](#With) expression below, if the corresponding `get` accessor is accessible
 
 ### Equality members
 
@@ -92,7 +94,7 @@ listed in the `anonymous_object_initializer`.
 A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
 
 On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
-sequence of assignments, each with a compiler-generated record property (see Properties, above) of the receiver 
+sequence of assignments, each with a compiler-generated record property (see [Properties](#Properties)) of the receiver 
 on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
 implicitly convertible to the type of the property.
 
--- a/proposals/records-wip.md
+++ b/proposals/records-wip.md
@@ -94,11 +94,11 @@ listed in the `anonymous_object_initializer`.
 A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
 
 On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
-sequence of assignments, each with a compiler-generated record property (see [Properties](#Properties)) of the receiver 
-on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
+sequence of assignments, each with an `initonly` property (see [Properties](#Properties)) of the `Clone`
+return type on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
 implicitly convertible to the type of the property.
 
 The evaluation of a `with` expression is equivalent to calling the `Clone` method exactly once,
-and then setting the backing field of each record property in the argument list to its corresponding
+and then setting the backing field of each `initonly` property in the argument list to its corresponding
 expression, in lexical order, using the result of the `Clone` method as the receiver.
 
--- a/proposals/records-wip.md
+++ b/proposals/records-wip.md
@@ -91,14 +91,16 @@ A `with` expression allows for "non-destructive mutation", designed to
 produce a copy of the receiver expression with modifications to properties
 listed in the `anonymous_object_initializer`.
 
-A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
+A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible
+parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
 
 On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
 sequence of assignments, each with an `initonly` property (see [Properties](#Properties)) of the `Clone`
 return type on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
 implicitly convertible to the type of the property.
 
-The evaluation of a `with` expression is equivalent to calling the `Clone` method exactly once,
-and then setting the backing field of each `initonly` property in the argument list to its corresponding
-expression, in lexical order, using the result of the `Clone` method as the receiver.
+The evaluation of a `with` expression calls the `Clone` method exactly once,
+and then sets the backing field of each `initonly` property in the argument list to 
+the value of its corresponding expression, in lexical order, using the result of the `Clone` method
+invocation as the receiver.
 
--- a/proposals/records-wip.md
+++ b/proposals/records-wip.md
@@ -99,3 +99,4 @@ of the left-hand side.
 The evaluation of a `with` expression is equivalent to calling the `Clone` method exactly once,
 and then setting the backing field of each record property in the argument list to its corresponding
 expression, in lexical order, using the result of the `Clone` method as the receiver.
+
--- a/proposals/records-wip.md
+++ b/proposals/records-wip.md
@@ -96,8 +96,8 @@ parameterless instance method called `Clone` whose return type must be the recei
 
 On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
 sequence of assignments, each with an `initonly` property (see [Properties](#Properties)) of the `Clone`
-return type on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
-implicitly convertible to the type of the property.
+return type on the left-hand side of the assignment (as a property invocation), and an arbitrary expression 
+on the right-hand side which is implicitly convertible to the type of the property.
 
 The evaluation of a `with` expression calls the `Clone` method exactly once,
 and then sets the backing field of each `initonly` property in the argument list to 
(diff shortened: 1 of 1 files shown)


gafter:
> A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the type of the receiver express type, or a base type thereof.

I have no idea what this means.

---
Refers to: proposals/records-wip.md:92 in e67167a. [](commit_id = e67167a4abc988361973ce6f483839f7a46b44d3, deletion_comment = False)

gafter:
> sequence of assignments with a compiler-generated record property of the receiver on the left-hand side of the

I don't know what this means.

---
Refers to: proposals/records-wip.md:95 in e67167a. [](commit_id = e67167a4abc988361973ce6f483839f7a46b44d3, deletion_comment = False)

proposals/records-wip.md
```diff
@@ -99,3 +99,4 @@ of the left-hand side.
 The evaluation of a `with` expression is equivalent to calling the `Clone` method exactly once,
 and then setting the backing field of each record property in the argument list to its corresponding
```
gafter, line 100: >backing field [](start = 21, length = 13)

The backing field is not typically accessible to the caller.

gafter, line 100: >record property [](start = 43, length = 15)

I don't know what a "record property" is. Also, the return type of the `Clone` method might not be a type that has the named property as a member.

gafter:
> A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the type of the receiver express type, or a base type thereof.

Record types, as defined herein, do not typically contain such a `Clone` method, so this section doesn't make sense in the context of this spec.

---
Refers to: proposals/records-wip.md:92 in e67167a. [](commit_id = e67167a4abc988361973ce6f483839f7a46b44d3, deletion_comment = False)

agocke:
I modified the phrasing to clarify that the return type of a Clone method must be the receiver type or a base type.

proposals/records-wip.md
agocke: Defined a record property above. The record property that is compiler-generated if one of equivalent signature does not exist.

agocke: It does if it's generated!

gafter: There is no requirement that it be a record type. Moreover, I don't know what it means for a type to be a record type from the point of view outside the type itself.

---
In reply to: [406952410](https://github.com/dotnet/csharplang/pull/3353#discussion_r406952410) [](ancestors = 406952410)

agocke:
> Record types, as defined herein, do not typically contain such a Clone method, so this section doesn't make sense in the context of this spec.

Seems well-defined to me. Records may implicitly define a Clone method in a future revision of the spec, but I haven't gotten there yet.

proposals/records-wip.md
agocke: If it's not a record type, there are no generated record members. It's just a consequence of the precondition.

proposals/records-wip.md
```diff
@@ -89,13 +89,14 @@ A `with` expression allows for "non-destructive mutation", designed to
...
-A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the type of the receiver express type, or a base type thereof.
+A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
 
 On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
-sequence of assignments with a compiler-generated record property of the receiver on the left-hand side of the
-assignment, and an arbitrary expression on the right-hand side which is implicitly convertible to the type
-of the left-hand side.
+sequence of assignments with a compiler-generated record property (see Properties, above) of the receiver 
+on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
+implicitly convertible to the type of the left-hand side.
```
gafter, line 97: >left-hand side [](start = 42, length = 14)

You mean the type of the property named on the left-hand-side?  (Only expressions have types)

proposals/records-wip.md
gafter: Not according to the spec above.  Per spec, an auto-property is generated, and there is no accessible backing field in that case.

---
In reply to: [406952450](https://github.com/dotnet/csharplang/pull/3353#discussion_r406952450) [](ancestors = 406952450)

gafter: Are you saying that the "record"ness of a type and the "generated record property"ness of a property, are visible to the language externally and across compilation boundaries?  If so, I would expect to see that in the implementation as well.

---
In reply to: [406953742](https://github.com/dotnet/csharplang/pull/3353#discussion_r406953742) [](ancestors = 406953742)

gafter:
> sequence of assignments with a compiler-generated record property of the receiver on the left-hand side of the

I think you need to identify the lookup rules and the error cases.

---
In reply to: [612224166](https://github.com/dotnet/csharplang/pull/3353#issuecomment-612224166) [](ancestors = 612224166)

---
Refers to: proposals/records-wip.md:95 in e67167a. [](commit_id = e67167a4abc988361973ce6f483839f7a46b44d3, deletion_comment = False)

proposals/records-wip.md
```diff
@@ -89,13 +89,14 @@ A `with` expression allows for "non-destructive mutation", designed to
...
 
 The evaluation of a `with` expression is equivalent to calling the `Clone` method exactly once,
 and then setting the backing field of each record property in the argument list to its corresponding
 expression, in lexical order, using the result of the `Clone` method as the receiver.
```
gafter, line 101: >using the result of the `Clone` method as the receiver [](start = 30, length = 54)

The `Clone` method might not have that property as a member (in the case that it returns a base type).

proposals/records-wip.md
agocke: That would seem to be true! However, the implementation is a prototype, and only implements a subset of the spec 😄

agocke: There is a backing field, and the semantics are defined st. it were as though the backing field were set. There's no requirement that the operation being performed is a legal C# assignment, so there's no requirement that the backing field is accessible.

gafter: I thought you were only specifying what you're implementing.

---
In reply to: [406956349](https://github.com/dotnet/csharplang/pull/3353#discussion_r406956349) [](ancestors = 406956349)

agocke: Alternatively, I could phrase this as "the property is `init`-only, but I don't have a full specification there either, so it seems like either way we'll end up with some specification gap.

This phrasing makes sense... it's just broader than will eventually be implemented.

proposals/records-wip.md
agocke: Yup! Fixed.

proposals/records-wip.md
gafter: You probably want to say some of this in the spec.

---
In reply to: [406956451](https://github.com/dotnet/csharplang/pull/3353#discussion_r406956451) [](ancestors = 406956451)

gafter:
I'm having a hard time making sense of the overall specification with just the pieces you're revealing. I understand you're only implementing some pieces, but without the missing pieces of the spec... the things that are in the spec you're providing don't fit together.

proposals/records-wip.md
agocke: > You probably want to say some of this in the spec. 

Fair! I'll try to clarify that where some of the existing language rules are being bypassed.

> Moreover, it isn't clear how this could be mapped onto verifiable IL.

IL verifiability isn't well-defined in the C# language specification, so I'll leave that for future work.

agocke:
Yup, I'll try to clarify this a bit more. It may be better to just write this using "init-only" terminology and then note the implementation gaps in the PR.

proposals/records-wip.md
agocke: Fixed. The properties should in the return type of `Clone`, not the receiver type

agocke:
Added some notes about `initonly` without going into too much detail.

I'm not clear what you mean by comments on the lookup rules. The lookup for `Clone()` is just straightforward method lookup, and the property lookup is equivalent to lookup for a property `get`

gafter:
We don't have a "lookup for a property get".  We just do a simple name lookup (or qualified name lookup).

proposals/records-wip.md
gafter: When the spec defines something as "equivalent to" something else, that includes legality of doing that something else in C#.  If instead you're defining the runtime semantics, just say that the backing field is assigned (no "equivalence" is involved).

---
In reply to: [406956451](https://github.com/dotnet/csharplang/pull/3353#discussion_r406956451) [](ancestors = 406956451)

proposals/records-wip.md
```diff
@@ -89,13 +91,16 @@ A `with` expression allows for "non-destructive mutation", designed to
 produce a copy of the receiver expression with modifications to properties
 listed in the `anonymous_object_initializer`.
 
-A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the type of the receiver express type, or a base type thereof.
```
gafter, line 92: >express [](start = 204, length = 7)

What does "express" mean here?

proposals/records-wip.md
gafter: As I understand the initonly feature, its backing field cannot be set directly by the caller. But its setter can be called during the initialization phase. If you intend the `with` expression to be part of the initialization phase, you need to specify where it begins, what value's initialization field it is the initialization phase *for*, and where it ends.

---
In reply to: [406950140](https://github.com/dotnet/csharplang/pull/3353#discussion_r406950140) [](ancestors = 406950140)

proposals/records-wip.md
```diff
@@ -59,6 +59,8 @@ For a record struct or a record class:
 
 * A public get-only auto-property is created. Its value is initialized during construction with the value of the corresponding primary constructor parameter. Each "matching" inherited abstract property's get accessor is overridden.
 
+  * This property is also `initonly`, meaning the backing field can be modified in the [with](#With) expression below, if the corresponding `get` accessor is accessible
```
333fred, line 62: Perhaps 
```suggestion
  * This property also has an `init` accessor, meaning the backing field can be modified in the [with](#With) expression below, if the `init` accessor is accessible
```
given that we're going with `init` being the keyword?

proposals/records-wip.md
```diff
@@ -89,13 +91,16 @@ A `with` expression allows for "non-destructive mutation", designed to
...
 
-A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible parameterless instance method called `Clone` whose return type must be the type of the receiver express type, or a base type thereof.
+A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible
+parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
 
 On the right hand side of the `with` expression is an `anonymous_object_initializer` with a
-sequence of assignments with a compiler-generated record property of the receiver on the left-hand side of the
-assignment, and an arbitrary expression on the right-hand side which is implicitly convertible to the type
-of the left-hand side.
+sequence of assignments, each with an `initonly` property (see [Properties](#Properties)) of the `Clone`
```
333fred, line 98:
```suggestion
sequence of assignments, each with an `init` property (see [Properties](#Properties)) of the `Clone`
```

proposals/records-wip.md
```diff
@@ -89,13 +91,16 @@ A `with` expression allows for "non-destructive mutation", designed to
...
+return type on the left-hand side of the assignment, and an arbitrary expression on the right-hand side which is 
+implicitly convertible to the type of the property.
+
+The evaluation of a `with` expression calls the `Clone` method exactly once,
+and then sets the backing field of each `initonly` property in the argument list to
```
333fred, line 103:
```suggestion
and then sets the backing field of each `init` property in the argument list to 
```
And perhaps it should be calling the `init` accessor instead of the backing field?

proposals/records-wip.md
agocke: Out of scope for now, I'm just trying to make records work, not user-declared `initonly` accessors.

PR #3361, opened 2020-04-15
Add draft spec for C# pattern-matching changes.

gafter:


proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+## Relational Patterns
+
+Relational patterns permit the programmer to express that an input value must satisfy a relational constraint when compared to a constant value:
+
+``` C#
+    public static LifeStage LifeStageAtAge(int age) => age switch
+    {
+        < 0 =>  LiftStage.Prenatal,
```
333fred, line 31:
```suggestion
        < 0 =>  LifeStage.Prenatal,
```

proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+        < 6 =>  LifeStage.EarlyChild,
+        < 12 => LifeStage.MiddleChild,
+        < 20 => LifeStage.Adolescent,
+        < 40 => LifeStage.EarlyAdult,
+        < 65 => LifeStage.MiddleAdult,
+        _ =>    LifeStage.LateAdult,
+    };
+```
+
+We imagine supporting `<`, `<=`, `>`, and `>=` patterns on all of the built-in types that support such binary relational operators with two operands of the same type in an expression. Specifically, we support all of these relational patterns for `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, and `decimal`.
```
333fred, line 43: Consider dropping the `imagine`, and be more concrete.

proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+    ;
+relational_pattern
+    : '<' relational_expression
+    | '<=' relational_expression
+    | '>' relational_expression
+    | '>=' relational_expression
+    ;
+```
+
+The expression is required to evaluate to a constant value.  It is an error if that constant value is `double.NaN` or `float.NaN`.  It is an error if the expression is a null constant and the relational operator is `<`, `<=`, `>`, or `>=`.
```
333fred, line 57: Is it worth specifying what would happen if you did `< double.NegativeInfinity` and related infinite constants? We could want to consider them a compile-time error?

proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+
+When the input is a type for which a suitable built-in binary relational operator is defined that is applicable with the input as its left operand and the given constant as its right operand, the evaluation of that operator is taken as the meaning of the relational pattern.  Otherwise we convert the input to the type of the expression using an explicit nullable or unboxing conversion.  It is a compile-time error if no such conversion exists.  The pattern is considered not to match if the conversion fails.  If the conversion succeeds then the result of the pattern-matching operation is the result of evaluating the expression `e OP v` where `e` is the converted input, `OP` is the relational operator, and `v` is the constant expression.
```
333fred, line 59: > explicit nullable or unboxing conversion

What if there's an implicit user-defined conversion to the appropriate type? For example, if you have a custom fractional struct, you could define a conversion operator to decimal, and it would be nice if you could then pattern match on that type. Given that implicit conversions are supposed to never lose information, it seems like this should be fine.

proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+
+Pattern *combinators* permit matching both of two different patterns using `and` (this can be extended to any number of patterns by the repeated use of `and`), either of two different patterns using `or` (ditto), or the *negation* of a pattern using `not`.
+
+A common use of a combinator will be the idiom
+
+``` c#
+if (e is not null) ...
+```
+
+More readable than the current idiom `e is object`, this pattern clearly expresses that one is checking for a non-null value.
```
333fred, line 71: Worth stating that this is a pure null check?

proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+## Open Issues with Proposed Changes
+
+### Syntax for relational operators
+
+Are `and`, `or`, and `not` some kind of contextual keyword?  If so, is there a breaking change (e.g. compared to their use as a designator in a *declaration-pattern*).
+
+Should we support some combination of declaration pattern along with a relational pattern?  For example,
+
+``` csharp
+if (o is int x <= 100) // x is an int with value < 100 here.
```
333fred, line 117: fwiw, looking at this code sample I really don't like it. I much prefer the `and` version.

proposals/patterns3.md
333fred: Actually, after considering for a bit, this is probably a very bad thing that would lead to inconsistent behavior.

proposals/patterns3.md
```diff
@@ -0,0 +1,253 @@
...
+Here, `0 or 1 or` adds nothing to the second case, as those values would have been handled by the first case.  This too deserves an error.
+
+``` csharp
+byte b = ...;
+int x = b switch { <100 => 0, 100 => 1, 101 => 2, >101 => 3 };
+```
+
+A switch expression such as this should be considered *exhaustive* (it handles all possible input values).
+
+In C# 8.0, a switch expression with an input of type `byte` is only considered exhaustive if it contains a final arm whose pattern matches everything (a *discard-pattern* or *var-pattern*).  Even a switch expression that has an arm for every distinct `byte` value is not considered exhaustive in C# 8.  In order to properly handle exhaustiveness of relational patterns, we will have to handle this case too.  This will technically be a breaking change, but one no user is likely to notice.
```
RikkiGibson, line 253:
```suggestion
In C# 8.0, a switch expression with an input of type `byte` is only considered exhaustive if it contains a final arm whose pattern matches everything (a *discard-pattern* or *var-pattern*).  Even a switch expression that has an arm for every distinct `byte` value is not considered exhaustive in C# 8.  In order to properly handle exhaustiveness of relational patterns, we will have to handle this case too.  This will technically be a breaking change, but no user is likely to notice.
```

proposals/patterns3.md
gafter: There is nothing special about that to specify.  If not combined with some other pattern in an `or`, such a switch case would be naturally unreachable.

proposals/patterns3.md
gafter: The `null` pattern is already a pure null check.  Placing it under a negation doesn't change that.  In any case, there is no specification of the concept of a pure null check, so there would be no term whose definition could be referenced.

proposals/patterns3.md
gafter: Me too, which is why we didn't do this.

proposals/patterns3.md
gafter: It sounds like you're quoting this paragraph verbatim. Is that what you intended?

---
In reply to: [409223574](https://github.com/dotnet/csharplang/pull/3361#discussion_r409223574) [](ancestors = 409223574)

gafter:
I think I've addressed all of the review comments.

proposals/patterns3.md
```diff
@@ -0,0 +1,273 @@
...
+
+``` csharp
+if (e is not int i) return;
+M(i); // is i definitely assigned here?
+```
+
+This does not work today because, for an *is-pattern-expression*, the pattern variables are considered *definitely assigned* only where the *is-pattern-expression* is true ("definitely assigned when true").
+
+Supporting this would be simpler (from the programmer's perspective) than also adding support for a negated-condition `if` statement.  Even if we add such support, programmers would wonder why the above snippet does not work.  On the other hand, the same scenario in a `switch` makes less sense, as there is no corresponding point in the program where *definitely assigned when false* would be meaningful.  Would we permit this in an *is-pattern-expression* but not in other contexts where patterns are permitted?  That seems irregular.
+
```
CyrusNajmabadi, line 201: note (from our talk today): i think we should support this.  users will be able to say any of:

```
if (e is int)
if (e is not int)
if (e is int i)
// but not
if (e is not int i)
```

Despite that they *can* write:

```
if (!(e is int)) -> if (e is not int)
// but not
if (!(e is int i)) -> if (e is not int i)
```

It's a very strange inconsistency.

gafter: We could do that but ONLY (1) for the is-pattern expression (no other places patterns could be used), and only (2) if the `not` is at the top level only. For other cases, there isn't any clear place that the pattern variables would be definitely assigned. If we did this, people would find these restrictions to be strange. I'm OK considering this relaxation as a separate proposal to be considered later.

333fred: Agreed with Cyrus. It may make the spec convoluted, but I think the regularity would be worth it. We should definitely put this in the open issues tracker to make sure we don't forget about it.

gafter: It is on the LDM agenda.

Issue #3323, opened 2020-03-30
Added: LDM Notes for March 30, 2020


agocke:
## March 30, 2020

1. Record Monday

[C# Language Design Notes for March 30, 2020](https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-03-30.md)

CyrusNajmabadi:
@agocke incorrect link?

333fred:
@CyrusNajmabadi I just fixed it.

CyrusNajmabadi:
Thanks!

YairHalberstadt:
> No separable value equality for non-records, right now. Default record equality is defined as field-based shallow equality.

Strongly agreed.

For me the most important aspects of records is that I can write something in one line as short as:

`record Request(int id, string url, data[] Data)` and have it just work. That I want inbuilt into the language, because **in scala I use it multiple times in almost every single file!**

I don't want records to be built of separable features because I would almost never use them. There's so many options once we step out of record land that it needs to be user customizable, and source generators seem the perfect way to do that.

qrli:
> Recognizing the modreq and allowing access to init-only members, and calling any necessary "validator" on construction, are pretty simple features for VB. 

I have never seen the "validator" feature design. Is it mentioned anywhere?

YairHalberstadt:
@qrli
https://github.com/jaredpar/csharplang/blob/record/proposals/recordsv3.md

YairHalberstadt:
Also, @Agocke, thanks for putting up the meeting notes so quickly - it's really appreciated!

Joe4evr:
Please consider allowing the use of `init` for non-readonly properties as well, since that would be the general purpose solution to #2328. I would really hate it if such a useful building block is only good for "Popsicle immutable" properties while cases where one works with data requiring a setter get left out in the cold *again*.

orthoxerox:
> Default record equality is defined as field-based shallow equality.

And I think this is absolutely fine. People who need sequence equality for list members can either override `Equals` or write their own struct wrapper around `List<T>` that defines `Equals` as sequence equality.

qrli:
Thanks for the link @YairHalberstadt 

After reading that, I feel the considered use cases for "validators" may be a bit less than I expected. Maybe it is the name, but to me it is not just "validators". It may also do:
- Precompute something and cache it in some mutable field
- Normalize the input
- Marshal data to be in a slightly different structure to be more efficient for read later

[...]

PR #3367, opened 2020-04-17
Init only proposal document

jaredpar:


proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+class Derived : Base
+{
+    internal readonly int DerivedField;
+    internal int DerivedProperty
+    {
+        get => DerivedField;
+        init
+        {
+            DerivedField = 42; // Okay
```
333fred, line 171: Consider assigning `Property` in this one as an example of what would be ok.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+    public Derived()
+    {
+        Property = 42; // Okay 
+        Field = 13; // Error Field is readonly
+    }
+}
+```
+
+When `init` is used in a virtual property then all the overrides must also
```
333fred, line 184: Consider adding an example of this.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+        {
+            Name = "Jared"
+        };
+        local.Name = "Jraed"; // Error
+    }
+}
+```
+
+Restrictions of this feature:
+- The `init` accessor can only be used instance properties
```
333fred, line 211:
```suggestion
- The `init` accessor can only be used on instance properties
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+- A property cannot contain both an `init` and `set` accessor
+- All overrides of a property must have `init` if the base had `init`. This rule
+also applies to interface implementation.
```
333fred, line 214: Interface implementations today allow you to only specify `get`, but add a `set` in the implementation. I think it's work allowing the same thing for specifying `init` in the interface but allowing `set` in the implementation.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+### Metadata encoding 
+Property `init` accessors will be emitted as a standard `set` accessor with
+the return type marked with a modreq of `IsInitOnly`. This is a new type
+which will have the following definition:
+
+```cs
+namespace System.Runtime.CompilerServices
+{
+    [AttributeUsage(AttributeTargets.All)]
+    public sealed class InitOnlyAttribute : Attribute
```
333fred, line 225: Take updates from https://github.com/dotnet/runtime/issues/34978#issuecomment-614845405.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+    {
+
+    }
+}
+```
+
+The compiler will match the attribute by full name. There is no requirement
```
333fred, line 232:
```suggestion
The compiler will match the type by full name. There is no requirement
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+that it appear in the core library. In the case there are mulitple attributes
```
333fred, line 233:
```suggestion
that it appear in the core library. If there are multiple attributes
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+by this name available then the compiler will pick the one defined in the core 
+library should one exist.
```
333fred, line 235: And if one does not?

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+## Questions
+
+### Breaking changes
+One of the main pivot points in how this feature is encoded will come down to
+the following question: 
+
+> Is it a binary breaking change to remove `init` from a `set`?
+
+Removing `init`, and thus making a field or property fully writable is never
```
333fred, line 245:
```suggestion
Replacing `init` with `set`, and thus making a field or property fully writable is never
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+a source breaking change. For fields it is never a binary breaking change
```
333fred, line 246: Consider removing `fields`, it's not in the actual proposal.

proposals/init.md
333fred: And everywhere you say `remove` below.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+Removing `init`, and thus making a field or property fully writable is never
+a source breaking change. For fields it is never a binary breaking change 
+either. Additionally removing it from a field is never a binary breaking 
+change. The only behavior up in question is whether or not this remains 
+true for a property. 
+
+If we want to make the removal of `init` from a property a compatible change
+then it will force our hand on the modreq vs. attributes decision below. If 
+one the other hand this is seen as a non-interesting scenario then this will
```
333fred, line 253:
```suggestion
then it will force our hand on the modreq vs. attributes decision below. If, 
on the other hand, this is seen as a non-interesting scenario then this will 
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+of the application.
+
+Using attributes to annotate the `set` accessor means that only compilers which
+understand the attribute will know to limit access to it. A compiler unaware 
+of `init` will see it as a simple read / write property and allow access.
+
+This would seemingly mean this decision is a choice between extra safety at 
+the expense of binary compatibility. Digging in a bit the extra safety is not
+exactly what it seems. It will not for instance protect against the following
+circumstances:
```
333fred, line 281:
```suggestion
exactly what it seems. It will not protect against the following circumstances:
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+1. Reflection over `public` members
+1. The use of `dynamic` 
+1. Compilers that don't recognize modreqs
+
+It should also be considered that when we complete the IL verification rules
```
333fred, line 287:
```suggestion
It should also be considered that, when we complete the IL verification rules 
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+the expense of binary compatibility. Digging in a bit the extra safety is not
+exactly what it seems. It will not for instance protect against the following
+circumstances:
+
+1. Reflection over `public` members
+1. The use of `dynamic` 
+1. Compilers that don't recognize modreqs
+
+It should also be considered that when we complete the IL verification rules 
+for .NET 5, `init` will be one of those rules. that means extra enforcement
```
333fred, line 288:
```suggestion
for .NET 5, `init` will be one of those rules. That means extra enforcement 
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+1. The use of `dynamic` 
+1. Compilers that don't recognize modreqs
+
+It should also be considered that when we complete the IL verification rules 
+for .NET 5, `init` will be one of those rules. that means extra enforcement 
+will be gained from simply verifying compilers emitting verifiable IL.
+
+The primary languages for .NET (C#, F# and VB) will all be updated to 
+recognize these `init` accessors. Hence the only realistic scenario here is 
+when a C# 9 compiler emits `init` properties and they are seen by a C# 8
```
333fred, line 293: Or an older VB compiler

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+The primary languages for .NET (C#, F# and VB) will all be updated to 
+recognize these `init` accessors. Hence the only realistic scenario here is 
+when a C# 9 compiler emits `init` properties and they are seen by a C# 8 
+compiler. That is the trade off to consider and weigh against binary 
+compatibility.
+
+Note: this discussion applies to the accessor members only, not to fields. There
+is no value to be gained by using a modreq on a field. The `init` feature for 
+fields is a relaxation of an existing rule. All existing compilers already
```
333fred, line 299: We're also not doing the feature.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+fields is a relaxation of an existing rule. All existing compilers already 
+support `readonly` and hence an attribute serves fine as a way to alert them
+that write access can be extended in certain circumstances.
+
+**Resolution**
+The feature will use a modreq to encode the property `init` setter. The
+compelling factors were (in no particular order):
+
+1. Desire to discourage older compilers from violating `init` semantics
+1. Desire to make adding or removing `init` in a `virtual` declaratio or
```
333fred, line 308:
```suggestion
* Desire to discourage older compilers from violating `init` semantics
* Desire to make adding or removing `init` in a `virtual` declaratio or 
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+binary compatible change it made the choice of using modreq straight forward.
+
+### init vs. initonly
+Syntax debate time.
+
+**Resolution**
+There was no other syntax which was overwelming favored in LDM. For the time 
+being we will be moving forward with the initial design laid out in this 
+document. This may change as the feature progresses and LDM members have time
+to digest the design here.
```
333fred, line 321: Consider talking about `init set` vs `init`, and us being open to making `init` a shorthand for `init set` in the future if we decide to make `init` a modifier for functions.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+init struct Point
+{
+    public int X;
+    public int Y;
+}
+```
+
+**Resolution** 
+This feature is too *cute* here and conflicts with the `readonly struct`
```
333fred, line 372: Consider removing the `too *cute*` part.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+    public int X;
+    public int Y;
+}
+```
+
+**Resolution** 
+This feature is too *cute* here and conflicts with the `readonly struct` 
+feature on which it is based. The `readonly struct` feature is simple in that 
+it applies `readonly` to all members: fields, methods, etc ... The 
+`init struct` feature would only apply to fields. This actually ends up making
```
333fred, line 375:
```suggestion
`init struct` feature would only apply to properties. This actually ends up making
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+it confusing for users. 
+
+Given that `init` is only valid on certain aspects of a type we rejected the 
+idea of having it as a type modifier.
+
+## Allow init as a field modifier
```
333fred, line 381: Consider moving this before the previous section, given that the decision was based on `init fields` not being a thing.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+In metadata these fields would be marked in the same way as `readonly` fields 
+but with an additional attribute or modreq to indicate they are `init` style
+fields. 
+
+**Resolution**
+LDM agrees this proposal is sound but overal the scenario felt disjoint from 
+properties. The decision was to proceed with properties only for now but this 
+will be reconsidered if there is significant customer feedback that justifies
+the scenario.
```
333fred, line 411: Consider also mentioning that this discussion was when we decided to make `readonly` fields settable in initers on the type.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+    public Name(string first, string last)
+    {
+        First = first;
+        Last = last;
+    }
+}
+```
+
+It is not a breaking change to use `init` in place of `readonly` here:
```
333fred, line 436: True for structs as well?

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+
+    public Name(string first, string last)
+    {
+        First = first;
+        Last = last;
+    }
+}
+```
+
+The same applies for changing a `get` only property to have an `init set`
```
333fred, line 452:
```suggestion
The same applies for changing a `get` only property to have an `init` 
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+        Last = last;
+    }
+}
+```
+
+The same applies for changing a `get` only property to have an `init set` 
+accessor.
+
+### IL verification
+When .NET Core decides to re-implement IL verify the rules will need to be
```
333fred, line 456:
```suggestion
When .NET Core decides to re-implement IL verification the rules will need to be 
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+adjusted to account for `init` members. This will need to be included in the 
+rule changes for non-mutating acess to `readonly` data.
+
+## init members
+The `init` modifier could be extended to apply to all instance members. This 
+would generalize the concept of `init` during object construction and allow
+types to declare helper methods that could partipate in the construction 
+process to initialize `init` fields and properties.
+
+Such members would have all the restricions that an `init set` accessor does
```
333fred, line 466:
```suggestion
Such members would have all the restricions that an `init` accessor does
```

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+The consumer can then use object initializers to create the object
+
+```cs
+var p = new Point() { X = 42, Y = 13 };
+```
+
+## Detailed Design
+
+### init members
+An init only property is declared by using the `init` accessor in place of the
```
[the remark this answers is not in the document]
jaredpar: Once I commit your edits I'll do another pass to make this more consistent.

proposals/init.md
```diff
@@ -0,0 +1,490 @@
...
+    public string FirstName { get; init; }
+    public string LastName { get; init; }
+}
+```
+
+An instance property containing an `init` accessor is considered settable in
+the following circumstances:
+
+- During an object initializer
+- Inside an instance constructor of the containing or derived type
```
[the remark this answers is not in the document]
jaredpar: Don't want to get into validators here since they don't have a concrete design. Think it will be more appropriate for validators to call out the expanded places `init` accessors can be called just as the init only proposal calls out the expanded places `readonly` fields can be mutated.

proposals/init.md
jaredpar: I also need to remove the use of attributes in this sentence

proposals/init.md
jaredpar: I want to leave the field discussion in because I think it's important for the entirety of the argument here. The argument is meant to be more general. I antincipate referring to it again when we discuss modreqs for future features. I will add a mention that fields aren't being done though.

proposals/init.md
jaredpar: Have a typo on `declaratio` I need to fix here.

proposals/init.md
jaredpar: I can definitely get the motivation for allowing a class to implement a `get` only property with `get; init;`. That follows from `set` and seems pretty reasonable. I'm less certain about allowing a `get; init;` property be implemented with `get; set;`. It would require us to generate an explicit implementation of the interface under the hood (modreq would break the signature match). That makes me hesitate. I think I'd need to see some user feedback to want to take that on.

proposals/init.md
jaredpar: Huh ... I changed this in my machine. Must not have saved before submitting.

proposals/init.md
jaredpar: Overall I think we should follow the same rules we have for looking up the `Extensions` attribute when you define an extension method. That has a lot of the same similarities here.

jaredpar:
Thanks for the feedback @333fred. Got through most of it but still have a bit left. Pushed updates for all the parts I got through.

proposals/init.md
jaredpar: This entire section is too field centric. Need to pretty much completely overhaul this.

jaredpar:
@333fred responded to all feedback now. One comment is still open for feedback from you but everything else I've cleaned up. Think I took pretty much all your suggestions. 😄

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+```cs
+class Complex
+{
+    readonly int Field1;
+    int Field2;
+    int Prop1 { get; init ; }
+    int Prop2
+    {
+        get => 42;
+        set
```
poke, line 141: This should be `init` right? Otherwise I don't get why those assignments would be valid.

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+those fields declared on the same type as the accessor. It cannot be used to 
+assign `readonly` fields in a base type. This rule ensures that type authors
+remain in control over the mutability behavior of their type. Developers who do
+not wish to utilize `init` cannot be impacted from other types choosing to
+do so:
+
+```cs
+class Base
+{
+    internal readonly int Field;
```
poke, line 161: Maybe make the field lower-case?

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+}
+
+class C2 : Base
+{
+    // Error: Property must have init to override Base.Property
+    public override int Property { get; set; }
+}
+```
+
+An `interface` declaration can also particpate in `init` style initalization
```
poke, line 215: Typo: “participate”

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+If we want to make the change of `init` to `set` a source and binary compatible
+change then it will force our hand on the modreq vs. attributes decision
+below because it will rule out modreqs as a soulution. If on the other hand
+this is seen as a non-interesting then this will make the modreq vs. attribute
+decision less impactful.
+
+**Resolution**
+This scenario is not seen as compelling by LDM.
+
+### Modreqs vs. attributes
```
poke, line 291: Do you have a link that explains what a modreq is? This is a new term for me personally.

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+`readonly` + an attribute there is no risk of older compilers mis-using the 
+field because they would already recognize `readonly`. Hence using a modreq here
+doesn't add any extra protection.
+
+**Resolution**
+The feature will use a modreq to encode the property `init` setter. The
+compelling factors were (in no particular order):
+
+* Desire to discourage older compilers from violating `init` semantics
+* Desire to make adding or removing `init` in a `virtual` declaratio or
```
poke, line 342: Typo “declaration”

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+**Resolution**
+There will be no warning on consumption of `init` fields and properties.
+
+LDM wants to have a broader discussion on the idea of required fields and
+properties. That may cause us to come back and reconsider our position on
+`init` members and validation.
+
+## Allow init as a field modifier
+In the same way `init` can server as a property accessor it could also serve as
```
poke, line 408: Typo: “serve”?

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+```cs
+class Complex
+{
+    readonly int Field1;
+    int Field2;
+    int Prop1 { get; init ; }
+    int Prop2
+    {
+        get => 42;
+        set
```
daniel-white, line 141: Shouldn’t this be `init`?

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+s.FirstName = "Jared"; // Error FirstName is readonly
+```
+
+In metadata these fields would be marked in the same way as `readonly` fields 
+but with an additional attribute or modreq to indicate they are `init` style
+fields. 
+
+**Resolution**
+LDM agrees this proposal is sound but overal the scenario felt disjoint from
```
poke, line 434: Typo: “overall”

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+    {
+        var d = new Derived() { Value = true; };
+    }
+}
+
+```
+
+At the point a `init` accessor is invoked the instance is known to be 
+in the open construction phase. Hence an `init` accessor is allowed to take 
+the following actions in addition to what a normal `set` accessor cand do:
```
daniel-white, line 127: can do?

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+`init struct` feature would only apply to properties. This actually ends up making
+it confusing for users. 
+
+Given that `init` is only valid on certain aspects of a type we rejected the 
+idea of having it as a type modifier.
+
+## Considerations
+
+### Compatibility
+The `init` feature is designed to be compatible with existin `get` only
```
poke, line 476: Typo: “existing”

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+### IL verification
+When .NET Core decides to re-implement IL verification the rules will need to be 
+adjusted to account for `init` members. This will need to be included in the 
+rule changes for non-mutating acess to `readonly` data.
+
+The IL verification rules will need to be broken into two parts: 
+
+1. Allowing `init` members to set a `readonly` field.
+1. Determining when an `init` member can be legally called.
```
poke, line 521: `2.`?

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+The first is a simple adjustmetn to the existing rules. The IL verifier can
```
poke, line 523: Typo: “adjustment”

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+the rule is straight forward. It should be legal to call `init` members when 
+the result of a `new` expression is still on the stack. That is until the 
+value has been stored in a local, array element or field or passed as an
+argument to another method it will still be legal to call `init` members. This
+ensures that once the result of the `new` expression is published to a named
+identifier (other than `this`) then it will no longer be legal to call `init`
+members. 
+
+The more complicated case though is when we mix `init` members, object
+initializers and `await`. That can cause the newly craeted object to be
```
poke, line 537: Typo: “created”

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+This has the potential advantage of allowing the use of modreq to enforce 
+correctness while maintaining binary compatibility. The implementation would
+roughly be the following:
+
+1. An `init` accessor is always emitted if there is a `set`. When not defined 
+by the developer it is simply a reference to `set`. 
+1. The set of a property in an object initializer will always use `init` if 
+present but fall back to `set` if it's missing.
+
+This means that a developer can always safely delete `init` from a property.
```
poke, line 576: This assumes that the property also has a setter, right? For init-only properties this would not work, right?

proposals/init.md
jaredpar: Fixed

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+    Name = await SomeMethod()
+};
+```
+
+Here the result of `new Student()` will be hoised into a state machine as a 
+field before the set of `Name` occurs. The compiler will need to mark such
+hoisted fields in a way that the IL verifier understands they're not user 
+accessible and hence doesn't violate the intended semantics of `init`.
+
+## init members
```
poke, line 552: Should this be a level 3 heading?

proposals/init.md
jaredpar: For design docs I usually leave all members as PascalCased. Feel like it makes the document easier to read.

proposals/init.md
jaredpar: Thanks! Noticed you added a lot of typo style suggestions. If you want you can always use the "Insert Suggestion" for these types of items. The upsides are I can just click to accept your suggestion and your GitHub alias ends up in the document history. 

The click to accept is great because it's much harder for me to insert a different typo when fixing the first :smile:

image

proposals/init.md
jaredpar: I use `1.` for all of my markdown lists. The actual markdown rendering engine will replace them with the correctly numbered lists. Given that having them all labeled as `1.` makes re-ordering elements in the document easier because I don't have to re-number after re-arranging.

proposals/init.md
jaredpar: Yeah it's a term very few understand. Probably only people who've had to deal with metadata encoding / decoding.

https://twitter.com/jaredpar/status/1247014490069020672

I don't have a good link to a document describing what a modreq or modopt is other than some obscure passages in the ECMA CLI document [(here)](https://twitter.com/ben_a_adams/status/1247016887600533505). Should probably write something up about it at some point. 

Probably the easiest way to think about them is they're like attributes on method signatures but they're actually a part of the signature. That its an attribute that must be repeated in the same position whenever a method is overridden or when a method implements an interface members. They have the further requirement than if a compiler doesn't understand the type inside a modreq then it should effectively ignore the containing member altogether.

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+This discussion primarily applies to members only, not to fields. While `init`
+fields were rejected by LDM they are still interesting to consider for the 
+modreq vs. attribute discussion. The `init` feature for fields is a relaxation
+of the existing restriction of `readonly`. That means if we emit the fields as
+`readonly` + an attribute there is no risk of older compilers mis-using the 
+field because they would already recognize `readonly`. Hence using a modreq here
+doesn't add any extra protection.
+
+**Resolution**
+The feature will use a modreq to encode the property `init` setter. The
```
HaloFour, line 338: Is this the appropriate place to comment about the decision?  I brought up [problems](https://github.com/dotnet/csharplang/issues/3323#issuecomment-616005297) with this approach in one of the issues for meeting notes which I think make `modreq` an unattractive option for encoding `init` setters.  I imagine the details of which will be a subject for a future language design meeting.

jaredpar: Let's use a separate issue to discuss this aspect. Want to keep this PR focused on recording the design + the current decision. I will ping the developer who's working on this feature @jcouv to make sure he sees that comment so we can make sure we get to the bottom of it.

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+The times above in which the `init` accessors are settable are collectively
+referred to in this document as the construction phase of the object.
+
+This means the `Student` class can be used in the following ways:
+
+```cs
+var s = new Student()
+{
+    FirstName = "Jared",
+    LastName = "Parosns",
```
haacked, line 90: Typo here

mgravell:
Hi; can I request clarification for what features, if any, will be discoverable / usable via reflection? In particular, I'm thinking of meta-programming code that pre-dates generators, and which may do any of the following today:

- discover properties with manual (non-auto) getters and setters  (not necessarily `public`) and call those via regular reflection during materialization; assume that `set` becomes an `init`: will the `init` be discoverable and callable?
- discover properties with `get`-only auto-properties and discover the underlying field via Roslyn implementation knowledge (yes, this happens): assume this becomes `get; init;` - will the backing field remain the same pattern for the old approach? is the `init` preferable, if it is discoverable and callable via regular reflection?
- all of the above with IL-emit in the same runtime?
- any it may be moot since there is not currently an "emit to disk" API, but are there any concerns re IL-verificatioj for runtime emit users? Would we need to retain the object purely on the stack? (which may not work well with complex scenarios), or (if verification is even a concern) would we need the same approach as `async`?
- on `async` - maybe allow marking a field/parameter/[local?] with an attribute (in CompilerServices): any such is considered "under construction" for the purposes of IL-emit? So: the backing field for hoisted async construction would be marked such

proposals/init.md
poke: I see! I already guessed that this was intentional when I saw it again a bit further down. Just makes it a bit odd when reading the source markdown 😄

proposals/init.md
poke: Yeah, sorry about that. I would have made edit suggestions if GitHub was allowing me to do so from the mobile website – I was sitting on the couch, reviewing this from my phone 😅🙈
If I knew before that I would have ended up with this many points, I maybe would have bothered to stand up and continue on my computer.
Sorry! 😅

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+can be set at the point of object creation but become effectively `get` only 
+once object creation has completed. This allows for a much more flexible
+immutable model in C#. 
+
+## Motivation
+The underlying mechanisms for building immutable data in C# haven't changed
+since 1.0. They remain:
+
+1. Declaring fields as `readonly`.
+1. Declaring properties that contain only a `get` accessor.
```
tiagocesar, line 15:
```suggestion
2. Declaring properties that contain only a `get` accessor.
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+    {
+        this.X = x;
+        this.Y = y;
+    }
+}
+```
+
+The `init` modifier makes immutable objects more flexible by allowing the
+caller to mutate the members during the act of construction. That means the
+object's immutable properties can participate in object initialzers and thus
```
tiagocesar, line 42:
```suggestion
object's immutable properties can participate in object initializers and thus
```

giuliano-barberi-tf:
Hi, will init properties also be taken into consideration when initializing NRTs somehow? It would be great if the caller was forced to specify a value for a non-nullable reference type, otherwise I'm guessing you would be required to still initialize it from the constructor or make init properties always nullable?

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+1. Call other `init` accessors available through `this`
+1. Assign `readonly` fields declared on the same type
+
+```cs
+class Complex
+{
+    readonly int Field1;
+    int Field2;
+    int Prop1 { get; init ; }
```
victorfrederiknielsen, line 137: whitespace after init

proposals/init.md
poke: Alright, I thought the exact opposite while reading this but I’m fine with it if this is like a design docs convention 😀

mgravell:
> One of the main pivot points in how this feature is encoded will come down to the following question: 
> > Is it a binary breaking change to replace `init` with `set`?

IMO that is the second question, since no `init` code currently exists. The first question is:

> Is it a binary breaking change to replace `set` with `init`?

i.e. can I change existing code that may have been fully settable (but perhaps using popsicle immutability internally) to code that expresses the new intent via `init`? It is understood that his may represent a build-breaking change when used outside of initialization, but is it a runtime breaking change? From the perspective of libraries, the runtime breaking is just as big - perhaps bigger - of a concern. If this is seen as a compelling case, then maybe an attribute is preferable over a modreq. However, in that scenario it may also be desirable to annotate the property automatically with `[ReadOnly(true)]`, which is used by many UI tools (via the `PropertyDescriptor` API) to determine mutability.

proposals/init.md
poke: I see, thanks for the explanation! So basically custom markers for the compiler and the runtime to handle stuff in a special way. Works for me :D
One final question: What does the “mod” in “modreq” and “modopt” stand for? Req is likely requirement and opt is option, but mod? “Modification” as in to make a modification on the generated IL?

proposals/init.md
poke: [As per above](https://github.com/dotnet/csharplang/pull/3367#discussion_r411008323) this is intentional.

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+to support construction as it does to declare the type. The bigger the type 
+the bigger the cost of this boiler plate:
+
+```cs
+struct Point
+{
+    public int X { get; }
+    public int Y { get; }
+
+    public Point(int X, int Y)
```
ViIvanov, line 32: Parameters should be small "x" and "y".

ViIvanov:
```suggestion
    public Point(int x, int y)
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+// 1. Use init 
+int Option1 { get; init; }
+// 2. Use init set
+int Option2 { get; init set; }
+// 3. Use initonly
+int Option3 { get; initonly; }
+```
+
+**Resolution**
+There was no syntax which was overwelmingly favored in LDM.
```
ViIvanov, line 362:
```suggestion
There was no syntax which was overwhelmingly favored in LDM.
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+{
+    public sealed class IsInitOnly
+    {
+
+    }
+}
+```
+
+The compiler will match the type by full name. There is no requirement
+that it appear in the core library. If there are mulitple types by this name
```
nycdotnet, line 259:
```suggestion
that it appear in the core library. If there are multiple types by this name
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+    public Name(string first, string last)
+    {
+        First = first;
+        Last = last;
+    }
+}
+```
+
+Adding `init` to these properties is a non-breaking change:
```
nycdotnet, line 497: This seems like a perfect lightbulb/analyzer suggestion.

GoldenCrystal:
My apologies if I missed a previous discussion about this, but I feel slightly unconfortable about the proposed syntax.
Not considering possible future evolutions mentioned in the proposal, in the context of get-only auto-properties, `init` and `get` seem somewhat redundant.
i.e. Would `init` alone be allowed, and what would it mean?

Have you considered a simpler alternative form where `init;` alone is a syntaxical replacement for `get;` with the broader meaning of `get; init;` from the current proposal?

e.g.
````csharp
// Property is only settable at initialization and in the constructor
public int PropA { init; }
// Property is only settable in the constructor
public int PropB { get; }
// Property is settable everywhere
public int PropC { get; set; }

// Can be read publicly, but only written-to inside the constructor or "internally" during object initialization
public int PropD { internal init; }
````

**Comment by poke** [comment]
@GoldenCrystal Set-only properties are allowed, so I don’t see why init-only properties, without a getter, shouldn’t be. So no, I wouldn’t say that a getter is implicit when you specify `init` on auto-properties. This is especially true because the get and init may have different access modifiers on them which makes the distinction between them very important.

File: proposals/init.md (hunk 31 above)
Comment (jaredpar) [reply proposals/init.md:215]: > Yeah, sorry about that. I would have made edit suggestions if GitHub was allowing me to do so from the mobile website – I was sitting on the couch, reviewing this from my phone

Not a problem. Mostly I wanted to make sure you were aware of the feature so you could get credit in the document history if that was interesting to you.

**Comment by GoldenCrystal** [comment]
> Set-only properties are allowed

Sure, set-only properties are allowed by the language spec, but:

- Their use is heavily discouraged: See the [official design guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property) that specifically mentions
  > ❌ DO NOT provide set-only properties or properties with the setter having broader accessibility than the getter.
  
  Generally, consider what it means & how bothersome it is to not be able to read back what you just wrote to a property…
  Given that properties are *intended* to be (mostly 😁) behaving as fields, that doesn't make a lot of sense. (Of course, I can't speak for the language designers, but I *think* that if they had to make the choice again today, they might not allow set-only properties at all)
- Set-only **auto**-properties are [not allowed at all](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAZgAJiAmcgYQFgAoAbyfPcooEsA7DcgArlm5XDAwBucgF8m0oA==). Anyway, what would be the point of having a property that you can write to but absolutely *never* ever read from? (Except by relying on reflection APIs of course)

If anything, I'd even argue that R/W auto-properties should be allowed to be declared as simply `T P { set; }` insetad of current `T P { get; set; }`, but that might a a bit off-topic here 😅

> This is especially true because the get and init may have different access modifiers on them which makes the distinction between them very important.

I addressed the idea of a different accessibility for implicit `get` and (explicit) `init` in the code sample above. 🙂 (Not saying it's perfect, though)

**Comment by jaredpar** [comment]
@mgravell 

> Hi; can I request clarification for what features, if any, will be discoverable / usable via reflection?

>  will the init be discoverable and callable?

Yes. The .NET API review notes for the `IsInitOnly` member haven't been published yet. But one particular discussion point was making sure that `init` members were discoverable via the reflection API. At the moment there will be no simple helper like `IsInitOnly` on the accessor but we made sure that a determined user could find the encoding via the existing APIs.

Whether or not to provide the more friendly APIs is a bit of a debate point right now. The upside is that it makes it easier to write meta programming tools but the downside is it has a tendency to veer reflection from being language neutral to being more C# centric. 

> will the backing field remain the same pattern for the old approach?

Yes there will be no change to how the backing field is encoded here. 

> all of the above with IL-emit in the same runtime?

Yes

> any it may be moot since there is not currently an "emit to disk" API, but are there any concerns re IL-verificatioj for runtime emit users? 

At the moment .NET Core doesn't have an IL verification story hence there is nothing to worry about right now. I did add a section in the document about what IL verification would look like for this feature but until we have an actual implementation and updated spec it's a bit speculative. 

> on async - maybe allow marking a field/parameter/[local?] with an attribute (in CompilerServices): any such is considered "under construction" for the purposes of IL-emit? S

If we every get an IL verification story with .NET Core then I think this will become a necessity for us. Until we have that though I suspect we won't emit the hoisted fields in any special way. It would make state machines bigger without providing any value. Or at least from the perspective of the C# language there is no benifit from doing this. Is there a meta programming consumption story that I'm missing here?

File: proposals/init.md [hunk 53]
```diff
@@ -0,0 +1,582 @@
...
+allow `init` to be a modifier in the property accessor list as well as a short
+hand for `init set`. Essentially the following two declarations would be
+identical.
+
+```cs
+int Property1 { get; init; }
+int Property1 { get; init set; }
+```
+
+The decision was made to move forward with `init` as a standalone accessor in
```
YairHalberstadt, line 378: If init is a standalone accessor, is there any desire to allow both init and set, but with different accessibility. For example you may want to allow setting a property at initialization via an object initializer instead of via a constructor, but allow setting it inside of the class.

In other words this pattern:

```csharp
class C {
    public C(int prop) => Prop = prop;
    public int Prop { get; private set; }


}
```

Could now be replaced with:

```csharp
class C {
    public int Prop { get; init; private set; }
}
```

poke: If I understand it correctly, as per section “Generate three accessors”, having `init` and `set` be separate was considered but the cost is apparently too much. So `init` is direct replacement for `set` if used.

YairHalberstadt: @poke
That was discussing this as an emit strategy. I'm asking whether there's a benefit to having this as a language feature. None of the benefits or drawbacks to that suggested emit strategy apply to this this suggestion.

proposals/init.md
HaloFour: I've submitted a [separate issue](https://github.com/dotnet/csharplang/issues/3376) to discuss this problem.  I also think I've found a solution which I included in that issue and doesn't deviate a great deal from this design.

proposals/init.md
```diff
@@ -372,58 +464,27 @@ init struct Point
...
-the scenario.
-
 ## Considerations
 
 ### Compatibility
-The `init` feature is designed to be compatible with existing data types. 
-Specifically it is meant to be a completely additive change for data which is
-`readonly` today but desires more flexible object creation semantics.
+The `init` feature is designed to be compatible with existin `get` only 
+properties. Specifically it is meant to be a completely addiditive change for
```
333fred, line 477:
```suggestion
properties. Specifically it is meant to be a completely additive change for 
```

proposals/init.md
333fred: "Mod" stands for modifier.

333fred:
> Not considering possible future evolutions mentioned in the proposal, in the context of get-only auto-properties, init and get seem somewhat redundant.

@GoldenCrystal you seem to be approaching this in exactly the opposite way from what we did in LDM, which is interesting to me. You propose `init` as a modifier to `get`, when what we view it as is a modifier to `set`. Specifically, it is a setter that has had it's valid scope restricted to a single location. In particular, your last example of differing accessibilities shows why that is a no-go in my book: there's no way to define an internal-only getter.

proposals/init.md
```diff
@@ -212,7 +212,7 @@ class C2 : Base
 }
 ```
 
-An `interface` declaration can also particpate in `init` style initalization 
+An `interface` declaration can also participate in `init` style initalization
```
333fred, line 215:
```suggestion
An `interface` declaration can also participate in `init` style initialization 
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
+Init Only Members
```
jcouv, line 1:
```suggestion
Init-Only Members
```
```suggestion
Init-Only Setters
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+## Motivation
+The underlying mechanisms for building immutable data in C# haven't changed
+since 1.0. They remain:
+
+1. Declaring fields as `readonly`.
+1. Declaring properties that contain only a `get` accessor.
+
+These mechanisms are effective at allowing the construction of immutable data
+but they do so by adding cost to the boiler plate code of types and opting
```
jcouv, line 18:
```suggestion
but they do so by adding cost to the boilerplate code of types and opting
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+    public Point(int X, int Y)
+    {
+        this.X = x;
+        this.Y = y;
+    }
+}
+```
+
+The `init` modifier makes immutable objects more flexible by allowing the
```
jcouv, line 40:
```suggestion
The `init` accessor makes immutable objects more flexible by allowing the
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+caller to mutate the members during the act of construction. That means the
+object's immutable properties can participate in object initializers and thus
+removes the need for all constructor boiler plate in the type. The `Point`
```
jcouv, line 43:
```suggestion
removes the need for all constructor boilerplate in the type. The `Point`
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+
+```cs
+class Student
+{
+    public string FirstName { get; init; }
+    public string LastName { get; init; }
+}
+```
+
+An instance property containing an `init` accessor is considered settable in
```
jcouv, line 74:
```suggestion
An instance property containing an `init` accessor is considered settable on `this` or `base` in
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+the following circumstances:
+
+- During an object initializer
+- Inside an instance constructor of the containing or derived type
```
jcouv, line 78:
```suggestion
- Inside an instance constructor of the containing or derived type, on `this` or `base`
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+- Inside the `init` accessor of any property on `this`
```
jcouv, line 79:
```suggestion
- Inside the `init` accessor of any property, on `this` or `base`
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
+Init Only Members
+=====
+
+## Summary
+This proposal adds the concept of init only properties to C#. These properties
```
jcouv, line 5:
```suggestion
This proposal adds the concept of init-only setters in properties and indexers to C#. These properties and indexers
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+    }
+}
+
+```
+
+At the point a `init` accessor is invoked the instance is known to be 
+in the open construction phase. Hence an `init` accessor is allowed to take 
+the following actions in addition to what a normal `set` accessor can do:
+
+1. Call other `init` accessors available through `this`
```
jcouv, line 129:
```suggestion
1. Call other `init` accessors available through `this` or `base`
```

proposals/init.md
```diff
@@ -0,0 +1,582 @@
...
+Property `init` accessors will be emitted as a standard `set` accessor with
+the return type marked with a modreq of `IsInitOnly`. This is a new type
+which will have the following definition:
+
+```cs
+namespace System.Runtime.CompilerServices
+{
+    public sealed class IsInitOnly
+    {
+
```
jcouv, line 253:
```suggestion
```

proposals/init.md
jcouv: We discussed whether `init` accessor is a separate thing (and should use a separate slot) than `set` accessor, but landed on just re-using the same slot and not allowing having both an `init` and `set`.

https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-04-13.md#init-only-members

proposals/init.md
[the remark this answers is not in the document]
jaredpar: Discussed to hyphen or not to hyphen earlier with @333fred  and decided against it.

proposals/init.md
333fred: Ajred specifically went through and settled on `init only` throughout the document.

jcouv: Missed that. Seems wrong though, but the hyphen is a nit, so I'm fine with whatever Jared picks.

GoldenCrystal:
> you seem to be approaching this in exactly the opposite way from what we did in LDM, which is interesting to me. You propose init as a modifier to get, when what we view it as is a modifier to set.

@333fred 
Sorry about that 😄
I read the proposal mostly with `readonly struct`s and get-only auto-properties in mind, as I thought that this was where the changes were the most interesting.
I do nevertheless understand that you're trying to implement a feature set that is coherent between auto and manual properties, and that `init` *technically* needs to be a setter.
(Please do not read my comments as trying to diminish the value of this proposal; I do think it is great overall 😊)

Not sure if I need to detail it, but my reasoning was the following:
Today, a property `int P { get; }` is already writable (from the constructor), despite only being indicated as `get`.
While maybe not technically true, we could very well consider this behavior to be entirely supported by `get;`, as opposed to `get {}` or `get =>`: Since `set;` also include the ability to write to the property from the constructor, this assertion cannot be observed to be false. (So it may as well be true 🤷‍♂️)

The only thing we currently miss from get-only auto-properties is the ability to be written from object initializers **in addition** to the type constructors, which this proposal intends to adress (hopefully on the way to C# records ?)

With that in mind, when I see `public int A { get; init; }`, I instinctively read it as `init;` being some kind of weird modifier that would mean "And you can **also** write it from object initializers"… And the funny thing is that, conceptually, this is mostly what it does.

Also, I do not know how to word it properly, but I found the `{ get; init; }` syntax to be a bit clunky. Adding 6 extra characters (5 without spaces) per property for a feature we'd like to have by default most of the time seems a bit much. 
I tend to think that initializable auto-properties will end up being used the most, so I'd really like the syntax to be optimized around it.
Hence, since `init;` alone wouldn't mean anything, that it extends `get;`, and that `get` already has a slightly different meaning in case of auto-properties, why not just omit `get;` and use only `init;`? (The same being true for `set;` but `get; set;` is already there 😕)

> In particular, your last example of differing accessibilities shows why that is a no-go in my book: there's no way to define an internal-only getter.

I think this is solved nicely if we consider that specifying the `get` accessor is optional when it can be infered. (That was my backup proposal anyway… 😁)

In the end, as far as auto-properties are concerned, we could see things in two different, yet not necessarily exclusive, ways:
* In terms of cabability, `get` ⊂ `init` ⊂ `set`
* `get` is mandatory in C# 8 and below; but it could as well be rendered optional. We only strictly need it to differentiate between a field and a get-only auto-property.

So, hmm…
Especially in the context of adding a different kind of property setter, could we make the `get;` optional in auto-properties?

That'd be a related, but totally different proposal, so I can open a separate issue to discuss this

333fred:
> In terms of cabability, get ⊂ init ⊂ set

See, that's actually not true at all. `get` and `set` are entirely orthogonal here: You can have `set`-only properties that cannot be read. We advise against creating them, true, but you can have them. It's an important point, though: nowhere in C# today do we have a mutability accessor or modifier that implies the member it is applied to is also readable as well as writable. If we were to allow `init` on its own, this would change; we'd have a thing, that changes where a member is able to be assigned (be an lvalue) also imply that it can be read (be an rvalue). There's also a matter of optics: when I'm reading the code, the word `init` does not imply anything about the readability of the code. Yes, you might say that `readonly` or a single `get` does not imply this either, but those are well-established patterns in the language and intuitively make sense, as _something_ needs to set an initial value.

proposals/init.md
jaredpar: > Not sure how to avoid repeating "properties or indexers" in the whole document.

I think it's okay to just refer properties in the document and call out that it also works for indexers in the details section.

proposals/init.md
jaredpar: Thanks. Fixed.

proposals/init.md
```diff
@@ -221,37 +248,42 @@ which will have the following definition:
...
-The compiler will match the attribute by full name. There is no requirement
-that it appear in the core library. In the case there are mulitple attributes
-by this name available then the compiler will pick the one defined in the core 
-library should one exist.
+The compiler will match the type by full name. There is no requirement
+that it appear in the core library. If there are mulitple types by this name
+then the compiler will tie break in the following order:
+
+1. The one defined in the project being compiled
+1. The one defined in corelib
```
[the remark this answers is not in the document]
jaredpar: It is essentially how the extension method resolution for the `ExtensionAttribute` works to my memory at least (and based on a small scale test). We can look that up though and make sure that is the case.

proposals/init.md
jaredpar: More specifically:

- modreq = modifier required: that is it is required that the compiler reading the entry understand the type in the modreq. That is because the type in the modreq implies special behavior that compilers are required to understand in order to handle the function correctly. Hence if the compiler does not understand it should discard that member.
- modopt = modifier optional: that is the compiler is not required to understand the type in the modopt. The type indicates behavior that can be beneficial to compilers who understand the behavior but knowledge is not required in order to correctly invoke the member.

proposals/init.md
jaredpar: The rules were meant to mention that defining `init` or `set` would end requiring both were emitted. Will update.

jaredpar:
Thanks for all the feedback!

proposals/init.md
```diff
@@ -0,0 +1,581 @@
...
+    int Prop1 { get; init ; }
+    int Prop2
+    {
+        get => 42;
+        init
+        {
+            Field1 = 13; // okay
+            Field2 = 13; // okay
+            Prop1 = 13; // okay
+        }
```
Carl-Hugo, line 146: Why do you want to add that, and what would be the advantage? To me, this is duplicating the job of a constructor, and I think this will only create less readable code. Maybe there is a use-case that I missed; if so, feel free to enlighten me.

333fred: I believe the motivation section of this proposal adequately outlines our reasoning: we want to support nominal, immutable construction patterns which we do not do today.

Carl-Hugo: > I believe the motivation section of this proposal adequately outlines our reasoning: we want to support nominal, immutable construction patterns which we do not do today.

Yes, I get the intent @333fred , which is a really good idea; less boilerplate is always welcome. I still remember the day you added auto-implemented properties; what a day!

But what is the difference between the following two blocks of code?

```csharp
init
{
    Field1 = 13; // okay
    Field2 = 13; // okay
    Prop1 = 13; // okay
}
```

```csharp
public SomeCtor()
{
    Field1 = 13; // okay
    Field2 = 13; // okay
    Prop1 = 13; // okay
}
```

Then what is the added value of that new way to initialize fields/props? (for C# users)

An intend is one thing, but does that help solve a problem or does it only create noise and possible confusion? Think about it, people will have one more place to put initialization logic in, does someone need that? If yes, who does? What use-case does this solve that was unsolvable before or that required more boilerplate?

333fred: Think about optional parameters in a constructor today: you have a location that can default them if not provided, or throw if a combination of provided parameters is invalid. This adds the same logic for properties that are inited by the caller, but then immutable afterwards.

Carl-Hugo: > Think about optional parameters in a constructor today: you have a location that can default them if not provided, or throw if a combination of provided parameters is invalid. This adds the same logic for properties that are inited by the caller, but then immutable afterwards.

When I think about that I see types that could require some kind of possible value/props/null/not null combo but without any way to know it in advance from the API of that class/struct. People can be creative 😉 

Then what about I add another property that initializes `init` and `readonly` props/fields as well?

```cs
class Complex
{
    readonly int Field1;
    int Field2;
    int Prop1 { get; init ; }

    int Prop2
    {
        get => 42;
        init
        {
            Field1 = 13; // okay
            Field2 = 13; // okay
            Prop1 = 13; // okay
        }
    }

    int Prop3
    {
        get => 123;
        init
        {
            Field1 = 3;
            Field2 = 3;
            Prop1 = 3;
        }
    }
}
```

In what order are those going to be initialized?

Wouldn't it be clearer to do the following?

```cs
class Complex
{
    readonly int Field1;
    int Field2;
    int Prop1 { get; init ; }
    int Prop2 { get; init; }
    int Prop3 { get; init; }

    public Complex()
    {
        Field1 = 13;
        Field2 = 13;
        Prop1 = 13;
        Prop2 = 42;
        Prop3 = 123;
    }
}
```

333fred: You shouldn't necessarily take syntax examples from a proposal that merely illustrate what things are legal where as a good API design. More than likely, most `init` properties would initialize one, and only one, readonly field. You can still write your example today, using `set` instead of `init`, and it provides a markedly worse experience if you wanted it to actually be an immutable object.

As to ordering, that's the thing here: this is about _nominal_ object creation. Nominal creation is unordered, that's the whole point. If you want to control the order in which properties are set, then you need to use the existing tools we have today in positional creation (namely, standard constructor arguments).

Imagine you had an example like this:

```cs
public class PathObject
{
    string ActualPath { get; init; }
    bool SupportsLongPath { get; init; }
    init
    {
        if (ActualPath == null || !supportsLongPath && ActualPath.Length >= 256)
        {
            throw new ArgumentException(nameof(ActualPath), $"Long path support not enabled, but path is {ActualPath.Length}!");
        }
    }
}
```

Imagine if you had to write that without `init` support? You literally can't do it. There's no way today to ensure that `ActualPath` was initialized by the caller. Further, for validation of the length of `ActualPath`, you can't do that either: you need to ensure that `SupportsLongPath` is set _before_ attempting to validate the path length. The only way to handle that correctly would be to use a constructor, which guarantees ordering. However, that brings burdens of its own. Let's say that this is a public class in a library that you publish publicly. You decide that you want to add support for hidden path object creation. But there's a problem: you've already published this class. If you want to avoid breaking your customers, you're going to have to maintain this 2-parameter constructor forever more, because removing it would break existing customer binaries. In the `init` world, this pain point is gone. You add a new init-only property, validate as necessary in the `init` body, and no one who compiled against a previous version of your library has to care about this new optional parameter that you've added. For a very real example of this type of pain, look at https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/CSharpCompilationOptions.cs#L787-L913. We cannot remove these constructors, as existing binaries depend on them, even the constructor marked as `Bad - DO NOT USE`. Those constructors will never be able to go away, unfortunately, but at least we won't have to add more of them.

Carl-Hugo: Good point, I could write my messy code in a `set`, so `init` is similar in that way (well, its an "initializable `readonly` property setter" after all).

That said, I did not see the "init constructor-like" that you used in your last sample anywhere in the proposal, maybe you should add that as it seems that will be needed.

If the initialization is unordered, how can you be sure that your constructor-like-init code block will be hit **after** the init setter of the other properties?

More on that, to make the usage clearer for users, maybe `afterinit` would be easier to understand for non-compiler/csharplang-guru? It would also point people in the direction of centralizing initialization logic instead of splitting it everywhere. Maybe a `beforeinit` could benefit centralizing initialization while `afterinit` benefits validation?

Or to make it less "magic" and more discoverable, what about an `IInitializable` interface that adds a `void AfterInit();` and a `void BeforeInit()` methods usable during the initialization process? Same idea here, you need additional logic, you implement the interface (or create the `afterinit` block) and code your initialization logic there; otherwise, you don't pollute your code with it.

_As I understand it, you seem to want to help people get into C#, so making usage and understanding of new (and existing) features as easy as possible for everyone would probably help that. Just for fun, ask people in what order are executed auto-property initializers and constructors to see how unclear it can be. With `init` you will be adding another layer of "confusion" over that._

As for breaking changes, a new major version should allow you to clean stuff up and introduce breaking changes. Everyone makes mistakes but having to carry all of the mistakes of everyone forever ain't helping anyone 😉

333fred: > That said, I did not see the "init constructor-like" that you used in your last sample anywhere in the proposal, maybe you should add that as it seems that will be needed.

It'll be coming in a separate proposal. It'll all part of a greater feature unity we're aiming for as part of records.

> If the initialization is unordered, how can you be sure that your constructor-like-init code block will be hit after the init setter of the other properties?

The compiler will be required to generate a call to the validator (straw-man name we're using for now) after an object initializer is run, so ordering won't matter.

Don't take the syntax I used as "How it will look exactly". A future proposal with more validator design will be coming, and we haven't talked much about the specific syntax beyond "We like the general idea".

Carl-Hugo: Good to know, thanks, I'll try to follow that up then.

proposals/init.md
jcouv: @jaredpar You may have missed some of my comments. The spec still doesn't mention indexers. I'll start a PR to correct some things.

---
In reply to: [411695374](https://github.com/dotnet/csharplang/pull/3367#discussion_r411695374) [](ancestors = 411695374)

jaredpar: Didn't miss that comment. I added a mention of that + an explicit example using indexers to drive home it was intended to be supported. What I apparently missed was saving and pushing that to the doc 😦 

Saw you already updated the doc though to mention they're included so I think we're fine.

proposals/init.md
```diff
@@ -0,0 +1,581 @@
...
+
+```cs
+int Property1 { get; init; }
+int Property1 { get; init set; }
+```
+
+The decision was made to move forward with `init` as a standalone accessor in
+the property accessor list.
+
+### Warn on failed init
```
domagojmedo, line 380: Any plans to add some option or another keyword that would force property to be set in initializer or constructor?

jcouv: We've touched on that topic in recent discussions, but haven't made any plans for that yet.

YairHalberstadt:
Will you be able to set an init only property from a nested object initializer?
Not to allow that would be inconsistent.

Allowing it would let you to set an initonly property of any arbitrary object:

```csharp
public class Foo
{
    public int Bar { get; init; }
}
public struct InitSetter<T>
{
    public InitSetter(T t) => Value = t;
    public T  Value { get; }
}

...

public void M(Foo foo)
{
     new InitSetter<Foo>(foo)
     {
          Value = 
          {
              Bar = 42,
          }
     }
}
```

alrz:
>Will you be able to set an init only property from a nested object initializer?


At that point the object is already created. So it's not really in the initialization context. I'd expect not.

PR #3368, opened 2020-04-17
Add metadata representation section for in and out

333fred:
Added the spec proposal discussed over email for representing `in` and `out` parameters.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -243,6 +243,26 @@ The better function member specification will be changed to include the followin
 
 This means that it is possible to overload on `void*` and a `delegate*` and still sensibly use the address-of operator.
 
+## Metadata representation of `in`, `out`, and `ref readonly` parameters and return types
+
+Function pointer signatures have no parameter flags location, so we must encode whether parameters and the return type are `in`, `out`, or `ref readonly` by using modreqs.
+
+### `in`
+
+We reuse `System.Runtime.InteropServices.InAttribute`, applied as a `modreq` to the ref specifier on a parameter or return type, to mean the following:
+* If applied to a parameter ref specifier, this parameter is treated as `in`.
+* If applied to the return type ref specifier, the return type is treated as `ref readonly`.
+
+### `out`
+
+We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to the ref specifier on a parameter type, to mean that the parameter is an `out` parameter.
+
+### Errors
+
+* It is an error to apply `OutAttribute` as a modreq to a return type.
+* It is an error to apply both `InAttribute` and `OutAttribute` as a modreq to a parameter type.
+* If either are specified via modopt, they are ignored.
+
 ## Open Issues
 
 ### NativeCallableAttribute


proposals/function-pointers.md
```diff
@@ -243,6 +243,26 @@ The better function member specification will be changed to include the followin
...
+* If applied to the return type ref specifier, the return type is treated as `ref readonly`.
+
+### `out`
+
+We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to the ref specifier on a parameter type, to mean that the parameter is an `out` parameter.
+
+### Errors
+
+* It is an error to apply `OutAttribute` as a modreq to a return type.
+* It is an error to apply both `InAttribute` and `OutAttribute` as a modreq to a parameter type.
```
tannergooding, line 263: Should this be `to a parameter type or return type`

proposals/function-pointers.md
```diff
@@ -243,6 +243,26 @@ The better function member specification will be changed to include the followin
 
 This means that it is possible to overload on `void*` and a `delegate*` and still sensibly use the address-of operator.
 
+## Metadata representation of `in`, `out`, and `ref readonly` parameters and return types
```
tannergooding, line 246: Do the rules for how the language converts a managed method signature to a function pointer signature need to be defined here?

That is, C# allows you to apply your own `[In]` and `[Out]` attributes (with certain limitations) today:
* `in` parameters and `ref readonly` return
  * Has the `IsReadOnly` attribute
  * Virtual methods additionally have a `modreq [In]` attribute
  * Sets the `in` built-in param attribute
  * Allows users to explicit specify `[In]`
  * Errors if the user specifies `[Out]`
* `out` parameters
  * Sets the `out` built-in param attribute
  * Allows users to explicitly specify `[Out]`
  * Errors if the user specifies `[In]`
* `ref` parameters and `ref` return
  * Allows users to explicitly specify `[In]`, which sets the `in` built-in param attribute
  * Allows users to explicitly specify `[In, Out]` which sets the `in` and `out` built-in param attributes
  * Errors if the user only specifies `[Out]`

I imagine the intent is it then behaves as the following when encountering a "ref-like" param in IL (`T&`):
* If it has the `IsReadOnly` attribute, the function-pointer signature will have `modreq [In]` for that parameter/return
* If it has the `out` built-in param attribute, the function pointer signature will have `modreq [Out]` for that parameter
* Otherwise, the function pointer signature will have no `modreq` for that parameter/return

Is that correct?

That is, it is my interpretation that it will ignore things which the language doesn't currently use to detect the "type"

proposals/function-pointers.md
333fred: No. It's always an error to have an `OutAttribute` on a return type, so I didn't bother spec'ing what happens when you apply both.

proposals/function-pointers.md
```diff
@@ -243,6 +243,26 @@ The better function member specification will be changed to include the followin
...
+
+Function pointer signatures have no parameter flags location, so we must encode whether parameters and the return type are `in`, `out`, or `ref readonly` by using modreqs.
+
+### `in`
+
+We reuse `System.Runtime.InteropServices.InAttribute`, applied as a `modreq` to the ref specifier on a parameter or return type, to mean the following:
```
jkotas, line 252: Since `in` maps to `IsReadOnlyAttribute` in regular signatures, would it make more sense to use `IsReadOnlyAttribute` here as well?

333fred: I wanted to keep using `InAttribute` because we don't emit `IsReadOnlyAttribute` as a modreq anywhere today, but we do for `InAttribute`.

333fred: I'm not opposed to using `IsReadOnlyAttribute` instead if we so desire.

jkotas: ok

proposals/function-pointers.md
333fred: We don't need to specify anything here. The language understands method signatures today, and method-function pointer conversions are already defined based on that understanding.

PR #3379, opened 2020-04-21
Extending partial methods

jaredpar:


RUSshy:
Will there be some methods available in the standard library to help with compile time reflection? or it is a non goal?

proposals/extending-partial-methods.md
```diff
@@ -0,0 +1,215 @@
...
+}
+
+partial class C
+{
+    // The definition of C.M
+    partial void M(string message) => Console.WriteLine(message);
+}
+```
+
+One behavior of `partial` methods is that when the definition is abscent then
```
jnm2, line 28:
```suggestion
One behavior of `partial` methods is that when the definition is absent then
```

proposals/extending-partial-methods.md
```diff
@@ -0,0 +1,215 @@
...
+    {
+        M(GetIt()); // Call to M and GetIt erased at compile time
+    }
+
+    string GetIt() => "Hello World";
+}
+```
+
+The original motivation for this feature was source generation in the form of 
+designer generated code. Users were constantly editting the generated code
```
jnm2, line 48:
```suggestion
designer generated code. Users were constantly editing the generated code 
```

jaredpar:
> Will there be some methods available in the standard library to help with compile time reflection? or it is a non goal?

Can you elaborate a bit on what you're asking here? Not quite sure what functionality you're referring to.

PR #3381, opened 2020-04-22
Added details and examples for null guard guidance

IEvangelist:
Addresses: https://github.com/dotnet/docs/issues/14491

### DIFF
--- a/proposals/csharp-8.0/nullable-reference-types.md
+++ b/proposals/csharp-8.0/nullable-reference-types.md
@@ -87,7 +87,7 @@ In the previous example, the `DoWork` function accepts a `Worker` and guards aga
 
 ### Express intent
 
-A more compelling use of the previous example is to express that the `Worker` parameter could be `null`, thus making the null guard more appropriate. Regardless, both null guards are still valid.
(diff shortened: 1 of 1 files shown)


proposals/csharp-8.0/nullable-reference-types.md
```diff
@@ -64,6 +64,46 @@ var y = b ? "Hello" : null; // string? or error
...
+
+    // Otherwise use worker argument
+}
+```
+
+In the previous example, the `DoWork` function accepts a `Worker` and guards against it potentially being `null`. If the `worker` argument is `null`, the `DoWork` function will `throw`. With nullable reference types, the code in the previous example makes the intent that the `Worker` parameter would *not* be `null`. If the `DoWork` function was a public API, such as a NuGet package or a shared library - as guidance you should leave null guards in place. As a public API, the only guarantee that a caller isn't passing `null` is to guard against it.
+
+### Express intent
+
+A more compelling use of the previous example is to express that the `Worker` parameter could be `null`, thus making the null guard more appropriate. Regardless, both null guards are still valid.
```
BillWagner, line 90: I'd add that the compiler helps enforce your intent here:

```suggestion
A more compelling use of the previous example is to express that the `Worker` parameter could be `null`, thus making the null guard more appropriate. If you remove the null guard in the following example, the compiler warns that you may be dereferencing null. Regardless, both null guards are still valid.
```

PR #3392, opened 2020-04-24
Module initializers

RikkiGibson:


proposals/module-initializers.md
```diff
@@ -0,0 +1,181 @@
...
+## Summary
+[summary]: #summary
+
+Although the .NET platform has a feature that directly supports writing initialization code for the assembly (technically, the module), it is not exposed in C#.  This is a rather niche scenario, but once you run into it the solutions appear to be pretty painful.  There are reports of [a number of customers](https://www.google.com/search?q=.net+module+constructor+c%23&oq=.net+module+constructor) (inside and outside Microsoft) struggling with the problem, and there are no doubt more undocumented cases.
+
+## Motivation
+[motivation]: #motivation
+
+- Enable libraries to do eager, one-time initialization when loaded, with minimal overhead and without the user needing to explicitly call anything
+- One particular pain point is that the runtime must do additional checks on usage of a type with a static constructor, in order to decide whether the static constructor needs to be run or not. This adds measurable overhead.
```
333fred, line 17:
```suggestion
- One particular pain point of current `static` constructor approaches is that the runtime must do additional checks on usage of a type with a static constructor, in order to decide whether the static constructor needs to be run or not. This adds measurable overhead.
```

proposals/module-initializers.md
```diff
@@ -0,0 +1,181 @@
...
+- Enable source generators to run some global initialization logic without the user needing to explicitly call anything
+  - (in which scenarios is a static constructor insufficient/undesirable for this?)
```
333fred, line 19: Seems like this a comment that should be expanded or deleted?

proposals/module-initializers.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Detailed design
+[design]: #detailed-design
+
+Add a tiny feature to support this without any explicit syntax, by having the C# compiler recognize a module attribute with a well-known name, like the following:
```
333fred, line 24:
```suggestion
The C# compiler will recognize the following attribute:
```

proposals/module-initializers.md
```diff
@@ -0,0 +1,181 @@
...
+internal static class MyModuleInitializer
+{
+    static MyModuleInitializer()
+    {
+        // put your module initializer here
+    }
+}
+```
+
+and the C# compiler would then emit a module constructor that causes the static constructor of the identified type to be triggered:
```
333fred, line 56: This seems counter to one of the primary motivations of the feature:
> - One particular pain point is that the runtime must do additional checks on usage of a type with a static constructor, in order to decide whether the static constructor needs to be run or not. **This adds measurable overhead.**

We should expand why this is an acceptable solution.

proposals/module-initializers.md
```diff
@@ -0,0 +1,181 @@
...
+``` c#
+void .cctor()
+{
+    // synthesize and call a dummy method with an unspeakable name,
+    // which will cause the runtime to call the static constructor
+    MyModuleInitializer.<TriggerClassConstructor>();
+}
+```
+
+This proposal naturally prevents the user from declaring more than one initializer class without adding any special language rules to accomplish that. It is very lightweight in that it uses existing syntax and semantic rules for attributes.
```
333fred, line 67: Why? What will happen if a user does? Please be precise.

proposals/module-initializers.md
```diff
@@ -0,0 +1,181 @@
...
+
+It may make it more difficult for users to realize that the static constructor they are looking at is the module initializer, because the attribute may be far away from the class that actually contains the module initializer.
```
333fred, line 69: This seems like it should be under Drawbacks.

proposals/module-initializers.md
RikkiGibson, line 19:
```suggestion
```

proposals/module-initializers.md
RikkiGibson: called out 'AllowMultiple = false' directly.

proposals/module-initializers.md
RikkiGibson: With this approach you would generally declare a type just for the purpose of holding your module initializer, and avoid referencing it outside of the `[ModuleInitializerAttribute]`. Added language to clarify this.

proposals/module-initializers.md
RikkiGibson: I chose to keep it with the description of our first design because I see it as a drawback specific to that design and not innate to the feature.

PR #3396, opened 2020-04-25
Update spec for record value equality

agocke:


proposals/records-wip.md
```diff
@@ -67,8 +67,10 @@ Record types produce synthesized implementations for the following methods:
 * `object.Equals(object)` override, unless it is sealed or user provided
 * `T Equals(T)` method, where `T` is the current type
 
-`T Equals(T)` is specified to perform value equality, comparing the property with same name as
-each primary constructor parameter to the corresponding property of the other type.
+`T Equals(T)` is specified to perform value equality such that `Equals` is
+true if and only if all the instance fields of the receiver type are equal
```
gafter, line 71: >equal [](start = 69, length = 5)

I don't know what `equals` means.  How is this possible, given that the fields of the base class might not be accessible?

gafter:
> class_declaration

This is ambiguous with the existing production, e.g. for `class A { }`. It is intended to replace it? If so the base clause is missing.

---
Refers to: proposals/records-wip.md:9 in 04e750a. [](commit_id = 04e750a6474d5ec5a530be7ac81198051dcb85e6, deletion_comment = False)

gafter:
>     : attributes? class_modifiers? 'partial'? 'class' identifier type_parameter_list?

Did you intend to augment the modifiers to include `data`?

---
Refers to: proposals/records-wip.md:10 in 04e750a. [](commit_id = 04e750a6474d5ec5a530be7ac81198051dcb85e6, deletion_comment = False)

gafter:
> The `attributes` non-terminal will also permit a new contextual attribute, `data`.

attribute or modifier?

---
Refers to: proposals/records-wip.md:30 in 04e750a. [](commit_id = 04e750a6474d5ec5a530be7ac81198051dcb85e6, deletion_comment = False)

proposals/records-wip.md
agocke: This is the exact same language the spec uses for anonymous types, so if it's underspecified, that's an existing problem :)

agocke:
Syntactic issues shouldn't be relevant to this PR, so I'll deal with them later

proposals/records-wip.md
gafter: No, anonymous types don't have a base class. In any case, some other existing text being underspecified isn't an excuse. I think you should just say `object.Equals(x, y)` (and mention as an aside that the implied boxing may be elided)

---
In reply to: [418282509](https://github.com/dotnet/csharplang/pull/3396#discussion_r418282509) [](ancestors = 418282509)

proposals/records-wip.md
```diff
@@ -67,8 +67,10 @@ Record types produce synthesized implementations for the following methods:
 * `object.Equals(object)` override, unless it is sealed or user provided
 * `T Equals(T)` method, where `T` is the current type
 
-`T Equals(T)` is specified to perform value equality, comparing the property with same name as
-each primary constructor parameter to the corresponding property of the other type.
+`T Equals(T)` is specified to perform value equality such that `Equals` is
```
gafter, line 70: >specified [](start = 17, length = 9)

Where can we find this specification you speak of? ;)

PR #3398, opened 2020-04-26
Tweaks to init-only speclet

jcouv:


proposals/init.md
```diff
@@ -75,6 +75,7 @@ An instance property containing an `init` accessor is considered settable in
 the following circumstances:
```
alrz, line 75: @jcouv should this include Attribute property arguments? `[Attr(InitOnlyProp = true)]`

PR #3425, opened 2020-05-04
Permit init-only initialization in attribute usages

alrz:
/cc @jcouv @agocke


alrz:
It's common to use `{ get; set; }` properties in order to use named paramters in attributes which shouldn't be needed with initonly.

jcouv:
I like it. I'll let @jaredpar confirm.

jaredpar:
Yes this seems completely reasonable.

PR #3441, opened 2020-05-08
Update the calling convention and UnmanagedCallersOnly attribute.

333fred:
This adds the changes to extensible calling conventions and support for UnmanagedCallersOnlyAttribute.

### DIFF
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -296,7 +296,7 @@ function pointer.
 
 ### UnmanagedCallersOnlyAttribute
 
-`UnamangedCallersOnlyAttribute` is an attribute used by the CLR to indicate that a method should be called with a specific
+`UnmanagedCallersOnlyAttribute` is an attribute used by the CLR to indicate that a method should be called with a specific
 calling convention. Because of this, we introduce the following support for working with the attribute:
 
 * It is an error to directly call a method annotated with this attribute from C#. Users must obtain a function pointer to
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -48,10 +48,10 @@ delegate* managed<int, int>;
 
 // This method will be invoked using whatever the default unmanaged calling convention on the runtime
 // platform is. This is platform and architecture dependent and is determined by the CLR at runtime.
-delegate* unmanged<int, int>;
+delegate* unmanaged<int, int>;
 
 // This method will be invoked using the cdecl calling convention
-delegate* unmanged[CallConvCdecl] <int, int>;
+delegate* unmanaged[CallConvCdecl] <int, int>;
 
 // This method will be invoked using the stdcall calling convention, and suppresses GC transition
 delegate* unmanaged[CallConvStdCall, CallConvSuppressGCTransition] <int, int>;
@@ -333,7 +333,7 @@ If lookup succeeds on all of the `identifier`s specified in an `unmanaged_callin
 signature. As a note, these rules mean that users cannot prefix these `identifier`s with `CallConv`, as that will result
 in looking up `CallConvCallConvVectorCall`.
 
-When interpreting metadata, the we first look at the `CallKind`. If it is anything other than `unmanaged ext`, we ignore all
+When interpreting metadata, we first look at the `CallKind`. If it is anything other than `unmanaged ext`, we ignore all
 `modopt`s on the return type for the purposes of determining the calling convention, and use only the `CallKind`. If the
 `CallKind` is `unmanaged ext`, we look at the modopts at the start of the function pointer type, taking the union of all
 types that meet the following requirements:
--- a/proposals/function-pointers.md
+++ b/proposals/function-pointers.md
@@ -88,7 +88,7 @@ Restrictions:
 - Custom attributes cannot be applied to a `delegate*` or any of its elements.
 - A `delegate*` parameter cannot be marked as `params`
 - A `delegate*` type has all of the restrictions of a normal pointer type.
-- Pointer arithmetic cannot be performed directly on pointer types.
+- Pointer arithmetic cannot be performed directly on function pointer types.
 
 ### Function pointer syntax
 
@@ -101,12 +101,32 @@ pointer_type
     ;
 
 funcptr_type
-    : 'delegate' '*' calling_convention_specifier? '<' (funcptr_parameter_modifier? type ',')* funcptr_return_modifier? return_type '>'
+    : 'delegate' '*' calling_convention_specifier? '<' funcptr_parameter_list funcptr_return_type '>'
     ;
 
 calling_convention_specifier
     : 'managed'
-    | 'unmanaged' ('[' identifier (',' identifier)* ']')?
+    | 'unmanaged' ('[' unmanaged_calling_convention ']')?
+    ;
+
+unmanaged_calling_convention
+    : 'Cdecl'
+    | 'Stdcall'
+    | 'Thiscall'
+    | 'Fastcall'
+    | identifier (',' identifier)*
+
+funptr_parameter_list
+    : (funcptr_parameter ',')*
+    ;
+
+funcptr_parameter
+    : funcptr_parameter_modifier? type
+    ;
+
+funcptr_return_type
+    : funcptr_return_modifier? return_type
+    ;
 
 funcptr_parameter_modifier
     : 'ref'
@@ -120,10 +140,9 @@ funcptr_return_modifier
     ;
 ```
 
-The `unmanaged` calling convention represents the default calling convention for native code on the current platform, and is encoded as winapi.
-All `identifer`s in the `unmanaged` case of the `calling_convention_specifier` must start with `CallConv`, and come from the
-`System.Runtime.CompilerServices` namespace in the program's core library (defined as one library that references no dependencies and defines
-`System.Object`).
+If no `calling_convention_specifier` is provided, the default is `managed`. The precise metadata encoding
+of the `calling_convention_specifier` and what `identifier`s are valid in the `unmanaged_calling_convention` is
+covered in [Metadata Representation of Calling Conventions](#Metadata-Representation-of-Calling-Conventions).
 
 ``` csharp
 delegate int Func1(string s);
@@ -173,7 +192,7 @@ unsafe class Util {
 ```
 
 In an unsafe context, a method `M` is compatible with a function pointer type `F` if all of the following are true:
-- `M` and `F` have the same number of parameters, and each parameter in `D` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter in `F`.
+- `M` and `F` have the same number of parameters, and each parameter in `M` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter in `F`.
 - For each value parameter (a parameter with no `ref`, `out`, or `in` modifier), an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `M` to the corresponding parameter type in `F`.
 - For each `ref`, `out`, or `in` parameter, the parameter type in `M` is the same as the corresponding parameter type in `F`.
 - If the return type is by value (no `ref` or `ref readonly`), an identity, implicit reference, or implicit pointer conversion exists from the return type of `F` to the return type of `M`.
@@ -183,16 +202,12 @@ In an unsafe context, a method `M` is compatible with a function pointer type `F
 
 In an unsafe context, an implicit conversion exists from an address-of expression whose target is a method group `E` to a compatible function pointer type `F` if `E` contains at least one method that is applicable in its normal form to an argument list constructed by use of the parameter types and modifiers of `F`, as described in the following.
 - A single method `M` is selected corresponding to a method invocation of the form `E(A)` with the following modifications:
-   - The arguments list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref`, `out`, or `in`) of the corresponding _formal\_parameter\_list_ of `D`.
+   - The arguments list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref`, `out`, or `in`) of the corresponding _funcptr\_parameter\_list_ of `F`.
    - The candidate methods are only those methods that are applicable in their normal form, not those applicable in their expanded form.
    - The candidate methods are only those methods that are static.
-- If the algorithm of Method invocations produces an error, then a compile-time error occurs. Otherwise, the algorithm produces a single best method `M` having the same number of parameters as `F` and the conversion is considered to exist.
+- If the algorithm of overload resolution produces an error, then a compile-time error occurs. Otherwise, the algorithm produces a single best method `M` having the same number of parameters as `F` and the conversion is considered to exist.
 - The selected method `M` must be compatible (as defined above) with the function pointer type `F`. Otherwise, a compile-time error occurs.
 - The result of the conversion is a function pointer of type `F`.
-
-An implicit conversion exists from an address-of expression whose target is a method group `E` to `void*` if there is only one static method `M` in `E`.
-If there is one static method, then the single best method from `E` is `M`.
-Otherwise, a compile-time error occurs.
 
 This means developers can depend on overload resolution rules to work in conjunction with the
 address-of operator:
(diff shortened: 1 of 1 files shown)


proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+### UnmanagedCallersOnlyAttribute
+
+`UnamangedCallersOnlyAttribute` is an attribute used by the CLR to indicate that a method should be called with a specific
+calling convention. Because of this, we introduce the following support for working with the attribute:
+
+* It is an error to directly call a method annotated with this attribute from C#. Users must obtain a function pointer to
+the method and then invoke that pointer.
+* It is an error to apply the attribute to anything other than a static method. The C# compiler will mark any non-static
+methods imported from metadata with this attribute as unsupported by the language.
+* It is an error to have non-blittable types as parameters or the return type of a method marked with the attribute.
```
tannergooding, line 306: The language doesn't, afaik, have a concept of "blittable" today; only "unmanaged". Is it introducing the former concept or using the latter?

333fred: I mean to write unmanaged, will fix.

proposals/function-pointers.md
```diff
@@ -36,15 +36,25 @@ of a `delegate*` will use `calli` where invocation of a `delegate` will use `cal
...
+delegate* unmanged<int, int>;
+
 // This method will be invoked using the cdecl calling convention
-delegate* cdecl<int, int>;
+delegate* unmanged[CallConvCdecl] <int, int>;
 
-// This method will be invoked using the stdcall calling convention
-delegate* stdcall<int, int>;
+// This method will be invoked using the stdcall calling convention, and suppresses GC transition
+delegate* unmanaged[CallConvStdCall, CallConvSuppressGCTransition] <int, int>;
```
tannergooding, line 57: I take it we are not allowing `unmanaged[StdCall, SuppressGCTransition]` then? I don't remember a decision being made here, but I may have missed that bit 😄

proposals/function-pointers.md
tannergooding: Will the same apply to `unmanaged` function pointers or just `UnmanagedCallersOnlyAttribute`?

proposals/function-pointers.md
333fred: I went for simplicity of implementation, I'll add a consideration note.

proposals/function-pointers.md
```diff
@@ -169,7 +178,7 @@ In an unsafe context, a method `M` is compatible with a function pointer type `F
 - For each `ref`, `out`, or `in` parameter, the parameter type in `M` is the same as the corresponding parameter type in `F`.
```
AaronRobinsonMSFT, line 178: `D` is mentioned above. Should that be updated to `F`?

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+* `System.Runtime.CompilerServices.CallConvStdcall`
+* `System.Runtime.CompilerServices.CallConvFastcall`
+* `System.Runtime.CompilerServices.CallConvThiscall`
+
+If the identifier specified for the `unmanaged` calling convention binds to one of these types, then the function pointer
+is emitted with the calling convention bit set to the appropriate value, and encodes the calling convention type as the first
+modopt on the return type of the function pointer. For any other `CallConv` type or combination of such types, the compiler
+emits a calling convention bit of Unmanaged (0x9), and encodes all `CallConv` types specified as the first modopts on the return
+type of the function pointer. What calling conventions and combinations are valid is determined by the runtime and platform of
+final execution, and no validation is performed that a valid combination of calling conventions has been specified beyond
```
AaronRobinsonMSFT, line 289: >What calling conventions and combinations are valid is determined by the runtime and platform of final execution

/cc @lambdageek

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+ensuring that all calling convention types come from the right namespace in the core library of the program. If the return
+type of a pointer is by `ref`, then all calling conventions will be encoded after the ref specifier.
+
+When interpreting a function pointer in metadata, all modopts that start with `CallConv` and come from the
+`System.Runtime.CompilerServices` namespace in the core library are considered to be part of the calling convention of the
+function pointer.
+
+### UnmanagedCallersOnlyAttribute
+
+`UnamangedCallersOnlyAttribute` is an attribute used by the CLR to indicate that a method should be called with a specific
```
jkotas, line 299:
```suggestion
`UnmanagedCallersOnlyAttribute` is an attribute used by the CLR to indicate that a method should be called with a specific
```

jkotas: Same typo is in the PR title

proposals/function-pointers.md
svick, line 57: It looks like there is a distinction between the calling convention (like `StdCall`) and calling convention "modifier" (like `SuppressGCTransition`), but this proposal ignores that.

Would it make sense to somehow perform basic checking based on that, so that e.g. the nonsensical `unmanaged[CallConvCdecl, CallConvStdCall]` did not compile? Even if the compiler didn't understand exactly which combinations of calling conventions and their modifiers are allowed, I think such basic checking could still be worthwhile.

tannergooding: It's not readily possible to know what combinations are valid and what are not or what will/wont be in the future. For example `thiscall` can be combined with basically any other convention is in some cases (`thiscall`+`stdcall` is the default for x86 COM, for example).

It's easier to just allow any combination and let the runtime decide.

proposals/function-pointers.md
```diff
@@ -78,6 +88,7 @@ Restrictions:
 - Custom attributes cannot be applied to a `delegate*` or any of its elements.
 - A `delegate*` parameter cannot be marked as `params`
 - A `delegate*` type has all of the restrictions of a normal pointer type.
+- Pointer arithmetic cannot be performed directly on pointer types.
```
AlekseyTs, line 91: >on pointer types [](start = 50, length = 16)

On **function** pointer types?

proposals/function-pointers.md
```diff
@@ -114,7 +121,9 @@ funcptr_return_modifier
 ```
 
 The `unmanaged` calling convention represents the default calling convention for native code on the current platform, and is encoded as winapi.
-All `calling_convention`s are contextual keywords when preceded by a `delegate*`.
+All `identifer`s in the `unmanaged` case of the `calling_convention_specifier` must start with `CallConv`, and come from the
+`System.Runtime.CompilerServices` namespace in the program's core library (defined as one library that references no dependencies and defines
```
AlekseyTs, line 125: Identifiers cannot really come from a namespace. Perhaps we should say that the identifier must match the name of a type in that namespace instead?

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+
+The following calling convention specifiers are specially recognized by the compiler:
+
+* `System.Runtime.CompilerServices.CallConvCdecl`
+* `System.Runtime.CompilerServices.CallConvStdcall`
+* `System.Runtime.CompilerServices.CallConvFastcall`
+* `System.Runtime.CompilerServices.CallConvThiscall`
+
+If the identifier specified for the `unmanaged` calling convention binds to one of these types, then the function pointer
+is emitted with the calling convention bit set to the appropriate value, and encodes the calling convention type as the first
```
AlekseyTs, line 285: >with the calling convention bit set to the appropriate value [](start = 11, length = 60)

We should be specific what is the value.

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+modopt on the return type of the function pointer. For any other `CallConv` type or combination of such types, the compiler
```
AlekseyTs, line 286: Regarding "as the first modopt". I do not believe there is such a requirement. I think we should simply say that the modopt is added to the set of custom modifiers at the beginning of the function pointer signature.

AlekseyTs, line 286: >on the return type [](start = 7, length = 18)

"on the return type" is also not accurate, the modifier can be on `ref` rather than on type.

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+emits a calling convention bit of Unmanaged (0x9), and encodes all `CallConv` types specified as the first modopts on the return
```
AlekseyTs, line 287: >as the first modopts on the return [](start = 94, length = 34)

Same comment here. And we shouldn't specify any relative order for modifiers because in case of overriding/implementing the syntax doesn't really define the order.

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+type of the function pointer. What calling conventions and combinations are valid is determined by the runtime and platform of
+final execution, and no validation is performed that a valid combination of calling conventions has been specified beyond
+ensuring that all calling convention types come from the right namespace in the core library of the program. If the return
+type of a pointer is by `ref`, then all calling conventions will be encoded after the ref specifier.
```
AlekseyTs, line 291: I do not believe this is accurate.

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+
+When interpreting a function pointer in metadata, all modopts that start with `CallConv` and come from the
```
AlekseyTs, line 293: >all modopts [](start = 50, length = 11)

I think this doesn't quite match the latest proposal discussed on the internal thread.

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+calling convention. Because of this, we introduce the following support for working with the attribute:
+
+* It is an error to directly call a method annotated with this attribute from C#. Users must obtain a function pointer to
+the method and then invoke that pointer.
+* It is an error to apply the attribute to anything other than a static method. The C# compiler will mark any non-static
+methods imported from metadata with this attribute as unsupported by the language.
+* It is an error to have non-blittable types as parameters or the return type of a method marked with the attribute.
+* It is an error to convert a method marked with the attribute to a delegate type.
+* The calling convention specified by the attribute contributes to the calling convention the compiler recognizes when
+a method is converted to a function pointer.
```
AlekseyTs, line 309: How exactly compiler recognizes them, what do they affect? How the attribute is defined?

proposals/function-pointers.md
```diff
@@ -169,7 +178,7 @@ In an unsafe context, a method `M` is compatible with a function pointer type `F
 - For each `ref`, `out`, or `in` parameter, the parameter type in `M` is the same as the corresponding parameter type in `F`.
 - If the return type is by value (no `ref` or `ref readonly`), an identity, implicit reference, or implicit pointer conversion exists from the return type of `F` to the return type of `M`.
 - If the return type is by reference (`ref` or `ref readonly`), the return type and `ref` modifiers of `F` are the same as the return type and `ref` modifiers of `M`.
-- The calling convention of `M` is the same as the calling convention of `F`.
+- The calling convention of `M` is the same as the calling convention of `F`. This includes both the calling convention bit, as well as any calling convention flags specified in the unmanaged identifier.
```
AlekseyTs, line 181: >This includes both the calling convention bit, as well as any calling convention flags specified in the unmanaged identifier. [](start = 78, length = 125)

I think this doesn't quite match the last proposal discussed on the internal thread. We need to specify precise rules for determining if calling conventions are compatible. Also, I am not sure if talking about that in syntactic terms ("calling convention flags specified in the unmanaged identifier") is the right approach.

AlekseyTs:
Done with review pass (iteration 2)

proposals/function-pointers.md
gafter: Instead of `come from` I suggest `bind to a type in`

---
In reply to: [422510659](https://github.com/dotnet/csharplang/pull/3441#discussion_r422510659) [](ancestors = 422510659)

proposals/function-pointers.md
AaronRobinsonMSFT: >Would it make sense to somehow perform basic checking based on that

@svick This concern was raised during design. The reason for not checking is an attempt to avoid having to update the compiler whenever a new combination is supported by the runtime. As @tannergooding mentions there are some combinations that could be described as a combination of multiple `CallConv*` types. Letting the implementing runtime decide what it supports avoids encoding a support matrix in more than one location (e.g. Roslyn and runtime) and lessens the time to market for consumption of those combinations.

The counter argument here is that deferring the error until run time is less than ideal for confidence in a compiled assembly. We can all appreciate that concern. The benefit of quicker adoption of desired combinations in an already advanced interop scenario outweighed the cost of degraded compile time validation. This is why the decision was made.

svick: @AaronRobinsonMSFT @tannergooding I get that hardcoding the full support matrix into the compiler is undesirable.

What I was proposing is that there would be two kinds of `CallConv`: "base" (like `Cdecl` or `StdCall`) and "modifier" (like `Thiscall` or `SuppressGCTransition`). An `unmanaged` function pointer could then have:

* Zero or one base `CallConv`.
* Zero or more modifier `CallConv`.

I think that this way, all valid combinations will still be allowed while many invalid combination will be disallowed. Or am I wrong?

Though thinking about it some more, I think it would actually be possible to encode the support matrix into the metadata of the corelib reference assembly: each base `CallConv` could have an attribute that specified the modifier `CallConv`s that it allows, or even the combination of modifier `CallConv` it allows.

Using this approach, the compiler can know the support matrix, even though altering the support matrix does not involve updating the compiler.

tannergooding: That would still require updating the compiler any time a new convention is released. Having an analyzer that flags potentially bad or unsupported combinations is likely a much better alternative as it is more trivially updated, shipped, and ignored (if desired).

The support might also differ across various runtime implementations. That is, while CoreCLR might support `X, Y, and Z`; you might have another runtime like `Unity` or a `CIL to LLVM` transpiler which supports a different set (or different combinations).

svick: @tannergooding

> That would still require updating the compiler any time a new convention is released. 

It wouldn't, but I forgot to explain that: the compiler would differentiate between base and modifier `CallConv` based on metadata. That could be represented by a base class of the `CallConv` class (e.g. `public class CallConvThiscall :  ModifierCallConv { … }`), or an attribute applied to the `CallConv` class.

> The support might also differ across various runtime implementations.

If they use the same corlib reference assembly, yes, that would be a problem.

jaredpar: > That would still require updating the compiler any time a new convention is released.

I've thought a lot about this and I keep asking myself the following: is that so bad? The compiler and runtime release as a pair now. Updating the calling convention is a small amount of work, or would be if we designed for this future. So is it really that bad if we said that every time the runtime adds a new calling convention that we have to change the compiler to support it?

proposals/function-pointers.md
jaredpar, line 293: Is this truly "all modopts" or is it ones that appear in specific locations?

proposals/function-pointers.md
jaredpar, line 291:
```suggestion
type of a pointer is by `ref`, then all calling conventions will be encoded after the `ref` specifier.
```

proposals/function-pointers.md
```diff
@@ -263,62 +272,41 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+type of the function pointer. What calling conventions and combinations are valid is determined by the runtime and platform of
+final execution, and no validation is performed that a valid combination of calling conventions has been specified beyond
+ensuring that all calling convention types come from the right namespace in the core library of the program. If the return
+type of a pointer is by `ref`, then all calling conventions will be encoded after the ref specifier.
+
+When interpreting a function pointer in metadata, all modopts that start with `CallConv` and come from the
+`System.Runtime.CompilerServices` namespace in the core library are considered to be part of the calling convention of the
+function pointer.
+
+### UnmanagedCallersOnlyAttribute
```
jaredpar, line 297: Consider listing the FQN of this type in this section.

proposals/function-pointers.md
jkotas: There are many runtime failure modes for interop that have complex conditions and that the compiler does not check for today. One example from many - the compiler does not check whether `MarshalAs` makes sense for given argument type in regular PInvokes. We could duplicate the runtime checks for all these in the compiler, but the value of it seems very low. I do not think that the support matrix for the unmanaged calling convention modifiers should be on different plan from the rest of interop.

proposals/function-pointers.md
davidwrighton: No, this isn't accurate. These modifiers must directly follow `ParamCount` in the method signature.

proposals/function-pointers.md
tannergooding: Thanks, it's mostly a convenience thing but I think it would help greatly with readability and with typing when you have hundreds of function pointers you are declaring/reviewing (such as DirectX which is all COM `ThisCall` + `StdCall`)...

333fred:
@AlekseyTs @davidwrighton @jkotas @jaredpar I've updated the proposal with the latest design decisions. Please let me know if there's any more feedback.

proposals/function-pointers.md
```diff
@@ -274,39 +289,105 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+and has no dependencies.
 
-`UnmanagedCallersOnlyAttribute` is an attribute used by the CLR to indicate that a method should be called with a specific
-calling convention. Because of this, we introduce the following support for working with the attribute:
+If lookup succeeds on all of the `identifier`s specified in an `unmanaged_calling_convention`, we encode the `CallKind` as
+`unmanaged ext`, and encode each of the resolved types in the set of `modopt`s at the beginning of the function pointer
+signature. As a note, these rules mean that users cannot prefix these `identifier`s with `CallConv`, as that will result
+in looking up `CallConvCallConvVectorCall`.
+
+When interpreting metadata, the we first look at the `CallKind`. If it is anything other than `unmanaged ext`, we ignore all
```
AaronRobinsonMSFT, line 336: ~~the we~~ -> "we"

proposals/function-pointers.md
```diff
@@ -274,39 +289,105 @@ We use `System.Runtime.InteropServices.OutAttribute`, applied as a `modreq` to t
...
+* The is defined in the core library, which is the library that references no other libraries and defines `System.Object`.
+* The type is defined in the `System.Runtime.CompilerServices` namespace.
+* The type starts with the prefix `CallConv`.
+
+ These represent the types that must be found when performing lookup on the `identifier`s in an
+`unmanaged_calling_convention` when defining a function pointer type in source.
+
+It is an error to attempt to use a function pointer with a `CallKind` of `unmanaged ext` if the target runtime does not
+support the feature. This will be determined by looking for the presence of the
+`System.Runtime.CompilerServices.RuntimeFeature.UnmanagedCallKind` constant. If this constant is present, the runtime is
```
AaronRobinsonMSFT, line 350: I thought the presence of the `UnmanagedCallersOnlyAttribute` was going to be enough here?

AaronRobinsonMSFT: I see the note below.

333fred: I don't particularly want to condition the support based on that attribute, as I think they're orthogonal and we shouldn't tie them together, but we can discuss that Thursday in design review :).

proposals/function-pointers.md
```diff
@@ -164,27 +192,23 @@ unsafe class Util {
...
+- `M` and `F` have the same number of parameters, and each parameter in `M` has the same `ref`, `out`, or `in` modifiers as the corresponding parameter in `F`.
 - For each value parameter (a parameter with no `ref`, `out`, or `in` modifier), an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `M` to the corresponding parameter type in `F`.
 - For each `ref`, `out`, or `in` parameter, the parameter type in `M` is the same as the corresponding parameter type in `F`.
 - If the return type is by value (no `ref` or `ref readonly`), an identity, implicit reference, or implicit pointer conversion exists from the return type of `F` to the return type of `M`.
 - If the return type is by reference (`ref` or `ref readonly`), the return type and `ref` modifiers of `F` are the same as the return type and `ref` modifiers of `M`.
-- The calling convention of `M` is the same as the calling convention of `F`.
+- The calling convention of `M` is the same as the calling convention of `F`. This includes both the calling convention bit, as well as any calling convention flags specified in the unmanaged identifier.
```
AlekseyTs, line 200: >This includes both the calling convention bit, as well as any calling convention flags specified in the unmanaged identifier. [](start = 78, length = 125)

I think we would want to expand on that. In particular, I think we should have a special handling of CallConvCDecl  modifier (and the like "special" legacy modifiers) in presence of a cdecl bit, as apposed to in presence of unmanaged bit (what ever the new bit we are going to use). One of the reasons, is that, I think, we would want to support overriding a method with a `cdecl with modifier` function pointer (C++ creates those) by a method with just `cdecl` function pointer in it. While looking for an override, we should make sure calling conventions match, once we find the matching override, the normal process will inherit the modifiers. Same goes for interface implementations.

AlekseyTs: I see that you expanded on that in a section below. Looks reasonable.

---
In reply to: [448095465](https://github.com/dotnet/csharplang/pull/3441#discussion_r448095465) [](ancestors = 448095465)

proposals/function-pointers.md
```diff
@@ -36,15 +36,25 @@ of a `delegate*` will use `calli` where invocation of a `delegate` will use `cal
...
+`System.Runtime.CompilerServices` namespace. These types must come from the program's core library, and the set of valid
+combinations is platform-dependent.
 
 ``` csharp
+//This method has a managed calling convention. This is the same as leaving the managed keyword off.
+delegate* managed<int, int>;
+
+// This method will be invoked using whatever the default unmanaged calling convention on the runtime
+// platform is. This is platform and architecture dependent and is determined by the CLR at runtime.
+delegate* unmanged<int, int>;
```
davidwrighton, line 51: unmanged should be unmanaged, both here, and in the next example.

PR #3451, opened 2020-05-12
Updated status of null coalescing assignment from proposed to implementation completed

stylesm:
The feature is available in C# 8. I can't see guidance on what should be updated in these docs to reflect that, so not sure if what I've changed is what should be changed!

### DIFF
--- a/proposals/csharp-8.0/null-coalescing-assignment.md
+++ b/proposals/csharp-8.0/null-coalescing-assignment.md
@@ -1,9 +1,9 @@
 # null coalescing assignment
 
-* [x] Proposed
-* [ ] Prototype: Not Started
-* [ ] Implementation: Not Started
-* [ ] Specification: Below
+* [ ] Proposed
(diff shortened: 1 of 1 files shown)


proposals/csharp-8.0/null-coalescing-assignment.md
```diff
@@ -1,9 +1,9 @@
 # null coalescing assignment
 
-* [x] Proposed
-* [ ] Prototype: Not Started
-* [ ] Implementation: Not Started
-* [ ] Specification: Below
+* [ ] Proposed
+* [ ] Prototype: Completed
+* [ ] Implementation: Completed
```
333fred, line 5: All the boxes should be checked.

stylesm: Thanks, done

333fred:
Thanks @stylesm!

PR #3459, opened 2020-05-14
Add doc for C# 6.0 struct autoprop initialization

gafter:


proposals/csharp-6.0/struct-autoprop-init.md
```diff
@@ -0,0 +1,57 @@
...
+        this = default(S); 
+        this.X = x; 
+        this.Y = y; 
+    } 
+} 
+```
+
+Unfortunately, both methods effectively disabled definite assignment analysis for instance fields for the given constructor. 
+
+C# 6.0 relaxes this rules by tracking assignments to instance auto-properties as if they were assignments to their backing fields, so the code in the beginning of this section is now valid.
```
MadsTorgersen, line 57: "this rules" => "this rule" or "these rules".

PR #3484, opened 2020-05-21
Improve writing in the top-level statements proposal

thomasio101:
I've made an effort to improve the spelling, grammar, and writing style in `proposals/top-level-statements.md`.

### DIFF
--- a/proposals/top-level-statements.md
+++ b/proposals/top-level-statements.md
@@ -124,7 +124,7 @@ static class $Program
 }
 ```
 
-At the same time an example like this:
+At the same time an example like this;
 ``` c#
 await System.Threading.Tasks.Task.Delay(1000);
 System.Console.WriteLine("Hi!");
@@ -142,7 +142,7 @@ static class $Program
 }
 ```
 
-An example like this:
+An example like this;
 ``` c#
 await System.Threading.Tasks.Task.Delay(1000);
 System.Console.WriteLine("Hi!");
@@ -162,7 +162,7 @@ static class $Program
 }
 ```
 
-And an example like this:
+And an example like this;
 ``` c#
 System.Console.WriteLine("Hi!");
 return 2;
--- a/proposals/top-level-statements.md
+++ b/proposals/top-level-statements.md
@@ -183,7 +183,7 @@ static class $Program
 ### Scope of top-level local variables and local functions
 
 Even though top-level local variables and functions are "wrapped" 
-into the generated entry point method, they should still be in scope throughout the program in
+into the generated entry point method, they should still be in-scope throughout the program in
 every compilation unit.
 For the purpose of simple-name evaluation, once the global namespace is reached:
 - First, an attempt is made to evaluate the name within the generated entry point method and 
--- a/proposals/top-level-statements.md
+++ b/proposals/top-level-statements.md
@@ -89,7 +89,7 @@ The method is designated as the entry point of the program. Explicitly declared 
 could be considered as entry point candidates are ignored. A warning is reported when that happens. It is
 an error to specify the `-main:<type>` compiler switch when there are top-level statements.
 
-The entry point method always has one formal parameter, ```string[] args```. The execution environment creates and passes a ```string[]``` argument containing the command-line arguments that were specified when the application was started. The ```string[]``` argument is never null, but it may have a length of zero if no command-line arguments were specified. The ‘args’ parameter is in scope  within top-level statements and is not in scope outside of them. Regular name conflict/shadowing rules apply.
+The entry point method always has one formal parameter; ```string[] args```. The execution environment creates and passes a ```string[]``` argument containing the command-line arguments that were specified when the application was started. The ```string[]``` argument is never null, but it may have a length of zero if no command-line arguments were specified. The ‘args’ parameter is in scope  within top-level statements and is not in scope outside of them. Regular name conflict/shadowing rules apply.
 
 Async operations are allowed in top-level statements to the degree they are allowed in statements within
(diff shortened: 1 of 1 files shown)


proposals/top-level-statements.md
```diff
@@ -81,15 +81,15 @@ static class Program
...
+an error to specify the `-main:<type>` compiler switch when there are top-level statements.
 
-The entry point method always has one formal parameter, ```string[] args```. The execution environment creates and passes a ```string[]``` argument containing the command-line arguments that were specified when the application was started. The ```string[]``` argument is never null, but it may have a length of zero if no command-line arguments were specified. The ‘args’ parameter is in scope  within top-level statements and is not in scope outside of them. Regular name conflict/shadowing rules apply.
+The entry point method always has one formal parameter; ```string[] args```. The execution environment creates and passes a ```string[]``` argument containing the command-line arguments that were specified when the application was started. The ```string[]``` argument is never null, but it may have a length of zero if no command-line arguments were specified. The ‘args’ parameter is in scope  within top-level statements and is not in scope outside of them. Regular name conflict/shadowing rules apply.
```
333fred, line 92: All of these changes to `;` are incorrect. The sentence is not ending, it is showing an example. Same with the example changes below.

proposals/top-level-statements.md
```diff
@@ -81,15 +81,15 @@ static class Program
...
 ```
 
-Note that the names "Program" and "Main" are used only for illustrations purposes, actual names used by
-compiler are implementation dependent and neither the type, nor the method can be referenced by name from
+Note that the names "Program" and "Main" are used only for illustration purposes, actual names used by the
+compiler are implementation-dependent and neither the type, nor the method can be referenced by name from
 source code.
 
 The method is designated as the entry point of the program. Explicitly declared methods that by convention 
-could be considered as an entry point candidates are ignored. A warning is reported when that happens. It is
```
333fred, line 89: Please revert the `an` change. A better grammatical change would be to remove the `as`, but it reads fine as is.

proposals/top-level-statements.md
```diff
@@ -183,7 +183,7 @@ static class $Program
 ### Scope of top-level local variables and local functions
 
 Even though top-level local variables and functions are "wrapped" 
-into the generated entry point method, they should still be in scope throughout the program in
+into the generated entry point method, they should still be in-scope throughout the program in
```
333fred, line 186: Please revert. `in` is a modifier on `scope`. Correct usage of a hyphen would be when there are multiple modifiers modifying the same noun.

proposals/top-level-statements.md
```diff
@@ -81,15 +81,15 @@ static class Program
 }
 ```
 
-Note that the names "Program" and "Main" are used only for illustrations purposes, actual names used by
-compiler are implementation dependent and neither the type, nor the method can be referenced by name from
+Note that the names "Program" and "Main" are used only for illustration purposes, actual names used by the
+compiler are implementation-dependent and neither the type, nor the method can be referenced by name from
```
333fred, line 85: Please remove the hyphen. `implementation` and `dependent` are not separate modifiers on `and`.

proposals/top-level-statements.md
thomasio101: `entry point candidates` is plural, thus `an` is invalid in this context, and, as such, I removed `an`.

proposals/top-level-statements.md
thomasio101: Please see the following example;  
`in-house software`  
Hyphenation is used in this situation due to the adjective `in-house` ascribing a single property, that of being developed internally, to the noun `software`.  
Similarly, in the following phrase;  
`in-scope variables`  
Hyphenation is used due to the adjective `in-scope` ascribing a single property, that of being in the scope, to the noun `variables`.

Thus, I decided to use hyphenation in the context of this commit as I assume the writer intended to describe the local variables and functions to be in the scope, and, thus, to be in-scope.

proposals/top-level-statements.md
thomasio101: I assume the writer intended to ascribe a single property, that of being dependent on the implementation, to `actual names used by the compiler`. Thus, I used the adjective `implementation-dependent`, **with hyphenation**, as it ascribes a single property to `actual names used by the compiler`.

proposals/top-level-statements.md
thomasio101: The use of `;` in the context of line 92 is correct, `one formal parameter` refers to, and directly precedes, `string[] args`, and, thus, a semicolon should be used to punctuate this.

proposals/top-level-statements.md
333fred: `scope` is a noun. It is the thing that the variables are in. The hyphen is not correct.

proposals/top-level-statements.md
333fred: We do not use a hyphen for this. Please remove.

proposals/top-level-statements.md
333fred: Fair, I misread this.

proposals/top-level-statements.md
333fred: It is not correct. Semicolons are used to break up 2 sentences; each sentence should be complete on its own. ```string[] args``` is not a complete sentence. A `:` would be acceptable, as is a `,` as written here. Please revert all conversions to `;`.

proposals/top-level-statements.md
agocke: Implementation-dependent is grammatically correct. I don't know if we have a style guide which says otherwise.

PR #3505, opened 2020-05-27
Update records-wip.md

agocke:


gafter:
> For each record parameter of a record type declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:

instance property?  Does the access on the inherited property matter?

---
Refers to: proposals/records-wip.md:56 in 69f18ad. [](commit_id = 69f18ad46fa0a44ad23351ee80ec7a4b3c5a976b, deletion_comment = False)

PR #3520, opened 2020-05-31
Add spec for nominal records

agocke:


YairHalberstadt:
Isn't this for positional records?

agocke:
I don't think so: the spec previously said that a record must have both a parameter list and a `data modifier`. Now it says it must have the `data` modifier and may optionally have a parameter list

PR #3527, opened 2020-05-31
Bring records proposal up to date 

agocke:
Try to bring the proposal in line with LDM decisions. Inheritance is somewhat described, but there
have been no affirmative decisions by LDM so this is still in flux

### DIFF
--- a/proposals/records.md
+++ b/proposals/records.md
@@ -32,7 +32,10 @@ a
 
 In addition to the members declared in the record body, a record type has additional synthesized members.
 Members are synthesized unless a concrete (non-abstract) member with a "matching" signature is either
-inherited or declared in the record body. The synthesized members are as follows:
+inherited or declared in the record body. Two members are considered matching if they have the same
+signature or would be considered "hiding" in an inheritance scenario.
+
+The synthesized members are as follows:
 
 ### Equality members
 
@@ -51,7 +54,7 @@ is present in the base type, the synthesized property overrides the base. If the
 `EqualityContract` is sealed or non-virtual, an error is produced.
 
 `T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
-all the instance fields declared in the receiver type are equal to the fields of the other type,
+all accessible instance fields in the receiver are equal to the fields of the parameter
 and `this.EqualityContract` equals `other.EqualityContract`.
 
 `object.Equals` performs the equivalent of
@@ -82,8 +85,7 @@ produced if the base type clone method is sealed.
 ## Positional record members
 
 In addition to the above members, records with a parameter list ("positional records") synthesize
-the following members, if a concrete (i.e. non-abstract) member with the same signature (or name
-if the member is a field or property) is not already present:
+additional members with the same conditions as the members above.
 
 ### Primary Constructor
 
--- a/proposals/records.md
+++ b/proposals/records.md
@@ -49,9 +49,9 @@ containing type:
 
 If either `object.GetHashCode()` or `object.Equals(object)` are sealed, an error is produced.
 
-`EqualityContract` is a virtual instance property which returns `typeof(T)`. If it
-is present in the base type, the synthesized property overrides the base. If the base
-`EqualityContract` is sealed or non-virtual, an error is produced.
+`EqualityContract` is a virtual instance property which returns `typeof(T)`. If the base type
+defines an `EqualityContract` it is overridden in the derived record. If the base `EqualityContract`
+is sealed or non-virtual, an error is produced.
 
 `T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
 all accessible instance fields in the receiver are equal to the fields of the parameter
@@ -65,14 +65,13 @@ override Equals(object o) => Equals(o as T);
 
 ### Copy and Clone members
 
-A record type contains two synthesized copying members if methods with the same
-signature are not already declared within the record type:
+A record type contains two synthesized copying members:
 
 * A protected constructor taking a single argument of the record type.
-* A public parameter-less virtual instance "clone" method with a compiler-reserved name
+* A public parameterless virtual instance "clone" method with a compiler-reserved name
 
 The protected constructor is referred to as the "copy constructor" and the synthesized
-body copies the values of all instance fields declared in the input type to the corresponding
+body copies the values of all accessible instance fields in the input type to the corresponding
 fields of `this`.
 
 The "clone" method returns the result of a call to a constructor with the same signature as the
@@ -146,10 +145,10 @@ in the `member_initializer_list`.
 A valid `with` expression has a receiver with a non-void type. The receiver type must contain an
 accessible synthesized record "clone" method.
 
-On the right hand side of the `with` expression is an `member_initializer_list` with a sequence
-of assignments to *identifier*, which must an accessible instance field or property of the return
+On the right hand side of the `with` expression is a `member_initializer_list` with a sequence
+of assignments to *identifier*, which must be an accessible instance field or property of the return
 type of the `Clone()` method.
 
-Each `member_initializer` is processed the same way as an assignment to the field or property
-target on the return value of the record clone method. The clone method is executed only once
+Each `member_initializer` is processed the same way as an assignment to a field or property
+access of the return value of the record clone method. The clone method is executed only once
 and the assignments are processed in lexical order.
--- a/proposals/records.md
+++ b/proposals/records.md
@@ -95,12 +95,9 @@ constructor and a constructor with the same signature already present in the cla
 
 At runtime the primary constructor
 
-1. assigns synthesized fields or properties (see [Properties](#properties)) with the value of a primary constructor
-   parameter of the same name
-
 1. executes the instance initializers appearing in the class-body
 
-2. invokes the base class constructor with the arguments provided in the `record_base` clause, if present
+1. invokes the base class constructor with the arguments provided in the `record_base` clause, if present
 
 
 ### Properties
@@ -111,7 +108,8 @@ member whose name and type are taken from the value parameter declaration.
 For a record:
 
 * A public `get` and `init` auto-property is created (see separate `init` accessor specification).
-  Each "matching" inherited abstract accessor is overridden.
+  Each "matching" inherited abstract accessor is overridden. The auto-property is initialized to
+  the value of the corresponding primary constructor parameter.
 
 ### Deconstruct
 
--- a/proposals/records.md
+++ b/proposals/records.md
@@ -1,54 +1,50 @@
 
-# Records Work-in-Progress
+# Records
 
-Unlike the other records proposals, this is not a proposal in itself, but a work-in-progress designed to record consensus design
-decisions for the records feature. Specification detail will be added as necessary to resolve questions.
+This proposal tracks the specification for the C# 9 records feature, as agreed to by the C#
+language design team.
 
-The syntax for a record is proposed to be added as follows:
+The syntax for a record is as follows:
 
 ```antlr
-class_declaration
-    : attributes? class_modifiers? 'partial'? 'class' identifier type_parameter_list?
-      parameter_list? type_parameter_constraints_clauses? class_body
+record_declaration
+    : attributes? class_modifier* 'partial'? 'record' identifier type_parameter_list?
+      parameter_list? record_base? type_parameter_constraints_clause* record_body
     ;
 
-struct_declaration
-    : attributes? struct_modifiers? 'partial'? 'struct' identifier type_parameter_list?
-      parameter_list? struct_interfaces? type_parameter_constraints_clauses? struct_body
+record_base
+    : ':' class_type argument_list?
+    | ':' interface_type_list
+    | ':' class_type argument_list? interface_type_list
     ;
 
-class_body
-    : '{' class_member_declarations? '}'
-    | ';'
-    ;
-
-struct_body
-    : '{' struct_members_declarations? '}'
+record_body
+    : '{' class_member_declaration* '}'
     | ';'
     ;
 ```
 
-The `attributes` non-terminal will also permit a new contextual attribute, `data`.
-
-A class (struct) declared with a parameter list or `data` modifier is called a record class (record struct), either of which is a record type.
-
-It is an error to declare a record type without the `data` modifier.
-
 ## Members of a record type
 
-In addition to the members declared in the class or struct body, a record type has the following additional members:
+In addition to the members declared in the record body, a record type has the following additional members:
 
 ### Equality members
 
-Record types produce synthesized implementations for the following methods:
+Record types produce synthesized implementations for the following methods, where `T` is the
+containing type:
 
-* `object.GetHashCode()` override, unless it is sealed or user provided
-* `object.Equals(object)` override, unless it is sealed or user provided
+* `object.GetHashCode()` override, unless it is sealed (error) or user provided
+* `object.Equals(object)` override, unless it is sealed (error) or user provided
 * `T Equals(T)` method, where `T` is the current type
+* `Type EqualityContract` get-only property
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is
-true if and only if all the instance fields declared in the receiver type
-are equal to the fields of the other type.
+`EqualityContract` is a virtual instance property which returns `typeof(T)`. If it
+is present in the base type, the synthesized property overrides the base. If the base
+is sealed or non-virtual, an error is produced.
+
+`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
+all the instance fields declared in the receiver type are equal to the fields of the other type,
+and `this.EqualityContract` equals `other.EqualityContract`.
 
 `object.Equals` performs the equivalent of
 
@@ -62,18 +58,24 @@ A record type contains two synthesized copying members if methods with the same
 signature are not already declared within the record type:
 
 * A protected constructor taking a single argument of the record type.
-* A public parameterless instance method called `Clone` which returns the record type.
+* A public parameter-less virtual instance "clone" method with a compiler-reserved name
 
 The protected constructor is referred to as the "copy constructor" and the synthesized
-body copies the values of all instance fields declared in the input type to the corresponding 
+body copies the values of all instance fields declared in the input type to the corresponding
 fields of `this`.
 
-The `Clone` method returns the result of a call to a constructor with the same signature as the
-copy constructor.
+The "clone" method returns the result of a call to a constructor with the same signature as the
+copy constructor. The return type of the clone method is the containing type, unless a virtual
+clone method is present in the base class. In that case, the return type is the current containing
+type if the "covariant returns" feature is supported and the override return type otherwise. The
+synthesized clone method is an override of the base type clone method if one exists. An error is
+produced if the base type clone method is sealed.
 
 ## Positional record members
 
-In addition to the above members, records with a parameter list ("positional records") have the following members:
+In addition to the above members, records with a parameter list ("positional records") synthesize
+the following members, if a concrete (i.e. non-abstract) member with the same signature (or name
+if the member is a field or property) is not already present:
 
 ### Primary Constructor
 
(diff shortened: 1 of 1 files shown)


proposals/records.md
```diff
@@ -1,75 +1,50 @@
...
 
 ### Equality members
 
-Record types produce synthesized implementations for the following methods:
+Record types produce synthesized implementations for the following methods, where `T` is the
+containing type:
 
-* `object.GetHashCode()` override, unless it is sealed or user provided
-* `object.Equals(object)` override, unless it is sealed or user provided
+* `object.GetHashCode()` override, unless it is sealed (error) or user provided
```
gafter, line 36: >it [](start = 42, length = 2)

Unless *what* is sealed?

proposals/records.md
```diff
@@ -1,75 +1,50 @@
...
+* `object.Equals(object)` override, unless it is sealed (error) or user provided
 * `T Equals(T)` method, where `T` is the current type
+* `Type EqualityContract` get-only property
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is
-true if and only if all the instance fields declared in the receiver type
-are equal to the fields of the other type.
+`EqualityContract` is a virtual instance property which returns `typeof(T)`. If it
+is present in the base type, the synthesized property overrides the base. If the base
```
gafter, line 42: >base [](start = 81, length = 4)

base method?

proposals/records.md
```diff
@@ -1,75 +1,50 @@
...
-1. initializes compiler-generated backing fields for the properties corresponding to the value parameters (if these properties are compiler-provided
-
-### Properties
-
-For each record parameter of a record type declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:
-
-For a record struct or a record class:
-
-* A public `get` and `init` auto-property is created (see separate `init` accessor specification). Its value is initialized during construction with the value of the corresponding primary constructor parameter. Each "matching" inherited abstract property's get accessor is overridden.
+In addition to the members declared in the record body, a record type has the following additional members:
```
gafter, line 29: >following additional members [](start = 78, length = 28)

Inherited members appear to be missing from this list.

proposals/records.md
```diff
@@ -83,14 +58,58 @@ A record type contains two synthesized copying members if methods with the same
 signature are not already declared within the record type:
```
gafter, line 58: >already [](start = 18, length = 7)

I don't understand what "already" means in this context.

proposals/records.md
```diff
@@ -83,14 +58,58 @@ A record type contains two synthesized copying members if methods with the same
...
 
 * A protected constructor taking a single argument of the record type.
-* A public parameterless instance method called `Clone` which returns the record type.
+* A public parameter-less virtual instance "clone" method with a compiler-reserved name
 
 The protected constructor is referred to as the "copy constructor" and the synthesized
-body copies the values of all instance fields declared in the input type to the corresponding 
+body copies the values of all instance fields declared in the input type to the corresponding
 fields of `this`.
```
gafter, line 65: >this`. [](start = 11, length = 6)

Just to be clear: I assume it is intentional that fields of the base are not copied. This will be extremely confusing to users.

proposals/records.md
```diff
@@ -114,13 +133,13 @@ A `with` expression allows for "non-destructive mutation", designed to
...
 
-A valid `with` expression has a receiver with a non-void type. The receiver type must contain an accessible
-parameterless instance method called `Clone` whose return type must be the receiver type, or a base type thereof.
+A valid `with` expression has a receiver with a non-void type. The receiver type must contain an
+accessible synthesized record "clone" method.
 
-On the right hand side of the `with` expression is an `member_initializer_list` with a
-sequence of assignments to *identifier*, which must an accessible instance field or property of the return
+On the right hand side of the `with` expression is an `member_initializer_list` with a sequence
+of assignments to *identifier*, which must an accessible instance field or property of the return
```
gafter, line 140: >must an [](start = 38, length = 7)

missing a word?

proposals/records.md
```diff
@@ -114,13 +133,13 @@ A `with` expression allows for "non-destructive mutation", designed to
...
 type of the `Clone()` method.
 
-Each `member_initializer` is processed the same way as an assignment to the field or property target on the return
-value of the `Clone()` method. The `Clone()` method is executed only once and the assignments are processed in
-lexical order.
+Each `member_initializer` is processed the same way as an assignment to the field or property
+target on the return value of the record clone method. The clone method is executed only once
```
gafter, line 144: I don't know what the juxtaposition of these two words means.

proposals/records.md
```diff
@@ -1,75 +1,50 @@
...
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is
-true if and only if all the instance fields declared in the receiver type
-are equal to the fields of the other type.
+`EqualityContract` is a virtual instance property which returns `typeof(T)`. If it
+is present in the base type, the synthesized property overrides the base. If the base
+is sealed or non-virtual, an error is produced.
+
+`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
+all the instance fields declared in the receiver type are equal to the fields of the other type,
```
gafter, line 46: >other type [](start = 85, length = 10)

Other type? I thought the types were the same.

gafter, line 46: >declared [](start = 24, length = 8)

To be clear, this does not appear to handle fields in the base. That will be extremely confusing to users.

proposals/records.md
```diff
@@ -1,75 +1,50 @@
...
-    : attributes? class_modifiers? 'partial'? 'class' identifier type_parameter_list?
-      parameter_list? type_parameter_constraints_clauses? class_body
-    ;
+The syntax for a record is as follows:
 
-struct_declaration
-    : attributes? struct_modifiers? 'partial'? 'struct' identifier type_parameter_list?
-      parameter_list? struct_interfaces? type_parameter_constraints_clauses? struct_body
+```antlr
+record_declaration
```
gafter, line 10: Nothing in this specification gives a hint as to whether the type resulting from this declaration is a reference type or value type.

proposals/records.md
agocke: Neither intentional nor unintentional -- I see two viable specifications here. One way is to try to set up a pattern of delegation, where if the base type is a record, we call `base.Equals`, and the other where we always override and check all types regardless.

LDM hasn't decided on anything so I haven't spent time detailing either one.

proposals/records.md
```diff
@@ -83,14 +58,58 @@ A record type contains two synthesized copying members if methods with the same
...
+type declaration. This is called the primary constructor for the type, and causes the implicitly
+declared default class constructor, if present, to be suppressed. It is an error to have a primary
+constructor and a constructor with the same signature already present in the class.
+
+At runtime the primary constructor
+
+1. executes the instance initializers appearing in the class-body; and then
+   invokes the base class constructor with the arguments provided in the `record_base` clause, if present
+
+1. assigns fields or properties declared in the record body with the value of a primary constructor
```
AlekseyTs, line 92: Shouldn't this be done before the base constructor is invoked, As done for any other field initializers?

proposals/records.md
```diff
@@ -83,14 +58,58 @@ A record type contains two synthesized copying members if methods with the same
...
+copy constructor. The return type of the clone method is the containing type, unless a virtual
+clone method is present in the base class. In that case, the return type is the current containing
+type if the "covariant returns" feature is supported and the override return type otherwise. The
+synthesized clone method is an override of the base type clone method if one exists. An error is
+produced if the base type clone method is sealed.
+
+## Positional record members
+
+In addition to the above members, records with a parameter list ("positional records") synthesize
+the following members, if a concrete (i.e. non-abstract) member with the same signature (or name
```
cston, line 77: >same signature [](start = 73, length = 14)

How do we handle inherited properties that are inaccessible, or `get` or `set` only.

And can the types, of inherited or explicit properties and record parameters, differ by `dynamic`/`object`, tuple element names, `n[u]int`/`System.[U]IntPtr`, or nullability?

proposals/records.md
```diff
@@ -83,14 +58,58 @@ A record type contains two synthesized copying members if methods with the same
...
+if the member is a field or property) is not already present:
```
cston, line 78: Are explicit fields/properties (inherited or defined on record type) allowed to differ by type, as long as the record parameter type is assignable to the field/property type?

proposals/records.md
```diff
@@ -1,75 +1,50 @@
...
+record_declaration
+    : attributes? class_modifier* 'partial'? 'record' identifier type_parameter_list?
+      parameter_list? record_base? type_parameter_constraints_clause* record_body
     ;
 
-class_body
-    : '{' class_member_declarations? '}'
-    | ';'
+record_base
+    : ':' class_type argument_list?
```
gafter, line 16: >argument_list [](start = 21, length = 13)

The spec doesn't explain the meaning of the `argument_list` for a record declaration without a `parameter_list`.

AlekseyTs: > The spec doesn't explain the meaning of the argument_list for a record declaration without a parameter_list.

@agocke I think we should make it clear that arguments are not allowed unless there is a parameter_list

proposals/records.md
agocke: Lemme know if this needs more specification but hopefully the "matching" rule resolves everything.

proposals/records.md
```diff
@@ -83,14 +58,58 @@ A record type contains two synthesized copying members if methods with the same
...
+
+A record type has a public constructor whose signature corresponds to the value parameters of the
+type declaration. This is called the primary constructor for the type, and causes the implicitly
+declared default class constructor, if present, to be suppressed. It is an error to have a primary
+constructor and a constructor with the same signature already present in the class.
+
+At runtime the primary constructor
+
+1. executes the instance initializers appearing in the class-body; and then
+   invokes the base class constructor with the arguments provided in the `record_base` clause, if present
```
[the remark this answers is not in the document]
agocke: Good question. I'm not sure if anything else needs to be spec'd but my read is: reference types must call a base constructor. If there are no arguments, no base constructor is called. Thus, an error should be produced.

proposals/records.md
```diff
@@ -1,75 +1,61 @@
...
-It is an error to declare a record type without both a parameter list and the `data` modifier.
+Record types are reference types, similar to a class declaration. It is an error for a record to provide
+a `record_base` `argument_list` if the `record_declaration` does not contain a `parameter_list`.
+a 
 
 ## Members of a record type
 
-In addition to the members declared in the class or struct body, a record type has the following additional members:
+In addition to the members declared in the record body, a record type has additional synthesized members.
+Members are synthesized unless a concrete (non-abstract) member with a "matching" signature is either
```
cston, line 34: "unless an accessible concrete ..."?

proposals/records.md
```diff
@@ -79,18 +64,60 @@ override Equals(object o) => Equals(o as T);
...
-signature are not already declared within the record type:
+A record type contains two synthesized copying members:
 
 * A protected constructor taking a single argument of the record type.
-* A public parameterless instance method called `Clone` which returns the record type.
+* A public parameterless virtual instance "clone" method with a compiler-reserved name
 
 The protected constructor is referred to as the "copy constructor" and the synthesized
-body copies the values of all instance fields declared in the input type to the corresponding 
+body copies the values of all accessible instance fields in the input type to the corresponding
```
AlekseyTs, line 73: >all accessible instance fields [](start = 26, length = 30)

Consider clarifying this, obviously all fields within a record are accessible.

proposals/records.md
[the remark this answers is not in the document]
AlekseyTs: >  I'm not sure if anything else needs to be spec'd but my read is: reference types must call a base constructor. If there are no arguments, no base constructor is called. Thus, an error should be produced.

I don't think there is anything special in this scenario by comparison to regular classes. A default constructor initializer is going to be synthesized. That can cause an error if, for example, base class doesn't have an accessible parameter-less constructor.    

---
In reply to: [433603013](https://github.com/dotnet/csharplang/pull/3527#discussion_r433603013) [](ancestors = 433603013)

proposals/records.md
AlekseyTs: > Shouldn't this be done before the base constructor is invoked, as done for any other field initializers? 

I believe LDM decided to mimic other initializers

---
In reply to: [433044003](https://github.com/dotnet/csharplang/pull/3527#discussion_r433044003) [](ancestors = 433044003)

proposals/records.md
```diff
@@ -79,18 +65,60 @@ override Equals(object o) => Equals(o as T);
...
+
+### Properties
+
+For each record parameter of a record type declaration there is a corresponding public property
+member whose name and type are taken from the value parameter declaration.
+
+For a record:
+
+* A public `get` and `init` auto-property is created (see separate `init` accessor specification).
+  Each "matching" inherited abstract accessor is overridden. The auto-property is initialized to
```
cston, line 110: Do we require the `abstract` property to match the expected signature? Could an accessor be missing, or non-public? Could the property be a type that is assignable from the record parameter type?

proposals/records.md
agocke: What about inherited private fields?

AlekseyTs: > What about inherited private fields?

Shouldn't base take care of this?

---
In reply to: [434061428](https://github.com/dotnet/csharplang/pull/3527#discussion_r434061428) [](ancestors = 434061428)

agocke: Not for the very first record in a hierarchy, inheriting from a user-written class.

AlekseyTs: > Not for the very first record in a hierarchy, inheriting from a user-written class.

That is why I think it would be good to clearly describe two distinct scenarios and what is the behavior of the constructor in each of them.

---
In reply to: [434063198](https://github.com/dotnet/csharplang/pull/3527#discussion_r434063198) [](ancestors = 434063198)

proposals/records.md
```diff
@@ -84,19 +94,29 @@ constructor and a constructor with the same signature already present in the cla
...
+1. executes the instance initializers appearing in the class-body
+
+1. invokes the base class constructor with the arguments provided in the `record_base` clause, if present
 
-1. initializes compiler-generated backing fields for the properties corresponding to the value parameters (if these properties are compiler-provided
 
 ### Properties
 
-For each record parameter of a record type declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration. If no concrete (i.e. non-abstract) property with a get accessor and with this name and type is explicitly declared or inherited, it is produced by the compiler as follows:
+For each record parameter of a record type declaration there is a corresponding public property
```
gafter, line 104: >public [](start = 80, length = 6)

synthesized?

proposals/records.md
```diff
@@ -84,19 +94,29 @@ constructor and a constructor with the same signature already present in the cla
...
+member whose name and type are taken from the value parameter declaration.
 
-For a record struct or a record class:
+For a record:
 
-* A public `get` and `init` auto-property is created (see separate `init` accessor specification). Its value is initialized during construction with the value of the corresponding primary constructor parameter. Each "matching" inherited abstract property's get accessor is overridden.
+* A public `get` and `init` auto-property is created (see separate `init` accessor specification).
+  Each "matching" inherited abstract accessor is overridden. The auto-property is initialized to
```
gafter, line 110: >overridden [](start = 49, length = 10)

But virtual (non-abstract) ones are not overridden?

proposals/records.md
```diff
@@ -1,54 +1,60 @@
...
-
-It is an error to declare a record type without the `data` modifier.
+Record types are reference types, similar to a class declaration. It is an error for a record to provide
+a `record_base` `argument_list` if the `record_declaration` does not contain a `parameter_list`.
 
 ## Members of a record type
 
-In addition to the members declared in the class or struct body, a record type has the following additional members:
+In addition to the members declared in the record body, a record type has additional synthesized members.
+Members are synthesized unless an accessible concrete (non-abstract) member with a "matching" signature is
```
gafter, line 33: >"matching" signature [](start = 83, length = 20)

The spec doesn't have a definition of signature for a property.

proposals/records.md
```diff
@@ -1,54 +1,60 @@
...
-It is an error to declare a record type without the `data` modifier.
+Record types are reference types, similar to a class declaration. It is an error for a record to provide
+a `record_base` `argument_list` if the `record_declaration` does not contain a `parameter_list`.
 
 ## Members of a record type
 
-In addition to the members declared in the class or struct body, a record type has the following additional members:
+In addition to the members declared in the record body, a record type has additional synthesized members.
+Members are synthesized unless an accessible concrete (non-abstract) member with a "matching" signature is 
+either inherited or declared in the record body. Two members are considered matching if they have the same
```
gafter, line 34: >inherited [](start = 7, length = 9)

An abstract member in a grandparent class is still inherited into the child class even if overridden (implemented) in the parent class due to transitivity of inheritance. You may want to be more precise about what conditions cause the synthesiis or its suppression.

proposals/records.md
```diff
@@ -84,19 +94,29 @@ constructor and a constructor with the same signature already present in the cla
...
+  Each "matching" inherited abstract accessor is overridden. The auto-property is initialized to
+  the value of the corresponding primary constructor parameter.
 
+### Deconstruct
+
+A positional record synthesizes a public void-returning method called Deconstruct with an out
+parameter declaration for each parameter of the primary constructor declaration. Each parameter
+of the Deconstruct method has the same type as the corresponding parameter of the primary
+constructor declaration. The body of the method assigns each parameter of the Deconstruct method
+to the value from an instance member access to a member of the same name.
```
RikkiGibson, line 119: What happens if there is a positional parameter with the same name as a property or field that is already present in the type? e.g.

```cs
record C(int X)
{
    public int X { get; init; }
}
```

It feels like we *shouldn't* generate a Deconstruct method in this case, for the same reason we don't automatically assign from the positional parameter to the explicitly declared property.

proposals/records.md
cston, line 33: Should an `abstract` property declared on the record type be allowed without error?
```C#
abstract data class C(object P)
{
    public abstract object P { get; }
}
```

proposals/records.md
HaloFour: Does that mean that patterns like `case Some(var value)` would not be possible?

RikkiGibson: Not sure I completely follow, but users are free to write whatever Deconstruct method is appropriate for their type. The compiler should refrain from generating the Deconstruct method if a method with the same signature exists already on the record type. Not sure what should happen if the Deconstruct() method is on the base type. I don't think that is specified here, so let's make sure we come to a conclusion on that question.

HaloFour: I'm sorry, I read that as not generating a `Deconstruct` for a positional record with a single parameter.  Nevermind.

agocke: I think we should generate a Deconstruct regardless. If the user doesn't want one, they can declare a private method with the same signature.

proposals/records.md
```diff
@@ -1,54 +1,60 @@
...
-are equal to the fields of the other type.
+If either `object.GetHashCode()` or `object.Equals(object)` are sealed, an error is produced.
+
+`EqualityContract` is a virtual instance property which returns `typeof(T)`. If the base type
+defines an `EqualityContract` it is overridden in the derived record. If the base `EqualityContract`
+is sealed or non-virtual, an error is produced.
+
+`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
+all accessible instance fields in the receiver are equal to the fields of the parameter
+and `this.EqualityContract` equals `other.EqualityContract`.
```
cston, line 57: The base class `Equals(T)` will ignore derived class fields Is that expected?

For instance, the following will print: `False, True`:
```C#
using System;

data class A(int X);

data class B(int X, int Y) : A(X);

class Program
{
    static void Main()
    {
        var b1 = new B(1, 2);
        var b2 = new B(1, 3);
        Console.WriteLine(b1.Equals(b2));
        Console.WriteLine(((A)b1).Equals(b2));
    }
}
```

agocke:
Merging what I have so far, we can continue iterating

proposals/records.md
theunrepentantgeek: There will be different `EqualityContract` values for `A` & `B` - so even if you are calling `A.Equals()`, it can detect that they are not the same type.

proposals/records.md
```diff
@@ -1,54 +1,60 @@
...
     ;
 ```
 
-The `attributes` non-terminal will also permit a new contextual attribute, `data`.
-
-A class (struct) declared with a parameter list or `data` modifier is called a record class (record struct), either of which is a record type.
-
-It is an error to declare a record type without the `data` modifier.
+Record types are reference types, similar to a class declaration. It is an error for a record to provide
+a `record_base` `argument_list` if the `record_declaration` does not contain a `parameter_list`.
```
andresmoschini, line 28: @agocke could you explain the rationale behind this restriction? (related discussion: https://github.com/dotnet/csharplang/issues/3795)

Thanks!

PR #3537, opened 2020-06-05
Update classes.md, added 'in' parameter_modifier -> grammar

aSchaumarow:
'in' parameter_modifier is added to grammar

### DIFF
--- a/spec/classes.md
+++ b/spec/classes.md
@@ -1809,7 +1809,8 @@ default_argument
     ;
 
 parameter_modifier
-    : 'ref'
+    : 'in'
+    | 'ref'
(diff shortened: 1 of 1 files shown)


svick:
[As explained here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/), this spec is for C# 6.0 and [the `in` modifier was only added in C# 7.2](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.2/readonly-ref), so this change is technically not correct. The spec should be updated and, as I understand it, that work is (slowly) being done, but this is not the right way to do it.

PR #3552, opened 2020-06-09
Update record equality specification

cston:


proposals/records.md
```diff
@@ -40,28 +40,35 @@ The synthesized members are as follows:
...
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
-all accessible instance fields in the receiver are equal to the fields of the parameter
-and `this.EqualityContract` equals `other.EqualityContract`.
+If the base implementations of any of the overridden members is sealed or non-virtual,
+or does not match the expected signature and accessibility, an error is produced.
 
-`object.Equals` performs the equivalent of
+`Equals(T other)` performs value equality of the instance fields in this type with the fields of `other`,
+combined with the value of `base.Equals(other)` if there is a base record.
```
AlekseyTs, line 61: Consider explicitly stating what it is combined with for the top-most record. Then, the next sentence might become unnecessary.

proposals/records.md
```diff
@@ -40,28 +40,35 @@ The synthesized members are as follows:
...
+The base record `Equals(T other)` includes `EqualityContract.Equals(other.EqualityContract)`.
 
+The overrides of `Equals(T other)` for the base methods, including `object.Equals(object other)`, perform the equivalent of:
 ```C#
-override Equals(object o) => Equals(o as T);
+public override bool Equals(object other) => Equals(other as T);
 ```
 
+`GetHashCode()` generates a hash code by combining the values of `GetHashCode()` of the instance fields in this type,
+combined with the value of `base.GetHashCode()` if there is a base record.
```
AlekseyTs, line 70: Similarly here, consider explicitly stating that for the top-most record it is combined with hash code of equality contract.

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
+```
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
-all accessible instance fields in the receiver are equal to the fields of the parameter
-and `this.EqualityContract` equals `other.EqualityContract`.
+If the base implementations of any of the overridden members is sealed or non-virtual,
+or does not match the expected signature and accessibility, an error is produced.
 
-`object.Equals` performs the equivalent of
+`Equals(T other)` performs value equality of the instance fields in this type with the fields of `other`.
```
gafter, line 61: >performs value equality [](start = 18, length = 23)

I don't know what "performs value equality" means.

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
+If there is a base record type, the field equality value is combined with `base.Equals(other)`.
+Otherwise the field equality value is combined with `EqualityContract.Equals(other.EqualityContract)`.
 
+The overrides of `Equals(T other)` for the base methods, including `object.Equals(object other)`, perform the equivalent of:
 ```C#
-override Equals(object o) => Equals(o as T);
+public override bool Equals(object other) => Equals(other as T);
 ```
 
+`GetHashCode()` generates a hash code by combining the values of `GetHashCode()` of the instance fields in this type.
```
gafter, line 70: >GetHashCode()` of the instance fields [](start = 66, length = 37)

So, if one of the fields is `null` a `NullReferenceException` is expected?

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
+derived record classes. The synthesized implementations return `typeof(T)`.
+```C#
+protected virtual Type EqualityContract { get; }
+```
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
-all accessible instance fields in the receiver are equal to the fields of the parameter
-and `this.EqualityContract` equals `other.EqualityContract`.
+If the base implementations of any of the overridden members is sealed or non-virtual,
+or does not match the expected signature and accessibility, an error is produced.
```
gafter, line 60: I suggest you reword this "It is an error if...".  Otherwise... is produced by what and when?

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
 
-`object.Equals` performs the equivalent of
+`Equals(T other)` performs value equality of the instance fields in this type with the fields of `other`.
+If there is a base record type, the field equality value is combined with `base.Equals(other)`.
```
gafter, line 62: >combined [](start = 60, length = 8)

combined how?

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
+Otherwise the field equality value is combined with `EqualityContract.Equals(other.EqualityContract)`.
```
gafter, line 63: >combined [](start = 38, length = 8)

Combined how?

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
 
+The overrides of `Equals(T other)` for the base methods, including `object.Equals(object other)`, perform the equivalent of:
 ```C#
-override Equals(object o) => Equals(o as T);
+public override bool Equals(object other) => Equals(other as T);
 ```
 
+`GetHashCode()` generates a hash code by combining the values of `GetHashCode()` of the instance fields in this type.
+If there is a base record type, the field hash code is combined with `base.GetHashCode()`.
```
gafter, line 71: >field hash code [](start = 36, length = 15)

"field hash code"?  Term not defined.

I recommend reorganizing as a bulleted list of things included in the hash code.

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
...
+```
+The record type implicitly implements `System.IEquatable<T>` where `T` is the containing type.
 
-* `object.GetHashCode()` override
-* `object.Equals(object)` override
-* `T Equals(T)` method, where `T` is the current type
-* `Type EqualityContract` get-only property
-
-If either `object.GetHashCode()` or `object.Equals(object)` are sealed, an error is produced.
+Derived record types also override the `Equals(T other)` method from each base record type.
```
gafter, line 50: >override the `Equals(T other)` method [](start = 26, length = 37)

What about implementing any other `System.IEquatable<U>` types?

proposals/records.md
```diff
@@ -40,28 +40,37 @@ The synthesized members are as follows:
 
 Record types produce synthesized implementations for the following methods, where `T` is the
 containing type:
+```C#
+public override int GetHashCode();
```
gafter, line 44: >GetHashCode [](start = 20, length = 11)

Are these methods intended to override the similar methods from `object`? If so, you need to say that, as there could be an intervening method hiding the one in `object`.

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
 
-* `object.GetHashCode()` override
-* `object.Equals(object)` override
-* `T Equals(T)` method, where `T` is the current type
-* `Type EqualityContract` get-only property
+Derived record types also override the `Equals(TBase other)` method from each base record type.
 
-If either `object.GetHashCode()` or `object.Equals(object)` are sealed, an error is produced.
+The record type implicitly implements `System.IEquatable<T>` where `T` is the containing type.
+Record types do not re-implement `System.IEquatable<TBase>` for any base type `TBase`.
```
gafter, line 54: >. [](start = 85, length = 1)

What if the user has the record type explicitly derive from `ISomething, System.IEquatable<ISomething>`? Is that implemented? Unless you say so, the answer would be no.

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
+```
 
-`T Equals(T)` is specified to perform value equality such that `Equals` is true if and only if
-all accessible instance fields in the receiver are equal to the fields of the parameter
-and `this.EqualityContract` equals `other.EqualityContract`.
+It is an error if the base implementations of any of the overridden members is sealed or non-virtual,
+or do not match the expected signature and accessibility.
 
-`object.Equals` performs the equivalent of
+`Equals(T other)` returns true if and only if each of the following terms are true:
```
gafter, line 65: >terms [](start = 68, length = 5)

- `other` is not null?

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
+- If there is a base record type, the value of `base.Equals(other)`; otherwise
+the value of `EqualityContract.Equals(other.EqualityContract)`.
 
+The overrides of `Equals(T other)` for the base methods, including `object.Equals(object other)`, perform the equivalent of:
 ```C#
-override Equals(object o) => Equals(o as T);
+public override bool Equals(object other) => Equals(other as T);
 ```
 
+`GetHashCode()` returns the sum of the following terms where each term may be multiplied by a constant:
```
gafter, line 76: >sum of the following terms where each term may be multiplied by a constant [](start = 28, length = 74)

I suspect this is too specific and technically prevents us from using another scheme. I recommend saying that the following values are input to an unspecified but deterministic function returning an `int`.

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
-Record types produce synthesized implementations for the following methods, where `T` is the
+Record types produce synthesized implementations of the following methods, where `T` is the
 containing type:
+```C#
+public override int GetHashCode();
+public override bool Equals(object other);
+public virtual bool Equals(T other);
+```
+`GetHashCode()` and `Equals(object other)` are overrides of the virtual methods in `System.Object`.
+Any methods on intermediate base classes that would hide those methods are ignored when overriding.
```
gafter, line 49: >are ignored when overriding [](start = 71, length = 27)

We should ensure we test this.

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
 ```C#
-override Equals(object o) => Equals(o as T);
+public override bool Equals(object other) => Equals(other as T);
 ```
 
+`GetHashCode()` returns the sum of the following terms where each term may be multiplied by a constant:
+- For each field declared in the record type, the value of
+`System.Collections.Generic.EqualityComparer<TN>.Default.GetHashCode(fieldN)` where `TN` is the field type, and
+- If there is a base record type, the value of `base.GetHashCode()`; otherwise
+the value of `EqualityContract.Equals(other.EqualityContract)`.
```
gafter, line 80: >Equals [](start = 31, length = 6)

`GetHashCode`?  It looks like some introductory text describing the behavior of `Equals` was lost.

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
+`System.Collections.Generic.EqualityComparer<System.Type>.Default.GetHashCode(EqualityContract)`.
```
gafter, line 81: >EqualityComparer [](start = 28, length = 16)

I think this is handled by the previous line. You shouldn't need to use an EqualityComparer to protect against null for this one.

proposals/records.md
gafter, line 80: >Equals(other.EqualityContract [](start = 31, length = 29)

`GetHashCode(`

proposals/records.md
```diff
@@ -38,30 +38,48 @@ The synthesized members are as follows:
...
+
```
jcouv, line 82: Should we call out that a user-defined `Equals` is used instead of a synthesized one, when one is present?

Issue #3194, opened 2020-02-12
[Proposal] Extension GetEnumerator


333fred:
# Extension GetEnumerator Recognition in `foreach`

* [x] Proposed
* [ ] Prototype: [Complete](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
* [ ] Implementation: [In Progress](https://github.com/dotnet/roslyn/BRANCH_NAME)
* [ ] Specification: [Not Started](pr/1)

Discussion in #600

## Summary
[summary]: #summary

Allow `foreach` loops to recognize an extension method `GetEnumerator` method that otherwise satisfies the `foreach` pattern, and loop over the expression when it would otherwise be an error.

## Motivation
[motivation]: #motivation

This will bring `foreach` inline with how other features in C# are implemented, including `async` and pattern-based deconstruction.

## Detailed design
[design]: #detailed-design

The spec change is relatively straightforward. We modify `The foreach statement` section to this text:

>The compile-time processing of a foreach statement first determines the ***collection type***, ***enumerator type*** and ***element type*** of the expression. This determination proceeds as follows:
>
>*  If the type `X` of *expression* is an array type then there is an implicit reference conversion from `X` to the `IEnumerable` interface (since `System.Array` implements this interface). The ***collection type*** is the `IEnumerable` interface, the ***enumerator type*** is the `IEnumerator` interface and the ***element type*** is the element type of the array type `X`.
>*  If the type `X` of *expression* is `dynamic` then there is an implicit conversion from *expression* to the `IEnumerable` interface ([Implicit dynamic conversions](conversions.md#implicit-dynamic-conversions)). The ***collection type*** is the `IEnumerable` interface and the ***enumerator type*** is the `IEnumerator` interface. If the `var` identifier is given as the *local_variable_type* then the ***element type*** is `dynamic`, otherwise it is `object`.
>*  Otherwise, determine whether the type `X` has an appropriate `GetEnumerator` method:
>    * Perform member lookup on the type `X` with identifier `GetEnumerator` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for an enumerable interface as described below. It is recommended that a warning be issued if member lookup produces anything except a method group or no match.
[...]

PR #3576, opened 2020-06-16
Add spec for extension GetEnumerator

333fred:
Added the spec for extension GetEnumerator. The only change from https://github.com/dotnet/csharplang/issues/3194 was the additional line about the refkind in overload resolution.

### DIFF
--- a/proposals/extension-getenumerator.md
+++ b/proposals/extension-getenumerator.md
@@ -43,6 +43,8 @@ The spec change is relatively straightforward. We modify `The foreach statement`
 >    * The ***collection type*** is `X`, the ***enumerator type*** is `E`, and the ***element type*** is the type of the `Current` property.
 >*  Otherwise, an error is produced and no further steps are taken.
 
+For `await foreach`, the rules are similarly modified. The only change that is required to that spec is removing the `Extension methods do not contribute.` line from the description, as the rest of that spec is based on the above rules with different names substituted for the pattern methods.
+
 ## Drawbacks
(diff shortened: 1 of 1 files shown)


YairHalberstadt:
Should there be a section about async foreach?
Also there is a difference between extension foreach and instance foreach in that the former accepts params and optional arguments and the latter doesn't. I can't see any indication of that in the spec.

333fred:
> Should there be a section about async foreach?

Probably, will add. 

> Also there is a difference between extension foreach and instance foreach in that the former accepts params and optional arguments and the latter doesn't. I can't see any indication of that in the spec.

It doesn't need to be explicit. Overload resolution with a single argument will find methods with default parameters.

YairHalberstadt:
I would expect it to be explicit for instance GetEnumerator then.

333fred:
> I would expect it to be explicit for instance GetEnumerator then.

@YairHalberstadt thanks to some sleuthing from @gafter, there's yet another spec violation for compat that we kept: https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Binder/ForEachLoopBinder.cs#L1006-L1016. It absolutely should be including those parameters, and it does for async enumerable.

333fred:
Note: the spec I'm referring to in the update is https://github.com/dotnet/roslyn/blob/master/docs/features/async-streams.md, not the one checked into csharplang. @jcouv has taken a work item to update the one in this repo.

PR #3581, opened 2020-06-17
Add constant interpolated strings

kevinsun-dev:
Add proposal spec for constant interpolated strings

### DIFF
--- a/proposals/constant_interpolated_strings.md
+++ b/proposals/constant_interpolated_strings.md
@@ -36,7 +36,7 @@ This proposal represents the next logical step for constant string generation, w
 ## Detailed design
 [design]: #detailed-design
 
-The following represent the updated specifications for constant expressions under this new proposal.
+The following represent the updated specifications for constant expressions under this new proposal. Current specifications from which this was directly based on can be found [here](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#constant-expressions).
 
 ### Constant Expressions
 
@@ -62,7 +62,7 @@ A constant expression must be the `null` literal or a value with one of  the fol
 *  The predefined `+`, `-`, `!`, and `~` unary operators.
 *  The predefined `+`, `-`, `*`, `/`, `%`, `<<`, `>>`, `&`, `|`, `^`, `&&`, `||`, `==`, `!=`, `<`, `>`, `<=`, and `>=` binary operators, provided each operand is of a type listed above.
 *  The `?:` conditional operator.
-*  *Interpolated strings `${}`, provided that all components are constant expressions of type `string` and all interpolated components lack format specifiers.*
+*  *Interpolated strings `${}`, provided that all components are constant expressions of type `string` and all interpolated components lack alignment and format specifiers.*
 
 The following conversions are permitted in constant expressions:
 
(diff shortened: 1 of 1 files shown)


proposals/constant_interpolated_strings.md
```diff
@@ -0,0 +1,121 @@
...
+*  Identity conversions
+*  Numeric conversions
+*  Enumeration conversions
+*  Constant expression conversions
+*  Implicit and explicit reference conversions, provided that the source of the conversions is a constant expression that evaluates to the null value.
+
+Other conversions including boxing, unboxing and implicit reference conversions of non-null values are not permitted in constant expressions. For example:
+```csharp
+class C {
+    const object i = 5;         // error: boxing conversion not permitted
```
jaredpar, line 78:
```suggestion
class C 
{
    const object i = 5;         // error: boxing conversion not permitted
```

Should keep formatting consistent for all samples in the doc

proposals/constant_interpolated_strings.md
```diff
@@ -0,0 +1,121 @@
...
+
+A *constant_expression* is an expression that can be fully evaluated at compile-time.
+
+```antlr
+constant_expression
+    : expression
+    ;
+```
+
+A constant expression must be the `null` literal or a value with one of  the following types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `object`, `string`, or any enumeration type. Only the following constructs are permitted in constant expressions:
```
jaredpar, line 51: Looks like this is coming mostly from the C# spec, should attribute that in the writing here. Basically talk about which section of which spec version you got this from.

proposals/constant_interpolated_strings.md
```diff
@@ -0,0 +1,121 @@
...
+*  References to `const` parameters or local variables
+*  Parenthesized sub-expressions, which are themselves constant expressions.
+*  Cast expressions, provided the target type is one of the types listed above.
+*  `checked` and `unchecked` expressions
+*  Default value expressions
+*  Nameof expressions
+*  The predefined `+`, `-`, `!`, and `~` unary operators.
+*  The predefined `+`, `-`, `*`, `/`, `%`, `<<`, `>>`, `&`, `|`, `^`, `&&`, `||`, `==`, `!=`, `<`, `>`, `<=`, and `>=` binary operators, provided each operand is of a type listed above.
+*  The `?:` conditional operator.
+*  *Interpolated strings `${}`, provided that all components are constant expressions of type `string` and all interpolated components lack format specifiers.*
```
jaredpar, line 65: Interpolated components have three parts: expression, alignment and format specifiers. The alignment portion isn't mentioned here. Is the intent to support it? My suggestion would be no for now unless we can find compelling reasons to do so.  

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

kevinsun-dev: Thanks, added further restrictions on alignment specifiers as well.

proposals/constant_interpolated_strings.md
kevinsun-dev: Added direct link to section of current spec.

PR #3583, opened 2020-06-18
Records: update "Equality members" spec section

cston:
Update spec section on "Equality members" based on LDM 2020-06-15 discussion and subsequent "working group" proposal.

### DIFF
--- a/proposals/records.md
+++ b/proposals/records.md
@@ -57,10 +57,10 @@ public virtual bool Equals(R? other);
 ```
 The synthesized `Equals(R?)` returns `true` if and only if each of the following are `true`:
 - `other` is not `null`, and
-- For each instance field in the record type that is not inherited, the value of
+- For each instance field `fieldN` in the record type that is not inherited, the value of
 `System.Collections.Generic.EqualityComparer<TN>.Default.Equals(fieldN, other.fieldN)` where `TN` is the field type, and
 - If there is a base record type, the value of `base.Equals(other)` (a non-virtual call to `public virtual bool Equals(Base? other)`); otherwise
-the value of `EqualityContract.Equals(other.EqualityContract)`.
+the value of `EqualityContract == other.EqualityContract`.
 
 If the record type is derived from a base record type `Base`, the record type includes a synthesized override of the strongly-typed `Equals(Base other)`.
 The synthesized override is `sealed`.
@@ -83,7 +83,7 @@ public override int GetHashCode();
 A warning is reported if one of `Equals(R?)` and `GetHashCode()` is explicitly declared but the other method is not explicit.
 
 The synthesized override of `GetHashCode()` returns an `int` result of a deterministic function combining the following values:
-- For each instance field in the record type that is not inherited, the value of
+- For each instance field `fieldN` in the record type that is not inherited, the value of
 `System.Collections.Generic.EqualityComparer<TN>.Default.GetHashCode(fieldN)` where `TN` is the field type, and
 - If there is a base record type, the value of `base.GetHashCode()`; otherwise
 the value of `System.Collections.Generic.EqualityComparer<System.Type>.Default.GetHashCode(EqualityContract)`.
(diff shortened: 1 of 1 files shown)


proposals/records.md
```diff
@@ -39,49 +39,119 @@ The synthesized members are as follows:
...
 ```
-
-It is an error if the base implementations of any of the overridden members is sealed or non-virtual,
-or do not match the expected signature and accessibility.
-
-`Equals(T other)` returns true if and only if each of the following terms are true:
+The synthesized `Equals(R?)` returns `true` if and only if each of the following are `true`:
 - `other` is not `null`, and
-- For each field declared in the record type, the value of
+- For each instance field in the record type that is not inherited, the value of
```
gafter, line 60: >instance field in [](start = 11, length = 17)

For each instance field `fieldN` in...

proposals/records.md
```diff
@@ -39,49 +39,119 @@ The synthesized members are as follows:
...
 ```C#
-public override bool Equals(object other) => Equals(other as T);
 ```
+A warning is reported if one of `Equals(R?)` and `GetHashCode()` is explicitly declared but the other method is not explicit.
 
-`GetHashCode()` returns the `int` result of a deterministic function taking the following values:
-- For each field declared in the record type, the value of
+The synthesized override of `GetHashCode()` returns an `int` result of a deterministic function combining the following values:
+- For each instance field in the record type that is not inherited, the value of
```
gafter, line 86: >For each instance field [](start = 2, length = 23)

For each instance field `fieldN` in...

proposals/records.md
```diff
@@ -39,49 +39,119 @@ The synthesized members are as follows:
...
 `System.Collections.Generic.EqualityComparer<TN>.Default.GetHashCode(fieldN)` where `TN` is the field type, and
 - If there is a base record type, the value of `base.GetHashCode()`; otherwise
 the value of `System.Collections.Generic.EqualityComparer<System.Type>.Default.GetHashCode(EqualityContract)`.
```
gafter, line 89: >value [](start = 4, length = 5)

Since you don't guard against a null `EqualityContract` in `Equals`, you needn't do so here. You should do so in both or neither.

proposals/records.md
```diff
@@ -39,49 +39,119 @@ The synthesized members are as follows:
...
+```C#
+class R1 : IEquatable<R1>
+{
+    public T1 P1 { get; set; }
+    protected virtual Type EqualityContract => typeof(R1);
+    public override bool Equals(object? obj) => Equals(obj as R1);
+    public virtual bool Equals(R1? other)
+    {
+        return !(other is null) &&
+            EqualityContract == other.EqualityContract &&
```
gafter, line 108: > == o [](start = 28, length = 5)

This is not `.Equals` as specified.

proposals/records.md
cston: Changed `Equals` to use `EqualityContract == other.EqualityContract`.

---
In reply to: [442963587](https://github.com/dotnet/csharplang/pull/3583#discussion_r442963587) [](ancestors = 442963587)

aredfox:
Please consider reopening this, to add an example on how users can override the behavior of Equals and GetHashCode?
Just to be clear, for users, I don't find it clear enough in the documentation how it is used. Can a clear example be documented how to override GetHashCode, and Equals? Or how to override the behavior?

As this is what "normal" C# devs would try to do, and would expect to do so in every class/struct, I feel documentation around this should be more precise on how to override the behavior, or if it simply is not possible.
image

PS, consider including this [example/gist ](https://gist.github.com/ericsampson/84672e210cc3b8c6e6db12af01aa25de) - it's a very clear way of communicating an example?

333fred:
This is a merged pull request, there's nothing to reopen. @BillWagner for the documentation ideas.

svick:
@333fred I believe that's already covered by https://github.com/dotnet/docs/issues/21031.

PR #3652, opened 2020-07-07
Disallow "Clone" members

jcouv:


proposals/records.md
```diff
@@ -38,6 +38,7 @@ Members are synthesized unless a member with a "matching" signature is declared
 an accessible concrete non-virtual member with a "matching" signature is inherited.
 Two members are considered matching if they have the same
 signature or would be considered "hiding" in an inheritance scenario.
+It is an error for a member of a record to be named "Clone".
```
AlekseyTs, line 41: Did we really decide to disallow all members named Clone, or should we allow methods named Clone as long as they have parameters?

HaloFour:
Out of curiosity is there a reason to disallow `Clone` methods?  Aren't records relying on an unspeakable name to clone the instance?  Is that changing or is this to prevent confusion or collision with a more general-purpose clone/wither solution intended for non-records?

jcouv:
@HaloFour We're reserving it as we may allow customization of the clone method, or implementation on non-record types in the future. Reserving that name keeps some options open. Notes: https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-07-01.md#Confirming-unspeakable-Clone-method-implications

proposals/records.md
jcouv: Notes: https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-07-01.md#Confirming-unspeakable-Clone-method-implications

We explicitly discussed methods and fields in LDM ("Clone" fields should be disallowed). For nested types, I chatted with Jared offline and I think we'll want the same behavior.

HaloFour:
ah ok, gotcha.  I assume that this means that a record cannot implement `ICloneable` unless they do so explicitly?

jcouv:
Correct. That's a good test scenario (I'll add to my PR).

proposals/records.md
333fred: I don't believe that particular variant of the question was raised in LDM. Personally, I'd prefer to disallow _all_ members, just in case we end up wanting to add some kind of parameter list to possibly support only cloning if an initer will actually change a field, but it's a followup we could raise with the LDM if we want to clarify.

PR #3667, opened 2020-07-12
Applying attributes to properties and fields that are created to back primary constructor parameters

AlekseyTs:


AlekseyTs:
@RikkiGibson, @MadsTorgersen, @cston, @jcouv Please review

PR #3670, opened 2020-07-12
Document ability to explicitly declare Deconstruct method.

AlekseyTs:


proposals/records.md
```diff
@@ -238,11 +238,13 @@ For a record:
...
 ### Deconstruct
 
-A positional record synthesizes a public void-returning method called Deconstruct with an out
+A positional record synthesizes a public void-returning instance method called Deconstruct with an out
 parameter declaration for each parameter of the primary constructor declaration. Each parameter
 of the Deconstruct method has the same type as the corresponding parameter of the primary
 constructor declaration. The body of the method assigns each parameter of the Deconstruct method
 to the value from an instance member access to a member of the same name.
+The method can be declared explicitly. It is an error if the explicit declaration does not match
+the expected signature or accessibility, or is static.
```
RikkiGibson, line 247: Do we still permit the user to write a Deconstruct method with a different number of parameters or different type of parameters?

333fred: Good question. I would expect to be able to declare my own `Deconstruct` that adds a nominal property, for example.

AlekseyTs: > Do we still permit the user to write a Deconstruct method with a different number of parameters or different type of parameters?

Yes.

AlekseyTs:
@MadsTorgersen, @cston, @jcouv Please review

proposals/records.md
333fred: What would be the error case for incorrect signature then? A conflicting accessibility/staticness would produce an error, but a different parameter type or number of parameters should produce no error, right?

AlekseyTs: > What would be the error case for incorrect signature then? 

Wrong error type, for example.

> A conflicting accessibility/staticness would produce an error, but a different parameter type or number of parameters should produce no error, right?

Correct.

333fred: Thanks for clarifying.

PR #3672, opened 2020-07-13
Disallow certain modifiers in record parameters

jcouv:
Here's the general section of the spec on method parameters, for context: https://github.com/dotnet/csharplang/blob/master/spec/classes.md#method-parameters (note it doesn't have `in` yet)

### DIFF
--- a/proposals/records.md
+++ b/proposals/records.md
@@ -27,7 +27,7 @@ record_body
 Record types are reference types, similar to a class declaration. It is an error for a record to provide
 a `record_base` `argument_list` if the `record_declaration` does not contain a `parameter_list`.
 
-Record parameters cannot use `ref`, `out` or `this` modifiers.
+Record parameters cannot use `ref`, `out` or `this` modifiers (but `in` and `params` are allowed).
 
 ## Inheritance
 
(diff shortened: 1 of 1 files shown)


proposals/records.md
```diff
@@ -27,6 +27,8 @@ record_body
 Record types are reference types, similar to a class declaration. It is an error for a record to provide
 a `record_base` `argument_list` if the `record_declaration` does not contain a `parameter_list`.
 
+Record parameters cannot use `ref`, `out` or `this` modifiers.
```
333fred, line 30: > (note it doesn't have in yet)

Does that mean we're allowing `in` here, or that we're not allowing `in` here and we'll need to update this when the spec updates? In either case, I think that should be explicitly clarified.

AlekseyTs, line 30: Should this be moved to the Primary Constructor section?

jcouv: `in` is allowed. If it were disallowed, we'd simply list it here as disallowed (no reason not to). I'd rather manage multiple copies of the list of allowed parameter modifiers as it continues to evolve.

jcouv: meant "avoid managing" (ie. the opposite of what I wrote ;-))

333fred: While I agree in principle, until we have an actual updated master list this just leaves it ambiguous. At some point we'll have to take this and turn it into an actual spec, and at that point we're going to have ask "well, was it intended to be allowed, or was it just not completely spec'ed and we'll do whatever the implementation does?

jcouv: Okay

PR #3674, opened 2020-07-13
Local functions and lambdas are outside init phase

jcouv:
Fixes https://github.com/dotnet/roslyn/issues/44704

### DIFF
--- a/proposals/init.md
+++ b/proposals/init.md
@@ -71,8 +71,8 @@ class Student
 }
 ```
 
-An instance property containing an `init` accessor is considered settable outside
-a local function or lambda in the following circumstances:
+An instance property containing an `init` accessor is considered settable in
+the following circumstances, except when in a local function or lambda:
 
 - During an object initializer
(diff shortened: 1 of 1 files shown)


proposals/init.md
```diff
@@ -71,8 +71,8 @@ class Student
 }
 ```
 
-An instance property containing an `init` accessor is considered settable in
-the following circumstances:
+An instance property containing an `init` accessor is considered settable outside
+a local function or lambda in the following circumstances:
```
333fred, line 75: The wording here is somewhat ambiguous: it doesn't explicitly forbid local functions or lambdas. I would instead phrase it as:

```suggestion
An instance property containing an `init` accessor is considered settable in
the following circumstances, except when in a local function or lambda:
```

As a side note, this might be a good argument for `init` methods, and allowing applying that to local functions :).

jcouv: I like it. Thanks

PR #3678, opened 2020-07-14
Clarify  accessibility requirements for explicitly declared record copy constructor.

AlekseyTs:


proposals/records.md
```diff
@@ -177,7 +177,7 @@ A record type contains two copying members:
 The purpose of the copy constructor is to copy the state from the parameter to the new instance being
 created. This constructor doesn't run any instance field/property initializers present in the record
 declaration. If the constructor is not explicitly declared, a protected constructor will be synthesized
-by the compiler.
+by the compiler. Explicitly declared copy constructor must be either public, or protected.
```
333fred, line 180:
```suggestion
by the compiler. An explicitly declared copy constructor must be either public or protected.
```

proposals/records.md
```diff
@@ -177,7 +177,8 @@ A record type contains two copying members:
 The purpose of the copy constructor is to copy the state from the parameter to the new instance being
 created. This constructor doesn't run any instance field/property initializers present in the record
 declaration. If the constructor is not explicitly declared, a protected constructor will be synthesized
-by the compiler. An explicitly declared copy constructor must be either public or protected.
+by the compiler. An explicitly declared copy constructor must be either public or protected, unless the
+record is sealed.
```
333fred, line 181: Is this intended to mean that a `sealed` record type can declare an `internal` copy constructor?

333fred: Also, why couldn't an `internal` record type declare an `internal` copy constructor?

AlekseyTs: > Is this intended to mean that a sealed record type can declare an internal copy constructor?

It is intended to mean that the restriction is only for non-sealed records.

> Also, why couldn't an internal record type declare an internal copy constructor?

Technically we could allow that, but right now I am going after simplest rule. A public constructor in internal record is effectively internal. So, important scenarios aren't really blocked.

333fred: > It is intended to mean that the restriction is only for non-sealed records.

I guess I don't understand the point of the rule then? Is it purely about inheritance, or about expected code behavior? For example, today we don't pick up `internal` instance `GetEnumerator` methods when binding `foreach`, in order to ensure that behavior is consistent across codebases. It feels like the behavior we have for copy constructors should be based on the same principles.

333fred: If this is a question we need to take for LDM next week, I'm happy to bring it.

AlekseyTs: > I guess I don't understand the point of the rule then?

"The first thing the constructor must do, is to call a copy constructor of the base, ..." Accessibility of the base copy constructor should allow that.

AlekseyTs: > It feels like the behavior we have for copy constructors should be based on the same principles.

I do not think rules for looking up of any member by name should apply here. We are not performing a lookup like that. The spec says the constructor is there, the spec says call it.

AlekseyTs: Also, I think there  is nothing wrong in hiding the constructor from consumers outside of inheritance hierarchy. In fact, the synthesized copy constructor is hidden, it is protected.

AlekseyTs:
@RikkiGibson, @cston, @jcouv, @MadsTorgersen Please review

AlekseyTs:

proposals/records.md
MadsTorgersen: A protected member in a sealed class produces a warning. I view Aleksey's exception for sealed classes as accounting for that; otherwise a user wouldn't be able to specify (without warning) a non-public copy constructor in a sealed class.

PR #3683, opened 2020-07-15
Add clarifications for instance "clone" method

AlekseyTs:


proposals/records.md
```diff
@@ -174,7 +174,7 @@ class R3 : R2, IEquatable<R3>
 A record type contains two copying members:
 
 * A constructor taking a single argument of the record type. It is referred to as a "copy constructor".
-* A synthesized public parameterless virtual instance "clone" method with a compiler-reserved name
+* A synthesized public parameterless instance "clone" method with a compiler-reserved name
```
RikkiGibson, line 177: Why remove 'virtual' here? Is the point to make the method virtual when the record is inheritable and make the method non-virtual when the record is sealed?

AlekseyTs: > Why remove 'virtual' here? Is the point to make the method virtual when the record is inheritable and make the method non-virtual when the record is sealed?

Yes, this is all stated later. Language doesn't allow declaring virtual methods in sealed classes. I think there is no reason to violate this for a synthesized method.

AlekseyTs:

PR #3696, opened 2020-07-16
Update unsafe-code.md

anurmatov:
It should be specified comparison operator `>=` (greater than or equal) instead of lambda `=>`


333fred:
Thanks @anurmatov!

PR #3700, opened 2020-07-16
Clarify expected accessibility of an EqualityContract property in a record directly inheriting from `object`.

AlekseyTs:


AlekseyTs:

proposals/records.md
```diff
@@ -48,9 +48,9 @@ The synthesized members are as follows:
 
 If the record is derived from `object`, the record type includes a synthesized readonly property equivalent to a property declared as follows:
 ```C#
-protected Type EqualityContract { get; };
+Type EqualityContract { get; };
 ```
-The property is `virtual` unless the record type is `sealed`.
+The property is `virtual` and `protected` unless the record type is `sealed`. Otherwise, the property is `private`.
```
333fred, line 53: Having unless and otherwise like this doesn't read particularly well. I'd suggest putting the exception first, like this:

```suggestion
The property is `private` if the record type is `sealed`. Otherwise, the property is `virtual` and `protected`.
```

proposals/records.md
```diff
@@ -48,9 +48,9 @@ The synthesized members are as follows:
 
 If the record is derived from `object`, the record type includes a synthesized readonly property equivalent to a property declared as follows:
 ```C#
-protected Type EqualityContract { get; };
+Type EqualityContract { get; };
 ```
-The property is `virtual` unless the record type is `sealed`.
+The property is `private` if the record type is `sealed`. Otherwise, the property is `virtual` and `protected`.
```
gafter, line 53: It is silly to have an `EqualityContract` property for a sealed record type that derives from `object`. It serves no purpose.

proposals/records.md
```diff
@@ -48,9 +48,9 @@ The synthesized members are as follows:
...
 The property can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`.
 
 If the record type is derived from a base record type `Base`, the record type includes a synthesized readonly property equivalent to a property declared as follows:
```
gafter, line 56: I don't think you intend that it be an auto-property. In particular, the specification needs to say what its implementation is.

AlekseyTs: > I don't think you intend that it be an auto-property. In particular, the specification needs to say what its implementation is.

There is a sentence below: "The synthesized property returns typeof(R) where R is the record type."

proposals/records.md
AlekseyTs: > It is silly to have an EqualityContract property for a sealed record type that derives from object. It serves no purpose.

There is a sentence below: "Can we omit EqualityContract if the record type is sealed and derives from System.Object?"

PR #3710, opened 2020-07-21
Only support public calling convention types.

333fred:


333fred:
Practically, this only means that the runtime cannot define an internal calling convention usable only from within corelib.

jkotas:
Just curious - is there a motivation for limiting this?

333fred:
> Just curious - is there a motivation for limiting this?

It simplifies implementation and we can't think of a motivating reason to have support.

AaronRobinsonMSFT:
> > Just curious - is there a motivation for limiting this?
> 
> It simplifies implementation and we can't think of a motivating reason to have support.

I would also imagine it simply puts effort on people to use `Reflection.Emit` or some other hackery to gain access to the type.

PR #3762, opened 2020-07-31
Add ToString and PrintMembers to records

jcouv:
@dotnet/roslyn-compiler for review. This is a proposal for shape of ToString synthesized members.

### DIFF
--- a/proposals/csharp-9.0/records.md
+++ b/proposals/csharp-9.0/records.md
@@ -201,27 +201,24 @@ If the "clone" method is not abstract, it returns the result of a call to a copy
 
 If the record is derived from `object`, the record includes a synthesized method equivalent to a method declared as follows:
 ```C#
-protected virtual void PrintMembers(System.StringBuilder builder, bool includeSeparator, bool includeName);
+protected virtual void PrintMembers(System.StringBuilder builder, bool includeSeparator);
 ```
 The method is `virtual` and `protected`.
 It is an error if the method is declared explicitly.
 
 The method:
-1. appends the record name to `builder` followed by " { ", if `includeName` is true
-2. appends a separator ", " to `builder` if `includeSeparator` is true and the record has public fields or properties,
-3. for each of the record's public field and property member, appends that member's name followed by " = " followed by the result of invoking `object.ToString()` on that member's value: `this.member.ToString()`, separated with ", ",
-4. appends " }" if `includeName` is true
+1. appends a separator ", " to `builder` if `includeSeparator` is true and the record has public fields or properties,
+2. for each of the record's public field and property member, appends that member's name followed by " = " followed by the result of invoking `object.ToString()` on that member's value: `this.member.ToString()`, separated with ", ",
 
 If the record type is derived from a base record `Base`, the record includes a synthesized override equivalent to a method declared as follows:
 ```C#
-protected override void PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName);
+protected override void PrintMembers(StringBuilder builder, bool includeSeparator);
 ```
 It is an error if the override is declared explicitly.
 
-The method appends the same contents to `builder` as described above, but calls `base.PrintMembers` between steps 3 and 4 with three arguments:
+The method appends the same contents to `builder` as described above, but also then calls `base.PrintMembers` with two arguments:
 1. its `builder` parameter,
-2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true,
-3. a false `includeName` argument.
+2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true.
 
 The record includes a synthesized method equivalent to a method declared as follows:
 ```C#
@@ -232,7 +229,9 @@ The method can be declared explicitly. It is an error if the explicit declaratio
 
 The synthesized method:
 1. creates a `StringBuilder` instance,
-2. invokes the record's "PrintMembers" method giving it the builder, `includeSeparator` as false, and `includeName` as true
+2. appends the record named to the builder, followed by " { ",
+3. invokes the record's "PrintMembers" method giving it the builder, 'includeSeparator` as false,
+4. appends " }",
 3. returns the builder's contents with `builder.ToString()`.
 
 For example, consider the following record types:
@@ -245,52 +244,40 @@ record R2(T1 P1, T2 P2, T3 P3) : R1(P1);
 For those record types, the synthesized printing members would be something like:
 
 ```C#
-class R1
+class R1 : IEquatable<R1>
 {
     public T1 P1 { get; init; }
     
-    protected virtual PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName)
+    protected virtual void PrintMembers(StringBuilder builder, bool includeSeparator)
     {
-        if (includeName)
-        {
-            builder.Append(nameof(R1));
-            builder.Append(" { ");
-        }
-        
         if (includeSeparator)
             builder.Append(", ");
             
         builder.Append(nameof(P1));
         builder.Append(" = ");
         builder.Append(this.P1.ToString());
-        
-        if (includeName)
-        {
-            builder.Append(" }");
-        }
     }
     
     public override string ToString()
     {
         var builder = new StringBuilder();
-        PrintMembers(builder, includeSeparator: false, includeName: true);
+        builder.Append(nameof(R1));
+        builder.Append(" { ");
+
+        PrintMembers(builder, includeSeparator: false);
+
+        builder.Append(" }");
         return builder.ToString();
     }
 }
 
-class R2 : R1
+class R2 : R1, IEquatable<R2>
 {
     public T2 P2 { get; init; }
     public T3 P3 { get; init; }
     
-    protected virtual PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName)
+    protected override void PrintMembers(StringBuilder builder, bool includeSeparator)
     {
-        if (includeName)
-        {
-            builder.Append(nameof(R1));
-            builder.Append(" { ");
-        }
-        
         if (includeSeparator)
             builder.Append(", ");
             
@@ -305,21 +292,22 @@ class R2 : R1
         builder.Append(this.P3.ToString());
         
         base.PrintMembers(builder, includeSeparator: true, includeName: false)
-        
-        if (includeName)
-        {
-            builder.Append(" }");
-        }
     }
     
     public override string ToString()
     {
         var builder = new StringBuilder();
-        PrintMembers(builder, includeSeparator: false, includeName: true);
+        builder.Append(nameof(R2));
+        builder.Append(" { ");
+
+        PrintMembers(builder, includeSeparator: false);
+
+        builder.Append(" }");
         return builder.ToString();
     }
 }
 ```
+
 
 ## Positional record members
 
--- a/proposals/csharp-9.0/records.md
+++ b/proposals/csharp-9.0/records.md
@@ -253,7 +253,7 @@ class R1 : IEquatable<R1>
             
         builder.Append(nameof(P1));
         builder.Append(" = ");
-        builder.Append(this.P1.ToString());
+        builder.Append(this.P1);
     }
     
     public override string ToString()
@@ -281,13 +281,13 @@ class R2 : R1, IEquatable<R2>
             
         builder.Append(nameof(P2));
         builder.Append(" = ");
-        builder.Append(this.P2.ToString());
+        builder.Append(this.P2);
         
         builder.Append(", ");
         
         builder.Append(nameof(P3));
         builder.Append(" = ");
-        builder.Append(this.P3.ToString());
+        builder.Append(this.P3);
         
         base.PrintMembers(builder, includeSeparator: true)
     }
--- a/proposals/csharp-9.0/records.md
+++ b/proposals/csharp-9.0/records.md
@@ -112,11 +112,11 @@ record R2(T1 P1, T2 P2) : R1(P1);
 record R2(T1 P1, T2 P2, T3 P3) : R2(P1, P2);
 ```
 
-For those record types, the synthesized members would be something like:
+For those record types, the synthesized equality members would be something like:
 ```C#
 class R1 : IEquatable<R1>
 {
-    public T1 P1 { get; set; }
+    public T1 P1 { get; init; }
     protected virtual Type EqualityContract => typeof(R1);
     public override bool Equals(object? obj) => Equals(obj as R1);
     public virtual bool Equals(R1? other)
@@ -134,7 +134,7 @@ class R1 : IEquatable<R1>
 
 class R2 : R1, IEquatable<R2>
 {
-    public T2 P2 { get; set; }
+    public T2 P2 { get; init; }
     protected override Type EqualityContract => typeof(R2);
     public override bool Equals(object? obj) => Equals(obj as R2);
     public sealed override bool Equals(R1? other) => Equals((object?)other);
@@ -152,7 +152,7 @@ class R2 : R1, IEquatable<R2>
 
 class R3 : R2, IEquatable<R3>
 {
-    public T3 P3 { get; set; }
+    public T3 P3 { get; init; }
     protected override Type EqualityContract => typeof(R3);
     public override bool Equals(object? obj) => Equals(obj as R3);
     public sealed override bool Equals(R2? other) => Equals((object?)other);
@@ -196,6 +196,130 @@ is the containing type and the method is virtual, unless the record is sealed or
 If the containing record is abstract, the synthesized clone method is also abstract.
 If the "clone" method is not abstract, it returns the result of a call to a copy constructor. 
 
+
+### Printing members: PrintMembers and ToString
+
+If the record is derived from `object`, the record includes a synthesized method equivalent to a method declared as follows:
+```C#
+protected virtual void PrintMembers(System.StringBuilder builder, bool includeSeparator, bool includeName);
+```
+The method is `virtual` and `protected`.
+It is an error if the method is declared explicitly.
+
+The method:
+1. appends the record name to `builder` followed by " { ", if `includeName` is true
+2. appends a separator ", " to `builder` if `includeSeparator` is true and the record has public fields or properties,
+3. for each of the record's public field and property member, appends that member's name followed by " = " followed by the result of invoking `object.ToString()` on that member's value: `this.member.ToString()`, separated with ", ",
+4. appends " }" if `includeName` is true
+
+If the record type is derived from a base record `Base`, the record includes a synthesized override equivalent to a method declared as follows:
+```C#
+protected override void PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName);
+```
+It is an error if the override is declared explicitly.
+
+The method appends the same contents to `builder` as described above, but calls `base.PrintMembers` between steps 3 and 4 with three arguments:
+1. its `builder` parameter,
+2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true,
+3. a false `includeName` argument.
+
+The record includes a synthesized method equivalent to a method declared as follows:
+```C#
+public override string ToString();
+```
+
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if either synthesized, or explicitly declared method doesn't override a property with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
+
+The synthesized method:
+1. creates a `StringBuilder` instance,
+2. invokes the record's "PrintMembers" method giving it the builder, `includeSeparator` as false, and `includeName` as true
+3. returns the builder's contents with `builder.ToString()`.
+
+For example, consider the following record types:
+
+``` csharp
+record R1(T1 P1);
+record R2(T1 P1, T2 P2, T3 P3) : R1(P1);
+```
+
+For those record types, the synthesized printing members would be something like:
+
+```C#
+class R1
+{
+    public T1 P1 { get; init; }
+    
+    protected virtual PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName)
+    {
+        if (includeName)
+        {
+            builder.Append(nameof(R1));
+            builder.Append(" { ");
+        }
+        
+        if (includeSeparator)
+            builder.Append(", ");
+            
+        builder.Append(nameof(P1));
+        builder.Append(" = ");
+        builder.Append(this.P1.ToString());
+        
+        if (includeName)
+        {
+            builder.Append(" }");
+        }
+    }
+    
+    public override string ToString()
+    {
+        var builder = new StringBuilder();
+        PrintMembers(builder, includeSeparator: false, includeName: true);
+        return builder.ToString();
+    }
+}
+
+class R2 : R1
+{
+    public T2 P2 { get; init; }
+    public T3 P3 { get; init; }
+    
+    protected virtual PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName)
+    {
+        if (includeName)
+        {
+            builder.Append(nameof(R1));
+            builder.Append(" { ");
+        }
+        
+        if (includeSeparator)
+            builder.Append(", ");
+            
+        builder.Append(nameof(P2));
+        builder.Append(" = ");
+        builder.Append(this.P2.ToString());
+        
+        builder.Append(", ");
+        
+        builder.Append(nameof(P3));
+        builder.Append(" = ");
+        builder.Append(this.P3.ToString());
+        
+        base.PrintMembers(builder, includeSeparator: true, includeName: false)
+        
+        if (includeName)
+        {
+            builder.Append(" }");
+        }
+    }
+    
+    public override string ToString()
+    {
+        var builder = new StringBuilder();
+        PrintMembers(builder, includeSeparator: false, includeName: true);
+        return builder.ToString();
+    }
+}
+```
 
 ## Positional record members
 
--- a/proposals/csharp-9.0/records.md
+++ b/proposals/csharp-9.0/records.md
@@ -53,19 +53,19 @@ If the record is derived from `object`, the record type includes a synthesized r
 Type EqualityContract { get; };
 ```
 The property is `private` if the record type is `sealed`. Otherwise, the property is `virtual` and `protected`.
-The property can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`.
+The property can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overriding it in a derived type and the record type is not `sealed`.
 
 If the record type is derived from a base record type `Base`, the record type includes a synthesized readonly property equivalent to a property declared as follows:
 ```C#
 protected override Type EqualityContract { get; };
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,130 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+        return builder.ToString();
+    }
+}
+
+class R2 : R1
+{
+    public T2 P2 { get; init; }
+    public T3 P3 { get; init; }
+    
+    protected virtual PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName)
```
cston, line 286: `override`

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,130 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+{
+    public T1 P1 { get; init; }
+    
+    protected virtual PrintMembers(StringBuilder builder, bool includeSeparator, bool includeName)
+    {
+        if (includeName)
+        {
+            builder.Append(nameof(R1));
+            builder.Append(" { ");
+        }
```
cston, line 258: Can `includeName` be handled directly in `ToString()` in each case?

jcouv: I had it that way in my first draft ;-) I'll change back.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+If the record is derived from `object`, the record includes a synthesized method equivalent to a method declared as follows:
+```C#
+protected virtual void PrintMembers(System.StringBuilder builder, bool includeSeparator);
+```
+The method is `virtual` and `protected`.
+It is an error if the method is declared explicitly.
+
+The method:
+1. appends a separator ", " to `builder` if `includeSeparator` is true and the record has public fields or properties,
+2. for each of the record's public field and property member, appends that member's name followed by " = " followed by the result of invoking `object.ToString()` on that member's value: `this.member.ToString()`, separated with ", ",
```
333fred, line 211:
```suggestion
2. for each of the record's public field and property members, appends that member's name followed by " = " followed by the result of invoking `object.ToString()` on that member's value: `this.member.ToString()`, separated with ", ",
```

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+```C#
+protected override void PrintMembers(StringBuilder builder, bool includeSeparator);
+```
+It is an error if the override is declared explicitly.
+
+The method appends the same contents to `builder` as described above, but also then calls `base.PrintMembers` with two arguments:
+1. its `builder` parameter,
+2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true.
+
+The record includes a synthesized method equivalent to a method declared as follows:
```
333fred, line 223: Consider specifying whether this is `Base` or all records, as it reads ambiguously currently.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+```C#
+public override string ToString();
+```
+
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if either synthesized, or explicitly declared method doesn't override a property with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
```
333fred, line 228:
```suggestion
The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if the method doesn't override a method with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
```

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+
+The synthesized method:
+1. creates a `StringBuilder` instance,
+2. appends the record named to the builder, followed by " { ",
```
333fred, line 232:
```suggestion
2. appends the record name to the builder, followed by " { ",
```

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+        builder.Append(" { ");
+
+        PrintMembers(builder, includeSeparator: false);
+
+        builder.Append(" }");
+        return builder.ToString();
+    }
+}
+```
+
```
333fred, line 310: Extra newline.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
 If the "clone" method is not abstract, it returns the result of a call to a copy constructor. 
 
 
+### Printing members: PrintMembers and ToString
+
+If the record is derived from `object`, the record includes a synthesized method equivalent to a method declared as follows:
+```C#
+protected virtual void PrintMembers(System.StringBuilder builder, bool includeSeparator);
```
RikkiGibson, line 204: should this method have a compiler-reserved (unspeakable) name?

proposals/csharp-9.0/records.md
RikkiGibson, line 211: If you include mutable properties in the ToString() representation, then don't you need to account for cycles in the object graph?

jcouv: Yes, the default `ToString` could cause problems in certain types. For types where that is the case, one should declare a custom `ToString` that adequately handles the situation.

proposals/csharp-9.0/records.md
jcouv: I think that would make sense. We don't expect anyone to call that method.

CyrusNajmabadi: `includeSeparator`?  taht seems really esoteric.  why do we need/want any sort of customization.  i dn't remember discussing that in the LDM at all for example.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+```
+The method is `virtual` and `protected`.
+It is an error if the method is declared explicitly.
+
+The method:
+1. appends a separator ", " to `builder` if `includeSeparator` is true and the record has public fields or properties,
```
CyrusNajmabadi, line 210: this feels unneeded.  why not just always include a trailing comma.  then you have the consistent, copyable:

```
Person
{
  Name = "Julien",
  Age = 99,
}
```

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+2. for each of the record's public field and property members, appends that member's name followed by " = " followed by the result of invoking `object.ToString()` on that member's value: `this.member.ToString()`, separated with ", ",
```
CyrusNajmabadi, line 211: is this more like objectdisplay?, or more like a jsut a sequence of tostring'ed members?  for example, if i have a string-member, will this be printed with quotes?

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+
+If the record type is derived from a base record `Base`, the record includes a synthesized override equivalent to a method declared as follows:
+```C#
+protected override void PrintMembers(StringBuilder builder, bool includeSeparator);
+```
+It is an error if the override is declared explicitly.
+
+The method appends the same contents to `builder` as described above, but also then calls `base.PrintMembers` with two arguments:
```
CyrusNajmabadi, line 219: shouldn't we have the base-members first?  that's how i would expect the default to be.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+1. its `builder` parameter,
+2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true.
```
CyrusNajmabadi, line 221: if we always append a comma, we never need to think about this.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+The record includes a synthesized method equivalent to a method declared as follows:
+```C#
+public override string ToString();
+```
+
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if the method doesn't override a method with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
+
+The synthesized method:
+1. creates a `StringBuilder` instance,
+2. appends the record name to the builder, followed by " { ",
```
CyrusNajmabadi, line 232: FQN, or just name?
What name is added if this is a generic record?  do we include type parameeters?

proposals/csharp-9.0/records.md
jnm2: The first warning sign that would remind me I needed to do this would be a stack overflow? Probably while debugging something else?

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+    public T1 P1 { get; init; }
+    
+    protected virtual void PrintMembers(StringBuilder builder, bool includeSeparator)
+    {
+        if (includeSeparator)
+            builder.Append(", ");
+            
+        builder.Append(nameof(P1));
+        builder.Append(" = ");
+        builder.Append(this.P1.ToString());
```
CyrusNajmabadi, line 258: P1 might be null, and thsi would throw.  we need to spec out the behavior for null-vals.

proposals/csharp-9.0/records.md
jcouv, line 211: :memo: will have to handle null values too. I think anonymous types print an empty value in that case.

jcouv: string members will not be printed with quotes.
See behavior of ToString on anonymous types ([sharplab](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQuAzAAQFUDCVA3mVazfvky2zwG4QIqADyoBeKpjgB3JlQCCYqgCIAFnAA26gPZKqAXwDc3Hq1yEAdHS2YoW9XHMB1BAEt4ACiHmAKloDKMK6YAObuAJRhRqQ8emR6QA))

CyrusNajmabadi:
https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQuAzAAQFUDCVA3mVazfvky2zwG4QIqADyoBeKpjgB3JlQCCYqgCJJMgEKyG4wlQC+SvQG5uPVrkIA6OgHtMUawBs4FgOoIAlvAAUQiwBVrAGUYD0wAcy8ASkjjUh5dMl0gA

THat's extremely ick.  Should we be taking a page from Go's book here: `%#v`

proposals/csharp-9.0/records.md
CyrusNajmabadi: THat seems like a really bad default (Esp. as we can't change this in teh future).  i get it for a debugger display, but the actual ToString of these types, it seems very insufficient and actually actively unhelpful for anyone who might use the ToString here.

jnm2: @CyrusNajmabadi But then you have to handle escaping the quotes or the you've only kicked the can down the road

CyrusNajmabadi: > @CyrusNajmabadi But then you have to handle escaping the quotes or then you've only kicked the can down the road

Yes :)  That's what 'go' does, what the debugger does, etc. etc.  taht seems like the sane approach to me if this is part of the public contract here.

proposals/csharp-9.0/records.md
RikkiGibson: IMO, it is not acceptable if a ToString that we generated causes a stack overflow. This would cause a stack overflow in the debugger on simply hovering on a record that indirectly refers to itself, for example.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
 If the "clone" method is not abstract, it returns the result of a call to a copy constructor. 
 
 
+### Printing members: PrintMembers and ToString
+
+If the record is derived from `object`, the record includes a synthesized method equivalent to a method declared as follows:
+```C#
+protected virtual void PrintMembers(System.StringBuilder builder, bool includeSeparator);
+```
+The method is `virtual` and `protected`.
```
jaredpar, line 206: Even if the record is sealed?

proposals/csharp-9.0/records.md
jaredpar: Not' sure if we doc this is in the spec or not but this method will need to remain forever in records for binary compat reasons. Once we ship it it's forever ;)

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+It is an error if the method is declared explicitly.
```
jaredpar, line 207: Another reason to make it an unspeakable name. It removes this entire scenario.

proposals/csharp-9.0/records.md
jaredpar, line 211: Why are we invoking `ToString` vs. letting the `StringBuilder.Append` method do it for us?

proposals/csharp-9.0/records.md
jaredpar: Why not just call `build.Append(this.P1)`? That takes the `null` problem out of the equation entirely because its' already naturally handled by `Append`

proposals/csharp-9.0/records.md
theunrepentantgeek: Having a **bool** parameter to specify that a separator is required seems questionable to me - I've seen an awful number of similar situations where using a **bool** effectively boxed in future directions, causing issues for maintenance. They're also obtuse to readers - I still have to look up what `.ConfigureAwait(false)` means even though I've written it hundreds of times.

I'd suggest either 

(a) Follow Steve McConnell's advice that a boolean parameter actually means you've got two methods disguised as one - and therefore split them out
(b) Replace the boolean with an enum (e.g. `PrintMembers(builder, Separator.Comma)`) to both make the call site readable and to allow for other options in the future.

proposals/csharp-9.0/records.md
jnm2: Avoiding boxing, maybe?

jaredpar: That is sensible and we could do this for value types. For reference types though seems better to feed them through.

proposals/csharp-9.0/records.md
jcouv: just name, without type parameters.

proposals/csharp-9.0/records.md
jcouv: @RikkiGibson What's a scenario for stack overflow, where the overflow would not also occur on one of the user properties (ie. also triggered when debugging and inspecting objects)?

proposals/csharp-9.0/records.md
jcouv: @CyrusNajmabadi Aside from anonymous types, we also have the precedent of tuples which do not quote or escape ([sharplab](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxAgrgOwD4AEBMAjALABQuAzAAQFUDCVA3mVazfvky2z7oQHR0A9pihCANnH4B1BAEt4ACkUAiABZxx4oSqRUALPgCURgNzdWAXzKWgA===)).
I'll stick with that precedent.

jcouv: Changed to rely on `StringBuilder.Append`

proposals/csharp-9.0/records.md
jcouv: Updated to use an unspeakable name

proposals/csharp-9.0/records.md
jcouv: Updated. The "print" method is private if the record derives from object and is sealed.

proposals/csharp-9.0/records.md
jcouv: Updated to use unspeakable "print" method.

proposals/csharp-9.0/records.md
jcouv: Changed to avoid calling ToString explicitly on value

proposals/csharp-9.0/records.md
jcouv: Okay, here's an example: `record List { public List Next { get; set; } public T Element { get; set; } }` ;-)

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,115 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+    public T1 P1 { get; init; }
+    
+    protected virtual void PrintMembers(StringBuilder builder, bool includeSeparator)
+    {
+        if (includeSeparator)
+            builder.Append(", ");
+            
+        builder.Append(nameof(P1));
+        builder.Append(" = ");
+        builder.Append(this.P1);
```
RikkiGibson, line 256: Is "pretty-printing" of members which are themselves record types a goal? If so, would it be required to pass something like an indent depth down to child properties which contain a PrintMembers method so they can print themselves "correctly"?

jcouv: We're bound to the `string ToString()` for purpose of this exercise, so I'd say that pretty-printing it's just not part of the ecosystem, even though I like it. We're not going to invent that infrastructure just because of records.

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,115 @@ If the containing record is abstract, the synthesized clone method is also abstr
 If the "clone" method is not abstract, it returns the result of a call to a copy constructor. 
 
 
+### Printing members: print and ToString methods
+
+If the record is derived from `object`, the record includes a synthesized "print" method equivalent to a method declared as follows, but with a compiler-reserved name:
+```C#
+void print(System.StringBuilder builder, bool includeSeparator);
+```
+The property is `private` if the record type is `sealed`. Otherwise, the property is `virtual` and `protected`.
```
svick, line 206:
```suggestion
The method is `private` if the record type is `sealed`. Otherwise, the method is `virtual` and `protected`.
```

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,115 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+```
+
+The method appends the same contents to `builder` as described above, but also then calls the base "print" method with two arguments:
+1. its `builder` parameter,
+2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true.
+
+The record includes a synthesized method equivalent to a method declared as follows:
+```C#
+public override string ToString();
+```
```
svick, line 224: Would it make sense to skip this for `abstract record`s? Or should this method be kept even for `abstract record`s, to prepare for the future where classes can inherit from records?

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,115 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if the method doesn't override a method with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
+
+The synthesized method:
+1. creates a `StringBuilder` instance,
```
svick, line 229: How strictly should this be interpreted? For example, could the compiler decide to implement this using a `StringBuilder` pool?

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,115 @@ If the containing record is abstract, the synthesized clone method is also abstr
 If the "clone" method is not abstract, it returns the result of a call to a copy constructor. 
 
 
+### Printing members: print and ToString methods
+
+If the record is derived from `object`, the record includes a synthesized "print" method equivalent to a method declared as follows, but with a compiler-reserved name:
```
cston, line 202: >"print" method equivalent to a method declared as follows, but with a compiler-reserved name [](start = 74, length = 92)

The spec should use the actual name since it is part of the contract, even if the name cannot be used from C#.

proposals/csharp-9.0/records.md
AlekseyTs, line 206: >property [](start = 4, length = 8)

method vs. property

AlekseyTs, line 206: >property [](start = 73, length = 8)

Same comment

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,115 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+The method appends the same contents to `builder` as described above, but also then calls the base "print" method with two arguments:
+1. its `builder` parameter,
+2. a true `includeSeparator` argument if either (a) the record has public fields or properties, or (b) its `includeSeparator` parameter was given as true.
+
+The record includes a synthesized method equivalent to a method declared as follows:
+```C#
+public override string ToString();
+```
+
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if the method doesn't override a method with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
```
AlekseyTs, line 226: >It is an error if the method doesn't override a method with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.). [](start = 251, length = 183)

I think this requirement should be stronger, i.e. similar to what we have for GetHashCode, for example. Specifically, something like this: "It is an error if either synthesized, or explicitly declared method doesn't override `object.ToString()` (for example, due to shadowing in intermediate base types, etc.)."

proposals/csharp-9.0/records.md
jcouv: We didn't do that for the "clone" method. I'm following that model.

proposals/csharp-9.0/records.md
jnm2: @jcouv Not sure I understood the significance of user properties, but doesn't stack overflow take down the process either way when you I open the locals window or expand an object and any of the displayed values is a self-referential record?

proposals/csharp-9.0/records.md
```diff
@@ -201,34 +201,38 @@ If the "clone" method is not abstract, it returns the result of a call to a copy
...
+3. for each of the record's printable members, appends that member's name followed by " = " followed by the member's value: `this.member`, separated with ", ",
+4. return true.
 
 The record includes a synthesized method equivalent to a method declared as follows:
 ```C#
 public override string ToString();
 ```
 
-The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if the method doesn't override a method with this signature in the record `Base` (for example, if the method is missing in the `Base`, or sealed, or not virtual, etc.).
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if either synthesized, or explicitly declared method doesn't override `object.ToString()` (for example, due to shadowing in intermediate base types, etc.).
```
333fred, line 230:
```suggestion
The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility, or if the explicit declaration doesn't allow overiding it in a derived type and the record type is not `sealed`. It is an error if either synthesized or explicitly declared method doesn't override `object.ToString()` (for example, due to shadowing in intermediate base types, etc.).
```

proposals/csharp-9.0/records.md
```diff
@@ -197,6 +197,118 @@ If the containing record is abstract, the synthesized clone method is also abstr
...
+    protected override void print(StringBuilder builder)
+    {
+        if (base.print(builder))
+            builder.Append(", ");
+            
+        builder.Append(nameof(P2));
+        builder.Append(" = ");
+        builder.Append(this.P2);
+        
+        builder.Append(", ");
```
tannergooding, line 289: Given that this will use the default `ToString` for each member, which typically uses the default culture, should this also use things like [TextInfo.ListSeparator](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.textinfo?view=netcore-3.1) for a culture aware format?

Speaking of globalization, is passing in a culture something that was discussed? The typical variants (as provided on "primitive like types in .NET") are:
```
public override string ToString()
public string ToString(string? format)
public string ToString(IFormatProvider? provider)
public string ToString(string? format, IFormatProvider? provider)
```

jcouv:
FYI, I'm going to revert to a speakable print method. Unspeakable print methods would throw a big wrench in our ability to derive records from classes. That seems unnecessary. I'll update the spec this afternoon.

Tagging @dotnet/roslyn-compiler

333fred:
> FYI, I'm going to revert to a speakable print method. Unspeakable print methods would throw a big wrench in our ability to derive records from classes. That seems unnecessary. I'll update the spec this afternoon.

That's a very good point, thanks Julien.

proposals/csharp-9.0/records.md
jcouv: @tannergooding The idea is to stick closely to `ToString` for anonymous types and tuples. They use comma as separators.
Even though the result of `ToString` is not C# (we're not quoting/escaping strings for instance), it is designed to look like C#. So I think commas make more sense than culture-dependent separators in this scenario.

tannergooding: 👍, just wanted to raise it as previous reviews by the libraries team have raised concerns over mixing culture sensitive numbers with `,` as it may lead to confusing semantics when read.

proposals/csharp-9.0/records.md
jcouv: @jnm2 I incorrectly thought that the overflow must have pre-existed the `ToString` on the record type (thus could not be blamed on the record's `ToString`). The `List` example above shows that's not the case.

The overflow problem exists with anonymous types and tuples also, but requires the help of a [wrapper type](https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4FgBQABAzAAgwJhwGF0BvdHSvffCq8tKpnANwgRwHcEIAHXxDgC8OAHZxOOAOo9+iABQBKANx1mrdjgjCxEnKRwA5OAA8YO7nwEcAvqsbqus6wDpjZnRHtrmGAIwuRAD2olBBADZwLjIAlvDyECo+NugpaOjY1NLOiGQ+mUHAAFZwAMbm7jDeDpQFLIgIMQAmcHh+AAw4ACpBAMowjaIA5ko+DI54AOxGpjAA/C49/YMjSTU4aTZAA=) to create loop.

We'll discuss more when LDM is back in session. If we restricted the printable members to only include public readonly/init-only fields and properties, we'd make this overflow problem much harder to hit.

proposals/csharp-9.0/records.md
jcouv: @svick If we decide to use a pool, we'll update the spec :-)

jcouv:
I'll merge this spec change shortly.  The unresolved follow-up questions were added to the [records test plan](https://github.com/dotnet/roslyn/issues/40726). Thanks for the review

- [ ] `ToString`: should we always print a trailing comma `R { Property1 = 42, Property2 = 43, }`?
- [ ] `ToString`: can we deal with stack overflow potential?
- [ ] `ToString`: should we try to quote/escape values, for example `R { IntProperty = 42, StringProperty = "hello" }`?
- [ ] `ToString`: should we omit `ToString` on abstract records?

PR #3823, opened 2020-08-25
Clarify the rules on UnmanagedCallersOnly

333fred:


333fred:
@elinor-fung @jkotas @jaredpar for review. I've updated the rules for UnmanagedCallersOnly following discussion on the implementation of https://github.com/dotnet/roslyn/pull/47008.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -358,9 +358,12 @@ the attribute:
 
 * It is an error to directly call a method annotated with this attribute from C#. Users must obtain a function pointer to
 the method and then invoke that pointer.
-* It is an error to apply the attribute to anything other than a static method. The C# compiler will mark any non-static
-methods imported from metadata with this attribute as unsupported by the language.
-* It is an error to have managed types as parameters or the return type of a method marked with the attribute.
+* It is an error to apply the attribute to anything other than an ordinary static method or ordinary static local function.
```
333fred, line 361: Previous wording technically permitted the attribute to be applied to operators. This forbids that.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -358,9 +358,12 @@ the attribute:
...
+The C# compiler will mark any non-static methods imported from metadata with this attribute as unsupported by the language.
+* It is an error to have managed types as parameters or the return type of a method marked with the attribute. This includes
+types with type parameters, even if those type parameters are constrained to `unmanaged` or substituted with `unmanaged` types.
```
333fred, line 364: The type parameter adjustment is to prevent things like `int?` being parameters to one of these methods.

jkotas, line 364: Just want to check: Is `Vector128<int>` is allowed? (I assume that the answer is yes.)

333fred: No, not by these rules. My understanding was that even something like this should be an error:
```cs
struct S2<T> where T : unmanaged { T Field; }
[UnmanagedCallersOnly]
void M(S2<int> param);
```

jkotas: There is nothing fundamental that prevents this from working and it works in .NET 5 today. I do not think we should be disallowing it in the compiler.

We have done work to enable blittable generic types in interop signatures in .NET 5: https://github.com/dotnet/runtime/pull/103

333fred:
@jkotas addressed feedback.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -358,9 +358,14 @@ the attribute:
 
 * It is an error to directly call a method annotated with this attribute from C#. Users must obtain a function pointer to
 the method and then invoke that pointer.
-* It is an error to apply the attribute to anything other than a static method. The C# compiler will mark any non-static
-methods imported from metadata with this attribute as unsupported by the language.
-* It is an error to have managed types as parameters or the return type of a method marked with the attribute.
+* It is an error to apply the attribute to anything other than an ordinary static method or ordinary static local function.
+The C# compiler will mark any non-static methods imported from metadata with this attribute as unsupported by the language.
```
jaredpar, line 362: Why just `non-static` methods? Why wouldn't this also apply to operators?

proposals/csharp-9.0/function-pointers.md
```diff
@@ -358,9 +358,14 @@ the attribute:
...
+* It is an error to have managed types as parameters or the return type of a method marked with the attribute. If the parameter
```
jkotas, line 363: I believe that the set of types allowed here should be same as the set of types that it is allowed to take address of in C#.

In other words, the set of types allowed for parameters and return types here should be exactly same as the set of the type for which the following compiles:
```
unsafe static void foo(TYPE x)
{
     void *p = &x;
}
```

Does the description match this?

333fred: That's the intent. I think I can simplify it a bit by just specifying `unmanaged_type`, but that's the goal of the proposal.

AaronRobinsonMSFT:
Lgtm

proposals/csharp-9.0/function-pointers.md
```diff
@@ -358,9 +361,13 @@ the attribute:
 
 * It is an error to directly call a method annotated with this attribute from C#. Users must obtain a function pointer to
 the method and then invoke that pointer.
-* It is an error to apply the attribute to anything other than a static method. The C# compiler will mark any non-static
-methods imported from metadata with this attribute as unsupported by the language.
-* It is an error to have managed types as parameters or the return type of a method marked with the attribute.
+* It is an error to apply the attribute to anything other than an ordinary static method or ordinary static local function.
```
tannergooding, line 364: Just wondering, but why are operators/properties restricted here? Is it just a language limitation, similar to binding against a signature that takes `explicit this` for instance methods?

333fred: Correct, this is a language limitation.

PR #3858, opened 2020-09-07
Update the example with the "hat" operator in csharp-8.0/ranges.md

ivanruski:
`array[^1]` is actually translated to `array[new Index(1, fromEnd: true).GetOffset(array.Length)]`

### DIFF
--- a/proposals/csharp-8.0/ranges.md
+++ b/proposals/csharp-8.0/ranges.md
@@ -89,7 +89,7 @@ Examples:
 ```csharp
 var array = new int[] { 1, 2, 3, 4, 5 };
 var thirdItem = array[2];    // array[2]
-var lastItem = array[^1];    // array[new Index(1, fromEnd: true)]
+var lastItem = array[^1];    // array[new Index(1, fromEnd: true).GetOffset(array.Length)]
 ```
 
 #### System.Range


proposals/csharp-8.0/ranges.md
```diff
@@ -89,7 +89,7 @@ Examples:
 ```csharp
 var array = new int[] { 1, 2, 3, 4, 5 };
 var thirdItem = array[2];    // array[2]
-var lastItem = array[^1];    // array[new Index(1, fromEnd: true)]
+var lastItem = array[^1];    // array[new Index(1, fromEnd: true).GetOffset(array.Length)]
```
jcouv, line 92: Looking at what the compiler actually produces for this scenario, I see that we emit `int num2 = array2[array2.Length - 1];`.
I don't know the history: is what should be document or is that a compiler optimization?

ivanruski: Yes, my mistake. I was testing this with an older version of the compiler.
If I am not mistaken `new Index(1, fromEnd: true).GetOffset(array.Length)]` was the behaviour of the hat operator when it was first implemented and [this commit](https://github.com/dotnet/roslyn/commit/5a24d130db3e7b624290799fe2552197eaf22280) changes the behaviour from `GetOffset` to `array.Length - 1`.

My initial motivation was just to keep the file ranges.md up to date with the actulal implementation.
Should I update the file once again or there is no need in doing that ?

PR #3872, opened 2020-09-10
Add issue templates

333fred:
Adds an issue template for full proposals, and adds a link to creating a new discussion with a description for when each should be chosen.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Open a Discussion
+    url: https://github.com/dotnet/csharplang/discussions/new
+    about: Please open new proposals as discussions if they are not fully specified.
--- a/.github/ISSUE_TEMPLATE/proposal_template.md
+++ b/.github/ISSUE_TEMPLATE/proposal_template.md
@@ -1,0 +1,48 @@
+---
+name: Language Proposal
+about: Fully specified language proposal
+title: "[Proposal]: [FEATURE_NAME]"
+---
+<!--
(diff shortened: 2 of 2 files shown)


MadsTorgersen (approved):

YairHalberstadt (commented):
Do we want everyone who opens an issue to fill in all of these sections? Is the assumption that if you can't fill in all of the sections it should be a discussion instead?

YairHalberstadt:
Very few people have the knowledge to write a suitable spec so this is going to push the vast majority of new issues towards discussions. Is that desired?

333fred:
> Do we want everyone who opens an issue to fill in all of these sections? Is the assumption that if you can't fill in all of the sections it should be a discussion instead?

Yep

> Very few people have the knowledge to write a suitable spec so this is going to push the vast majority of new issues towards discussions. Is that desired?

Yes, that is what we are experimenting with. If we find that it ends up being a bad experience we can backtrack here, but the goal is that open issues reflect things that the LDM would actually consider, making our process more visible to the community. Issues are a bad fit for the types of open-ended, lengthy discusssions that need to happen to generate a fully specified proposal.

stakx:
Looking at the template, I am a little surprised that the checkbox for _Specification_ is only the last one. Shouldn't the specification typically precede implementation?

yaakov-h:
Seemingly not for C# 7.0, 7.1, 7.2, 7.3, 8.0 and 9.0.

stakx:
Uhm, now I'm not so sure whether my post above was clear enough. I wasn't concerned so much with consistency of the template with existing C# 7, 8, and 9 issues, but with _process_: shouldn't specs be written (started, at least) before the implementation? Obviously, the order of some checkboxes isn't going to enforce a sequential, process, but they _can_ suggest what the desired process looks like.

333fred:
We check in speclets in the /proposal folder (you can find the ones for C# 8 and 9 in there, the 7 ones are less complete unfortunately), but that checkbox is referrring to actually integrating into the ECMA spec and being checked into the /spec folder.

stakx:
I see. Makes perfect sense then to do that last, since updates to ECMA-334/5 are probably a slow & lengthy process. Thanks for helping me understand.

Issue #3884, opened 2020-09-12
[Proposal]: Allow init-only auto-properties in readonly structs


HaloFour:
# Allow init-only auto-properties in readonly structs

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

## Summary
[summary]: #summary

I propose that C# relax the restriction around the types of auto-properties supported on readonly structs to allow init-only auto-properties.  Said properties are effectively immutable, only permitting mutation during initialization.  They are also backed by a readonly-field.

## Motivation
[motivation]: #motivation

Supporting init-only auto-properties in readonly structs would make it easier to create said types via initializer.  This would also enable "with" support for readonly structs in C# 10.0.

## Detailed design
[design]: #detailed-design

The compiler will relax the imposed restrictions on what types of auto-properties are allowed to be declared in a readonly struct to support auto-properties with an `init` accessor:

```cs
public readonly struct Point
{
    public int X { get; init; }
    public int Y { get; init; }
}

var point = new Point { X = 25, Y = 30 };
var anotherPoint = point with { Y = 35 };
```

## Drawbacks
[drawbacks]: #drawbacks

TBD

## Alternatives
[alternatives]: #alternatives

TBD

## Unresolved questions
[unresolved]: #unresolved-questions

TBD

## Design meetings

N/A

333fred:
https://github.com/dotnet/roslyn/pull/47652 :wink:

333fred:
Note: I didn't open the issue on csharplang directly because I don't believe it actually requires changing the init spec.

HaloFour:
Is it worth leaving this proposal open to ensure that the spec includes verbiage that `init` is allowed here?

333fred:
Probably

pkulikov:
@HaloFour @333fred 
Now, C# compiler allows to use `init` with **not** auto-properties as well. The following compiles:

```csharp
public readonly struct A
{
    private readonly int x;
    public A(int n) => x = n;
    public int X { get => x; init => x = value; }
}
```

However, note the preceding example assigns to the **field** in the constructor. Assignment to the property doesn't compile:
```csharp
public readonly struct A
{
    private readonly int x;
    public A(int n) => X = n;
    public int X { get => x; init => x = value; }
}
```

The error: `error CS0188: The 'this' object cannot be used before all of its fields have been assigned`

The variant with auto-property compiles:
```csharp
public readonly struct A
[...]

PR #3915, opened 2020-09-18
Add `init` accessors in `readonly struct`s and `readonly` properties

333fred:
Fixes https://github.com/dotnet/csharplang/issues/3884.

### DIFF
--- a/proposals/csharp-9.0/init.md
+++ b/proposals/csharp-9.0/init.md
@@ -247,7 +247,7 @@ also applies to interface implementation.
 `init` accessors (both auto-implemented accessors and manually-implemented
 accessors) are permitted on properties of `readonly struct`s, as well as
 `readonly` properties. `init` accessors are not permitted to be marked
-`readonly` themselves.
+`readonly` themselves, in both `readonly` and non-`readonly` `struct` types.
 
 ```cs
 readonly struct ReadonlyStruct1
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/init.md
```diff
@@ -242,6 +242,27 @@ Restrictions of this feature:
 - All overrides of a property must have `init` if the base had `init`. This rule
 also applies to interface implementation.
 
+### Readonly structs
+
+`init` accessors (both auto-implemented accessors and manually-implemented
```
pkulikov, line 247: As for manually-implemented accessors, please see https://github.com/dotnet/csharplang/issues/3884#issuecomment-714987528

proposals/csharp-9.0/init.md
```diff
@@ -242,6 +242,27 @@ Restrictions of this feature:
...
+accessors) are permitted on properties of `readonly struct`s, as well as
+`readonly` properties. `init` accessors are not permitted to be marked
+`readonly` themselves.
```
jaredpar, line 250: Is the last statement true for all `struct` types or only non-readonly `struct`? Think we should clarify that point.

jaredpar (commented):
Minor feedback

333fred:
@jaredpar addressed feedback.

jaredpar (approved):

PR #3934, opened 2020-09-22
We should Inform people about appropriate use of analyzers

AartBluestoke:
as per https://github.com/dotnet/csharplang/discussions/3928

The initial suggestion was to point people to examples and a tutorial, but feedback suggested a direct link to this documentation would be more appropriate.

( the reason my initial suggestion didn't link that documentation was that it has a poor seo rank) 
searches that do not find the official documentation:
-"writing roslyn analyzers","c# analyzers" "write a c# analyzer", "c# analyzer", "write roslyn analyzer", "write visual studio code analyzer", "create visual studio code analyzer"

searches that find it:
- "roslyn analyzers"


AartBluestoke:
original links were :
-- Microsoft outlines how to do so at https://devblogs.microsoft.com/dotnet/how-to-write-a-roslyn-analyzer/
-- some sample Analyzers can be found at https://github.com/dotnet/roslyn/blob/master/docs/analyzers/Analyzer%20Samples.md
-- Microsoft's recommended FxCop analyzers can be found at https://github.com/dotnet/roslyn-analyzers

README.md
```diff
@@ -30,6 +30,8 @@ Discussions that are short and stay on topic are much more likely to be read. If
 - Stick to the topic of the discussion. If a comment is tangential, or goes into detail on a subtopic, start a new discussion and link back.
 - Is your comment useful for others to read, or can it be adequately expressed with an emoji reaction to an existing comment?
 
+Language modifications which prevent specific syntax from occurring can probably be achieved with [an analyzer](https://docs.microsoft.com/en-us/visualstudio/extensibility/getting-started-with-roslyn-analyzers). Such changes are likely to be rejected by the language design committee to prevent increased language complexity.
```
333fred, line 33:
```suggestion
Language proposals which prevent specific syntax from occurring can be achieved with [a Roslyn analyzer](https://docs.microsoft.com/en-us/visualstudio/extensibility/getting-started-with-roslyn-analyzers). Proposals that only attempt to make existing syntax optionally illegal will be rejected by the language design committee to prevent increased language complexity.
```

333fred (changes requested):
Some suggestions to make the language a bit more firm here. We want to actively discourage this type of request, so I think being a little more firm on the wording would be good.

YairHalberstadt:
Shouldn't this be in the issue/discussion template rather than the Readme?

333fred:
> Shouldn't this be in the issue/discussion template rather than the Readme?

It can be in both.

333fred (approved):
LGTM. @MadsTorgersen, can you also take a look?

scottdorman (approved):
I think this is good.

MadsTorgersen (approved):
Great addition.

333fred:
Thanks @AartBluestoke!

Issue #1147, opened 2017-11-25
allow references to structs to be stored in fields of ref structs


lucasmeijer:
Unity is a game engine that uses C# extensively.  

We're in need of a way to have a struct that contains one or more pointers to other structs that we want to be able to read/write from/to.  We do this with unsafe code today:

    struct Color { float r,g,b; }
    
    unsafe struct Group
    {
      Color* _color;
      ref Color color => ref *_color;
    }

But we would love to do this without unsafe code, and would like to suggest a feature where it's allowed to store references to structs in fields of ref structs, that can guarantee that these pointers don't escape onto the heap:

    struct Color { float r,g,b; }

    ref struct Group
    {
        ref Color color;
    }

Joe4evr:
I feel like you're trying too hard to use `ref struct` everywhere. Always keep in mind:
![](https://i.imgur.com/Vo20UsT.png)

jnm2:
@Joe4evr I don't think that comment is necessarily apropos.

VSadov:
ref fields are not directly expressible in IL.

You can very close to the desired behavior if there is a public type like: https://github.com/dotnet/corert/blob/796aeaa64ec09da3e05683111c864b529bcc17e8/src/System.Private.CoreLib/src/System/ByReference.cs 

The bigger problem (and one of the reasons the type is not public) is tracking of life times. 
It is assumed right now that that methods that take a `ref` and return a `ref-like` cannot return that same ref wrapped in a ref-like. Allowing/assuming such wrapping would make use of `ref` and `ref-likes` very difficult. 

Such wrapping could be ok though, if wrapped ref is guaranteed to refer to a heap location. There are ideas how this could be allowed via some specialized syntax like:

```C#
ByReference<int> r = ref inArray[0];  // ok since array element is definitely on the heap
```

Just wondering, if the feature comes with "only on the heap" restriction, would it be acceptable limitation?

joeante:
>You can very close to the desired behavior if there is a public type like: 

That is possible but we intend to use this for simplifying our API's significantly for our upcoming new component system. We want the syntax to be straightforward, and typing .Value is quite annoying for what is a very common pattern in that Component system.

For our use case:
* The referenced data is allocated from C++ via malloc
* Group in the above example is a ref struct, thus can't be written to any classes or other persistent data

For our particular use case, we can live with any restriction on what you can assign to it. Internally we are patching a pointer to the data in by writing to memory at an offset relative to the adress of the Group struct.

[...]

PR #3936, opened 2020-09-23
Proposal: Low level struct improvements

jaredpar:
This design covers adding the following features to C# for low level
struct performance improvements:

- Allowing `ref struct` to contain `ref` fields.
- Allowing `struct` and `ref struct` to return `ref` to their fields.
- Allowing safe fixed buffers for unmanaged and managed types.

Note: the section on providing parameter escape notations is incomplete
at this time. I will send a PR in the future once I finalize this
portion. Feel free to still leave comments though.

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -749,7 +749,7 @@ Misc Notes:
 `System.Runtime.CompilerServices` namespace.
 
 ### Safe fixed size buffers
-The language will relax the restrictions on fixed sized arrays such that the 
+The language will relax the restrictions on fixed sized arrays such that they 
 can be declared in safe code and the element type can be managed or unmanaged. 
 This will make types like the following legal:
 
@@ -879,7 +879,7 @@ struct RS
 
         // Okay: The parameter here is not a part of the calculated "must match"
         // set because it can't be returned hence this is legal.
-        p.Set(span);
+        rs2.Set(span);
 
         // Error: the *safe-to-escape* scope of 'rs2' is the top scope of this
         // method
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -1,0 +1,972 @@
+Low Level Struct Improvements
+=====
+
+## Summary
+This proposal is an aggregation of several different proposals for `struct` 
+performance improvements. The goal being a design which takes into account the
+various proposals to create a single overarching feature set for `struct` 
+improvements.
+
+## Motivation
+Over the last few releases C# has added a number of low level performance 
+features to the language: `ref` returns, `ref struct`, function pointers, 
+etc. ... These enabled .NET developers to create write highly performant code
+while continuing to leverage the C# language rules for type and memory safety.
+It also allowed the creation of fundamental performance types in the .NET 
+libraries like `Span<T>`.
+
+As these features have gained traction in the .NET ecosystem developers, both
+internal and external, have been providing us with information on remaining 
+friction points in the ecosystem. Places where they still need to drop to 
+`unsafe` code to get their work, or require the runtime to special case types
+like `Span<T>`. 
+
+This proposal aims to address many of these concerns by building on top of our
+existing low level features. Specifically it aims to:
+
+- Allow `ref struct` types to declare `ref` fields.
+- Allow the runtime to fully define `Span<T>` using the C# type system and 
+remove special case type like `ByReference<T>`
+- Allow `struct` types to return `ref` to their fields.
+- Allow the declaration of safe `fixed` buffers for managed and unmanaged types
+in `struct`
+
+## Detailed Design 
+The rules for `ref struct` safety are defined in the 
+[span-safety document](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md). 
+This document will describe the required changes to this document as a result of
+this proposal. Once accepted as an approved feature these changes will be
+incorporated into that document.
+
+### Provide ref fields
+The language will allow developers to declare `ref` fields inside of a 
+`ref struct`. This can be useful for example when encapsulating large 
+mutable `struct` instances or defining high performance types like `Span<T>`
+in libraries besides the runtime.
+
+Today `ref` fields accomplished in the runtime by using the `ByReference<T>` 
+type which the runtime treats effective as a `ref` field. This means though 
+that only the runtime repository can take full advantage of `ref` field like 
+behavior and all uses of it require manual verification of safety. Part of the 
+[motivation for this work](https://github.com/dotnet/runtime/issues/32060) is
+to remove `ByReference<T>` and use proper `ref` fields in all code bases. 
+The challenging part about allowing `ref` fields declarations though comes in
+defining rules such that `Span<T>` can be defined using `ref` fields without 
+breaking compatibility with existing code. 
+
+Before diving into the problems here it should be noted that `ref` fields only
+require a small number of targeted changes to our existing span safety rules. In
+some cases it's not even to support new features but to rationalize our existing
+`Span<T>` usage of `ref` data. This section of the proposal is quite involved
+though because I feel it's important to communicate the "why" of these changes
+in as much detail as possible and providing supporting samples. This is to both 
+ensure the changes are sound as well as giving future developers a better 
+understanding of the choices made here.
+
+To understand the challenges here let's first consider how `Span<T>` will look
+once `ref` fields are supported.
+
+```cs
+// This is the eventual definition of Span<T> once we add ref fields into the
+// language
+readonly ref struct Span<T>
+{
+    ref readonly T _field;
+    int _length;
+
+    // This constructor does not exist today however will be added as a
+    // part of changing Span<T> to have ref fields. It is a convenient, and
+    // safe, way to create a length one span over a stack value that today 
+    // requires unsafe code.
+    public Span(ref T value)
+    {
+        ref _field = ref value;
+        _length = 1;
+    }
+}
+```
+
+The constructor defined here presents a problem because its return values must
+necessarily have restricted lifetimes for many inputs. Consider that if a 
+local parameter is passed by `ref` into this constructor that the returned 
+`Span<T>` must have a *safe-to-escape* scope of the local's declaration scope.
+
+```cs
+Span<int> CreatingAndReturningSpan()
+{
+    int i = 42;
+
+    // This must be an error in the new design because it stores stack 
+    // state in the Span. 
+    return new Span<int>(ref i);
+
+    // This must be legal in the new design because it is legal today (it 
+    // cannot store stack state)
+    return new Span<int>(new int[] { });
+}
+```
+
+At the same time it is legal to have methods today which take a `ref` parameter 
+and return a `Span<T>`.  These methods bear a lot of similarity to the newly 
+added `Span<T>` constructor: take a `ref`, return a `Span<T>`. However the
+lifetime of the return value of these methods is never restricted by the inputs.
+The existing span safety rules consider such values as effectively always 
+*safe-to-escape* outside the enclosing method.
+
+```cs
+class ExistingScenarios
+{
+    Span<T> CreateSpan<T>(ref T p)
+    {
+        // The implementation of this method is irrelevant. From the point of
+        // the consumer the returned value is always safe to return.
+        ... 
+    }
+
+    Span<T> Examples<T>(ref T p, T[] array)
+    {
+        // Legal today
+        return CreateSpan(ref p); 
+
+        // Legal today, must remain legal
+        T local = default;
+        return CreateSpan(ref local);
+
+        // Legal for any possible value that could be used as an argument
+        return CreateSpan(...);
+    }
+}
+```
+
+The reason that all of the above samples are legal is because in the 
+existing design there is no way for the return `Span<T>` to store a reference 
+to the input state of the method call. This is because the span safety rules 
+explicitly depend on `Span<T>` not having a constructor which takes a `ref` 
+parameter and stores it as a field. 
+
+```cs
+class ExistingAssumptions
+{
+    Span<T> CreateSpan<T>(ref T p)
+    {
+        // !!! Cannot happen today !!!
+        // The existing span safety rules specifically call out that this method
+        // cannot exist hence they can assume all returns from CreateSpan are
+        // safe to return.
+        return new Span<T>(ref p);
+    }
+}
+```
+
+The rules we define for `ref` fields must ensure the `Span<T>` constructor 
+properly restricts the *safe-to-escape* scope of constructed objects in the 
+cases it captures `ref` state. At the same time it must ensure that we don't 
+break the existing consumption rules for methods like `CreateSpan<T>`. 
+
+```cs
+class GoalOfRefFields
+{
+    Span<T> CreateSpan<T>(ref T p)
+    {
+        // ERROR: the existing consumption rules for CreateSpan believe this 
+        // can never happen hence we must continue to enforce that it cannot
+        return new Span<T>(ref p);
+
+        // Okay: this is legal today
+        return new Span<int>(new int[] { });
+    }
+
+    Span<int> ConsumptionCompatibility()
+    {
+        // Okay: this is legal today and must remain legal.
+        int local = 42;
+        return CreateSpan(ref local);
+
+        // Okay: the arguments don't actually matter here. Literally any value 
+        // could be passed to this method and the return of it would still be 
+        // *safe-to-escape* outside the enclosing method. 
+        return CreateSpan(...);
+    }
+}
+```
+
+This tension between allowing constructors such as `Span<T>(ref T field)` and 
+ensuring compatibility with `ref struct` returning methods like `CreateSpan<T>` 
+is a key pivot point in the design of `ref` fields.
+
+To do this we will change the escape rules for a constructor invocation, which
+today are the same as method invocation, on a `ref struct` that **directly** 
+contains a `ref` field as follows:
+- If the constructor contains any `ref struct`, `ref`, `out` or `in` parameters 
+then the *safe-to-escape* of the return will be the current scope
+- Else the *safe-to-escape* will be the outside the method scope
+
+Lets examine these rules in the context of samples to better understand their
+impact.
+
+```cs
+ref struct RS
+{
+    ref int _field;
+
+    public RS(int[] array, int index)
+    {
+        ref _field = ref array[index];
+    }
+
+    public RS(ref int i)
+    {
+        ref _field = ref i;
+    }
+
+    static RS CreateRS(ref int i)
+    {
+        // The implementation of this method is irrelevant to the safety rule
+        // examples below. The returned value is always *safe-to-escape* outside
+        // the enclosing method scope
+    }
+
+    static RS RuleExamples(ref int i, int[] array)
+    {
+        var rs1 = new RS(ref i);
+
+        // ERROR by bullet 1: the safe-to-escape scope of 'rs1' is the current
+        // scope.
+        return rs1; 
+
+        var rs2 = new RS(array, 0);
+
+        // Okay by bullet 2: the safe-to-escape scope of 'rs2' is outside the 
+        // method scope.
+        return rs2; 
+
+        int local = 42;
+
+        // ERROR by bullet 1: the safe-to-escape scope is the current scope
+        return new RS(ref local);
+        return new RS(ref i);
+
+        // Okay because rules for method calls have not changed. This is legal
+        // today hence it must be legal in the presence of ref fields.
+        return CreateRS(ref local);
+        return CreateRS(ref i);
+    }
+}
+```
+
+It is important to note that for the purposes of the rule above any use of 
+constructor chaining via `this` is considered a constructor invocation. The
+result of the chained constructor call is considered to be returning to the 
+original constructor hence *safe-to-escape* rules come into play. That is 
+important in avoiding unsafe examples like the following:
+
+```cs
+ref struct RS1
+{
+    ref int _field;
+    public RS1(ref int p)
+    {
+        ref _field = ref p;
+    }
+}
+
+ref struct RS2
+{
+    RS1 _field;
+    public RS2(RS1 p)
+    {
+        // Okay
+        _field = p;
+    }
+
+    public RS2(ref int i)
+    {
+        // ERROR: The *safe-to-escape* scope of the constructor here is the 
+        // current method scope while the *safe-to-escape* scope of `this` is
+        // outside the current method scope hence this assignment is illegal
+        _field = new RS1(ref i);
+    }
+
+    public RS2(ref int i)  
+        // ERROR: the *safe-to-escape* return of :this the current method scope
+        // but the 'this' parameter has a *safe-to-escape* outside the current
+        // method scope
+        : this(new RS1(ref i))
+    {
+
+    }
+}
+```
+
+The limiting of the constructor rules to just `ref struct` that directly contain
+ `ref` field is another important compatibility concern. Consider that the 
+majority of `ref struct` defined today indirectly contain `Span<T>` references. 
+That mean by extension they will indirectly contain `ref` fields once `Span<T>`
+adopts `ref` fields. Hence it's important to ensure the *safe-to-return* rules
+of constructors on these types do not change. That is why the restrictions
+must only apply to types that directly contain a `ref` field.
+
+Example of where this comes into play.
+
+```cs
+ref struct Container
+{
+    LargeStruct _largeStruct;
+    Span<int> _span;
+
+    public Container(in LargeStruct largeStruct, Span<int> span)
+    {
+        _largeStruct = largeStruct;
+        _span = span;
+    }
+}
+```
+
+Much like the `CreateSpan<T>` example before the *safe-to-escape* return of the 
+`Container` constructor is not impacted by the `largeStruct` parameter. If the
+new constructor rules were applied to this type then it would break 
+compatibility with existing code. The existing rules are also sufficient for 
+existing constructors to prevent them from simulating `ref` fields by storing
+them into `Span<T>` fields.
+
+```cs
+ref struct RS4
+{
+    Span<int> _span;
+
+    public RS4(Span<int> span)
+    {
+        // Legal today and the rules for this constructor invocation 
+        // remain unchanged
+        _span = span;
+    }
+
+    public RS4(ref int i)
+    {
+        // ERROR. Bullet 1 of the new constructor rules gives this newly created
+        // Span<T> a *safe-to-escape* of the current scope. The 'this' parameter
+        // though has a *safe-to-escape* outside the current method. Hence this
+        // is illegal by assignment rules because it's assigning a smaller scope
+        // to a larger one.
+        _span = new Span(ref i);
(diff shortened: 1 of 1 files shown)


VSadov:
I like this!!!  👍

JamesNK:
I think I have a situation where this would be useful. Would it solve this highlighted problem? - https://github.com/protocolbuffers/protobuf/blob/a690227398d307419d45b1f352494a176873897b/csharp/src/Google.Protobuf/ParseContext.cs#L77-L80

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+Members which contain the `[RefEscapes]` attribute cannot be used to implement 
+interface members. This would hide the lifetime nature of the member at 
+the `interface` call site and would lead to incorrect lifetime calculations.
+
+To account for this change the "Parameters" section of the span safety document
+will be updated to include the following:
+
+- If the parameter is the `this` parameter of a `struct` type, it is 
+*ref-safe-to-escape* to the top scope of the enclosing method unless the 
+method is annotated with `[RefEscapes]` in which case it is *ref-safe-to-escape*
```
VSadov, line 743: I think the ref-escape of `[RefEscapes]` members will have to match val-escape of the container. 
If we are permitting returning refs to fields, we must ensure the containing variable is not dead by the time the ref is used.

I.E accessing ref-returning members on such structs would work as if accessing a field.

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+type is `ref T`. The indexer will be annotated with the `[RefEscapes]` attribute
+as the implementation will be returning fields of the declaring type. The 
+accessibility of the member will match the accessibility on the `fixed` field.
+
+For example, the signature of the indexer for `CharBuffer.Data` will be the 
+following:
+
+```cs
+[RefEscapes]
+internal ref char <>DataIndexer(int index) => ...;
```
VSadov, line 793: I would suggest also adding ordinary byval getter and setter and select which one to use based on the same rules as with ordinary arrays. (Arrays implicitly have 3 index accessors - get, set and address).

The reasons for why this would be convenient are:
- evaluation order of arguments in indexed assignments will match ordinary arrays and common collections. I.E. you will get `IndexOutOfRange` after evaluating RHS, for example. 
It is a small thing but consistency could be nice when switching between inline arrays and ordinary arrays.
- byval accessors may compose better with features like `await`. 
Although I think the byref accessor in this case could be spillable too, there is some simplicity when spilling values. Even for ordinary arrays it is a bit simpler.

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+The language will relax the restrictions on fixed sized arrays such that the 
+can be declared in safe code and the element type can be managed or unmanaged. 
+This will make types like the following legal:
+
+```cs
+internal struct CharBuffer
+{
+    internal fixed char Data[128];
+}
+```
```
VSadov, line 761: Is there a story if I want to use a buffer as a local?  Or I will have to wrap in a struct?

I have seen some use of `stackalloc` in corefx where a fixed sized buffer local would fit better. 
Typically small buffers of fixed size, used in a loop to parse something. 
stackallocs are not very friendly to use in loops.

VSadov: (Example: https://github.com/dotnet/runtime/pull/34149)

YairHalberstadt:
To me the rules around how I can or can't use a ref field in a ref struct are extremely detailed and arcane. I can't imagine remembering them all, and I like remembering language details :-) I'm worried that for any particular use case that doesn't work straight out the box developers are going to waste hours fumbling around trying to find a way around it.

I think either of the following would be much more palatable:

1. Take the breaking change. If a method accepts a ref, and returns a ref struct or vice versa, the returned scope is equivalent to the passed in scope.

2. Any method which gets or sets a ref field in a struct must be marked with a modreq. Any method which uses those must be marked with a modreq. For methods marked with this modreq, if a method accepts a ref, and returns a ref struct or vice versa, the returned scope is equivalent to the passed in scope. It is a language decision whether or not you need to explicitly mark this via an attribute since it is a breaking change, but at least the error message is dead simple - add this attribute.

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+To account for this change the "Parameters" section of the span safety document
+will be updated to include the following:
+
+- If the parameter is the `this` parameter of a `struct` type, it is 
+*ref-safe-to-escape* to the top scope of the enclosing method unless the 
+method is annotated with `[RefEscapes]` in which case it is *ref-safe-to-escape*
+outside the enclosing method.
+
+Misc Notes:
+- A member marked as `[RefEscapes]` can not implement an `interface` method.
```
tannergooding, line 747: Is this true even if the interface is annotated as `RefEscapes`?

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+            }
+
+            return ref _list[index];
+        }
+    }
+}
+```
+
+Members which contain the `[RefEscapes]` attribute cannot be used to implement 
+interface members. This would hide the lifetime nature of the member at
```
davkean, line 735: And overrides?

YairHalberstadt:
After discussing with @tannergooding on discord, I think what's more important to me from a usability perspective than the actual rules, are the diagnostics which going along with them.

For example, let's say I write:

```csharp
  public Span<int> M(ref int x) {
        return new Span<int>(ref x);
    }
```

I think it's important that the error message is really simple and points me to the correct solution:

> parameter 'ref int x' must be marked with `[RefEscapes]`

Instead of:

> `Span<int>(ref x)` cannot be returned from `M` since `x` has a scope greater than the current method.

Similiarly if I then use M:

```csharp
public Span<int> UseM(ref int x)
{
    return M(ref x);
}
```

The error message should be:

> Parameter 'ref int x' must be marked with `[RefEscapes]`

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+        return CreateSpan(...);
+    }
+}
+```
+
+This tension between allowing constructors such as `Span<T>(ref T field)` and 
+ensuring compatibility with `ref struct` returning methods like `CreateSpan<T>` 
+is a key pivot point in the design of `ref` fields.
+
+To do this we will change the escape rules for a constructor invocation, which
```
YairHalberstadt, line 197: Why not instead of changing the rules for ref struct constructors (which is a breaking change), mark them with [RefEscapes] explicitly, and consider `this` to be the return type of a constructor?

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+    public ref T this[int index]
+    {
+        [RefEscapes]
+        get
+        {
+            switch (index)
+            {
+                case 0: return ref _item1;
+                case 1: return ref _item2;
+                case 1: return ref _item3;
```
Ruikuan, line 701:
```suggestion
                case 2: return ref _item3;
```

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+method is annotated with `[RefEscapes]` in which case it is *ref-safe-to-escape*
+outside the enclosing method.
+
+Misc Notes:
+- A member marked as `[RefEscapes]` can not implement an `interface` method.
+- The `RefEscapesAttribute` will be defined in the 
+`System.Runtime.CompilerServices` namespace.
+
+### Safe fixed size buffers
+The language will relax the restrictions on fixed sized arrays such that the
```
Ruikuan, line 752:
```suggestion
The language will relax the restrictions on fixed sized arrays such that they 
```

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+        _field = p; 
+    }
+
+    static RS M(ref RS rs1, [DoesNotEscape]RS rs2)
+    {
+        Span<int> span = stackalloc int[] { 42 };
+
+        // Okay: The parameter here is not a part of the calculated "must match"
+        // set because it can't be returned hence this is legal.
+        p.Set(span);
```
Ruikuan, line 882:
```suggestion
        rs2.Set(span);
```

Ruikuan (commented):
fix typo

proposals/low-level-struct-improvements.md
Joe4evr: Well, the only virtuals that exist in structs are `GetHashCode()` and `Equals(object)`.....

davkean: Indeed.

jaredpar: Given the exact structure of `object` today I don't think it's a problem to have `[RefEscapes]` on `override`. There is no method on `object` that has a `ref` return hence it can't really feed into this problem. 

Generally though the compiler does not make assumptions about `object` (or most other types). Hence we do need to consider a hypothetical corelib where a `ref` returning method is added to `object`. Lets say the following:

```cs
class Object 
{
  public virtual ref int GetIt() => ...;
}
```

Even in this case I believe it would be safe for a `struct` to `override` this method and return `ref` to its internal state. That is because there are only two ways in which this method can be invoked:

1. Through a reference to the `struct`. In that case the rules outlined here would come into effect and everything would be safe. 
1. Through a reference to `object`. At that point the `struct` is no longer on the stack because it's a boxed object. Hence the compiler interpretation of the `ref` as being *safe-to-escape* outside the enclosing method scope is correct because it's returning a `ref` to the boxed object, not a stack value. 

Even after writing that justification out I'm unsure if it should still be allowed. It feels cute and doesn't really open up any scenarios. Going to think on this a bit and see how I feel later in the day.

proposals/low-level-struct-improvements.md
jaredpar: Agree but I believe that falls out from the proposed rules. Essentially this is meant to change the *ref-safe-to-escape* scope for `this` when the `[RefEscapes]` attribute is applied. that will require an update to the `Parameters` section of the span safety rules. That should then feed into the existing `Field` rules for determining appropriate scopes.

proposals/low-level-struct-improvements.md
jaredpar: This is a good point. Another benefit of going with the three accessor pattern is that it will benefit language interop. Think these could be structured as normal named indexers that *should* just work with existing languages.

proposals/low-level-struct-improvements.md
jaredpar: I had not thought about this case but I can see the value that it would provide. Will think on this a bit and see if it's easy enough to make work.

proposals/low-level-struct-improvements.md
tannergooding: Would this mean users could define:
```csharp
public int X
{
    get => ...
    set => ...
    ref => ...
}
```
?

So far the framework has "worked around" this by defining `Value` and `ValueRef` versions of the property. But the above would allow you to extend an existing property and to light up as possible, which would be amazing

jaredpar: > Would this mean users could define

No, this would be an implementation detail of safe `fixed` buffers. The idea being make them as array-like as possible.

proposals/low-level-struct-improvements.md
jaredpar: My intent for `[RefEscapes]` when writing this proposal was that should apply to `struct` only. Having it work on `struct` and `interface` is fine though; it's straight forward to add consistent rules here. I didn't add `interface` originally because the scenario seemed pretty limited. Then again I have an other proposal in the works to expand the use of `ref struct` into generics so it's possible this will be more common. 

Think I'll change this so that it can apply to `struct` or `interface` and see how that conversation goes.

proposals/low-level-struct-improvements.md
jaredpar: These rules only apply to constructors on `ref struct` which directly contain `ref` fields. Given that `ref` fields don't exist today this can't be a breaking change because there are no constructors that they presently apply to.

jaredpar:
Thanks @Ruikuan !

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+    public ref T this[int index] => ...; 
+}
+```
+
+This directly exposes the `ref` state inside `Span<T>` and the span safety 
+rules account for this. Whether that was implemented as `ByReference<T>` or `ref`
+fields is immaterial to those rules. As a part of allowing `ref` fields though 
+we must define their rules such that they fit into the existing consumption 
+rules for `ref struct`. Specifically this must account for the fact that it's 
+legal *today* for a `ref struct` to return it's `ref` state as `ref` to the
```
jnm2, line 384:
```suggestion
legal *today* for a `ref struct` to return its `ref` state as `ref` to the 
```

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+    return ref s2[0];
+}
+```
+
+To account for `ref` fields the *ref-safe-to-escape* rules for fields will be 
+adjusted to the following:
+
+> An lvalue designating a reference to a field, e.F, is *ref-safe-to-escape* (by reference) as follows:
+> - If `F` is a `ref` field and `e` is `this`, it is *ref-safe-to-escape* from the enclosing method.
+> - Else if `F` is a `ref` field it's *ref-safe-to-escape* scope is the *safe-to-escape* scope of `e`.
```
jnm2, line 428:
```suggestion
> - Else if `F` is a `ref` field its *ref-safe-to-escape* scope is the *safe-to-escape* scope of `e`.
```

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+> - Else if `e` is of a reference type, it is *ref-safe-to-escape* from the enclosing method.
+> - Else it's *ref-safe-to-escape* is taken from the *ref-safe-to-escape* of `e`.
```
jnm2, line 430:
```suggestion
> - Else its *ref-safe-to-escape* is taken from the *ref-safe-to-escape* of `e`.
```

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+```
+
+Misc Notes:
+- A `ref` field can only be declared inside of a `ref struct` 
+- A `ref` field cannot be declared `static`
+- A `ref` field can only be `ref` assigned in the constructor of the declaring
+type.
+- The reference assembly generation process must preserve the presence of a 
+`ref` field inside a `ref struct` 
+- A `ref readonly struct` must declare it's `ref` fields as `ref readonly`
```
jnm2, line 603:
```suggestion
- A `ref readonly struct` must declare its `ref` fields as `ref readonly`
```

proposals/low-level-struct-improvements.md
tannergooding: Except for `Span<T>`, right?

proposals/low-level-struct-improvements.md
```diff
@@ -0,0 +1,972 @@
...
+is a key pivot point in the design of `ref` fields.
+
+To do this we will change the escape rules for a constructor invocation, which
+today are the same as method invocation, on a `ref struct` that **directly** 
+contains a `ref` field as follows:
+- If the constructor contains any `ref struct`, `ref`, `out` or `in` parameters 
+then the *safe-to-escape* of the return will be the current scope
+- Else the *safe-to-escape* will be the outside the method scope
+
+Lets examine these rules in the context of samples to better understand their
```
jnm2, line 204:
```suggestion
Let's examine these rules in the context of samples to better understand their
```

jnm2 (commented):
Yay for English

jaredpar:
@JamesNK 

> I think I have a situation where this would be useful. Would it solve this highlighted problem? -

Not as presently written. The rules are pretty restrictive, they assert that any use of a constructor that has a `ref` parameter on a `ref struct` with a `ref` field is always limited to the current method scope. Hence in this case you wouldn't be able to assign that into `ctx` here. 

In this case though the invocation is safe. It's passing a `ref` to a known heap location hence there is no reason to restrict the return. An earlier version of the doc had this exception but I pulled it out because I thought it might be too much of a special case. This is the second time though this has come up so I'm leaning towards finding a way to add it back.

jaredpar:
@YairHalberstadt 

> I think what's more important to me from a usability perspective than the actual rules, are the diagnostics which going along with them.

Agree. Good diagnostics will be important.

tannergooding:
Just wanted to confirm something that was discussed in Discord (as its my understanding this is the case).

`RefEscapes` would allow the following code to work (it is illegal today):
```csharp
public Span<int> M([RefEscapes]ref int x) {
    return new Span<int>(ref x);
}
```

While `DoesNotEscape` would allow the following, currently also illegal:
```csharp
public ref int M([DoesNotEscape] Span<int> x, int[] y)
{
    return ref y[0];
}

public ref int X(int[] y)
{
    return ref M(stackalloc int[5], y);
}
```

In the case of the latter, `DoesNotEscape` would do basic analysis to validate it doesn't escape (I'm not sure if any validation could be done or should be done for `RefEscapes`).

Would these be just attributes, `modreq`, `modopt`?
Would these attributes only be valid on `in`, `out`, ref` and `ref struct`?

jaredpar:
Thanks @jnm2. With only three typo suggestions before you, I was beginning to worry that maybe I'd gotten better at English. 😄

jaredpar:
@tannergooding 

> RefEscapes would allow the following code to work (it is illegal today):

No it would not. The `[RefEscapes]` attribute can only be applied to instance members of `struct` (explicitly called out in the proposal). It cannot be applied to individual parameters of members. Further the `[RefEscapes]` attribute only impacts the *ref-safe-to-escape* scope of `this`. That in turn causes a tiny tweak to the *ref-safe-to-escape* rules for method invocations returning a `ref` value. What your discussing requires changes to the *safe-to-escape* return of all methods.

Essentially the behavior you are looking for here is changing how *safe-to-escape* is interpreted for method calls. That is a very different feature than `[RefEscapes]`. Sure it's possible to add such annotations to the language but my intuition is that it's not really worth it. 

One thing to remember is that the rules today around how this return pattern is calculated came in large part to our own usability feedback implementing `Span<T>`. If we were starting from scratch today I think that we would very likely end up with the exact same rules because it optimizes for the majority use case. 

> While DoesNotEscape would allow the following, currently also illegal:

Correct. This is the exact intended purpose of `DoesNotEscape`. 

If you have real world examples of this pattern I would be eager to see them. It will help ensure the feature is solving the intended problem as well as making me feel more confident that it's worth adding to the language.

> Would these be just attributes, modreq, modopt?

`modreq` 

> Would these attributes only be valid on in, out, ref and ref struct?

Applicability:

- `RefEscapes` applicable to `struct` instance members only 
- `DoesNotEscape` likely applicable to parameters and instance members. This section is still a work in progress so less certainty in all the places it will apply

jnm2:
@jaredpar I think the usability of the English language is highly questionable :D

tannergooding:
> The [RefEscapes] attribute can only be applied to instance members of struct (explicitly called out in the proposal). It cannot be applied to individual parameters of members.

Ah, I see. I had seen it applied to the `getter` and misinterpreted that.

> RefEscapes applicable to struct instance members only

So this is Methods that have a return, or fields and such as well? And it basically just indicates that the `ref` return is going to be `this` or a subset of `this`?

> Essentially the behavior you are looking for here is changing how safe-to-escape is interpreted for method calls. That is a very different feature than [RefEscapes].

Hmmm, I had interpreted it as changing how safe-to-escape is interpreted because that seems like a useful feature.
Being able to annotate that a `ref` is `this` allows ref returns to non ref-fields. However, being able to annotate that a `ref` escapes allows things like `MemoryMarshal.CreateSpan` to exist and to exist "safely" without leaving a hole around something that can't otherwise be expressed.

jaredpar:
@tannergooding 

> So this is Methods that have a return, or fields and such as well? And it basically just indicates that the ref return is going to be this or a subset of this?

It indicates that the `ref` return is *potentially* returning state inside of `this`. That effectively makes the following edit to [method invocation rules](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md#method-invocation) for calculating *ref-safe-to-escape* when the target method is annotated with `[RefEscapes]` 

- the ref-safe-to-escape of all ref and out argument expressions ~~(excluding the receiver)~~

This is a very specific and narrow fix because it's a narrow and specific problem. :smile:

> Being able to annotate that a ref is this allows ref returns to non ref-fields

No it does not. I understand on the surface they seem related but in terms of how the span safety rules work they are very different problems and require different solutions. The `[RefEscapes]` proposal is all about changing how *ref-safe-to-escape* validation works (and it's a very targeted change). The behavior you are asking for is about how *safe-to-escape* returns are calculated. Attempting to use one solution both unnecessarily complicates the space and very likely creates other side effects that you almost certainly do not want.

I think the name `[RefEscapes]` is adding to the confusion here. Think of the name as `[ThisRefEscapes]` instead. 

> However, being able to annotate that a ref escapes allows things like MemoryMarshal.CreateSpan to exist and to exist "safely" without leaving a hole around something that can't otherwise be expressed.

Annotating that a `ref` escapes though is somewhat rhetorical because `ref` can escape today, and they can escape by `ref`:

```cs
ref int Identity(ref int p) => ref p;
```

The issue is more that `ref` cannot be smuggled in the proposal. Or rather a `ref` that sneakily escapes by `ref` inside a value. 

Again though, this is solvable with another annotation. The issue is that when looking at the issues and samples the extra complexity did not seem warranted. All of the features in this proposal line up to specific customer asks based on samples, real scenarios, etc ... The same does not exist for this request. Further the existing rules, and the decisions we made about them originally WRT usability, provide reasonable evidence this is not actually needed as a feature.

If you feel passionately about this I think the way to move it forward is with samples, evidence, etc ....

tannergooding:
> If you feel passionately about this I think the way to move it forward is with samples, evidence, etc ....

Not super passionate and I don't think I have any specific examples today. I'd have to go back through the various hacks I'm using right now and double-check.

I think I have a better understanding of the differences now though, so thanks! 

> I think the name [RefEscapes] is adding to the confusion here. Think of the name as [ThisRefEscapes] instead.

Any reason why we couldn't use this name instead, to help reduce confusion?

jaredpar:
@tannergooding 

> Any reason why we couldn't use this name instead, to help reduce confusion?

Because once proposals are published they're immutable. 😄 

Joking aside it's one of the changes I'm considering.

gdkchan:
Haven't seen anything about multi-dimensional fixed buffers in the proposal, but it would be useful too.
Example:
```cs
struct Matrix4x4
{
    public fixed float M[4][4];
}
```
It would be equivalent to the following C struct:
```c
typedef struct Matrix4x4 {
    float M[4][4];
}
```
I'm not sure how one would pass it around though. According to the current proposal, I believe the safe method to pass the fixed buffers around is passing as `Span<T>`. However there's no multi-dimensional `Span<T>`, so it wouldn't work for this case.

Thealexbarney:
> Haven't seen anything about multi-dimensional fixed buffers in the proposal, but it would be useful too.

Multi-dimensional arrays could work without relatively major changes, but only if referencing a single element is the only thing allowed.

Example:
```cs
struct MyStruct
{
    public fixed int MyArray[4][5][6];
}

int element = myStruct.MyArray[2][3][4];
```

Could be equivalent to (Bounds checks skipped for brevity):
```cs
struct MyStruct
{
    public fixed int MyArray[120];

    public int GetMyArray(int dim1, int dim2, int dim3)
    {
        return MyArray[dim1 * 5 * 6 + dim2 * 6 + dim3];
    }
}

int element = myStruct.GetMyArray(2, 3, 4);
```

Getting a single dimension of the array would be more difficult. Getting a span from the least significant dimension wouldn't be difficult:

```cs
public Span<int> Get(int dim1, int dim2)
{
    return MyArray.Slice(dim1 * 5 * 6 + dim2 * 6, 6);
}
```

Getting higher dimensions might be done by encoding the dimension sizes in the type. C# can't do anything like `MultiDim<int, 4, 5, 6>`, so the only way I see would be to generate a struct for each dimension. However the types would get ugly rather quickly, not to mention that these compiler-generated types would need to be accessible by user-facing code.

Pseudocode example:
```cs
public ref struct Array4_5_6
{
    private Span<int> _span;

    public Array5_6 GetDim2(int index)
    {
        return new Array5_6(_span.Slice(index * 5 * 6, 5 * 6));
    }

    public Span<int> GetDim3(int dim1, int dim2)
    {
        return _span.Slice(dim1 * 5 * 6 + dim2 * 6, 6);
    }
}

public ref struct Array5_6
{
    private Span<int> _span;

    public Span<int> GetDim2(int index)
    {
        return _span.Slice(index * 6, 6);
    }

    public int Get(int dim1, int dim2)
    {
        return _span[dim1 * 6 + dim2];
    }
}
```

jaredpar:
@gdkchan 

I had not considered multi-dimensional array support here. The idea of this proposal was essentially to extend the existing support for fixed sized buffers to include supporting all element types and not just unmanaged types. Multi-dimensional support is essentially expanding the scope of the original feature quite a bit. Given that it's existed for the entirety of C# and there hasn't been a big push to add this support I'm fairly skeptical that it's worth the extra complexity. 

Added it to the open issue section though so we can discuss in LDM when we discuss the rest of the feature (this specific section hasn't been covered yet).

jaredpar:
Thanks for the feedback everyone! Have updated the proposal based on various discussions here. Particularly:

1. Renamed `[RefEscapes]` to `[ThisRefEscapes]`  to make the semantics of the feature much clearer. 
1. Made constructors and `ref` field assignment more flexible when the values being used are known to refer to the heap. Scenarios provided by @JamesNK  and @davidwrighton made me feel the extra complexity in the rules are justified (and really it's actually a fairly small change). 
1. Responded to several feedback points by @VSadov on how `fixed` buffers should be emitted to make them more closely resemble existing arrays.

proposals/low-level-struct-improvements.md
jaredpar: Added this to the "Open Issues" section of the design so we can discuss if the complexity is worth it in LDM. This section actually hasn't been discussed in LDM yet so this can just be a part of that discussion.

jaredpar:
Thanks again for the feedback everyone. At this point the *initial* design of `ref` field, `[ThisRefEscapes]` and safe `fixed` buffers is a good place. The `ref` field portion has largely been presented to LDM while the `[ThisRefEscapes]` + safe `fixed` buffers should be presented this week / next week. Likely another round of edits will follow after that. 

Merging now because I want to make progress on the `[DoesNotEscape]` section. Using a separate PR for that as I noted that section should not be considered in this PR as it was decidedly incomplete.

YairHalberstadt:
@jaredpar is there any championed issue for this proposal?

Thanks

333fred:
There are a few. https://github.com/dotnet/csharplang/issues/1147 is fixed fields.

YairHalberstadt:
Would it be useful to have a single one for this proposal? Are there any for returning a field of a struct by ref?

jaredpar:
I've been using #1147 as the primary championed issue here.

Shpendicus:
@jaredpar is something of this possible by current (.NET 6.0 version, VS2022) ?Would be nice to hear :)

PR #3938, opened 2020-09-25
Update explicit enumeration conversions for native integers

Youssef1313:
Related to: https://github.com/dotnet/roslyn/issues/48035

cc: @cston

### DIFF
--- a/spec/conversions.md
+++ b/spec/conversions.md
@@ -245,7 +245,7 @@ The explicit numeric conversions possibly lose information or possibly cause exc
 
 The explicit enumeration conversions are:
 
-*  From `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, or `decimal` to any *enum_type*.
+*  From `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `char`, `float`, `double`, or `decimal` to any *enum_type*.
 *  From any *enum_type* to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, or `decimal`.
 *  From any *enum_type* to any other *enum_type*.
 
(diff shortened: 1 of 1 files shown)


spec/conversions.md
```diff
@@ -245,7 +245,7 @@ The explicit numeric conversions possibly lose information or possibly cause exc
 
 The explicit enumeration conversions are:
 
-*  From `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, or `decimal` to any *enum_type*.
+*  From `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `char`, `float`, `double`, or `decimal` to any *enum_type*.
 *  From any *enum_type* to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, or `decimal`.
```
cston, line 249: "From any *enum_type* to ... `nint`, `nuint`, ..."

Youssef1313: Thanks for review and spotting this. I'll update it and also revise the implicit numeric conversions as they're outdated too.

spec/conversions.md
```diff
@@ -50,8 +50,8 @@ The implicit numeric conversions are:
 *  From `byte` to `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `float`, `double`, or `decimal`.
 *  From `short` to `int`, `long`, `float`, `double`, or `decimal`.
 *  From `ushort` to `int`, `uint`, `long`, `ulong`, `float`, `double`, or `decimal`.
-*  From `int` to `long`, `float`, `double`, or `decimal`.
-*  From `uint` to `long`, `ulong`, `float`, `double`, or `decimal`.
+*  From `int` to `nint`, `long`, `float`, `double`, or `decimal`.
+*  From `uint` to `nuint`, `long`, `ulong`, `float`, `double`, or `decimal`.
```
Youssef1313, line 54: These updates made due to the fact that the following compiles without error:

```
int x = 5;
nint y = x;

uint a = 5;
nuint b = a;
```

Correct me if that's not correct.

svick:
This is the draft spec for **C# 6.0**, so I think it shouldn't be updated with C# 9.0 features. Instead, the page that should be updated is [the `nint` proposal](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/native-integers).

Youssef1313:
@svick Thanks for catching this. I opened #3941

PR #3955, opened 2020-09-30
Improve title Update README.md

shubhamb008:


333fred:
Thanks for the suggestion. However, unless @MadsTorgersen would like to add this, I think the readme title is fine as it currently is.

333fred:
We're going to leave this as is. Thanks though.

PR #3961, opened 2020-10-02
Parameter does not escape annotation

jaredpar:
This updates the low level struct proposal section on annotations for
parameters that do not escape. This is intended to remove low level
friction points that result from using `ref struct` as arguments to
instance methods on `ref struct`.

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -961,7 +961,7 @@ void WriteData(ReadOnlySpan<char> data)
 ```
 
 This pattern is fairly common across .NET code and it works just fine when 
-`ref struct` are not involved. Once users adopt `ref struct` though it forces them
+a `ref struct` is not involved. Once users adopt `ref struct` though it forces them
 to change their patterns here and often they just resort to `unsafe` to work
 around the limitations here.
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -975,7 +975,7 @@ parameter.
 ```cs
 class C
 {
-    static Span<int> M1(Span<int> p1, [DoesNotEsacpe] Span<int> p2)
+    static Span<int> M1(Span<int> p1, [DoesNotEscape] Span<int> p2)
     {
         // Okay: the *safe-to-escape* here is still outside the enclosing scope
         // of the current method.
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -876,65 +885,167 @@ struct RS
...
+        data = buffer.Slice(0, length);
     }
 
-    static RS M(ref RS rs1, [DoesNotEscape]RS rs2)
+    WriteDataCore(data);
+}
+```
+
+This pattern is fairly common across .NET code and it works just fine when 
+`ref struct` are not involved. Once users adopt `ref struct` though it forces them
```
333fred, line 964:
```suggestion
a `ref struct` is not involved. Once users adopt `ref struct` though it forces them
```

proposals/low-level-struct-improvements.md
```diff
@@ -995,6 +1106,53 @@ struct Dimensions
...
+{
+    // Okay
+    span = stackalloc byte[length];
+}
+```
+
+This only works for `Span<T>` though, there is no general purpose mechanism for
+`ref struct` values. However the `[DoesNotEscape]` attribute provides exactly 
+the semantics that are desired here. If we decide in the future to allow 
+attributes to apply to local functions it would provide immediate relief to this
```
333fred, line 1153: We allow attributes to apply to local functions.

jaredpar: Blah, I meant to say local variables .

proposals/low-level-struct-improvements.md
```diff
@@ -876,65 +885,167 @@ struct RS
...
+`[DoesNotEscape]`. This can be applied to parameters of any type or instance 
+members defined on `ref struct`. When applied to parameters the *safe-to-escape*
+and *ref-safe-to-escape* scope will be the current method scope. When applied to
+instance members of `ref struct` the same limitation will apply to the `this`
+parameter.
+
+```cs
+class C
+{
+    static Span<int> M1(Span<int> p1, [DoesNotEsacpe] Span<int> p2)
```
jkotas, line 978:
```suggestion
    static Span<int> M1(Span<int> p1, [DoesNotEscape] Span<int> p2)
```

proposals/low-level-struct-improvements.md
jaredpar, line 1153:
```suggestion
attributes to apply to local variables it would provide immediate relief to this
```

PR #3991, opened 2020-10-10
Update classes.md method reference

jessejjohnson:
Corrected method name reference within _Hiding_  section.

### DIFF
--- a/spec/classes.md
+++ b/spec/classes.md
@@ -1017,7 +1017,7 @@ class Test
     }
 }
 ```
-shows a nested class `M` that hides the method `M` defined in `Base`.
+shows a nested class `M` that hides the method `F` defined in `Base`.
 
 #### this access
 


spec/classes.md
```diff
@@ -1017,7 +1017,7 @@ class Test
     }
 }
 ```
-shows a nested class `M` that hides the method `M` defined in `Base`.
+shows a nested class `M` that hides the method `F` defined in `Base`.
```
CyrusNajmabadi, line 1020: i don't believe this change is correct.  the current text is accurate.  the nested class 'M' in 'Derived' does hide the method 'M' in 'Base'.

333fred: Agreed. This is correct as written.

jessejjohnson: My bad, **you’re both absolutely correct**. Was wizzing through and the `class` to `method` hiding must have threw me off 🤦‍♂️ Thanks for the quick review 🙌

PR #4036, opened 2020-10-19
Add proposals/csharp-8.0/shadowing-in-nested-functions.md

cston:


333fred (approved):

proposals/csharp-8.0/shadowing-in-nested-functions.md
```diff
@@ -1,3 +1,22 @@
...
+﻿# Name shadowing in nested functions
 
-In C# 8.0, we added a feature that permits parameters and locals in lambdas and local functions to use names that hide/shadow the names of locals or parameters from the enclosing scope. This is a placholder for its specification.
+## Summary
+
+Permit variable names in lambdas and local functions to reuse (and shadow) names from the enclosing scope.
+
+## Detailed design
+
+With `-langversion:8`, names of locals, local functions, parameters, type parameters, and range variables within a lambda or local function can reuse names of locals, local functions, parameters, type parameters, and range variables from an enclosing scope. The name in the nested scope hides the symbol of the same name from the enclosing scope within the nested scope.
```
jaredpar, line 9: I think we need to clarify the meaning of "scope" here. I believe this is meant to refer to the logical method scope however term "scope" in the compiler at least typically refers to any set of `{ }`. Concretely this languaeg could be read to indicate the following program is legal (it is not)

```cs
int i = 42;
if (i > 13)
{
  int i = 12;
}
```

jaredpar (approved):

PR #4043, opened 2020-10-20
Add proposals/csharp-9.0/unconstrained-type-parameter-annotations.md

cston:
Proposal for annotations on unconstrained type parameters: `T?` and `where T : default`

### DIFF
--- a/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
+++ b/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
@@ -11,14 +11,21 @@ static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }
 
 In C#8, `?` annotations could only be applied to type parameters that were explicitly constrained to value types or reference types.
 In C#9, `?` annotations can be applied to any type parameter, regardless of constraints.
+Unless a type parameter is explicitly constrained to value types, `?` annotations can only be applied within a nullable-enabled context (typically `#nullable enable`).
 
 If a type parameter `T` is substituted with a reference type, then `T?` represents a nullable instance of that reference type.
-If `T` is substituted with a value type, then `T?` is represents an instance of `T`.
+If `T` is substituted with a value type, then `T?` represents an instance of `T`.
 ```C#
 var s1 = new string[0].FirstOrDefault();  // string? s1
 var s2 = new string?[0].FirstOrDefault(); // string? s2
 var i1 = new int[0].FirstOrDefault();     // int i1
 var i2 = new int?[0].FirstOrDefault();    // int? i2
+```
+
+If `T` is substituted with an annotated type `U?`, then `T?` represents the annotated type `U?` rather than `U??`.
+```C#
+var u1 = new U[0].FirstOrDefault();  // U? u1
+var u2 = new U?[0].FirstOrDefault(); // U? u2
 ```
 
 For return values, `T?` is equivalent to `[MaybeNull]T`.
--- a/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
+++ b/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
@@ -4,7 +4,7 @@
 
 Allow nullable annotations for type parameters that are not constrained to value types or reference types: `T?`.
 ```C#
-static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }`
+static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }
 ```
 
 ## `?` annotation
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
```diff
@@ -0,0 +1,67 @@
...
+static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }
+```
+
+## `?` annotation
+
+In C#8, `?` annotations could only be applied to type parameters that were explicitly constrained to value types or reference types.
+In C#9, `?` annotations can be applied to any type parameter, regardless of constraints.
+
+If a type parameter `T` is substituted with a reference type, then `T?` represents a nullable instance of that reference type.
+If `T` is substituted with a value type, then `T?` is represents an instance of `T`.
```
pkulikov, line 16:
```suggestion
If `T` is substituted with a value type, then `T?` represents an instance of `T`.
```

pkulikov, line 16: Question: in case of a value type, `T?` represents an instance of `T` or the corresponding nullable `T?` type?

pkulikov, line 16: What if `T` is a nullable value type?

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
```diff
@@ -0,0 +1,67 @@
...
+```C#
+static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }
+```
+
+## `?` annotation
+
+In C#8, `?` annotations could only be applied to type parameters that were explicitly constrained to value types or reference types.
+In C#9, `?` annotations can be applied to any type parameter, regardless of constraints.
+
+If a type parameter `T` is substituted with a reference type, then `T?` represents a nullable instance of that reference type.
```
jaredpar, line 15: May want to clarify that this is the case even if `T` is used in source with an annotation. Essentially make it clear that when `T` is substituted with `string?` a `T?` becomse `string?` not `string??`

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
jaredpar: @pkulikov do you mean what if `T` is substituted with a nullable value type (say `int?`)? In that case then the result is `int?`. Essentially in the case where the type argument is a value type then the `?` is effectively dropped from the type parameter.

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
```diff
@@ -0,0 +1,67 @@
...
+If `T` is substituted with a value type, then `T?` is represents an instance of `T`.
+```C#
+var s1 = new string[0].FirstOrDefault();  // string? s1
+var s2 = new string?[0].FirstOrDefault(); // string? s2
+var i1 = new int[0].FirstOrDefault();     // int i1
+var i2 = new int?[0].FirstOrDefault();    // int? i2
```
jaredpar, line 21: One case not covered in the samples or text is what happens when `T` is substituted with a type parameter `U` which is not definitely constrained to a reference or value type. Essentially 

``` 
T? M1<T>() => throw null!;
void M2<U>()
{
  var local1 = M1<U>();   // what is the type here?
  var local2 = M1<U?>(); // what is the type here?
}
```

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
```diff
@@ -0,0 +1,67 @@
...
+
+Allow nullable annotations for type parameters that are not constrained to value types or reference types: `T?`.
+```C#
+static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }
+```
+
+## `?` annotation
+
+In C#8, `?` annotations could only be applied to type parameters that were explicitly constrained to value types or reference types.
+In C#9, `?` annotations can be applied to any type parameter, regardless of constraints.
```
jaredpar, line 13: Does the annotation context impact how substitution works in these scenarios?

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
pkulikov: > do you mean what if `T` is substituted with a nullable value type (say `int?`)? In that case then the result is `int?`

@jaredpar yes, thanks!

> Essentially in the case where the type argument is a value type then the `?` is effectively dropped from the type parameter.

That's a good clarification, thank you.

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+If `T` is substituted with an annotated type `U?`, then `T?` represents the annotated type `U?` rather than `U??`.
+```C#
+var u1 = new U[0].FirstOrDefault();  // U? u1
+var u2 = new U?[0].FirstOrDefault(); // U? u2
+```
+
+For return values, `T?` is equivalent to `[MaybeNull]T`.
+For argument values, `T?` is equivalent to `[AllowNull]T`.
```
jaredpar, line 36: Should specify that this applies to inheritance scenarios. Essentially that they can be interchanged there. Really important for code versioning across assembly boundaries.

proposals/csharp-9.0/unconstrained-type-parameter-annotations.md
```diff
@@ -0,0 +1,78 @@
...
+```C#
+static T? FirstOrDefault<T>(this IEnumerable<T> collection) { ... }
+```
+
+## `?` annotation
+
+In C#8, `?` annotations could only be applied to type parameters that were explicitly constrained to value types or reference types.
+In C#9, `?` annotations can be applied to any type parameter, regardless of constraints.
+
+Unless a type parameter is explicitly constrained to value types, annotations can only be applied within a `#nullable enable` context.
```
jaredpar, line 15: Occurs to me the doc doesn't state what the behavior is when the type is instantiated in an oblivious context. How does that impact the type?

jaredpar (approved):

jcouv (commented):
LGTM Thanks

PR #4060, opened 2020-10-22
Allow semi-colon after block body for record declarations

jcouv:
Since we allow `class C { };`...
Tagging @dotnet/roslyn-compiler for review

### DIFF
--- a/proposals/csharp-9.0/records.md
+++ b/proposals/csharp-9.0/records.md
@@ -19,7 +19,7 @@ record_base
     ;
 
 record_body
-    : '{' class_member_declaration* '}'
+    : '{' class_member_declaration* '}' ';'?
     | ';'
     ;
 ```


YairHalberstadt:
Can we allow this for all classes?

Now that source generators are a thing this is increasingly common, e.g. with https://github.com/martinothamar/WrapperValueObject.

If primary constructors become a thing I think this might become very common, combined with source generators.

svick:
@YairHalberstadt As @jcouv mentioned, semicolon after a class declaration is already valid, see [SharpLab](https://sharplab.io/#v2:C4LgbgpgTglgZgTwD4AEBMACAwgWAFADe+GJG6GAchgQL4Dc+9QA), or [the spec](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#class-declarations). Or am I missing something?

YairHalberstadt:
Oops, I thought this was after the primary constructor and without a body 🤦

333fred (approved):

PR #4104, opened 2020-11-05
Refresh NRT spec

jcouv:
This is a first pass.

Several things aren't yet covered:
- invocations and arguments (including `ref` and `out` ),
- attributes,
- type substitution rules,
- difference between "maybe null" and "maybe default", 
- array syntax, 
- warnings for overrides, implementations, and delegate conversions,
- null tests, pure null tests, and learning from them,
- special methods (`CompareExchange`),
- conversions (including user-defined conversions),
- property/indexer input versus output,
- and more (feel free to comment with area that we should expand).

Tagging @MadsTorgersen @jaredpar

### DIFF
--- a/proposals/csharp-9.0/nullable-reference-types-specification.md
+++ b/proposals/csharp-9.0/nullable-reference-types-specification.md
@@ -58,7 +58,7 @@ Nullable reference types and nullable type parameters cannot occur in the follow
 
 A warning is given on a `nullable_reference_type` and `nullable_type_parameter` in a *disabled* nullable annotation context.
 
-### Nullable and nonnullable class constraints
+### `class` and `class?` constraint
 
 The `class` constraint has a nullable counterpart `class?`:
 
@@ -69,9 +69,9 @@ primary_constraint
     ;
 ```
 
-A type parameter constrained with `class` (in an *enabled* annotation context) must be a nonnullable reference type.
+A type parameter constrained with `class` (in an *enabled* annotation context) must be instantiated with a nonnullable reference type.
 
-A type parameter constrained with `class?` (or `class` in a *disabled* annotation context) may either be a nullable or nonnullable reference type.
+A type parameter constrained with `class?` (or `class` in a *disabled* annotation context) may either be instantiated with a nullable or nonnullable reference type.
 
 A warning is given on a `class?` constraint in a *disabled* annotation context.
 
@@ -88,7 +88,7 @@ primary_constraint
 
 ### `default` constraint
 
-The `default` constraint can be used on a method override or explicit implementation to disambiguate `T?` meaning "nullable type parameter" from "nullable value type" (`Nullable<T>`).
+The `default` constraint can be used on a method override or explicit implementation to disambiguate `T?` meaning "nullable type parameter" from "nullable value type" (`Nullable<T>`). Lacking the `default` constraint a `T?` syntax in an override or explicit implementation will be interpreted as `Nullable<T>`
 
 See https://github.com/dotnet/csharplang/blob/master/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md#default-constraint
 
@@ -265,7 +265,7 @@ tracked_expression
 
 Where the identifiers denote fields or properties.
 
-The null state for tracked variables is "not null" in unreachable code.
+The null state for tracked variables is "not null" in unreachable code. This follows other decisions around unreachable code like considering all locals to be definitely assigned.
 
 ***Describe null state transitions similar to definite assignment***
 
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -49,29 +69,75 @@ primary_constraint
     ;
 ```
 
+A type parameter constrained with `class` (in an *enabled* annotation context) must be a nonnullable reference type.
```
jaredpar, line 72:
```suggestion
A type parameter constrained with `class` (in an *enabled* annotation context) must be instantiated with a nonnullable reference type.
```

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -49,29 +69,75 @@ primary_constraint
...
+
+A type parameter constrained with `class?` (or `class` in a *disabled* annotation context) may either be a nullable or nonnullable reference type.
```
jaredpar, line 74:
```suggestion
A type parameter constrained with `class?` (or `class` in a *disabled* annotation context) may either be instantiated with a nullable or nonnullable reference type.
```

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -36,9 +56,9 @@ Nullable reference types cannot occur in the following positions:
 - as the `type` in an `is_expression`, a `catch_clause` or a `type_pattern`
 - as the `interface` in a fully qualified interface member name
 
-A warning is given on a `nullable_reference_type` where the nullable annotation context is disabled.
+A warning is given on a `nullable_reference_type` and `nullable_type_parameter` in a *disabled* nullable annotation context.
 
-### Nullable class constraint
+### Nullable and nonnullable class constraints
```
jaredpar, line 61:
```suggestion
### `class` and `class?` constraint
```

This matches the style of the `notnull` constraint section below

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -49,29 +69,75 @@ primary_constraint
     ;
 ```
 
+A type parameter constrained with `class` (in an *enabled* annotation context) must be a nonnullable reference type.
+
+A type parameter constrained with `class?` (or `class` in a *disabled* annotation context) may either be a nullable or nonnullable reference type.
+
+A warning is given on a `class?` constraint in a *disabled* annotation context.
```
jaredpar, line 76: May be worth calling out explicitly that `class` is okay here but `class?` is not. Think it's clearer that way.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -49,29 +69,75 @@ primary_constraint
...
+```antlr
+primary_constraint
+    : ...
+    | 'notnull'
+    ;
+```
+
+### `default` constraint
+
+The `default` constraint can be used on a method override or explicit implementation to disambiguate `T?` meaning "nullable type parameter" from "nullable value type" (`Nullable<T>`).
```
jaredpar, line 91:
```suggestion
The `default` constraint can be used on a method override or explicit implementation to disambiguate `T?` meaning "nullable type parameter" from "nullable value type" (`Nullable<T>`). Lacking the `default` constraint a `T?` syntax in an override or explicit implementation will be interpreted as `Nullable<T>`
```

Think this gets the summary of the issue here and saves some clicks for readers

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -202,6 +265,8 @@ tracked_expression
 
 Where the identifiers denote fields or properties.
 
+The null state for tracked variables is "not null" in unreachable code.
```
jaredpar, line 268:
```suggestion
The null state for tracked variables is "not null" in unreachable code. This follows other decisions around unreachable code like considering all locals to be definitely assigned.
```

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -291,39 +358,25 @@ The null state of the following expression forms is always "not null":
...
-If the type of `E` is a nullable reference type `C?` and the null state of `E` is "not null" then the type inferred for `x` is `C`. Otherwise, the inferred type is the type of `E`.
-
-The nullability of the type inferred for `x` is determined as described above, based on the annotation context of the `var`, just as if the type had been given explicitly in that position.
-
-### Type inference for `var?`
+### Nested functions
 
-The type inferred for local variables declared with `var?` is independent of the null state of the initializing expression.
+Nested functions (lambdas and local functions) are treated like methods, except in regards to their captured variables.
+The default state of a captured variable inside a lambda or local function is the intersection of the nullable state
```
jaredpar, line 364:
```suggestion
The initial state of a captured variable inside a lambda or local function is the intersection of the nullable state
```

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -291,39 +358,25 @@ The null state of the following expression forms is always "not null":
...
+of the variable at all the "uses" of that nested function. A use of a function is either a call to that function, or
+where it is converted to a delegate.
```
jaredpar, line 366:
```suggestion
of the variable at all the "uses" of that nested function or lambda. A use of a local function is either a call to that 
function, or where it is converted to a delegate. A use of a lambda is the point at which it is defined in source.
```

jaredpar (approved):

jaredpar (approved):

jaredpar:
@jcouv are we good to merge here?

PR #4126, opened 2020-11-11
NRT spec updates

jaredpar:
This makes the following updates to the NRT spec

- Examples for scenarios that commonly trip up customers
- Clarification on when expressions have the null state "maybe default"
  vs. "maybe null"
- Minor langauge cleanup

### DIFF
--- a/proposals/csharp-9.0/nullable-reference-types-specification.md
+++ b/proposals/csharp-9.0/nullable-reference-types-specification.md
@@ -400,9 +400,9 @@ If `B` denotes the base type of the enclosing type, `base.I` has the same null s
 - Else if the type is a type parameter then it has the null state "maybe default"
 - Else it has the null state "maybe null"
 
-### Null-conditional expressions
+### Null-conditional expressions ?.
 
-A `null_conditional_expression` has the null state based on the properties of expression type. Note that this refers to the type of the `null_conditional_expression`, not the original type of the member being invoked:
+A `null_conditional_expression` has the null state based on the expression type. Note that this refers to the type of the `null_conditional_expression`, not the original type of the member being invoked:
 
 - If the type is a *nullable* value type then it has the null state "maybe null"
 - Else if the type is a *nullable* type parameter then it has the null state "maybe default"
@@ -412,9 +412,12 @@ A `null_conditional_expression` has the null state based on the properties of ex
 
 If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for the type of the user-defined conversion. Otherwise:
 
-- If `T` is a *nullable* type in the form `U?` where `U` is a type parameter then the null state is "maybe default"
-- Else If `T` is a *nullable* type then the null state is "maybe null"
-- Else the null state is the same as the null state of `E`
+- If `T` is a *nonnullable* value type then `T` has the null state "not null"
+- Else if `T` is a *nullable* value type then `T` has the null state "maybe null"
+- Else if `T` is a *nullable* type in the form `U?` where `U` is a type parameter then `T` has the null state "maybe default"
+- Else if `T` is a *nullable* type, and `E` has null state "maybe null" or "maybe default", then `T` has the null state "maybe null"
+- Else if `T` is a type parameter, and `E` has null state "maybe null" or "maybe default", then `T` has the null state "maybe default"
+- Else `T` has the same null state as `E`
 
 ### Unary and binary operators
 
@@ -428,9 +431,9 @@ The null state of `await E` is the default null state of its type.
 
 ### The `as` operator
 
-The null state of an `E as T` expression depends first on properties of the type `T`. If the type *nonnullable* then the null state is "not null". Otherwise the null state depends on the conversion from the type of `E` to type `T`:
+The null state of an `E as T` expression depends first on properties of the type `T`. If the type of `T` is *nonnullable* then the null state is "not null". Otherwise the null state depends on the conversion from the type of `E` to type `T`:
 
-- If the conversion Identity, boxing, implicit reference or implicit nullable has the same null state as `E`
+- If the conversion is an identity, boxing, implicit reference, or implicit nullable conversion, then the null state is the null state of `E`
 - Else if `T` is a type parameter then it has the null state "maybe default"
 - Else it has the null state "maybe null"
 
--- a/proposals/csharp-9.0/nullable-reference-types-specification.md
+++ b/proposals/csharp-9.0/nullable-reference-types-specification.md
@@ -293,9 +293,9 @@ The null state of an expression is derived from its form and type, and from the 
 
 ### Literals
 
-The null state of a `null` literal is "maybe null". 
+The null state of a `null` literal depends on the target type of the expression. If the target type is a type parameter constrained to a reference type then it's "maybe default". Otherwise it is "maybe null".
 
-Teh null state of a `default` literal depends on the target type of the `default` literal. A `default` literal with target type `T` has the same null state as the `default(T)` expression.
+The null state of a `default` literal depends on the target type of the `default` literal. A `default` literal with target type `T` has the same null state as the `default(T)` expression.
 
 The null state of any other literal is "not null".
 
@@ -319,9 +319,10 @@ if (person.FirstName is not null)
 
 // The return of an invocation is not a tracked expression hence the member_access
 // of the return is also not tracked
-if (GetPerson().FirstName is not null)
+if (GetAnonymous().FirstName is not null)
 {
-    Use(GetPerson().FirstName);
+    // Warning: Cannot convert null literal to non-nullable reference type.
+    Use(GetAnonymous().FirstName);
 }
 
 void Use(string s) 
@@ -393,18 +394,18 @@ If `B` denotes the base type of the enclosing type, `base.I` has the same null s
 
 ### Default expressions
 
-The `default(T)` has the null state based on the properties of the type `T`:
+`default(T)` has the null state based on the properties of the type `T`:
 
-- If the type is a known value type then it has the null state "not null"
+- If the type is a *nonnullable* type then it has the null state "not null"
 - Else if the type is a type parameter then it has the null state "maybe default"
 - Else it has the null state "maybe null"
 
 ### Null-conditional expressions
 
-A `null_conditional_expression` has the null state based on the properties of expression type:
+A `null_conditional_expression` has the null state based on the properties of expression type. Note that this refers to the type of the `null_conditional_expression`, not the original type of the member being invoked:
 
-- If the type is a known value type then it has the null state "maybe null"
-- Else if the type is a type parameter then it has the null state "maybe default"
+- If the type is a *nullable* value type then it has the null state "maybe null"
+- Else if the type is a *nullable* type parameter then it has the null state "maybe default"
 - Else it has the null state "maybe null"
 
 ### Cast expressions
@@ -417,7 +418,7 @@ If a cast expression `(T)E` invokes a user-defined conversion, then the null sta
 
 ### Unary and binary operators
 
-If a unary or binary operator invokes an user-defined operator that is declared with one or more attributes for special null behavior, the null state is determined by those attributes. Otherwise the null state of the expression is the default null state of its type.
+If a unary or binary operator invokes an user-defined operator then the null state of the expression is the default null state for the type of the user-defined operator. Otherwise it is the null state of the expression.
 
 ***Something special to do for binary `+` over strings and delegates?***
 
--- a/proposals/csharp-9.0/nullable-reference-types-specification.md
+++ b/proposals/csharp-9.0/nullable-reference-types-specification.md
@@ -250,9 +250,27 @@ The nullability of a type argument or of a constraint does not impact whether th
 
 Every expression in a given source location has a *null state*, which indicated whether it is believed to potentially evaluate to null. The null state is either "not null", "maybe null", or "maybe default". The null state is used to determine whether a warning should be given about null-unsafe conversions and dereferences.
 
+The distinction between "maybe null" and "maybe default" is subtle and applies to type parameters. The distinction is that a type parameter `T` which has the state "maybe null" means the value is in the domain of legal values for `T` however that legal value may include `null`. Where as a "maybe default" means that the value may be outside the legal domain of values for `T`. 
+
+Example: 
+
+```c#
+// The value `t` here has the state "maybe null". It's possible for `T` to be instantiated
+// with `string?` in which case `null` would be within the domain of legal values here. The 
+// assumption though is the value provided here is within the legal values of `T`. Hence 
+// if `T` is `string` then `null` will not be a value, just as we assume that `null` is not
+// provided for a normal `string` parameter
+void M<T>(T t)
+{
+    // There is no guarantee that default(T) is within the legal values for T hence the 
+    // state *must* be "maybe-default" and hence `local` must be `T?`
+    T? local = default(T);
+}
+```
+
 ### Null tracking for variables
 
-For certain expressions denoting variables or properties, the null state is tracked between occurrences, based on assignments to them, tests performed on them and the control flow between them. This is similar to how definite assignment is tracked for variables. The tracked expressions are the ones of the following form:
+For certain expressions denoting variables, fields or properties, the null state is tracked between occurrences, based on assignments to them, tests performed on them and the control flow between them. This is similar to how definite assignment is tracked for variables. The tracked expressions are the ones of the following form:
 
 ```antlr
 tracked_expression
@@ -275,7 +293,11 @@ The null state of an expression is derived from its form and type, and from the 
 
 ### Literals
 
-The null state of a `null` and `default` literals is "maybe default". The null state of any other literal is "not null".
+The null state of a `null` literal is "maybe null". 
+
+Teh null state of a `default` literal depends on the target type of the `default` literal. A `default` literal with target type `T` has the same null state as the `default(T)` expression.
+
+The null state of any other literal is "not null".
 
 ### Simple names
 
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
 
 The null state of a query expression is the default null state of its type.
 
+*Additional work needed here*
+
 ### Assignment operators
 
 `E1 = E2` and `E1 op= E2` have the same null state as `E2` after any implicit conversions have been applied.
 
-### Unary and binary operators
```
jaredpar, line 338: Just moved this closer to the cast section as I think they are going to have similar language when we are done here.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -275,7 +293,11 @@ The null state of an expression is derived from its form and type, and from the
 
 ### Literals
 
-The null state of a `null` and `default` literals is "maybe default". The null state of any other literal is "not null".
+The null state of a `null` literal is "maybe null". 
+
+Teh null state of a `default` literal depends on the target type of the `default` literal. A `default` literal with target type `T` has the same null state as the `default(T)` expression.
```
cston, line 298: >Teh [](start = 0, length = 3)

Typo

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+if (person.FirstName is not null)
+{
+    Use(person.FirstName);
+}
+
+// The return of an invocation is not a tracked expression hence the member_access
+// of the return is also not tracked
+if (GetPerson().FirstName is not null)
+{
+    Use(GetPerson().FirstName);
```
cston, line 324: >; [](start = 30, length = 1)

Consider adding `// warning: argument may be null`

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+    // ...
+}
+
+public class Person
+{
+    public string? FirstName { get; set; }
+    public string? LastName { get; set; }
+
+    private static Person s_anonymous = new Person();
+    public static Person GetAnonymous() => s_anonymous;
```
cston, line 338: >GetAnonymous [](start = 25, length = 12)

Should this name match `GetPerson()` above?

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+```
+
 ### Base access
 
 If `B` denotes the base type of the enclosing type, `base.I` has the same null state as `((B)this).I` and `base[E]` has the same null state as `((B)this)[E]`.
 
 ### Default expressions
 
-`default(T)` has the null state "not null" if `T` is known to be a nonnullable value type. Otherwise it has the null state "maybe default".
+The `default(T)` has the null state based on the properties of the type `T`:
```
cston, line 396: >The [](start = 0, length = 3)

Consider dropping "The" unless this refers to a specific `default(T)`.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+A `null_conditional_expression` has the null state based on the properties of expression type:
+
+- If the type is a known value type then it has the null state "maybe null"
+- Else if the type is a type parameter then it has the null state "maybe default"
+- Else it has the null state "maybe null"
 
 ### Cast expressions
 
-If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for its type. Otherwise, if `T` is *nullable* then the null state is "maybe null". Otherwise the null state is the same as the null state of `E`.
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for the type of the user-defined conversion. Otherwise:
```
cston, line 412: >If a cast expression `(T)E` [](start = 0, length = 27)

It looks like this section assumes `E` is an expression that is maybe null (or maybe default). If so, we should make that explicit.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
 ### Base access
 
 If `B` denotes the base type of the enclosing type, `base.I` has the same null state as `((B)this).I` and `base[E]` has the same null state as `((B)this)[E]`.
 
 ### Default expressions
 
-`default(T)` has the null state "not null" if `T` is known to be a nonnullable value type. Otherwise it has the null state "maybe default".
+The `default(T)` has the null state based on the properties of the type `T`:
+
+- If the type is a known value type then it has the null state "not null"
```
cston, line 398: >If the type is a known value type then it has the null state "not null" [](start = 2, length = 71)

If the type is a non-nullable value type, then it has the null state "not null". If the type is a nullable value type, it has the null state "maybe null".

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+- Else it has the null state "maybe null"
 
 ### Cast expressions
 
-If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for its type. Otherwise, if `T` is *nullable* then the null state is "maybe null". Otherwise the null state is the same as the null state of `E`.
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for the type of the user-defined conversion. Otherwise:
 
-***This needs upddating***
+- If `T` is a *nullable* type in the form `U?` where `U` is a type parameter then the null state is "maybe default"
+- Else If `T` is a *nullable* type then the null state is "maybe null"
```
cston, line 415: If `T` is non-nullable value type, then it has the null state "not null".
Else if `T` is a nullable value type, it has the null state "maybe null".
Else if `T` is type parameter, then it has the null state "maybe default". (That is not quite what is implemented. The implementation uses "maybe null" when the type parameter is a reference type.)

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+- Else if the type is a type parameter then it has the null state "maybe default"
+- Else it has the null state "maybe null"
 
 ### Null-conditional expressions
 
-A `null_conditional_expression` has the null state "maybe null".
+A `null_conditional_expression` has the null state based on the properties of expression type:
+
+- If the type is a known value type then it has the null state "maybe null"
+- Else if the type is a type parameter then it has the null state "maybe default"
```
cston, line 407: If the type is a non-nullable value type, then it has the null state "not null".
Else if the type is a nullable value type, then it has the null state "maybe null".
Else if the type is type parameter, then it has the null state "maybe default". (That is not quite what is implemented. The implementation uses "maybe null" when the type parameter is a reference type.)

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -275,7 +293,11 @@ The null state of an expression is derived from its form and type, and from the
 
 ### Literals
 
-The null state of a `null` and `default` literals is "maybe default". The null state of any other literal is "not null".
+The null state of a `null` literal is "maybe null".
```
RikkiGibson, line 296: I don't think this is true in the case of `T M<T>() where T : class? => null;`

proposals/csharp-9.0/nullable-reference-types-specification.md
RikkiGibson, line 398:
```suggestion
- If the type is a known non-nullable value type then it has the null state "not null"
```

For example, `default(int?)` has the null state "maybe null".

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: I did not have that assumption here.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+// The result of an invocation_expression is not tracked
+if (GetText() is not null)
+{
+    // Warning: Converting null literal or possible null value to non-nullable type.
+    string s = GetText();
+    // Warning: Dereference of a possibly null reference.
+    Use(s);
+}
+
+// Nullable friendly pattern
```
jcouv, line 359: Not sure what is the purpose/intent for this example. The TODO list includes having a section on null and pure null tests. Patterns could be covered there.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
 ### Null-conditional expressions
 
-A `null_conditional_expression` has the null state "maybe null".
+A `null_conditional_expression` has the null state based on the properties of expression type:
+
+- If the type is a known value type then it has the null state "maybe null"
+- Else if the type is a type parameter then it has the null state "maybe default"
+- Else it has the null state "maybe null"
 
 ### Cast expressions
```
jcouv, line 410: For casts that don't involve user-defined conversions, there are two rules worth mentioning:
- warning for casting a dangerous value to a non-nullable type `(string)null`
- if the type is nullable, then the state of the cast expression is maybe-null/default. `(string?)""` produces a maybe-null state

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,154 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for the type of the user-defined conversion. Otherwise:
 
-***This needs upddating***
+- If `T` is a *nullable* type in the form `U?` where `U` is a type parameter then the null state is "maybe default"
+- Else If `T` is a *nullable* type then the null state is "maybe null"
+- Else the null state is the same as the null state of `E`
+
+### Unary and binary operators
+
+If a unary or binary operator invokes an user-defined operator that is declared with one or more attributes for special null behavior, the null state is determined by those attributes. Otherwise the null state of the expression is the default null state of its type.
```
jcouv, line 420: >declared with one or more attributes for special null behavior [](start = 71, length = 62)

I'm not sure why mention attributes here, but not in the user-defined conversion section above. I would suggest to keep attributes out to a separate section, saying that they can modify return types of methods and operators.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Good catch.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Yep I didn't update that correctly.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: The intent of this example is to demonstarte that method invocations are different than properties / fields in that their results are not tracked.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Chose slightly different formatting to match other parts of the spec but yes agree

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: The first "If" should be "maybe null" though correct? The return of a `?.` where the return type is a *nonnullable* value type will be `Nullable<T>`

jaredpar: Think we're confusing which type we are talking about here. In a `?.` expression there is no possibility of a *nonnullable* value type. I will make the spec be clearer that we are referring to the type of the null conditional expression, not the method being invoked in the conditional.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: @cston and I chatted offline and cleared up that this isn't an assumption here.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: I agree. We should keep this section free of discussion about attributes and have a separate section detailing how attributes impact the return types of all methods. That can then be generally applied to every case where we discuss return and parameter types elsewhere in the spec.

jaredpar:
Thanks @jcouv, @cston and @RikkiGibson. Updated based on your feedback.

proposals/csharp-9.0/nullable-reference-types-specification.md
jcouv: I think that's already achieved by the previous sample (under `// The result of an invocation_expression is not tracked`)

---
In reply to: [521708462](https://github.com/dotnet/csharplang/pull/4126#discussion_r521708462) [](ancestors = 521708462)

jcouv (approved):

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+`default(T)` has the null state based on the properties of the type `T`:
+
+- If the type is a *nonnullable* type then it has the null state "not null"
+- Else if the type is a type parameter then it has the null state "maybe default"
+- Else it has the null state "maybe null"
 
 ### Null-conditional expressions
 
-A `null_conditional_expression` has the null state "maybe null".
+A `null_conditional_expression` has the null state based on the properties of expression type. Note that this refers to the type of the `null_conditional_expression`, not the original type of the member being invoked:
```
cston, line 405: >based on the properties of expression type [](start = 51, length = 42)

"based on the properties of ..." is a little confusing. How about just "based on the expression type"?

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+- Else it has the null state "maybe null"
 
 ### Cast expressions
 
-If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for its type. Otherwise, if `T` is *nullable* then the null state is "maybe null". Otherwise the null state is the same as the null state of `E`.
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for the type of the user-defined conversion. Otherwise:
+
+- If `T` is a *nullable* type in the form `U?` where `U` is a type parameter then the null state is "maybe default"
+- Else If `T` is a *nullable* type then the null state is "maybe null"
+- Else the null state is the same as the null state of `E`
```
cston, line 417: >Else [](start = 2, length = 4)

Else if `T` is a non-nullable value type, the null state is "not null".

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+- Else if the type is a *nullable* type parameter then it has the null state "maybe default"
+- Else it has the null state "maybe null"
 
 ### Cast expressions
 
-If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for its type. Otherwise, if `T` is *nullable* then the null state is "maybe null". Otherwise the null state is the same as the null state of `E`.
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state of the expression is the default null state for the type of the user-defined conversion. Otherwise:
+
+- If `T` is a *nullable* type in the form `U?` where `U` is a type parameter then the null state is "maybe default"
+- Else If `T` is a *nullable* type then the null state is "maybe null"
```
cston, line 416: >If [](start = 7, length = 2)

"if" (lowercase "i")

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Cleared this up offline with @cston. The issue here to think through is unboxing conversions. That necessitates the first rule here.

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+***Something special to do for binary `+` over strings and delegates?***
 
 ### Await expressions
 
 The null state of `await E` is the default null state of its type.
 
 ### The `as` operator
 
-An `as` expression has the null state "maybe null".
+The null state of an `E as T` expression depends first on properties of the type `T`. If the type *nonnullable* then the null state is "not null". Otherwise the null state depends on the conversion from the type of `E` to type `T`:
```
cston, line 431: >type *nonnullable* [](start = 93, length = 18)

"type of `T` is *nonnullable*"

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
+
+- If the conversion Identity, boxing, implicit reference or implicit nullable has the same null state as `E`
```
cston, line 433: >If the conversion Identity, boxing, implicit reference or implicit nullable has the same null state as `E` [](start = 2, length = 106)

Perhaps "If the conversion is an identity, boxing, implicit reference, or implicit nullable conversion, then the null state is the null state of `E`"

cston, line 433: >If the conversion Identity, boxing, implicit reference or implicit nullable has the same null state as `E` [](start = 2, length = 106)

Perhaps "If the conversion is an identity, implicit reference, or implicit nullable conversion, then the null state is the null state of `E`"

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
 If `B` denotes the base type of the enclosing type, `base.I` has the same null state as `((B)this).I` and `base[E]` has the same null state as `((B)this)[E]`.
 
 ### Default expressions
 
-`default(T)` has the null state "not null" if `T` is known to be a nonnullable value type. Otherwise it has the null state "maybe default".
+`default(T)` has the null state based on the properties of the type `T`:
+
+- If the type is a *nonnullable* type then it has the null state "not null"
+- Else if the type is a type parameter then it has the null state "maybe default"
+- Else it has the null state "maybe null"
```
RikkiGibson, line 401: - If the type is a *non-nullable value type* then it has the null state "not null"
- Else if the type is a type parameter that may accept either nullable reference type or non-nullable reference type arguments then it has the null state "maybe default"
- Else it has the null state "maybe null"

(it is feeling like we might want to have a shorthand for such type parameters which require this third state. Call them Heisenparameters if you want.)

proposals/csharp-9.0/nullable-reference-types-specification.md
```diff
@@ -285,62 +307,155 @@ If a `simple_name` is not classified as a value, its null state is "not null". O
...
 
 ### Null-conditional expressions
```
RikkiGibson, line 403: Consider using "?. (null-conditional) expression" as the section title a la [the rules for definite assignment](https://github.com/dotnet/csharplang/blob/master/spec/variables.md#-null-coalescing-expressions) to remind readers what this means :)

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Agreed but I wanted to try and keep every section as self contained as possible. Wanted to basically have supporting examples, when relevant, for the relevant sections.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Why do you think me second bullet is incorrect here? Every case I could think of has "maybe default". The restrictions you added aren't necessary as far as I can see. Can you add a counter example to show why this isneeded?

Note: I've chosen *nonnullable* over *non-nullable* as that is the spelling the "Nullability of types" section uses.

proposals/csharp-9.0/nullable-reference-types-specification.md
jaredpar: Clarified this.

proposals/csharp-9.0/nullable-reference-types-specification.md
RikkiGibson: re: bullet 1: `string` is a nonnullable reference type, yet the flow state of `default(string)` is maybe-null. Maybe I overlooked the definition of *nonnullable type*?

re: bullet 2:
```cs
#nullable enable
void M<T>() where T : class
{
    T? t = default(T); // has maybe-null state
}
```

In the above scenario, we don't give a "maybe-default" state because there is no need to distinguish "maybe null, but is a valid value of type T" from "maybe null, and possibly an invalid value of type T" (maybe-default). Since we don't allow, for example, type parameter constraints that allows `string?` but not `string`, we could say that this state is required for any type parameter which accepts a nullable reference type argument.

Alternatively, we could collapse things down to say that "maybe default" states can be implicitly raised to "maybe null" states when the type doesn't require the usage of a "maybe default" state, but in this case, we might as well eliminate bullet 2 entirely and say in bullet 3: *else it has the null state "maybe default"*.

cston (approved):

PR #4233, opened 2020-12-15
Fix typo 'retconn'

jnm2:
'Retcon' is an abbreviation for "retroactive continuity." The word 'continuity' doesn't start 'conn.'

A second 'n' would be added as usual for 'retconned' and 'retconning.'


333fred:
> A second 'n' would be added as usual for 'retconned' and 'retconning.'

Or, as in this case, when accidentally hitting 'n' twice 😉

PR #4234, opened 2020-12-15
Disallow record with ambiguous constructors

jcouv:


proposals/csharp-9.0/records.md
```diff
@@ -370,6 +370,8 @@ ordinary constructors today.
 
 A warning is produced if a parameter of the primary constructor is not read.
 
+An error is produced if the parameter list has a single parameter and its type is the record type.
```
AlekseyTs, line 373: I do not think any spec changes are required. We are fixing a bug of failing to detect declaration of methods with duplicate signature. No need to state that explicitly.

PR #4307, opened 2021-01-06
Spec for record structs

jcouv:
We'll queue this up for discussion in LDM in the coming weeks.

My biggest concern is with allowing field initializers when a primary constructor (ie. with parameters) is present, since default values for the fields are still possible.
This and other smaller open questions are listed at the bottom of the document.

FYI @dotnet/roslyn-compiler @MadsTorgersen @BillWagner @dotnet/csharplangdesign

### DIFF
--- a/proposals/record-structs.md
+++ b/proposals/record-structs.md
@@ -301,4 +301,5 @@ See https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-0
 - why did we disallow unsafe types in instance fields in records? I assume we also want to disallow in record structs.
 - `with` on generics? (may affect the design for record structs)
 - should we also allow `record ref struct`?
-
+- how are we going to allow user-defined positional init-only properties?
+- should we also allow `readonly record struct`?
--- a/proposals/record-structs.md
+++ b/proposals/record-structs.md
@@ -304,5 +304,5 @@ See https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-0
 - confirm that we want to disallow members named "Clone".
 - why did we disallow unsafe types in instance fields in records? I assume we also want to disallow in record structs.
 - `with` on generics? (may affect the design for record structs)
-- confirm we won't allow `record ref struct`.
-
+- confirm we won't allow `record ref struct` (issue with `IEquatable<RefStruct>` and ref fields)
+- confirm implementation of equality members. Alternative is that synthesized `bool Equals(R other)`, `bool Equals(object? other)` and operators all just delegate to `ValueType.Equals`.
--- a/proposals/record-structs.md
+++ b/proposals/record-structs.md
@@ -77,7 +77,7 @@ The record struct includes a synthesized override equivalent to a method declare
 public override bool Equals(object? obj);
 ```
 It is an error if the override is declared explicitly. 
-The synthesized override returns `other is R && Equals((R)other)` where `R` is the record struct.
+The synthesized override returns `other is R temp && Equals(temp)` where `R` is the record struct.
 
 The record struct includes a synthesized override equivalent to a method declared as follows:
 ```C#
@@ -100,8 +100,8 @@ struct R1 : IEquatable<R1>
 {
     public T1 P1 { get; set; }
     public T1 P1 { get; set; }
-    public override bool Equals(object? obj) => obj is R1 && Equals((R1)obj);
-    public virtual bool Equals(R1 other)
+    public override bool Equals(object? obj) => obj is R1 temp && Equals(temp);
+    public bool Equals(R1 other)
     {
         return
             EqualityComparer<T1>.Default.Equals(P1, other.P1) &&
@@ -162,7 +162,7 @@ struct R1 : IEquatable<R1>
     public T1 P1 { get; set; }
     public T2 P2 { get; set; }
 
-    protected virtual bool PrintMembers(StringBuilder builder)
+    private bool PrintMembers(StringBuilder builder)
     {
         builder.Append(nameof(P1));
         builder.Append(" = ");
@@ -237,14 +237,14 @@ For a record struct:
 
 * A public `get` and `set` auto-property is created.
   An inherited `abstract` property with matching type is overridden.
-  It is an error if the inherited property does not have `public` overridable `get` and `set` accessors.
+  It is an error if the inherited property does not have `public` `get` and `set` accessors.
   The auto-property is initialized to the value of the corresponding primary constructor parameter.
   Attributes can be applied to the synthesized auto-property and its backing field by using `property:` or `field:`
   targets for attributes syntactically applied to the corresponding record struct parameter.  
 
 ### Deconstruct
 
-A positional record struct with at least one parameter synthesizes a public void-returning instance method called Deconstruct with an out
+A positional record struct with at least one parameter synthesizes a public void-returning instance method called `Deconstruct` with an out
 parameter declaration for each parameter of the primary constructor declaration. Each parameter
 of the Deconstruct method has the same type as the corresponding parameter of the primary
 constructor declaration. The body of the method assigns each parameter of the Deconstruct method
@@ -300,4 +300,5 @@ See https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-0
 - confirm that we want to disallow members named "Clone".
 - why did we disallow unsafe types in instance fields in records? I assume we also want to disallow in record structs.
 - `with` on generics? (may affect the design for record structs)
+- should we also allow `record ref struct`?
 
--- a/proposals/record-structs.md
+++ b/proposals/record-structs.md
@@ -24,6 +24,8 @@ See https://github.com/dotnet/csharplang/blob/master/spec/structs.md
 But instance field declarations for a record struct are permitted to include variable initializers when there is a primary constructor.
 (TODO: need to confirm, since default values still possible)
 
+Record structs cannot use `ref` modifier.
+
 At most one partial type declaration of a partial record struct may provide a `parameter_list`.
 The `parameter_list` may not be empty.
 
@@ -235,9 +237,11 @@ member whose name and type are taken from the value parameter declaration.
 
 For a record struct:
 
-* A public `get` and `set` auto-property is created.
+* A public `get` and `init` auto-property is created if the record struct has `readonly` modifier, `get` and `set` otherwise.
+  Both kinds of set accessors (`set` and `init`) are considered "matching". So the user may declare an init-only property
+  in place of a synthesized mutable one.
   An inherited `abstract` property with matching type is overridden.
-  It is an error if the inherited property does not have `public` `get` and `set` accessors.
+  It is an error if the inherited property does not have `public` `get` and `set`/`init` accessors.
   The auto-property is initialized to the value of the corresponding primary constructor parameter.
   Attributes can be applied to the synthesized auto-property and its backing field by using `property:` or `field:`
   targets for attributes syntactically applied to the corresponding record struct parameter.  
@@ -300,6 +304,5 @@ See https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-0
 - confirm that we want to disallow members named "Clone".
 - why did we disallow unsafe types in instance fields in records? I assume we also want to disallow in record structs.
 - `with` on generics? (may affect the design for record structs)
-- should we also allow `record ref struct`?
-- how are we going to allow user-defined positional init-only properties?
-- should we also allow `readonly record struct`?
+- confirm we won't allow `record ref struct`.
+
--- a/proposals/record-structs.md
+++ b/proposals/record-structs.md
@@ -306,4 +306,5 @@ public record Derived(int Field);
 - confirm we won't allow `record ref struct` (issue with `IEquatable<RefStruct>` and ref fields)
 - confirm implementation of equality members. Alternative is that synthesized `bool Equals(R other)`, `bool Equals(object? other)` and operators all just delegate to `ValueType.Equals`.
 - confirm that we want to allow field initializers when there is a primary constructor. Do we also want to allow parameterless struct constructors while we're at it (the Activator issue was apparently fixed)?
+- how much do we want to say about `Combine` method?
 
--- a/proposals/record-structs.md
+++ b/proposals/record-structs.md
@@ -1,0 +1,301 @@
+# Record structs
+
+The syntax for a record struct is as follows:
+
+```antlr
+struct_record_declaration
+    : attributes? struct_modifier* 'partial'? 'record' 'struct' identifier type_parameter_list?
+      parameter_list? struct_interfaces? type_parameter_constraints_clause* record_struct_body
+    ;
+
+record_struct_body
+    : struct_body
+    | ';'
+    ;
+```
+
+Record struct types are value types, like other struct types. They implicitly inherit from the class `System.ValueType`.
+The modifiers and members of a record struct are subject to the same restrictions as those of structs
+(accessibility on type, modifiers on members, parameterless instance constructors,
+`base(...)` instance constructor initializers, definite assignment for `this` in constructor, destructors, ...).
+
+See https://github.com/dotnet/csharplang/blob/master/spec/structs.md
+
+But instance field declarations for a record struct are permitted to include variable initializers when there is a primary constructor.
+(TODO: need to confirm, since default values still possible)
+
+At most one partial type declaration of a partial record struct may provide a `parameter_list`.
+The `parameter_list` may not be empty.
+
+Record struct parameters cannot use `ref`, `out` or `this` modifiers (but `in` and `params` are allowed).
+
+## Members of a record struct
+
+In addition to the members declared in the record struct body, a record struct type has additional synthesized members.
+Members are synthesized unless a member with a "matching" signature is declared in the record struct body or
+an accessible concrete non-virtual member with a "matching" signature is inherited.
+Two members are considered matching if they have the same
+signature or would be considered "hiding" in an inheritance scenario.
+It is an error for a member of a record struct to be named "Clone".
+It is an error for an instance field of a record struct to have an unsafe type.
+
+A record struct is not permitted to declare a destructor.
+
+The synthesized members are as follows:
+
+### Equality members
+
+The synthesized equality members are similar as in a record class (`Equals` for this type, `Equals` for `object` type, `==` and `!=` operators for this type),\
+except for the lack of `EqualityContract`, null checks or inheritance.
+
+The record struct implements `System.IEquatable<R>` and includes a synthesized strongly-typed overload of `Equals(R other)` where `R` is the record struct.
+The method is `public`.
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility.
+
+If `Equals(R other)` is user-defined (not synthesized) but `GetHashCode` is not, a warning is produced.
+
+```C#
+public bool Equals(R other);
+```
+
+The synthesized `Equals(R)` returns `true` if and only if for each instance field `fieldN` in the record struct
+the value of `System.Collections.Generic.EqualityComparer<TN>.Default.Equals(fieldN, other.fieldN)` where `TN` is the field type is `true`.
+
+The record struct includes synthesized `==` and `!=` operators equivalent to operators declared as follows:
+```C#
+public static bool operator==(R r1, R r2)
+    => r1.Equals(r2);
+public static bool operator!=(R r1, R r2)
+    => !(r1 == r2);
+```
+The `Equals` method called by the `==` operator is the `Equals(R other)` method specified above. The `!=` operator delegates to the `==` operator. It is an error if the operators are declared explicitly.
+
+The record struct includes a synthesized override equivalent to a method declared as follows:
+```C#
+public override bool Equals(object? obj);
+```
+It is an error if the override is declared explicitly. 
+The synthesized override returns `other is R && Equals((R)other)` where `R` is the record struct.
+
+The record struct includes a synthesized override equivalent to a method declared as follows:
+```C#
+public override int GetHashCode();
+```
+The method can be declared explicitly.
+
+A warning is reported if one of `Equals(R)` and `GetHashCode()` is explicitly declared but the other method is not explicit.
+
+The synthesized override of `GetHashCode()` returns an `int` result of a deterministic function combining the values of `System.Collections.Generic.EqualityComparer<TN>.Default.GetHashCode(fieldN)` for each instance field `fieldN` with `TN` being the type of `fieldN`.
+
+For example, consider the following record struct:
+```C#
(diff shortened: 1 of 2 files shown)


proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+struct R1 : IEquatable<R1>
+{
+    public T1 P1 { get; init; }
+    public T2 P2 { get; init; }
+
+    protected virtual bool PrintMembers(StringBuilder builder)
+    {
+        builder.Append(nameof(P1));
+        builder.Append(" = ");
+        builder.Append(this.P1); // or builder.Append(this.P1); if P1 has a value type
```
chsienki, line 167: Am I missing something here, what is the difference when `P1` has a value type?

333fred: Should be `P1.ToString()` when a value type.

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+Record struct types are value types, like other struct types. They implicitly inherit from the class `System.ValueType`.
+The modifiers and members of a record struct are subject to the same restrictions as those of structs
+(accessibility on type, modifiers on members, parameterless instance constructors,
+`base(...)` instance constructor initializers, definite assignment for `this` in constructor, destructors, ...).
+
+See https://github.com/dotnet/csharplang/blob/master/spec/structs.md
+
+But instance field declarations for a record struct are permitted to include variable initializers when there is a primary constructor.
```
333fred, line 24: Personally, I would say no, they can't, you need to define a constructor, or use the primary constructor body (which we also need to add to the language).

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+- https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-05.md#record-struct-primary-constructor-defaults
+- https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-07.md#record-struct-syntax
+
+# Open questions
+
+- should we disallow a user-defined constructor with a copy constructor signature?
+- confirm that we want to keep PrintMembers design (separate method returning `bool`)
+- confirm that we want to disallow members named "Clone".
+- why did we disallow unsafe types in instance fields in records? I assume we also want to disallow in record structs.
+- `with` on generics? (may affect the design for record structs)
```
333fred, line 300: `record ref struct`?

333fred: And if we do allow such a thing, can you use `ref` in the primary constructor to declare a `ref` field in conjunction with the addition of `ref` fields?

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+Record struct parameters cannot use `ref`, `out` or `this` modifiers (but `in` and `params` are allowed).
+
+## Members of a record struct
+
+In addition to the members declared in the record struct body, a record struct type has additional synthesized members.
+Members are synthesized unless a member with a "matching" signature is declared in the record struct body or
+an accessible concrete non-virtual member with a "matching" signature is inherited.
+Two members are considered matching if they have the same
+signature or would be considered "hiding" in an inheritance scenario.
+It is an error for a member of a record struct to be named "Clone".
```
333fred, line 39: Are these supposed to appear as separate lines? If so, you need to either put two trailing spaces on the previous line, or separate with an extra newline.

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+It is an error for an instance field of a record struct to have an unsafe type.
+
+A record struct is not permitted to declare a destructor.
+
+The synthesized members are as follows:
+
+### Equality members
+
+The synthesized equality members are similar as in a record class (`Equals` for this type, `Equals` for `object` type, `==` and `!=` operators for this type),\
```
333fred, line 48: I thought that the intent was _not_ to synthesize an equals implementation, but instead work with the runtime to improve the implementation for all structs? https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-09-30.md#struct-equality covers this: we wanted being a record to only opt into _new_ functionality such as `IEquatable<T>`.

333fred: Much of this section seems incorrect to me based on that understanding.

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+            EqualityComparer<T2>.Default.GetHashCode(P2));
+    }
+}
+```
+
+### Printing members: PrintMembers and ToString methods
+
+The record struct includes a synthesized method equivalent to a method declared as follows:
+```C#
+private bool PrintMembers(System.Text.StringBuilder builder);
```
333fred, line 125: `PrintMembers` seems like overkill for a record struct, as the point was to support inheritance for record classes. I guess this would make it resilient if we ever wanted to add struct inheritance, but we also removed EqualityContract from equality so that's broken anyway. We should consider simplifying this since we don't need the abstraction.

proposals/record-structs.md
333fred:
```suggestion
        builder.Append(this.P1); // or builder.Append(this.P1.ToString()); if P1 has a value type
```

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+        builder.Append("}");
+        return builder.ToString();
+    }
+}
+```
+
+## Positional record struct members
+
+In addition to the above members, record structs with a parameter list ("positional records") synthesize
+additional members with the same conditions as the members above.
```
333fred, line 195: I don't know what this means. What conditions? Some of the conditions apply, some do not, and you don't reference a specific section so I don't really know what this means.

proposals/record-structs.md
333fred: Actually, thinking about this a bit more, maybe I can be convinced either way.

proposals/record-structs.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+### Properties
+
+For each record struct parameter of a record struct declaration there is a corresponding public property
+member whose name and type are taken from the value parameter declaration.
+
+For a record struct:
+
+* A public `get` and `init` auto-property is created (see separate `init` accessor specification).
```
333fred, line 236: We did not decide on `init` properties, we decided on `set` properties: https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-05.md#record-struct-primary-constructor-defaults.

333fred (changes requested):
Some of these points don't match LDM decisions and need to be revised, and I think we're complicating the generated code for implementation details on our part (ie, including PrintMembers because it would make impl easier for us, but result in worse code for the record struct).

proposals/record-structs.md
jcouv: The way I understand the notes is that `record struct` is a way of opting into `IEquatable` and also strongly-typed `Equals`. It is an all-or-nothing opt-in.
```
Next, we looked at whether we should expose new equality operators and strongly-typed Equals methods on struct types, as well as implementing IEquatable<T>. We again came to the conclusion that, for all existing struct types, it would be too costly in metadata (and a potential breaking change for exposing a strongly-typed Equals method or IEquatable<T>) to do this for all types. However, we do think that a gesture for opting into this generation would be useful. Given that, we considered whether it was useful to have these be more granular gestures, ie if a type could just opt-in to generating IEquatable<T> without the equality features. For these scenarios, we feel that the need just isn't there, and that it should be an all-or-nothing opt-in.
```

Did I misunderstand?

---
In reply to: [553000880](https://github.com/dotnet/csharplang/pull/4307#discussion_r553000880) [](ancestors = 553000880)

proposals/record-structs.md
jcouv: It's listed as an open question. I can go either way too.

---
In reply to: [553002951](https://github.com/dotnet/csharplang/pull/4307#discussion_r553002951) [](ancestors = 553002951)

proposals/record-structs.md
jcouv: Yes. I did consider it and also called it out as an open question.
The reason I kept `PrintMembers` is for more symmetry between `class` and `struct` records. You can convert from one to another with less change, and the mental model of the APIs you can customize is more consistent.

---
In reply to: [553001779](https://github.com/dotnet/csharplang/pull/4307#discussion_r553001779) [](ancestors = 553001779)

proposals/record-structs.md
jcouv: This is from the original records spec. I think it refers to the intro blob:
```
Members are synthesized unless a member with a "matching" signature is declared in the record struct body or
an accessible concrete non-virtual member with a "matching" signature is inherited.
```

Note: I've kept the inheritance here, since technically ValueType can be your own implementation.

---
In reply to: [553002512](https://github.com/dotnet/csharplang/pull/4307#discussion_r553002512) [](ancestors = 553002512)

proposals/record-structs.md
jcouv: Thanks. I'd missed that.

---
In reply to: [553003576](https://github.com/dotnet/csharplang/pull/4307#discussion_r553003576) [](ancestors = 553003576)

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+the value of `System.Collections.Generic.EqualityComparer<TN>.Default.Equals(fieldN, other.fieldN)` where `TN` is the field type is `true`.
+
+The record struct includes synthesized `==` and `!=` operators equivalent to operators declared as follows:
+```C#
+public static bool operator==(R r1, R r2)
+    => r1.Equals(r2);
+public static bool operator!=(R r1, R r2)
+    => !(r1 == r2);
+```
+The `Equals` method called by the `==` operator is the `Equals(R other)` method specified above. The `!=` operator delegates to the `==` operator. It is an error if the operators are declared explicitly.
```
CyrusNajmabadi, line 73: out of curiosity, why?  seems to go against our view that we'll defer to the user if they provide it.

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+
+The record struct includes a synthesized override equivalent to a method declared as follows:
+```C#
+public override bool Equals(object? obj);
+```
+It is an error if the override is declared explicitly. 
+The synthesized override returns `other is R && Equals((R)other)` where `R` is the record struct.
```
CyrusNajmabadi, line 80:
```suggestion
The synthesized override returns `other is R temp && Equals(temp)` where `R` is the record struct.
```

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+
+The record struct includes a synthesized override equivalent to a method declared as follows:
+```C#
+public override int GetHashCode();
+```
+The method can be declared explicitly.
+
+A warning is reported if one of `Equals(R)` and `GetHashCode()` is explicitly declared but the other method is not explicit.
+
+The synthesized override of `GetHashCode()` returns an `int` result of a deterministic function combining the values of `System.Collections.Generic.EqualityComparer<TN>.Default.GetHashCode(fieldN)` for each instance field `fieldN` with `TN` being the type of `fieldN`.
```
CyrusNajmabadi, line 90: Can/should we doc if deterministic has guarantees across multiple invocations of the compiler, or new versions of the compiler? Or if it is just deterministic within a single invocation of the compiler?

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+    }
+    public static bool operator==(R1 r1, R1 r2)
+        => r1.Equals(r2);
+    public static bool operator!=(R1 r1, R1 r2)
+        => !(r1 == r2);    
+    public override int GetHashCode()
+    {
+        return Combine(
+            EqualityComparer<T1>.Default.GetHashCode(P1),
+            EqualityComparer<T2>.Default.GetHashCode(P2));
```
CyrusNajmabadi, line 118: should we mention that 'Combine' is not defined in the spec, but just has to be deterministic within an invocation of the program?

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+
+For this record struct, the synthesized printing members would be something like:
+
+```C#
+struct R1 : IEquatable<R1>
+{
+    public T1 P1 { get; set; }
+    public T2 P2 { get; set; }
+
+    protected virtual bool PrintMembers(StringBuilder builder)
```
CyrusNajmabadi, line 165: remove protected/virtual

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Properties
+
+For each record struct parameter of a record struct declaration there is a corresponding public property
+member whose name and type are taken from the value parameter declaration.
+
+For a record struct:
+
+* A public `get` and `set` auto-property is created.
+  An inherited `abstract` property with matching type is overridden.
```
CyrusNajmabadi, line 239: doesn't apply afaict.

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+  It is an error if the inherited property does not have `public` overridable `get` and `set` accessors.
```
CyrusNajmabadi, line 240: overridable seems wrong here.

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+  The auto-property is initialized to the value of the corresponding primary constructor parameter.
+  Attributes can be applied to the synthesized auto-property and its backing field by using `property:` or `field:`
+  targets for attributes syntactically applied to the corresponding record struct parameter.  
+
+### Deconstruct
+
+A positional record struct with at least one parameter synthesizes a public void-returning instance method called Deconstruct with an out
```
CyrusNajmabadi, line 247:
```suggestion
A positional record struct with at least one parameter synthesizes a public void-returning instance method called `Deconstruct` with an out
```

proposals/record-structs.md
```diff
@@ -0,0 +1,303 @@
...
+parameter declaration for each parameter of the primary constructor declaration. Each parameter
+of the Deconstruct method has the same type as the corresponding parameter of the primary
+constructor declaration. The body of the method assigns each parameter of the Deconstruct method
+to the value from an instance member access to a member of the same name.
+The method can be declared explicitly. It is an error if the explicit declaration does not match
+the expected signature or accessibility, or is static.
+
+# Allow `with` expression on structs
+
+It is now valid for the receiver in a `with` expression to have a struct type.
```
CyrusNajmabadi, line 257: perhaps good to mention that this is not restricted to just `record struct` types.

proposals/record-structs.md
jcouv: If you mean why it is an error for declaring those operators explicitly, it's for consistency with record class types.

---
In reply to: [553562440](https://github.com/dotnet/csharplang/pull/4307#discussion_r553562440) [](ancestors = 553562440)

proposals/record-structs.md
jcouv: `System.ValueType` can be custom. The compiler will handle this scenario one way or another, so seemed okay to call out in the spec.

---
In reply to: [553564798](https://github.com/dotnet/csharplang/pull/4307#discussion_r553564798) [](ancestors = 553564798)

proposals/record-structs.md
jcouv: We didn't for records (which this spec is largely based on), but I don't remember if we discussed that question then. 
In general, I think we try to make as few promises as possible with regards to hashing. Does `Combine` make such promise (stable hash over multiple version)?

---
In reply to: [553563047](https://github.com/dotnet/csharplang/pull/4307#discussion_r553563047) [](ancestors = 553563047)

proposals/record-structs.md
jcouv: Meh. `struct type` seems clear enough (the rest of the spec says `record struct`).
Also, I don't have a good term for non-record structs. "Regular" structs, "non-record" structs, ...?

---
In reply to: [553565377](https://github.com/dotnet/csharplang/pull/4307#discussion_r553565377) [](ancestors = 553565377)

proposals/record-structs.md
jcouv: Good question for sure. 
I'll capture that as an open issue.

---
In reply to: [552998468](https://github.com/dotnet/csharplang/pull/4307#discussion_r552998468) [](ancestors = 552998468)

proposals/record-structs.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+### Properties
+
+For each record struct parameter of a record struct declaration there is a corresponding public property
+member whose name and type are taken from the value parameter declaration.
+
+For a record struct:
+
+* A public `get` and `set` auto-property is created.
```
orthoxerox, line 238: What if it's a readonly record struct? It will cause CS8341, the property will have to be get-only.

proposals/record-structs.md
CyrusNajmabadi: >  I think we try to make as few promises as possible 

Agreed.  bt i would go further.  I would explicitly state that we do *not* promise this.  As opposed to being silent on this.

proposals/record-structs.md
jcouv: Good catch. I need to think about this more.
There's also a difficulty with allowing an init-only property declaration when the expected/generated signature is `get/set`...

proposals/record-structs.md
```diff
@@ -0,0 +1,305 @@
...
+```C#
+record struct R1(T1 P1, T2 P2);
+```
+
+For this record struct, the synthesized equality members would be something like:
+```C#
+struct R1 : IEquatable<R1>
+{
+    public T1 P1 { get; set; }
+    public T1 P1 { get; set; }
```
yaakov-h, line 102: Did you mean:?
```suggestion
    public T2 P2 { get; set; }
```

proposals/record-structs.md
333fred: It's not how I understood the result of that meeting and not how I intended the notes to be written :). Making it a record exposes more surface area, but the surface area should all still be forwarding back the built-in struct equality functionality and not be synthesizing its own handling (or at least, that's what I took from our debates).

proposals/record-structs.md
333fred: I'd have it be an open question then. I can see the argument, but given that the only use for these members is to support inheritance I think it's not worth the overhead.

proposals/record-structs.md
```diff
@@ -298,4 +304,5 @@ See https://github.com/dotnet/csharplang/blob/master/meetings/2020/LDM-2020-10-0
 - confirm that we want to disallow members named "Clone".
 - why did we disallow unsafe types in instance fields in records? I assume we also want to disallow in record structs.
 - `with` on generics? (may affect the design for record structs)
+- confirm we won't allow `record ref struct`.
```
333fred, line 307: Probably worth adding a note of the elements we talked about in the team meeting today around `IEquatable<T>` and ref fields.

proposals/record-structs.md
jcouv: Yes. I already have this open question to cover this: "- confirm that we want to keep PrintMembers design (separate method returning `bool`)"

---
In reply to: [553689611](https://github.com/dotnet/csharplang/pull/4307#discussion_r553689611) [](ancestors = 553689611)

333fred (approved):

proposals/record-structs.md
```diff
@@ -0,0 +1,308 @@
...
+
+But instance field declarations for a record struct are permitted to include variable initializers when there is a primary constructor.
+(TODO: need to confirm, since default values still possible)
+
+Record structs cannot use `ref` modifier.
+
+At most one partial type declaration of a partial record struct may provide a `parameter_list`.
+The `parameter_list` may not be empty.
+
+Record struct parameters cannot use `ref`, `out` or `this` modifiers (but `in` and `params` are allowed).
```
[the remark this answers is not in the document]
jcouv: Fields cannot have pointer types (that would break our method of comparison, requiring special handling).
"It is an error for an instance field of a record struct to have an unsafe type."

---
In reply to: [557613324](https://github.com/dotnet/csharplang/pull/4307#discussion_r557613324) [](ancestors = 557613324)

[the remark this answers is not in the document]
jcouv: That's covered in the "properties" section. We grab the name and type (`in` is not part of the type).

"For each record struct parameter of a record struct declaration there is a corresponding public property member whose name and type are taken from the value parameter declaration."

---
In reply to: [557615389](https://github.com/dotnet/csharplang/pull/4307#discussion_r557615389) [](ancestors = 557615389)

proposals/record-structs.md
```diff
@@ -0,0 +1,308 @@
...
+The synthesized members are as follows:
+
+### Equality members
+
+The synthesized equality members are similar as in a record class (`Equals` for this type, `Equals` for `object` type, `==` and `!=` operators for this type),\
+except for the lack of `EqualityContract`, null checks or inheritance.
+
+The record struct implements `System.IEquatable<R>` and includes a synthesized strongly-typed overload of `Equals(R other)` where `R` is the record struct.
+The method is `public`.
+The method can be declared explicitly. It is an error if the explicit declaration does not match the expected signature or accessibility.
```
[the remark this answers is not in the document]
jcouv: "The method" refers to at most one method that matches `Equals(R other)`. Here were essentially saying that  the return type and accessibility have to match our expectations.
So it's possible to declare some other method like `bool Equals(string s)`.

---
In reply to: [557619948](https://github.com/dotnet/csharplang/pull/4307#discussion_r557619948) [](ancestors = 557619948)

proposals/record-structs.md
```diff
@@ -0,0 +1,308 @@
...
+
+In addition to the members declared in the record struct body, a record struct type has additional synthesized members.
+Members are synthesized unless a member with a "matching" signature is declared in the record struct body or
+an accessible concrete non-virtual member with a "matching" signature is inherited.
+Two members are considered matching if they have the same
+signature or would be considered "hiding" in an inheritance scenario.
+
+It is an error for a member of a record struct to be named "Clone".
+
+It is an error for an instance field of a record struct to have an unsafe type.
```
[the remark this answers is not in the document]
jcouv: I think "unsafe type" is acceptable here. It's consistent with the rest of the spec.

The section on anonymous types (which we're following the rules of) says: "It is also a compile-time error for the expression to have an unsafe type."
[https://github.com/dotnet/csharplang/blob/a3d67d141501101af9dbb9e8192ca8b4c717e316/spec/expressions.md#anonymous-object-creation-expressions](https://github.com/dotnet/csharplang/blob/a3d67d141501101af9dbb9e8192ca8b4c717e316/spec/expressions.md#anonymous-object-creation-expressions)

On the other hand we don't call `unsafe class C { }` an "unsafe type", only a type with a modifier that introduces an unsafe context.
[https://github.com/dotnet/csharplang/blob/a3d67d141501101af9dbb9e8192ca8b4c717e316/spec/unsafe-code.md#unsafe-code](https://github.com/dotnet/csharplang/blob/a3d67d141501101af9dbb9e8192ca8b4c717e316/spec/unsafe-code.md#unsafe-code)

---
In reply to: [557618252](https://github.com/dotnet/csharplang/pull/4307#discussion_r557618252) [](ancestors = 557618252)

PR #4310, opened 2021-01-07
Function pointer type inference

333fred:
I've taken a stab at creating a spec for function pointer type inference, basing it off the rules for constructed and delegate type inference. I'm unsure that I actually got the wording right that reflects the actual rules I'm implementing in https://github.com/dotnet/roslyn/pull/50249, so careful review is appreciated.

### DIFF
--- a/proposals/csharp-9.0/function-pointers.md
+++ b/proposals/csharp-9.0/function-pointers.md
@@ -303,7 +303,7 @@ https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#exact-infer
 
 The following sub-bullet is added as a case to bullet 2:
 
-> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<V2..Vk, V1>`, and the calling convention of `V`
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<U2..Uk, U1>`, and the calling convention of `V`
 is identical to `U`, and the refness of `Vi` is identical to `Ui`.
 
 #### Lower-bound inferences
--- a/proposals/csharp-9.0/function-pointers.md
+++ b/proposals/csharp-9.0/function-pointers.md
@@ -303,9 +303,12 @@ https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#lower-bound
 
 The following case is added to bullet 3:
 
-> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a unique type `delegate*<U2..Uk, U1>` such that `U` (or, if `U` is a type parameter,
-it's effective base class or any member of its effective interface set) is identical to, inherits from (directly or indirectly), implements (directly or
-indirectly), or has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`.
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` has an implicit
+function pointer conversion to `delegate*<U2..Uk, U1>`.
+
+The first bullet of inference from `Ui` to `Vi` is modified to:
+
+> * If `Ui` is not known to be a reference type or a function pointer type then an _exact inference_ is made
 
 Then, added to the inference from `Ui` to `Vi`:
 
@@ -323,9 +326,12 @@ https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound
 
 The following case is added to bullet 3:
 
-> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and there is a unique type `delegate*<V2..Vk, V1>` and `V` a class, struct, interface, delegate type,
-or function pointer type which is identical to, inherits from (directly or indirectly), implements (directly or indirectly), or has an implicit function pointer
-conversion to `delegate*<U2..Uk, U1>`.
+> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and there is a unique type `delegate*<V2..Vk, V1>` and `V` is a function pointer type which 
+has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`.
+
+The first bullet of inference from `Ui` to `Vi` is modified to:
+
+> * If `Ui` is not known to be a reference type or a function pointer type then an _exact inference_ is made
 
 Then added to the inference from `Ui` to `Vi`:
 
--- a/proposals/csharp-9.0/function-pointers.md
+++ b/proposals/csharp-9.0/function-pointers.md
@@ -297,20 +297,28 @@ The following bullet is added between bullets 2 and 3:
 > * If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution
 of `E` with the types `T1..Tk` yields a single method with return type `U`, then a _lower-bound inference_ is made from `U` to `Tb`.
 
+#### Exact inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#exact-inferences
+
+The following sub-bullet is added as a case to bullet 2:
+
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<V2..Vk, V1>`
+
 #### Lower-bound inferences
 
 https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#lower-bound-inferences
 
 The following case is added to bullet 3:
 
-> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` has an implicit
-function pointer conversion to `delegate*<U2..Uk, U1>`.
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to
+`delegate*<U2..Uk, U1>`.
 
 The first bullet of inference from `Ui` to `Vi` is modified to:
 
 > * If `Ui` is not known to be a reference type or a function pointer type then an _exact inference_ is made
 
-Then, added to the inference from `Ui` to `Vi`:
+Then, added after the 3rd bullet of inference from `Ui` to `Vi`:
 
 > * Otherwise, if `V` is `delegate*<V2..Vk, V1>` then inference depends on the i-th parameter of `delegate*<V2..Vk, V1>`:
 >    * If V1:
@@ -326,14 +334,13 @@ https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound
 
 The following case is added to bullet 3:
 
-> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and there is a unique type `delegate*<V2..Vk, V1>` and `V` is a function pointer type which 
-has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`.
+> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and `V` is a function pointer type which is identical to `delegate*<V2..Vk, V1>`.
 
 The first bullet of inference from `Ui` to `Vi` is modified to:
 
 > * If `Ui` is not known to be a reference type or a function pointer type then an _exact inference_ is made
 
-Then added to the inference from `Ui` to `Vi`:
+Then added after the 3rd bullet of inference from `Ui` to `Vi`:
 
 > * Otherwise, if `U` is `delegate*<U2..Uk, U1>` then inference depends on the i-th parameter of `delegate*<U2..Uk, U1>`:
 >    * If U1:
--- a/proposals/csharp-9.0/function-pointers.md
+++ b/proposals/csharp-9.0/function-pointers.md
@@ -294,7 +294,7 @@ https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#output-type
 
 The following bullet is added between bullets 2 and 3:
 
-> * If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution
+> * If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1...Tk` and return type `Tb`, and overload resolution
 of `E` with the types `T1..Tk` yields a single method with return type `U`, then a _lower-bound inference_ is made from `U` to `Tb`.
 
 #### Exact inferences
@@ -334,7 +334,7 @@ Then, added after the 3rd bullet of inference from `Ui` to `Vi`:
 
 https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound-inferences
 
-The following case is added to bullet 3:
+The following case is added to bullet 2:
 
 > * `U` is a function pointer type `delegate*<U2..Uk, U1>` and `V` is a function pointer type which is identical to `delegate*<V2..Vk, V1>`, and the
 calling convention of `U` is identical to `V`, and the refness of `Ui` is identical to `Vi`.
--- a/proposals/csharp-9.0/function-pointers.md
+++ b/proposals/csharp-9.0/function-pointers.md
@@ -268,6 +268,75 @@ The better function member specification will be changed to include the followin
 
 This means that it is possible to overload on `void*` and a `delegate*` and still sensibly use the address-of operator.
 
+### Type Inference
+
+In unsafe code, the following changes are made to the type inference algorithms:
+
+#### Input types
+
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,75 @@ The better function member specification will be changed to include the followin
...
+
+#### Lower-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#lower-bound-inferences
+
+The following case is added to bullet 3:
+
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a unique type `delegate*<U2..Uk, U1>` such that `U` (or, if `U` is a type parameter,
+it's effective base class or any member of its effective interface set) is identical to, inherits from (directly or indirectly), implements (directly or
+indirectly), or has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`.
```
AlekseyTs, line 308: >or has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`. [](start = 13, length = 74)

This looks suspicious. If `V` doesn't refer to an unfixed type parameter, there is nothing to infer. Otherwise, how can we check if a conversion exists? I think we should simply perform the inference with the goal to end up with `delegate*<U2..Uk, U1>` implicitly convertible to `delegate*<V2..Vk, V1>` after the inference, if that is at all possible. Therefore, we should probably state that calling conventions should match and ref modifiers in all positions should match. And then we simply perform inference through types in each position as specified below.

AlekseyTs, line 308: > unique type `delegate*<U2..Uk, U1>` such that `U` (or, if `U` is a type parameter,
it's effective base class or any member of its effective interface set) is identical to, inherits from (directly or indirectly), implements (directly or
indirectly)

I think almost of this is applicable to a function pointer type:
- it is not a class, thus cannot be inherited from or be an effective base class
- it is not an interface, thus cannot be implemented

If I remember correctly, pointers cannot be type parameters. If that is the case, probably no need to deal with constraints at all. And we will be left with just: "`delegate*<U2..Uk, U1>` such that `U` is identical to"

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,75 @@ The better function member specification will be changed to include the followin
...
+
+#### Upper-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound-inferences
+
+The following case is added to bullet 3:
+
+> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and there is a unique type `delegate*<V2..Vk, V1>` and `V` a class, struct, interface, delegate type,
+or function pointer type which is identical to, inherits from (directly or indirectly), implements (directly or indirectly), or has an implicit function pointer
+conversion to `delegate*<U2..Uk, U1>`.
```
AlekseyTs, line 328: >`U` is a function pointer type `delegate*<U2..Uk, U1>` and there is a unique type `delegate*<V2..Vk, V1>` and `V` a class, struct, interface, delegate type,
or function pointer type which is identical to, inherits from (directly or indirectly), implements (directly or indirectly), or has an implicit function pointer
conversion to `delegate*<U2..Uk, U1>`.

Similar comments as above.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,75 @@ The better function member specification will be changed to include the followin
...
+> If `E` is an address-of method group and `T` is a function pointer type then the return type of `T` is an output type of `E` with type `T`.
+
+#### Output type inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#output-type-inferences
+
+The following bullet is added between bullets 2 and 3:
+
+> * If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution
+of `E` with the types `T1..Tk` yields a single method with return type `U`, then a _lower-bound inference_ is made from `U` to `Tb`.
```
AlekseyTs, line 298: >and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`

I am assuming that overload resolution is the one specified in "Allow address-of to target methods" section. Perhaps we should clarify that.

AlekseyTs:

proposals/csharp-9.0/function-pointers.md
333fred: I don't believe we need to be specific. We're in an unsafe context, so the overload resolution changes from before apply.

proposals/csharp-9.0/function-pointers.md
333fred: I don't believe we need to state anything about ref and calling conventions: if `delegate*<U2..Uk, U1>` is implicitly convertible to `delegate*<V2..Vk, V1>`, then all the existing conversion rules apply, including those. I do thing you're right about the rest though.

333fred: > `delegate*<U2..Uk, U1>` such that `U` is identical to

I believe this needs to be "`delegate*<U2..Uk, U1>` such that `U` has an implicit function pointer conversion to", because we want to perform lower-bound inference from each Vi to Ui, not exact inference.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,81 @@ The better function member specification will be changed to include the followin
...
+
+#### Lower-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#lower-bound-inferences
+
+The following case is added to bullet 3:
+
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` has an implicit
+function pointer conversion to `delegate*<U2..Uk, U1>`.
```
AlekseyTs, line 307: It looks like `U` is not defined and it is not clear how to determine if it has any conversion to `delegate*<U2..Uk, U1>`.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,81 @@ The better function member specification will be changed to include the followin
...
+>        * If the parameter is by reference, then an _exact inference_ is made.
+
+#### Upper-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound-inferences
+
+The following case is added to bullet 3:
+
+> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and there is a unique type `delegate*<V2..Vk, V1>` and `V` is a function pointer type which 
+has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`.
```
AlekseyTs, line 330: > and there is a unique type `delegate*<V2..Vk, V1>` and `V` is a function pointer type which 
has an implicit function pointer conversion to `delegate*<U2..Uk, U1>`.

A couple of points:
- It is not clear how to classify conversion form a type `V` which presumably still refers to unfixed type parameters. For example, how to classify conversion from `delegate*<T>` To `delegate*<SomeClass>`?
- It looks like the relationship between `delegate*<V2..Vk, V1>` and `V` is not established.

AlekseyTs:

333fred:
@AlekseyTs I believe I've addressed the feedback.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,88 @@ The better function member specification will be changed to include the followin
...
+> * If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution
+of `E` with the types `T1..Tk` yields a single method with return type `U`, then a _lower-bound inference_ is made from `U` to `Tb`.
+
+#### Exact inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#exact-inferences
+
+The following sub-bullet is added as a case to bullet 2:
+
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<V2..Vk, V1>`
```
AlekseyTs: >* `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<V2..Vk, V1>` [](start = 2, length = 115)

As we discussed offline, I think it would be god to clarify that calling convention and ref-ness should match. Same for lower an upper inference

AlekseyTs: >`delegate*<V2..Vk, V1>` [](start = 94, length = 23)

> `U` is a function pointer type `delegate*<V2..Vk, V1>`

Should this be `delegate*<U2..Uk, U1>`?

AlekseyTs:
Done with review pass (commit 5)

333fred:
@AlekseyTs addressed feedback.

AlekseyTs (approved):
LGTM (commit 6)

AlekseyTs:
This is not blocking, but I think we need to bring the proposal to LDM for an approval.

333fred:
> This is not blocking, but I think we need to bring the proposal to LDM for an approval.

I think we can do that over email, I'll send one.

333fred:
@AlekseyTs made the change we discussed offline.

AlekseyTs (approved):
LGTM (commit 7)

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
 
 This means that it is possible to overload on `void*` and a `delegate*` and still sensibly use the address-of operator.
 
+### Type Inference
+
+In unsafe code, the following changes are made to the type inference algorithms:
```
jcouv, line 273: >In unsafe code,  [](start = 0, length = 16)

Is "In unsafe code" necessary?

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
...
+
+> If `E` is an address-of method group and `T` is a function pointer type then the return type of `T` is an output type of `E` with type `T`.
+
+#### Output type inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#output-type-inferences
+
+The following bullet is added between bullets 2 and 3:
+
+> * If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution
```
jcouv, line 297: >T1..Tk [](start = 98, length = 6)

nit: T1...Tk with 3 dots (here and below) to match original spec

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
...
+
+The following case is added to bullet 3:
+
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to
+`delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
+
+The first bullet of inference from `Ui` to `Vi` is modified to:
+
+> * If `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be
+>   a function pointer type or a reference type, then an _exact inference_ is made
```
jcouv, line 321: >then an _exact inference_ is made [](start = 49, length = 33)

If you don't mind, could we walk through this together tomorrow?

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
...
+
+Then, added after the 3rd bullet of inference from `Ui` to `Vi`:
+
+> * Otherwise, if `V` is `delegate*<V2..Vk, V1>` then inference depends on the i-th parameter of `delegate*<V2..Vk, V1>`:
+>    * If V1:
+>        * If the return is by value, then a _lower-bound inference_ is made.
+>        * If the return is by reference, then an _exact inference_ is made.
+>    * If V2..Vk:
+>        * If the parameter is by value, then an _upper-bound inference_ is made.
+>        * If the parameter is by reference, then an _exact inference_ is made.
```
jcouv, line 331: >by reference [](start = 31, length = 12)

Should we distinguish `out`/`ref` from `in`? (also in upper-bound section)

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
...
+
+#### Upper-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound-inferences
+
+The following case is added to bullet 3:
```
jcouv, line 337: Should be bullet 2

proposals/csharp-9.0/function-pointers.md
jcouv: Also, I'd like to clarify what "Conversions between delegate* types is done based on their signature including the calling convention." means (line 63).

---
In reply to: [561598552](https://github.com/dotnet/csharplang/pull/4310#discussion_r561598552) [](ancestors = 561598552)

jcouv (commented):
Done with review pass (iteration 7)

proposals/csharp-9.0/function-pointers.md
333fred: Yes, all of the unsafe rules only apply in unsafe code.

proposals/csharp-9.0/function-pointers.md
333fred: > Should we distinguish out/ref from in? (also in upper-bound section)

No. `in`/`out`/`ref` are only distinguished by modopts, they must be invariant.

proposals/csharp-9.0/function-pointers.md
333fred: > If you don't mind, could we walk through this together tomorrow?

No. I can do Friday though :P

proposals/csharp-9.0/function-pointers.md
333fred: No, bullet 2 is about nullable value types. This is correct.

333fred: Ah, wait, this was upper-bounds, not lower bounds. You're correct.

jcouv (approved):
LGTM Thanks (iteration 8)

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
...
+is identical to `U`, and the refness of `Vi` is identical to `Ui`.
+
+#### Lower-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#lower-bound-inferences
+
+The following case is added to bullet 3:
+
+> * `V` is a function pointer type `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to
+`delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
```
MadsTorgersen, line 316: Why "there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to
`delegate*<U2..Uk, U1>`", not just "`U` is a function pointer type `delegate*<U2..Uk, U1>`"?

MadsTorgersen: "the calling convention of `V` is identical to `U`" is a bit imprecise. Either "`V` and `U` have identical calling conventions" or "the calling convention of `V` is identical to the calling convention of `U`"

MadsTorgersen, line 316: "the refness of `Vi` is identical to `Ui`" same comment plus make clear that it is each of the `Vi`, e.g.: "the refness of each `Vi` is identical to the refness of the corresponding `Ui`" or something like that.

MadsTorgersen: Also, is refness a word? I believe we have another term for it, though I would have to dig to find it.

proposals/csharp-9.0/function-pointers.md
```diff
@@ -268,6 +268,92 @@ The better function member specification will be changed to include the followin
...
+>        * If the parameter is by reference, then an _exact inference_ is made.
+
+#### Upper-bound inferences
+
+https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#upper-bound-inferences
+
+The following case is added to bullet 2:
+
+> * `U` is a function pointer type `delegate*<U2..Uk, U1>` and `V` is a function pointer type which is identical to `delegate*<V2..Vk, V1>`, and the
+calling convention of `U` is identical to `V`, and the refness of `Ui` is identical to `Vi`.
```
MadsTorgersen, line 340: Same comments as above, mutatis mutandum.

PR #4351, opened 2021-01-22
Update async-method-builders.md

jcouv:


proposals/async-method-builders.md
```diff
@@ -30,17 +36,31 @@ We need a way to have an individual async method opt-in to a specific builder.
...
+    {
+        /// <summary>Initializes the <see cref="AsyncMethodBuilderOverrideAttribute"/>.</summary>
+        /// <param name="builderType">The <see cref="Type"/> of the associated builder.</param>
+        public AsyncMethodBuilderOverrideAttribute(Type builderType) => BuilderType = builderType;
+
+        // for scoped application (use property for targetReturnType? have compiler check that it's provided)
+        public AsyncMethodBuilderOverrideAttribute(Type builderType, Type targetReturnType) => ...
+
+        /// <summary>Gets the <see cref="Type"/> of the associated builder.</summary>
+        public Type BuilderType { get; }
```
stephentoub, line 60: If we're adding that second ctor above (which is more of a P1), we'd probably also want a property here returning it.

proposals/async-method-builders.md
```diff
@@ -80,52 +100,131 @@ static ValueTask<int> ExampleAsync()
...
 ```C#
-public AsyncMethodBuilderAttribute(Type builderType, params object[] createArguments);
+public struct AsyncPooledBuilderWithSize4
+{
+    AsyncPooledBuilder Create() => AsyncPooledBuilder.Create(4);
+}
 ```
-If any such arguments are specified, the compiler would expect the builder to have a `Create` method that could bind with those arguments, e.g. if a developer used:
+
+Note that we need the emitted code to allow a different type being returned from `Create` method:
```
stephentoub, line 126: I think we should call this out as its own thing.  It's useful regardless of which of these options is chosen, basically just making it possible to author a builder factory.

proposals/async-method-builders.md
```diff
@@ -80,52 +100,131 @@ static ValueTask<int> ExampleAsync()
...
-Such support would enable custom builders to be parameterized per call site, without requiring the builder to perform complicated and expensive reflection.
 
 #### P2: Enable at the module (and type?) level as well
 
 A developer that wants to using a specific custom builder for all of their methods can do so by putting the relevant attribute on each method.  But we could also enable attributing at the module or type level, in which case every relevant method within that scope would behave as if it were directly annotated, e.g.
 ```C#
-[module: PoolingAsyncValueTaskMethodBuilder]
-[module: PoolingAsyncValueTaskMethodBuilder<>]
+[module: AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder), typeof(ValueTask)]
+[module: AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder<>), typeof(ValueTask<>)]
```
stephentoub, line 227: I'd move the second ctor from earlier to be part of this section.

stephentoub (approved):

PR #4438, opened 2021-02-12
Fix Typo

hugoppp:
iff -> if


theunrepentantgeek:
"Iff" may be what's intended - it means "if and only if"

333fred (approved):

333fred:
> "Iff" may be what's intended - it means "if and only if"

This proposal was written by @jaredpar: best to assume all typos are actually typos 😉

333fred:
Thanks @hugoppp!

PR #4448, opened 2021-02-17
Remove static delegates proposal

333fred:
The static delegates proposal was superceded by function pointers.


333fred:
@jaredpar for review.

jaredpar:
Why are we deleting vs. moving to rejected or archive? That would be my expectation.

sharwell:
If we keep it in a rejected folder, it would be good to have a link to the new one:

> 🔗 Superseded by [issue]

333fred:
> Why are we deleting vs. moving to rejected or archive? That would be my expectation.

Because I forgot we had such a folder.

333fred:
@jaredpar updated.

jaredpar (approved):

PR #4449, opened 2021-02-18
Add proposal for Global Using Directive

AlekseyTs:


333fred (approved):
Might want to add some of the open bikeshedding questions, like "Should using or global be first", but otherwise looks good to me as a first checkin. At some point we should also add the rest of the proposal format (motivation, discussion of alternatives, etc).

Issue #338, opened 2017-03-24
Declaration of ref/out parameters in lambdas without typename


gafter:
@ViIvanov commented on [Sat Feb 07 2015](https://github.com/dotnet/roslyn/issues/303)

Hello! 

I have a little suggestion for C# language. 
Let us have delegate like this:

``` cs
delegate T Parse<T>(string text);
```

and we want to create an instance:

``` cs
Parse<int> parse = text => Int32.Parse(text);
```

All is OK. What about ref/out parameters in a delegate?

``` cs
delegate bool TryParse<T>(string text, out T result);
```

when we want to create an instance…

``` cs
TryParse<int> parse1 = (string text, out int result) => Int32.TryParse(text, out result);
```

…we shoud to specify types on parameters.

Why is this required? What about a syntax below:

``` cs
TryParse<int> parse2 = (text, out result) => Int32.TryParse(text, out result);
```

?

---

@mikedn commented on [Sat Feb 07 2015](https://github.com/dotnet/roslyn/issues/303#issuecomment-73400299)

In this particular case you don't need to write any of the parameter stuff, it's just:

``` C#
TryParse<int> parse1 = Int32.TryParse;
```

---

@alanfo commented on [Sun Feb 08 2015](https://github.com/dotnet/roslyn/issues/303#issuecomment-73428069)

This would be a worthwhile improvement to type inference, in my view.

As `ref` and `out` are _full_ keywords, they can't possibly be type names and so there appears no reason why the compiler won't be able to infer the type from the delegate signature.

---

@paulomorgado commented on [Sun Feb 08 2015](https://github.com/dotnet/roslyn/issues/303#issuecomment-73445227)

I'm not sure I understand what you are proposing.

Would you care to elaborate a bit more?

---

@ViIvanov commented on [Mon Feb 09 2015](https://github.com/dotnet/roslyn/issues/303#issuecomment-73494521)

@paulomorgado Of course! Let us see a code below:

``` cs
delegate T Parse<T>(string text);
delegate bool TryParse<T>(string text, out T result);

static void Main() {
  // We can create an instance of Parse<int> like this:
  Parse<int> parseHex1 = (string text) => Int32.Parse(text, NumberStyles.HexNumber);
  // or like this (and, I think, this way is more simple and more readable):
  Parse<int> parseHex2 = text => Int32.Parse(text, NumberStyles.HexNumber);

  // To create an instance of TryParse<int> delegate
[...]

PR #4451, opened 2021-02-19
Add proposal for lambda improvements

cston:
Support lambdas with attributes, explicit return type, and natural type.

### DIFF
--- a/proposals/csharp-10.0/lambda-improvements.md
+++ b/proposals/csharp-10.0/lambda-improvements.md
@@ -1,7 +1,10 @@
 ﻿# Lambda improvements
 
 ## Summary
-Support lambdas with attributes, explicit return type, and natural type.
+Proposed changes:
+1. Allow lambdas with attributes
+2. Allow lambdas with explicit return type
+3. Infer a natural delegate type for lambdas and method groups
 
 ## Motivation
 Support for attributes on lambdas would provide parity with methods and local functions.
@@ -32,13 +35,13 @@ f = [MyAttribute] delegate { return 1; };         // syntax error
 f = delegate ([MyAttribute] int x) { return x; }; // syntax error
 ```
 
-Attributes on the lambda or lambda parameters will be emitted to metadata on the method that maps to the lambda. (If the lambda does not require a closure class, the lambda is emitted as a method on the containing type. Otherwise the lambda is as a method on the generated closure class.)
+Attributes on the lambda or lambda parameters will be emitted to metadata on the method that maps to the lambda.
 
 ## Explicit return type
 An explicit return type may be specified after the parameter list.
 ```csharp
 f = () : T => default;              // () : T
-f = x : int => 1;                   // <unknown> : int
+f = x : short => 1;                 // <unknown> : short
 f = (ref int x) : ref int => ref x; // ref int : ref int
 f = static _ : void => { };         // <unknown> : void
 ```
@@ -49,10 +52,10 @@ f = delegate : int { return 1; };         // syntax error
 f = delegate (int x) : int { return x; }; // syntax error
 ```
 
-## Natural type
+## Natural delegate type
 A lambda expression has a natural type if the parameters types are explicit and either the return type is explicit or there is a common type from the natural types of all `return` expressions in the body. Otherwise there is no natural type.
 
-A method group has a natural type if the method group contains a single method and the method or reduced extension method has no unbound type parameters.
+A method group has a natural type if the method group contains a single method and the method has no unbound type parameters.
 
 Lambdas or method groups with natural types can be used as initializers in `var` declarations.
 
@@ -96,7 +99,39 @@ var fA = (ref char c) => ref c; // delegate ref char <anonymous2><char, char>(re
 var fB = (ref int i) => ref i;  // delegate ref int <anonymous2><int, int>(ref int <unnamed>);
 ```
 
-_Issue: Is it necessary to use `System.Action<>` and `System.Func<>` for the delegate types or should the compiler always generate an anonymous delegate type? There is an observable difference since the anonymous delegate types are not co- or contra-variant._
+_Issue: Should the compiler always generate anonymous delegate types rather than using `System.Action<>` and `System.Func<>`? There are observable differences: 1. the anonymous delegate types are not co- or contra-variant; and 2. using a synthesized delegate type would mean the following assignment to `y` would fail: `var x = () => { }; Action y = x;`._
+
+### Implicit conversion to `System.Delegate`
+A consequence of inferring a natural type is that lambda expressions and method groups with natural type are implicitly convertible to `System.Delegate`.
+```csharp
+static void Invoke(Func<string> f) { }
+static void Invoke(Delegate d) { }
+
+static string GetString() => "";
+static int GetInt() => 0;
+
+Invoke(() => "");  // Invoke(Func<string>)
+Invoke(() => 0);   // Invoke(Delegate) [new]
+
+Invoke(GetString); // Invoke(Func<string>)
+Invoke(GetInt);    // Invoke(Delegate) [new]
+```
+
+To avoid a breaking change, overload resolution will be updated to prefer lambda and method group conversions that do not use the natural type.
+_The example below demonstrates the tie-breaking rule for lambdas. Is there an equivalent example for method groups?_
+```csharp
+static void Execute(Expression<Func<string>> e) { }
+static void Execute(Delegate d) { }
+
+static string GetString() => "";
+static int GetInt() => 0;
+
+Execute(() => "");  // Execute(Expression<Func<string>>) [tie-breaker]
+Execute(() => 0);   // Execute(Delegate)
+
+Execute(GetString); // Execute(Delegate)
+Execute(GetInt);    // Execute(Delegate)
+```
 
 ## Syntax
 
--- a/proposals/csharp-10.0/lambda-improvements.md
+++ b/proposals/csharp-10.0/lambda-improvements.md
@@ -102,7 +102,7 @@ The delegate type for the lambda or method group and parameter types `P1, ..., P
 
 `modopt()` or `modreq()` in the method group signature are ignored in the corresponding delegate type.
 
-If synthesized delegate types are required, the compiler will attempt to reuse delegate types across multiple use sites. The compiler could generate generic delegate types parameterized by parameter types and return type similar to the generic types synthesized for anonymous types. But reuse might be limited to simply reusing delegate types when lambda or method group signatures match exactly.
+If two lambda expressions or method groups in the same compilation require synthesized delegate types with the same parameter types and modifiers and the same return type and modifiers, the compiler will use the same synthesized delegate type.
 
 Lambdas or method groups with natural types can be used as initializers in `var` declarations.
 
--- a/proposals/csharp-10.0/lambda-improvements.md
+++ b/proposals/csharp-10.0/lambda-improvements.md
@@ -13,6 +13,7 @@ Support for explicit return types would provide symmetry with lambda parameters 
 Allowing explicit return types would also provide control over compiler performance in nested lambdas where overload resolution must bind the lambda body currently to determine the signature.
 
 A natural type for lambda expressions and method groups will allow more scenarios where lambdas and method groups may be used without an explicit delegate type, including as initializers in `var` declarations.
+
 Requiring explicit delegate types for lambdas and method groups has been a friction point for customers, and has become an impediment to progress in ASP.NET with recent work on [MapAction](https://github.com/dotnet/aspnetcore/pull/29878).
 
 [ASP.NET MapAction](https://github.com/dotnet/aspnetcore/pull/29878) without proposed changes (`MapAction()` takes a `System.Delegate` argument):
@@ -59,7 +60,13 @@ f = [MyAttribute] delegate { return 1; };         // syntax error
 f = delegate ([MyAttribute] int x) { return x; }; // syntax error
 ```
 
-Attributes on the lambda or lambda parameters will be emitted to metadata on the method that maps to the lambda.
+Attributes on the lambda expression or lambda parameters will be emitted to metadata on the method that maps to the lambda.
+
+In general, customers should not depend on how lambda expressions and local functions map from source to metadata. How lambdas and local functions are emitted can, and has, changed between compiler versions.
+
+The changes proposed here are targeted at the `Delegate` driven scenario.
+It should be valid to inspect the `MethodInfo` associated with a `Delegate` instance to determine the signature of the lambda expression or local function including any explicit attributes and additional metadata emitted by the compiler such as default parameters.
+This allows teams such as ASP.NET to make available the same behaviors for lambdas and local functions as ordinary methods.
 
 ## Explicit return type
 An explicit return type may be specified after the parameter list.
@@ -84,7 +91,9 @@ The natural type is a delegate type where the parameter types are the explicit l
 - if the lambda has no return expressions, the return type is `void` or `System.Threading.Tasks.Task` if `async`;
 - if the common type from the natural type of all `return` expressions in the body is the type `R0`, the return type is `R0` or `System.Threading.Tasks.Task<R0>` if `async`.
 
-A method group has a natural type if the method group contains a single method and the method has no unbound type parameters.
+A method group has a natural type if the method group contains a single method.
+
+_A method group might refer to extension methods. Normally method group resolution searches for extension methods lazily, only iterating through successive namespace scopes until extension methods are found that match the target type. But to determine that a method group contains a single method the compiler may need to search all namespace scopes. To minimize unnecessary binding, perhaps natural type should be calculated only in cases where there is no target type - that is, only calculate the natural type in cases where it is needed._
 
 The delegate type for the lambda or method group and parameter types `P1, ..., Pn` and return type `R` is:
 - if any parameter or return value is not by value, or there are more than 16 parameters, or any of the parameter types or return are not valid type arguments (say, `(int* p) => { }`), then the delegate is a synthesized `internal` anonymous delegate type with signature that matches the lambda or method group, and with parameter names `arg1, ..., argn` or `arg` if a single parameter;
@@ -94,8 +103,6 @@ The delegate type for the lambda or method group and parameter types `P1, ..., P
 `modopt()` or `modreq()` in the method group signature are ignored in the corresponding delegate type.
 
 If synthesized delegate types are required, the compiler will attempt to reuse delegate types across multiple use sites. The compiler could generate generic delegate types parameterized by parameter types and return type similar to the generic types synthesized for anonymous types. But reuse might be limited to simply reusing delegate types when lambda or method group signatures match exactly.
-
-The anonymous delegate types will not be co- or contra-variant unlike the delegates constructed from `System.Action<>` and `System.Func<>`.
 
 Lambdas or method groups with natural types can be used as initializers in `var` declarations.
 
@@ -117,6 +124,15 @@ var f7 = "".F1; // System.Action
 var f8 = F2;    // System.Action<string> 
 ```
 
+The synthesized delegate types are implicitly co- and contra-variant.
+```csharp
+var fA = (IEnumerable<string> e, ref int i) => { }; // void DA$(IEnumerable<string>, ref int);
+fA = (IEnumerable<object> e, ref int i) => { };     // ok
+
+var fB = (IEnumerable<object> e, ref int i) => { }; // void DB$(IEnumerable<object>, ref int);
+fB = (IEnumerable<string> e, ref int i) => { };     // error: parameter type mismatch
+```
+
 ### Implicit conversion to `System.Delegate`
 A consequence of inferring a natural type is that lambda expressions and method groups with natural type are implicitly convertible to `System.Delegate`.
 ```csharp
@@ -131,6 +147,14 @@ Invoke(() => 0);   // Invoke(Delegate) [new]
 
 Invoke(GetString); // Invoke(Func<string>)
 Invoke(GetInt);    // Invoke(Delegate) [new]
+```
+
+If a natural type cannot be inferred, there is no implicit conversion to `System.Delegate`.
+```csharp
+static void Invoke(Delegate d) { }
+
+Invoke(Console.WriteLine); // error: cannot to 'Delegate'; multiple candidate methods
+Invoke(x => x);            // error: cannot to 'Delegate'; no natural type for 'x'
 ```
 
 To avoid a breaking change, overload resolution will be updated to prefer lambda and method group conversions that do not use the natural type.
@@ -166,6 +190,10 @@ lambda_parameter
   | (attribute_list* modifier* type)? identifier equals_value_clause?
   ;
 ```
+
+_Does the `: type` return type syntax introduce ambiguities with `?:` that cannot be resolved easily?_
+
+_Should we allow attributes on parameters without explicit types, such as `([MyAttribute] x) => { }`? (We don't allow modifiers on parameters without explicit types, such as `(ref x) => { }`.)_
 
 ## Design meetings
 
--- a/proposals/csharp-10.0/lambda-improvements.md
+++ b/proposals/csharp-10.0/lambda-improvements.md
@@ -1,14 +1,18 @@
 ﻿# Lambda improvements
 
 ## Summary
-
 Support lambdas with attributes, explicit return type, and natural type.
 
-## Attributes
-### Motivation
+## Motivation
 Support for attributes on lambdas would provide parity with methods and local functions.
 
-### Design
+Support for explicit return types would provide symmetry with lambda parameters where explicit types can be specified.
+Allowing explicit return types would also provide control over compiler performance in nested lambdas where overload resolution must bind the lambda body currently to determine the signature.
+
+A natural type for lambda expressions and method groups will allow more scenarios where lambdas and method groups may be used without an explicit delegate type, including as initializers in `var` declarations.
+Requiring explicit delegate types for lambdas and method groups has been a friction point for customers, and has become an impediment to progress in ASP.NET with recent work on [MapAction](https://github.com/dotnet/aspnetcore/pull/29878).
+
+## Attributes
 Attributes may be added to lambda expressions.
 ```csharp
 f = [MyAttribute] x => x;          // [MyAttribute]lambda
@@ -28,13 +32,9 @@ f = [MyAttribute] delegate { return 1; };         // syntax error
 f = delegate ([MyAttribute] int x) { return x; }; // syntax error
 ```
 
+Attributes on the lambda or lambda parameters will be emitted to metadata on the method that maps to the lambda. (If the lambda does not require a closure class, the lambda is emitted as a method on the containing type. Otherwise the lambda is as a method on the generated closure class.)
+
 ## Explicit return type
-### Motivation
-Support for explicit return types would provide symmetry with lambda parameters where explicit types can be specified.
-
-Allowing explicit return types would also provide control over compiler performance in nested lambdas where overload resolution must bind the lambda body currently to determine the signature.
-
-### Design
 An explicit return type may be specified after the parameter list.
 ```csharp
 f = () : T => default;              // () : T
@@ -50,21 +50,17 @@ f = delegate (int x) : int { return x; }; // syntax error
 ```
 
 ## Natural type
-### Motivation
-A natural type for lambda expressions and method groups will allow more scenarios where lambdas and method groups may be used without an explicit delegate type, including in `var` declarations.
-
-### Design
 A lambda expression has a natural type if the parameters types are explicit and either the return type is explicit or there is a common type from the natural types of all `return` expressions in the body. Otherwise there is no natural type.
 
 A method group has a natural type if the method group contains a single method and the method or reduced extension method has no unbound type parameters.
+
+Lambdas or method groups with natural types can be used as initializers in `var` declarations.
 
 If the lambda or method group has no more than 16 parameters and no return value, and all parameters are passed by value, the natural type will be `delegate void System.Action<P1, ..., Pn>(P1, ..., Pn)` where `P1, ..., Pn` are the lambda parameter types.
 
 If the lambda or method group has no more than 16 parameters and a by-value return type, and all parameters are passed by value, the natural type will be `delegate R System.Func<P1, ..., Pn, R>(P1, ..., Pn)` where `P1, ..., Pn` are the lambda parameter types, and `R` is the lambda return type.
 
-Otherwise the natural type will be an `internal` anonymous `delegate` type with a signature matching the lambda signature. The names of the anonymous delegate types and the names of the parameters are unspeakable.
-
-Lambdas or method groups with natural types can be used as initializers in `var` declarations.
+Otherwise the natural type will be a synthesized `internal` anonymous `delegate` type with a signature that matches the lambda or method group.
 
 ```csharp
 var f1 = () => default;        // error: no natural type
@@ -72,18 +68,35 @@ var f2 = x => { };             // error: no natural type
 var f3 = x => x;               // error: no natural type
 var f4 = () => 1;              // System.Func<int>
 var f5 = () : string => null;  // System.Func<string>
-var f6 = (ref int x) => ref x; // delegate ref int <anonymous>(ref int <unnamed>);
 ```
 
 ```csharp
 static void F1() { }
-static void F1(this string s) { }
+static void F1<T>(this T t) { }
 static void F2(this string s) { }
 
-var f7 = F1;    // error: multiple methods
-var f8 = "".F1; // System.Action
-var f9 = F2;    // System.Action<string> 
+var f6 = F1;    // error: multiple methods
+var f7 = "".F1; // System.Action
+var f8 = F2;    // System.Action<string> 
 ```
+
+### Anonymous delegate type
+If synthesized delegate types are required, the compiler generates generic anonymous delegate types that are shared across all anonymous delegates in the module that have the same number of parameters and same parameter ref kinds.
+
+The names of the synthesized delegate types and the names of the parameters are unspeakable.
+
+The anonymous delegate types are not co- or contra-variant unlike the delegates constructed from `System.Action<>` and `System.Func<>`.
+
+The natural type of a method group does not include any `modopt()` or `modreq()` on the method group.
+
+`fA` and `fB` share a common generic type `internal delegate ref R <anonymous2><P1>(ref P1 <unnamed>)` below:
+```csharp
+var f9 = (ref int i) => { };    // delegate void <anonymous1><int>(ref int <unnamed>);
+var fA = (ref char c) => ref c; // delegate ref char <anonymous2><char, char>(ref char <unnamed>);
+var fB = (ref int i) => ref i;  // delegate ref int <anonymous2><int, int>(ref int <unnamed>);
+```
+
+_Issue: Is it necessary to use `System.Action<>` and `System.Func<>` for the delegate types or should the compiler always generate an anonymous delegate type? There is an observable difference since the anonymous delegate types are not co- or contra-variant._
 
 ## Syntax
 
--- a/proposals/csharp-10.0/lambda-improvements.md
+++ b/proposals/csharp-10.0/lambda-improvements.md
@@ -93,7 +93,7 @@ The natural type is a delegate type where the parameter types are the explicit l
 
 A method group has a natural type if the method group contains a single method.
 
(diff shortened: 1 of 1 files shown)


proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+  ;
+
+lambda_parameters
+  : lambda_parameter
+  | '(' (lambda_parameter (',' lambda_parameter)*)? ')'
+  ;
+
+lambda_parameter
+  : identifier
+  | (attribute_list* modifier* type)? identifier equals_value_clause?
```
CyrusNajmabadi, line 115: couold we do: `attribute_list? (modifier* type)? identifier equals_value_clause?` instead?

it seems like `([x] a) => ...` is reasonable, and doesn't require the user to have to provide full types for lambdas that want attributes.

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+f = ([MyAttribute] int x) => x;  // [MyAttribute]x
+```
+
+Attributes are not supported for anonymous methods declared with `delegate { }` syntax.
+```csharp
+f = [MyAttribute] delegate { return 1; };         // syntax error
+f = delegate ([MyAttribute] int x) { return x; }; // syntax error
+```
+
+Attributes on the lambda or lambda parameters will be emitted to metadata on the method that maps to the lambda. (If the lambda does not require a closure class, the lambda is emitted as a method on the containing type. Otherwise the lambda is as a method on the generated closure class.)
```
333fred, line 35: Do we have existing guarantees this strong about how lambdas are emitted?

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Explicit return type
+An explicit return type may be specified after the parameter list.
+```csharp
+f = () : T => default;              // () : T
+f = x : int => 1;                   // <unknown> : int
```
333fred, line 41: Might be a better example if it's a type we couldn't infer, like `short`.

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+Explicit return types are not supported for anonymous methods declared with `delegate { }` syntax.
+```csharp
+f = delegate : int { return 1; };         // syntax error
+f = delegate (int x) : int { return x; }; // syntax error
+```
+
+## Natural type
+A lambda expression has a natural type if the parameters types are explicit and either the return type is explicit or there is a common type from the natural types of all `return` expressions in the body. Otherwise there is no natural type.
+
+A method group has a natural type if the method group contains a single method and the method or reduced extension method has no unbound type parameters.
```
333fred, line 55: Do we need to explicitly mention reduced extension methods? I think just "the method" would be fine?

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+static void F1<T>(this T t) { }
+static void F2(this string s) { }
+
+var f6 = F1;    // error: multiple methods
+var f7 = "".F1; // System.Action
+var f8 = F2;    // System.Action<string> 
+```
+
+### Anonymous delegate type
+If synthesized delegate types are required, the compiler generates generic anonymous delegate types that are shared across all anonymous delegates in the module that have the same number of parameters and same parameter ref kinds.
```
333fred, line 84: And parameter types, return type, return type ref kind.

333fred: Unless you're saying that these anonymous types are generic? Not sure from reading this.

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+
+The names of the synthesized delegate types and the names of the parameters are unspeakable.
+
+The anonymous delegate types are not co- or contra-variant unlike the delegates constructed from `System.Action<>` and `System.Func<>`.
```
333fred, line 88: This seems possibly unnecessary. If I'm reading correctly that these anonymous types are generic, it feels like we should make value params and returns contra/covariant, as appropriate to ease other friction points.

proposals/csharp-10.0/lambda-improvements.md
333fred: I'm of 2 minds on this:

1. We don't allow you specify just the `ref` today. If you want more signature than just a name, you have to have the whole signature. This decision is in line with that.
2. https://github.com/dotnet/csharplang/issues/338 is championed and would remove that restriction, so why add a similar restriction here?

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+var fB = (ref int i) => ref i;  // delegate ref int <anonymous2><int, int>(ref int <unnamed>);
+```
+
+_Issue: Is it necessary to use `System.Action<>` and `System.Func<>` for the delegate types or should the compiler always generate an anonymous delegate type? There is an observable difference since the anonymous delegate types are not co- or contra-variant._
+
+## Syntax
+
+```antlr
+lambda_expression
+  : attribute_list* modifier* lambda_parameters (':' type)? '=>' (block | body)
```
333fred, line 105: The `:` does feel like the natural thing to do, but I'm concerned that there's going to be ambiguities with `?:` expressions we have today. Are we certain this is a syntax form that we can support?

333fred: Might easier to parse if we put it inside the `()`? Like `var a = (: string) => "";`? Looks a bit ugly, but I don't believe it's ambiguous.

proposals/csharp-10.0/lambda-improvements.md
CyrusNajmabadi: Exactly :)  but even if we don't do #338, i wouldn't perpetuate this.  :)

proposals/csharp-10.0/lambda-improvements.md
CyrusNajmabadi: i hate that form @333fred  :)   

THe potential ambiguity would be something like:

`x ? (y) : ...`

is that one of these new lambdas on the RHS of the `?` or is this a parenthesized expression?  Given that this is a legal start, it might be painful to disambiguate.

--

Note: supplying the return type explicitly is one of hte least interesting parts of this proposal for me.  I wouldn't gate it on that, and i'd be happy to drop it and come up with an independent solution there.

333fred: > i hate that form @333fred :)

I didn't say I _liked_ my form either :P. I just don't think that `:` postfixes are resolvable.

proposals/csharp-10.0/lambda-improvements.md
alrz, line 55: >A method group has a natural type if the method group contains a single method

See https://github.com/dotnet/csharplang/discussions/129#discussioncomment-65296 and https://github.com/dotnet/csharplang/discussions/129#discussioncomment-65297. Does this address those issues?

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+Explicit return types are not supported for anonymous methods declared with `delegate { }` syntax.
+```csharp
+f = delegate : int { return 1; };         // syntax error
+f = delegate (int x) : int { return x; }; // syntax error
+```
+
+## Natural type
+A lambda expression has a natural type if the parameters types are explicit and either the return type is explicit or there is a common type from the natural types of all `return` expressions in the body. Otherwise there is no natural type.
```
YairHalberstadt, line 53: Should there be a natural type for async lambdas? If so, `Task` or `ValueTask`?

proposals/csharp-10.0/lambda-improvements.md
jaredpar: No we do not. This would be a change from previous versions of C#. The goal is to mirror the decisions and language we made for local functions.

proposals/csharp-10.0/lambda-improvements.md
jaredpar: Ah good catch yes there should. We should spell out the rules here in the spec to make it clear that's supported

proposals/csharp-10.0/lambda-improvements.md
jaredpar: Also should mention `async`-ness.

> Unless you're saying that these anonymous types are generic? 

In practice the implementation would be generic, just as anonymous types are today in C# and anonymous types + anonymous delegates are in VB. As for what the spec says, I would re-use the language in the C# spec for anonymous types.

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+
+```csharp
+static void F1() { }
+static void F1<T>(this T t) { }
+static void F2(this string s) { }
+
+var f6 = F1;    // error: multiple methods
+var f7 = "".F1; // System.Action
+var f8 = F2;    // System.Action<string> 
+```
```
jaredpar, line 81: One part I think we should elaborate on is that this specifically would allow us to call methods which take a `Delegate` as a parameter with arguments that are just lambda expressions. For example: 

```c#
void M(Delegate d) { ... }
M(() => 42); // okay
M(x: int => 13); // okay because natural type takes over 
```

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Anonymous delegate type
+If synthesized delegate types are required, the compiler generates generic anonymous delegate types that are shared across all anonymous delegates in the module that have the same number of parameters and same parameter ref kinds.
+
+The names of the synthesized delegate types and the names of the parameters are unspeakable.
```
halter73, line 86: I'm not sure I understand this. Does this mean that `Delegate.Method.GetParameters()` won't return `ParameterInfo`s with the actual `Name` of the parameters for lambdas requiring synthesized delegate types? The parameter names are important to MapAction(). Or is this referring to some other parameter name?

jaredpar: > The parameter names are important to MapAction(). 

Then don't we have a more fundamental problem? If the parameter names matter then any re-use of the `Action` and `Func` types will not work because we can't change the names of them.

halter73: Considering we already get preserved parameter names when we explicitly cast a method group to the appropriate `Action` or `Func` I assumed this is something that was stored in the `Action` or `Func` instance rather than baked into the types themselves.

mburbea: I think this isn't a problem as the synthesized delegate type may have generic names, but `Delegate.Method.GetParameters()` will still have the correct names.
```e.g.
Func<int,int> foo = (int horse)=>horse;
Console.WriteLine(foo.Method.GetParameters()[0].Name)
```
prints "horse".

HaloFour: > I assumed this is something that was stored in the `Action` or `Func` instance rather than baked into the types themselves.

All delegates track the runtime method handle of the target method, which is what you're getting when you reference the `Method` property.  That's how you get the parameter names of the target method instead of the placeholder parameter names of the `Func<...>` delegates themselves, e.g. `arg1`, arg2`, etc.

davidfowl: Right. This works fine, we care about the target method, not the delegate parameters.

davidfowl:
@halter73 and I were just discussion something we want to make sure is called out in the spec wrt type inference and overload resolution. Consider the following overloads:

```C#
public static class EndpointRouteBuilderExtensions
{
    public static IEndpointConventionBuilder MapGet(this IEndpointRouteBuilder routes, string pattern, Func<HttpContext, Task> requestDelegate)
    {
        return null;
    }

    public static IEndpointConventionBuilder MapGet(this IEndpointRouteBuilder routes, string pattern, Delegate @delegate)
    {
        return null;
    }
}

public class Startup
{
    public void Configure(IApplicationBuilder app)
    {
        app.UseRouting();

        app.UseEndpoints(endpoints =>
        {
            // This should pick the `Func<HttpContext, Task>` overload
            endpoints.MapGet("/hello", context => context.Response.WriteAsync("Hello World")); 

            // This should pick the Delegate overload
            endpoints.MapGet("/hello2", (int c) => Task.CompletedTask);
        });
    }
}

```

Normally before lambdas have a natural type the second overload wouldn't be possible but now it is and it should choose the right overload appropriately (though I know overloading is a source of complication in the language)

jaredpar:
@davidfow indeed, was chatting with @cston this morning about adding similar style samples to the spec to make the updates clear here

HaloFour:
Regarding natural lambda types, does that come with any relaxation when passing a lambda to a method that accepts a different delegate type that has a compatible signature?  If not it seems like you couldn't really pass those synthetically created lambdas anywhere without casting to the explicit delegate type which would make the feature a lot less useful, IMO.

davidfowl:
> Regarding natural lambda types, does that come with any relaxation when passing a lambda to a method that accepts a different delegate type that has a compatible signature? If not it seems like you couldn't really pass those synthetically created lambdas anywhere without casting to the explicit delegate type which would make the feature a lot less useful, IMO.

So to be clear, you're asking about this scenario:


```C#
void Main()
{
    var f = (Span<int> s) => s[0] + 1; // Infers:   delegate int CustomDelegate$0(Span<int> s);
    Call(f);
}

public delegate int CompatibleButDifferent(Span<int> s);

void Call(CompatibleButDifferent del)
{
    d(new byte[100]);
}
```

Feels like a much rarer scenario to me. It would matter in places where a Func/Action type couldn't be used so that means, byref types (because of generics), out and ref parameters (not sure what else besides those).

HaloFour:
> Feels like a much rarer scenario to me. It would matter in places where a Func/Action type couldn't be used 

Or with any API that doesn't use `Func<...>`/`Action<...>` delegate types, which isn't terribly uncommon.  In all but the simplest of situations I prefer to define my own delegates as I can explicitly name the parameters and offer XML documentation.

davidfowl:
> Or with any API that doesn't use Func<...>/Action<...> delegate types, which isn't terribly uncommon. In all but the simplest of situations I prefer to define my own delegates as I can explicitly name the parameters and offer XML documentation.

I get it, I don't think its as pervasive anymore since the dawn of Func and Action. I hear you on the parameter types and documentation as well (though those don't show up super well in the IDE today). ASP.NET does this today for the RequestDelegate and it means this wouldn't work:

```C#
public delegate Task RequestDelegate(HttpContext context);

void Configure(IApplicationBuilder app)
{
  var f = (HttpContext context) => context.Response.WriteAsync("Hello World"); // infers Func<HttpContext, Task>
  app.Run(f); // Fails to convert Func<HttpContext, Task> to RequestDelegate
}
```

This does seem related to the overall set of features here but could be generalized as structurally compatible delegate conversions should work. I think that goes against the mantra that delegates are type safe .

https://github.com/dotnet/runtime/issues/4331 (ah look it's your issue 😉 )

HaloFour:
@davidfowl 

> I think that goes against the mantra that delegates are type safe .

Which goes against the design of `Action<...>` and `Func<...>` in general where the only relevant aspect of the type is the signature.

Discounting the prevalence of `Action<...>`/`Func<...>` the concept makes no sense with generated unspeakable delegates as it is impossible to write an API that can accept them.  They only seem useful with APIs that accept `Delegate`, which are much less common beasts, especially such APIs designed to accept delegates that use `ref` parameters or other such things that prevent them from reusing the existing `Action<...>`/`Func<...>` delegates.

Anyway the argument against lambda inference in the past has always been that you need to know the type of the delegate so that you can actually do something with the instance of the delegate, like pass it to a method, and that the compiler doesn't (and shouldn't) have special knowledge of `Func<...>` or `Action<...>` over any other delegate.  Whether or not you agree with the second point it seems very awkward to have a language feature like this without trying to address these points specifically.

davidfowl:
> Anyway the argument against lambda inference in the past has always been that you need to know the type of the delegate so that you can actually do something with the instance of the delegate, like pass it to a method, and that the compiler doesn't (and shouldn't) have special knowledge of Func<...> or Action<...> over any other delegate. Whether or not you agree with the second point it seems very awkward to have a language feature like this without trying to address these points specifically.

Trying to solve both during the same timeframe seem fine but I don't think they are coupled and one can exist without the other.

HaloFour:
I'll also reference a previous proposal/discussion I posted on the subject of inferred lambdas: https://github.com/dotnet/csharplang/discussions/1694

What I like about the approach is that the compiler doesn't take a position on the actual type generated until it is used, at which point the compiler knows which delegate type to use, if one at all.  If the lambda is used as a delegate anywhere in the method then the compiler will emit the appropriate display class and create/cache the delegate instance by type at the point of first usage.  The cost would be a single allocation for the display class and one allocation for each delegate type referencing the lambda.  If the lambda is not used as a delegate at all then the compiler can emit the more optimized machinery for a local function instead.

jaredpar:
@HaloFour 

>...  the compiler doesn't (and shouldn't) have special knowledge of Func<...> or Action<...> over any other delegate.

The compiler already has special knowledge of `Func<>` and `Action<>` up to 16 generic arguments

http://sourceroslyn.io/#Microsoft.CodeAnalysis/WellKnownTypes.cs,3bb7bf76580e8f66

> What I like about the approach is that the compiler doesn't take a position on the actual type generated until it is used, at which point the compiler knows which delegate type to use, if one at all. 

This is essentially an entire new kind of type inference though. Essentially delay the decision of  `var` until we see how the local is actually used in production. That is a feature that would be generally useful, not just for lambdas. Consider:

```c# 
var local = 42;
M(out local); 
void M(out long l) { ... }
```

That would benefit from the same type of delayed type inference. As such it seems, while useful and **very** challenging to implement, an orthogonal concept. 

> This does seem related to the overall set of features here but could be generalized as structurally compatible delegate conversions should work. I think that goes against the mantra that delegates are type safe .
> dotnet/runtime#4331 (ah look it's your issue 😉 )

This is definitely the type of feature I'd like to see. 

> Feels like a much rarer scenario to me. It would matter in places where a Func/Action type couldn't be used so that means, byref types (because of generics), out and ref parameters (not sure what else besides those). 

Short term this is likely a better avenue to approach. For example: 

- Conisder expanding the set of existing delegates to support standard `out` patterns (last parameter is `out`)
- Allow `ref struct` to be used as generic type arguments in `delegate`. This is safe and even the lose proposals I've written for `where T: ref struct` keep coming back to the conclusion that we should assume all `delegate` generic parameters are marked this way. Need to do some more validation but it's worth investigating.

HaloFour:
@jaredpar 

> This is essentially an entire new kind of type inference though.

Possibly, although I'm not arguing for taking it that far as `1` has a type, whereas a lambda expression does not.

> Short term this is likely a better avenue to approach.

HaloFour:
I'd suggest that the ASP.NET API use cases can all be solved without adding inference of lambda variables to the language.  If the plan is to allow conversion from method group to `Delegate` in C# 10 then the same could be done with lambda expressions.  That would enable both of the following:

```cs
app.UseEndpoints(endpoints =>
{
    Task hello1(int x) => Task.CompletedTask;
    endpoints.MapGet("/hello1", hello1);

    endpoints.MapGet("/hello2", (int c) => Task.CompletedTask);   
});
```

Because of this my opinion is that it isn't necessary to explore inference of lambda variables at this time.  I think having to choose between `Action<...>`/`Func<...>` or an unspeakable synthetic delegate type limits the functionality of that feature too much.  One of the arguments made in Discord about wanting such a variable to behave like a variable and not like a local method is to have reference identity, which I can understand, however these delegate types are already incompatible with event handlers and coercing them to `EventHandler<T>`-compatible delegates would break that identity.

I'd rather see this space re-explored with the appropriate runtime support for signature equivalence and/or a more flexible approach rather than implementing something now that could not be changed without observable breaking changes in the future.

davidfowl:
I don't understand how suggesting a runtime feature changes any of this...

jaredpar:
> Because of this my opinion is that it isn't necessary to explore inference of lambda variables at this time. 

It's a natural outcome of the other decisions here. Solving the core problem here, general catch all overload for method group and lambda conversions, necessitates us defining an inherent delegate type for the method groups and lambda. There is certainly healthy debates to be had around how we choose this type: always generating a new delegate, always re-using existing ones or a mix. At the end of the day though a concrete delegate type is defined as the natural type of the expression. 

At that point allowing `var` is just a natural extension of the rest of the work. The entire reason it wasn't allowed in the first place is we didn't define a natural type for the value. Now the expressions have natural types, they can convert to `Delegate` which is a more interesting prospect.

mburbea:
Would this be a breaking change?
```cs
void M(Expression<Action> expr)=> Console.WriteLine(nameof(Expression));
void M(Delegate d) => Console.WriteLine(nameof(Delegate));
...
M(()=>Console.WriteLine());
```
In this scenario would it still print Expression or would it now complain about ambiguity?

jaredpar:
@mburbea 

It will still print out `"Expression"`. One of the tie breaking rules will be adjusted to prefer the conversion to a concrete delegate type over the inferred on. The exact rule we change still needs to be decided on (likely better conversion from expression). That will be done once we get through the initial approval process.

HaloFour:
@jaredpar 

> At that point allowing `var` is just a natural extension of the rest of the work. The entire reason it wasn't allowed in the first place is we didn't define a natural type for the value. Now the expressions have natural types, they can convert to `Delegate` which is a more interesting prospect.

I disagree.  You can achieve conversion between a lambda expression and `Delegate` without having to define a natural type for all lambda expressions.  That would achieve the narrow ASP.NET API use case without the severe limitations it would impose on the usefulness of local inference of lambda expressions for the majority of use cases, e.g. event handlers and lots of BCL APIs.

jaredpar:
@HaloFour 

>  You can achieve conversion between a lambda expression and Delegate without having to define a natural type for all lambda expressions. 

Conversion to `Delegate` requires a type be decided on for lambdas / method groups. Call it natural type, random type, etc ... but a type must be decide in the absence of any target type. That is a fundamental requirement for this work. Making it the "natural type" of the expression is fitting into existing C# constructs vs. creating a new one which is specific to lambdas / method groups.

>  without the severe limitations it would impose on the usefulness of local inference of lambda expressions for the majority of use cases,

I find the use of "severe limitations" here to be very misleading to the reader. The limitations here are no different than the limitations posed for all use of `var` in C#. The argument about inferring at the site of `var` not working for later use cases applies to existing C# scenarios. 

```c# 
var x = 42;
M(ref x); // Error: cause we picked different type at var 
void M(ref long l) { } 
```

The answer for this case, as well as for the lambda case, is to use an explicit type at the declaration of the local if the natural type of the expression isn't in line with later uses cases. 

Overall though for pretty much every sample you've provided where `var` doesn't work, the code is better off written as a local function in C# 10. That requires less keystrokes, has the "pick the type at the use" flexibility and doesn't provide a false sense of object identity which would be necessary to implement the flexible conversions you're advocating.

HaloFour:
@jaredpar 

> Conversion to `Delegate` requires a type be decided on for lambdas / method groups. Call it natural type, random type, etc ... but a type must be decide in the absence of any target type. That is a fundamental requirement for this work. Making it the "natural type" of the expression is fitting into existing C# constructs vs. creating a new one which is specific to lambdas / method groups.

And specifically where the conversion is to `Delegate` I agree.  The actual type of the delegate is much less important in that case.  But I don't think that needs to bleed over into `var` inference for lambdas.

> I find the use of "severe limitations" here to be very misleading to the reader.

Your own example was the need for reference identity for use with event handlers.  Except that this solution is entirely incompatible with event handlers as they are never `Action<...>` or `Func<...>`.

I think what's going to end up happening is that developers fall into the pit of failure of declaring an inferred lambda due to the language encouraging them to do so and then passing `Invoke` as a method group to those other APIs and as event handlers, which incurs additional allocations/overhead plus loses value identity.  Would also break any API that expects to be any to interrogate the `Method` of the delegate like the ASP.NET router examples above.

> Overall though for pretty much every sample you've provided where `var` doesn't work, the code is better off written as a local function in C# 10.

And my assertion is that every sample suggesting the use of `var` would also be better off written as a local function since you would also have conversion from method groups to `System.Delegate`.

proposals/csharp-10.0/lambda-improvements.md
HaloFour: How about `(int x, int y => int)` ?

proposals/csharp-10.0/lambda-improvements.md
cston: > Unless you're saying that these anonymous types are generic?

Yes, the anonymous delegate types are generic with type parameters for parameter and return types. The type of a specific lambda expression or method group would be a constructed type of the shared anonymous delegate type.

---
In reply to: [579440707](https://github.com/dotnet/csharplang/pull/4451#discussion_r579440707) [](ancestors = 579440707)

halter73:
```

**Comment by halter73** [comment]
Can we also add support optional parameters in lambdas? It would be useful for endpoints like the following and would be natural to developers used to defining similar actions inside of MVC controllers.

```c#
endpoints.MapGet("/products", ([FromQuery] int pageIndex = 0, [FromService] ProductsDb db) =>
	db.GetProductsAsync(pageIndex));
```

Tangentially related: It looks like local functions lose default parameters in the compiler generated method. This could cause problems for developers trying to use local functions instead of lambdas with MapAction. I filed a separate issue for that at https://github.com/dotnet/roslyn/issues/51518 Thanks @Kahbazi for finding this!

jaredpar:
> Tangentially related: It looks like local functions lose default parameters in the compiler generated method. This could cause problems for developers trying to use local functions instead of lambdas with MapAction. 

This is something we should be tracking in this proposal. Part of doing attributes on the emitted methods rationalizing that it's important, and okay, for these members to be seen from metadata. That would include logic like emitting our optional member metadata.

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,172 @@
...
+f = ([MyAttribute] int x) => x;  // [MyAttribute]x
+```
+
+Attributes are not supported for anonymous methods declared with `delegate { }` syntax.
+```csharp
+f = [MyAttribute] delegate { return 1; };         // syntax error
+f = delegate ([MyAttribute] int x) { return x; }; // syntax error
+```
+
+Attributes on the lambda or lambda parameters will be emitted to metadata on the method that maps to the lambda.
```
jaredpar, line 62: Think we should elaborate on this point a bit. In the past we have taken a very hard stance that we don't want customers taking a dependency on how lambdas and local funcs map from source to metadata. The structure of how the are emitted can, and has, changed between compiler versions. This has, and will continue, to break customers that attempt to take a dependency on how source maps to metadata. 

The changes we are making here are more targeted at the `Delegate` driven scenario. It's completely valid to look at the `MethodInfo` attached to a `Delegate` and dig into its properties. Emitting the explicit attributes to metadata here, as well as any other supporting attribute in code or emitted by compiler (items like default params), is intended to support this type of usage by customers. It allows for teams like ASP.NET to make the same types of behaviors on lambdas that they do for normal functions as well as allowing for features like trimming to handle lambdas appropriately. 

It does not change the overall stance that customer should not depend on how items like lambdas map from source to metadata .

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,172 @@
...
+static string GetString() => "";
+static int GetInt() => 0;
+
+Invoke(() => "");  // Invoke(Func<string>)
+Invoke(() => 0);   // Invoke(Delegate) [new]
+
+Invoke(GetString); // Invoke(Func<string>)
+Invoke(GetInt);    // Invoke(Delegate) [new]
+```
+
```
jaredpar, line 135: Think we need a couple of counter examples here showing where method group and lambda conversions will not work after this change. Specifically:

1. Cases where there are multiple candidates in the method group which means we can't infer a natural type
2. Cases where the return type of the lambda cannot be inferred hence the lambda does not have a natural type

proposals/csharp-10.0/lambda-improvements.md
cston: I've recorded the possible ambiguity as an issue for now.

---
In reply to: [579504322](https://github.com/dotnet/csharplang/pull/4451#discussion_r579504322) [](ancestors = 579504322)

proposals/csharp-10.0/lambda-improvements.md
cston: Added an open issue for now.

---
In reply to: [579450645](https://github.com/dotnet/csharplang/pull/4451#discussion_r579450645) [](ancestors = 579450645)

proposals/csharp-10.0/lambda-improvements.md
cston: Added open issue.

---
In reply to: [579441314](https://github.com/dotnet/csharplang/pull/4451#discussion_r579441314) [](ancestors = 579441314)

proposals/csharp-10.0/lambda-improvements.md
cston: Good catch, thanks. I've added an open issue that determining the natural type of a method group may require searching for all extension method candidates.

---
In reply to: [579579245](https://github.com/dotnet/csharplang/pull/4451#discussion_r579579245) [](ancestors = 579579245)

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,200 @@
...
+Support for explicit return types would provide symmetry with lambda parameters where explicit types can be specified.
+Allowing explicit return types would also provide control over compiler performance in nested lambdas where overload resolution must bind the lambda body currently to determine the signature.
+
+A natural type for lambda expressions and method groups will allow more scenarios where lambdas and method groups may be used without an explicit delegate type, including as initializers in `var` declarations.
+
+Requiring explicit delegate types for lambdas and method groups has been a friction point for customers, and has become an impediment to progress in ASP.NET with recent work on [MapAction](https://github.com/dotnet/aspnetcore/pull/29878).
+
+[ASP.NET MapAction](https://github.com/dotnet/aspnetcore/pull/29878) without proposed changes (`MapAction()` takes a `System.Delegate` argument):
+```csharp
+[HttpGet("/")] Todo GetTodo() => new(Id: 0, Name: "Name");
```
davidfowl, line 21: @halter73 we should get these samples updated.

proposals/csharp-10.0/lambda-improvements.md
bernd5: What about `(int a, int b) -> int`?
We have `->` only for pointers and it makes it more clear - at least to me - that is specifies the return type.

davidfowl: Doesn't that end up looking like this?

```C#
var f = (int a, int b) -> int => a + b;
```
VS

```C#
var f = (int a, int b) : int => a + b;
```

Not sure about the -> followed by the =>. It does give me haskell vibes though...

bernd5: In this case I dislike the second "arrow", too. 
Perhaps the arrow could be omitted in case of explicit return type followed by a block syntax - which means that we could write for example:

```
var fkt = (int a, int b) -> int
{
    return a + b;
}
```

For simple expression lambdas the return type is not so important - but for complex lambdas this proposal is a very huge improvement.

proposals/csharp-10.0/lambda-improvements.md
agocke: iirc this is a huge perf hit for some lambdas due to the `this` shuffle with delegates. remember that we only ever emit local functions like this when they're not converted to delegates

proposals/csharp-10.0/lambda-improvements.md
halter73: Yeah. One issue with that is the current updated samples rely heavily on conventions for parameter binding and no longer use an attribute for specifying the route pattern or HTTP method. That said, all that parameter attributes can still be optionally applied.

```C#
app.MapPut("/todos/{id}", async ([FromRoute(Name = "id")] int identifier, [FromBody] Todo inputTodo, [FromServices] TodoDb db) =>
{
    var todo = await db.Todos.FindAsync(identifier);

    if (todo is null) return NotFound();

    todo.Title = inputTodo.Title;
    todo.IsComplete = inputTodo.IsComplete;

    await db.SaveChanges();

    return NoContent();
});
```

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,200 @@
...
+[ASP.NET MapAction](https://github.com/dotnet/aspnetcore/pull/29878) with attributes and natural types for lambda expressions:
+```csharp
+app.MapAction([HttpGet("/")] () => new Todo(Id: 0, Name: "Name"));
+app.MapAction([HttpPost("/")] ([FromBody] Todo todo) => todo);
+```
+
+## Attributes
+Attributes may be added to lambda expressions.
+```csharp
+f = [MyAttribute] x => x;          // [MyAttribute]lambda
```
halter73, line 46: This is the form that I think should not be allowed because it could plausibly mean either `[MyAttribute] (x) => x` or `([MyAttribute] x) => x`. I think we should require parens around the argument list if you're using attributes.

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,200 @@
...
+app.MapAction(PostTodo);
+```
+
+[ASP.NET MapAction](https://github.com/dotnet/aspnetcore/pull/29878) with attributes and natural types for lambda expressions:
+```csharp
+app.MapAction([HttpGet("/")] () => new Todo(Id: 0, Name: "Name"));
+app.MapAction([HttpPost("/")] ([FromBody] Todo todo) => todo);
+```
+
+## Attributes
```
jmarolf, line 43: So excited to change all of our roslyn tests to lambdas :) Right now they can only be local functions :(

333fred:
Proposal needs to be moved from the proposals/csharp-10.0 folder to proposals. The 10.0 folder will be filled when we know what actually shipped.

cston:
> Proposal needs to be moved from the proposals/csharp-10.0 folder to proposals.

Sounds good. Perhaps we can move the proposal after merging, to avoid breaking the links to existing comments in the this PR.

---
In reply to: [791597365](https://github.com/dotnet/csharplang/pull/4451#issuecomment-791597365) [](ancestors = 791597365)

proposals/csharp-10.0/lambda-improvements.md
cston: Updated the comment for re-using synthesized delegate types.

---
In reply to: [579984320](https://github.com/dotnet/csharplang/pull/4451#discussion_r579984320) [](ancestors = 579984320)

proposals/csharp-10.0/lambda-improvements.md
333fred: @cston is there an open question in this doc for that?

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,200 @@
...
+Attributes may be added to lambda expressions.
+```csharp
+f = [MyAttribute] x => x;          // [MyAttribute]lambda
+f = [MyAttribute] (int x) => x;    // [MyAttribute]lambda
+f = [MyAttribute] static x => x;   // [MyAttribute]lambda
```
333fred, line 48: This form is also probably included in the question about parameter parenthesization

proposals/csharp-10.0/lambda-improvements.md
```diff
@@ -0,0 +1,200 @@
...
+  | '(' (lambda_parameter (',' lambda_parameter)*)? ')'
+  ;
+
+lambda_parameter
+  : identifier
+  | (attribute_list* modifier* type)? identifier equals_value_clause?
+  ;
+```
+
+_Does the `: type` return type syntax introduce ambiguities with `?:` that cannot be resolved easily?_
```
333fred, line 194: Consider including the alternative forms we discussed in LDM.

333fred (approved):
LGTM (commit 9). My remaining comments can be resolved in a follow up, if you prefer. In that followup, please update the link in https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-03-03.md#natural-type-for-lambdas to point to the checked in document, rather than this PR.

jaredpar (approved):

proposals/csharp-10.0/lambda-improvements.md
cston: I'll update the syntax based on the LDM and subsequent meetings in a separate PR.

---
In reply to: [588782316](https://github.com/dotnet/csharplang/pull/4451#discussion_r588782316) [](ancestors = 588782316)

cston:
> In that followup, please update the link in [https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-03-03.md#natural-type-for-lambdas](https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-03-03.md#natural-type-for-lambdas) to point to the checked in document, rather than this PR.

Will do, thanks.

---
In reply to: [605683764](https://github.com/dotnet/csharplang/pull/4451#issuecomment-605683764) [](ancestors = 605683764)

PR #4452, opened 2021-02-20
Add proposal for using aliases allowing type parameters and additional target types

cston:


proposals/csharp-10.0/using-alias-improvements.md
```diff
@@ -0,0 +1,117 @@
...
+
+The same name may be used for multiple aliases with distinct arity.
+```C#
+using MyDictionary = System.Collections.Generic.Dictionary<string, object>;
+using MyDictionary<T> = System.Collections.Generic.Dictionary<string, T>; // ok
+using MyDictionary<U> = System.Collections.Generic.Dictionary<string, U>; // error: 'MyDictionary<>' already defined
+```
+
+A partially bound type is not valid.
+The `using` aliases below are both valid, and the use of `MyList<>` is valid because the expansion `List<>` is an unbound type, but the use of `MyDictionary<>` invalid because `Dictionary<string,>` is a partially bound type.
```
333fred, line 34: Consider adding an example of a valid use of `MyDictionary`. This section is still starting with the words "A partially bound type is not valid.", which doesn't really address the confusion.

proposals/csharp-10.0/using-alias-improvements.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+### Constraints
+If type parameter constraints cannot be declared explicitly, then constraints in the alias target are verified at the use site only.
+```C#
+using Option<T> = System.Nullable<T>; // ok
+static void F<T>(Option<T> t) { ... } // error: 'T' must be value type in 'Nullable<T>'
+```
+
+However, without explicit constraints, the compiler will treat `T?` as `Nullable<T>` in an alias target which means nullable reference types cannot be used with type parameters in aliases.
```
333fred, line 50: Should we even permit `T?` at all? Or if we do, is this line even true? If we view these as macros, then `T?` will mean whatever `T?` would mean to the final expansion, not anything predefined like `Nullable<T>`. For example, `using MyMap<T> = Dictionary<T, T?>;` would cause different expansions for `int` and `string`, where one would be mapped to `int?` and one would be mapped to `string?`.

proposals/csharp-10.0/using-alias-improvements.md
```diff
@@ -0,0 +1,117 @@
...
+
+The alternative is to allow explicit type parameter constraints, perhaps using the same syntax as generic type and method constraint clauses.
+```C#
+using MaybeNull<T> = T? where T : class;
+using Option<T> = System.Nullable<T> where T : struct;
+```
+
+Explicit constraints would also allow restricting the use of the target type.
```
333fred, line 58: Consider referencing the LDM today, notes are up: https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-02-22.md

sylveon:
Is there any plan to make `using`s be namespace scoped? My current issue with `using`s is their globality. A library may be interested in shipping some aliases, but these could be intrusive. It could be affecting source code you vendored in, for example.

So making them namespace scoped would solve that issue, and allow to scope these aliases to only files importing that namespace (or using the full name for the alias), effectively making usage of a library's `using`s opt-in, only working in source files which actually use that library. Globality by default breaks (or makes it way easier to break) separation of concern that multiple source files provide.

A concrete example is TerraFX, which wants to start shipping global `using`s and keep the codegen 1:1 with C, so we'd have a struct named `tagRECT` and then a global `using` named `RECT` (because that's how the WinAPI headers work). I definitely do not want the name `RECT` to be taken over in all my source files, and would prefer confining it to only files doing interop (by importing the relevant namespace).

CyrusNajmabadi:
@sylveon I've hid your post for being off topic to this pr.  If you would like other features added to the language, please use github-discussions to open the topic. Thanks.

sylveon:
But I was told by @tannergooding to raise this concern here...

CyrusNajmabadi:
I don't think proposals are the place to tack on general questions/requests for other features. That's what discussions are for. 

A mechanism to export an alias is an entirely new thing and needs to be discussed and potentially championed first.

PR #4459, opened 2021-02-24
Add proposal for explicit parameterless struct constructors

cston:


proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,102 @@
...
+x.ToString(); // error: use of unassigned local (unless type is empty)
+y.ToString(); // error: use of unassigned local (unless type is empty)
+z.ToString(); // error: use of unassigned local (unless type is empty)
+```
+
+Array allocation ignores any parameterless constructor and generates zeroed elements.
+A warning is reported if there is an accessible parameterless constructor.
+```csharp
+var a = new NoConstructor[1];      // ok
+var b = new PublicConstructor[1];  // warning: constructor ignored
```
jnm2, line 59: Is a warning for this specific case needed, given the expectation that array creation never runs constructors for elements? Or conversely, should there be a warning for:

```cs
var b = new ClassWithPublicConstructor[1];
```

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+`default(S)` ignores any parameterless constructor and generates a zeroed `S`.
+```csharp
+NoConstructor x = default;      // ok
+PublicConstructor y = default;  // ok
+PrivateConstructor z = default; // ok
+```
+
+For a type parameter `T` with a `new()` constraint, `new T()` is emitted as a call to `System.Activator.CreateInstance<T>()` which ignores any explicit parameterless constructor.
```
HaloFour, line 32: My understanding is that this bug in the CLR was fixed:

https://github.com/dotnet/coreclr/pull/10778

jnm2: Also found this from a recent LDM:

https://github.com/dotnet/csharplang/blob/master/meetings/2021/LDM-2021-01-27.md#field-initializers

> We're still interested in doing this: the `Activator.CreateInstance` bug was in a version of the framework that is long out of support at this point, and we have universal agreement behind the idea.

proposals/csharp-10.0/parameterless-struct-constructors.md
HaloFour: Also seems a tad extraneous given that C# has always supported the consumption of structs with parameterless constructors and the behavior here will not change, but it's understandable to be explicit given how exceptionally unlikely a C# dev will have run into such a struct before.

jnm2: Makes sense. Because the idea that newing up an array would execute constructors runs so counter to how I think about types both with and without default constructors, it would be nice if this was a separate warning code so that I could suppress just the new array scenario without suppressing the other checks.

jnm2: Where I'm coming from is wanting to treat structs more like classes in this regard, since one of the main reasons I don't go with a struct these days is if I want a default constructor. So when I picture what it will be like when structs with parameterless constructors are common, I want to reason about constructors the same way.
But the difference between structs and classes will never be completely dismissable. If I think about the `new PublicConstructor[1]` case in terms of filling the array with Formatter.GetUninitializedObject, suddenly warnings feel really good.

So I get that, but where I still land is that there's really no other way to new up an array of structs that have parameterless constructors, even if you do immediately fill the array with valid instances right after you create the array. I'm picturing that there will be a lot of rote suppressions in my future.

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
```
333fred, line 32:
```suggestion
```

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
+A warning is reported if there is an accessible parameterless constructor.
+
+_Should `Activator.CreateInstance()` call the parameterless constructor if available and accessible?_
```
333fred, line 35:
```suggestion
```

It does.

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
+```csharp
+static T Create<T>() where T : new() => new T();
+
+x = Create<NoConstructor>();      // ok
+y = Create<PublicConstructor>();  // warning: constructor ignored
```
333fred, line 40:
```suggestion
y = Create<PublicConstructor>();  // ok
```

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
+z = Create<PrivateConstructor>(); // ok
```
333fred, line 41: This should warn.

proposals/csharp-10.0/parameterless-struct-constructors.md
333fred: This warning doesn't feel correct. We don't warn for classes, the same should apply to structs.

333fred (commented):
Activator stuff needs to be corrected.

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
+# Parameterless struct constructors
+
+## Summary
+
+Support explicit parameterless constructors for value types.
```
jaredpar, line 5:
```suggestion
Support parameterless constructors for value types.
```

In the case where you have a field initializer the parameterless constructor is implicitly defined

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Proposal
+
+If the parameterless constructor is less accessible than the value type, a warning is reported and the constructor is ignored when inaccessible.
```
jaredpar, line 9: Does such a `struct` satisfy the `new()` constraint?

proposals/csharp-10.0/parameterless-struct-constructors.md
jaredpar: Think we should have an open issue section at the bottom to follow up with CLR on this point.

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Array allocation ignores any parameterless constructor and generates zeroed elements.
+A warning is reported if there is an accessible parameterless constructor.
+```csharp
+var a = new NoConstructor[1];      // ok
+var b = new PublicConstructor[1];  // warning: constructor ignored
+var c = new PrivateConstructor[1]; // ok
+```
+
```
jaredpar, line 62: Couple of details think we need to make explicit here:

- If a `struct` has no fields and a paremeterless constructor do we consider a local of that type to be definitely assigned? 
- Can parameterless `struct` be applied to all `struct` types: `ref` and `readonly` specifically 
-

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,106 @@
...
+### Instance field initializers
+_[Adapted from [classes.md#instance-variable-initializers](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#instance-variable-initializers).]_
+When an instance constructor has no constructor initializer `this(...)`, that constructor implicitly performs the initializations specified in the _variable_initializers_ of the instance fields.
+This corresponds to a sequence of assignments that are executed immediately upon entry to the constructor and before the implicit invocation of the direct base class.
```
jaredpar, line 66: There is no `base` class in this scenario but we should specify the behavior when the parameterless ctor chains to another ctor. When do initializers execute in that scenario?

ufcpp:
```cs
// `= new()` is equivalent to `= default` and emitted as `.param[1] = nullref`
void m(S x = new S()) { }
struct S
{
    // If `S` has a parameterless constructor, should `new S()` in optional parameters be error or warning?
}
```

proposals/csharp-10.0/parameterless-struct-constructors.md
333fred: I did some testing of this, and I don't believe that it should. `Activator.CreateInstance` does not work if the struct constructor is declared as `internal` or `private`, regardless of whether the containing type is `internal` or not. It will _only_ call `public` constructors. We enforce this for reference types today:

```cs
M<C>(); // error CS0310: 'C' must be a non-abstract type with a public parameterless constructor in order to use it as parameter 'T' in the generic type or method 'M<T>()'

void M<T>() where T : new()
{
    _ = new T();
}

internal class C {
    internal C() { }
}
```

proposals/csharp-10.0/parameterless-struct-constructors.md
333fred: Scratch that, this should error, like we do for reference types today. It will throw at runtime. As an example:

```il
.assembly 'test' {}

.class private auto ansi abstract sealed beforefieldinit '<Program>$'
    extends [mscorlib]System.Object
{
    .custom instance void [mscorlib]System.Runtime.CompilerServices.CompilerGeneratedAttribute::.ctor() = (
        01 00 00 00
    )
    // Methods
    .method private hidebysig static 
        void '<Main>$' (
            string[] args
        ) cil managed 
    {
        // Method begins at RVA 0x2050
        // Code size 17 (0x11)
        .maxstack 8
        .entrypoint

        IL_0000: ldtoken S
        IL_0005: call class [mscorlib]System.Type [mscorlib]System.Type::GetTypeFromHandle(valuetype [mscorlib]System.RuntimeTypeHandle)
        IL_000a: call object [mscorlib]System.Activator::CreateInstance(class [mscorlib]System.Type)
        IL_000f: pop
        IL_0010: ret
    } // end of method '<Program>$'::'<Main>$'

} // end of class <Program>$

.class private sequential ansi sealed beforefieldinit S
    extends [mscorlib]System.ValueType
{
    .pack 0
    .size 1
    // Methods
    .method assembly hidebysig specialname rtspecialname 
        instance void .ctor () cil managed 
    {
        // Method begins at RVA 0x2062
        // Code size 17 (0x11)
        .maxstack 8

        IL_0000: ldarg.0
        IL_0001: call instance void [mscorlib]System.ValueType::.ctor()
        IL_0006: ldstr "Hello world!"
        IL_000b: call void [mscorlib]System.Console::WriteLine(string)
        IL_0010: ret
    } // end of method S::.ctor
} // end of class S
```

Prints:

```
Unhandled Exception: System.MissingMethodException: No parameterless constructor defined for this object.
   at System.RuntimeTypeHandle.CreateInstance(RuntimeType type, Boolean publicOnly, Boolean noCheck, Boolean& canBeCached, RuntimeMethodHandleInternal& ctor, Boolean& bNeedSecurityCheck)
   at System.RuntimeType.CreateInstanceSlow(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, StackCrawlMark& stackMark)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at <Program>$.<Main>$(String[] args)
```

333fred: (The sample works if you change `assembly` to `public` on the ctor, if that wasn't clear).

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -2,101 +2,139 @@
...
+Instance fields must be definitely assigned in struct instance constructors that do not have a `this()` initializer (see [struct constructors](https://github.com/dotnet/csharplang/blob/master/spec/structs.md#constructors)).
+
+Definite assignment is required for explicit parameterless constructors as well.
+Definite assignment _is not required_ for synthesized parameterless constructors.
 ```csharp
-struct NoConstructor
+struct S0
 {
+    object x = null;
+    object y;
```
333fred, line 42: This feels wrong when compared with `S1`. We should be consistent here: if you initialize one field, you must initialize them all. Or drop the initialization requirement in `S1`.

cston:
Good catch, thanks.

---
In reply to: [784731182](https://github.com/dotnet/csharplang/pull/4459#issuecomment-784731182) [](ancestors = 784731182)

proposals/csharp-10.0/parameterless-struct-constructors.md
cston: Yes, thanks for catching this.

---
In reply to: [581598691](https://github.com/dotnet/csharplang/pull/4459#discussion_r581598691) [](ancestors = 581598691)

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -14,6 +14,7 @@ As with [class field initializers](https://github.com/dotnet/csharplang/blob/mas
 
 ### Constructors
 A struct may declare a parameterless instance constructor.
+The struct can be declared as `struct`, `readonly struct`, or `ref struct`.
```
jaredpar, line 17: Given that we're about to introduce `record struct` perhaps we should flip this to say it's valid for all kinds of `struct` including ... and put the current list.

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,157 @@
...
+
+If the struct does not declare a parameterless instance constructor, and the struct has no fields with variable initializers, the struct (see [struct constructors](https://github.com/dotnet/csharplang/blob/master/spec/structs.md#constructors)) ...
+> implicitly has a parameterless instance constructor which always returns the value that results from setting all value type fields to their default value and all reference type fields to null.
+
+If the struct does not declare a parameterless instance constructor, and the struct has field initializers, a `public` parameterless instance constructor is synthesized.
+The parameterless constructor is synthesized even if all initializer values are zeros.
+
+### Modifiers
+A parameterless instance constructor may be less accessible than the struct.
```
jaredpar, line 26:
```suggestion
A parameterless instance constructor may be less accessible than the containing `struct`.
```

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,157 @@
...
+
+A parameterless instance constructor may be declared `extern`.
```
jaredpar, line 28: Can they be `partial`? 

I think we should start out with a base statement here, essentially something along the lines of 

> A parameterless struct constructor can have the same set of modifiers as other struct instance constructors. The only exceptions are ... 

Assuming there are any exceptions (can't think of any right now).

proposals/csharp-10.0/parameterless-struct-constructors.md
jaredpar: Might want to add that this has consequencs for otherr use cases that will be detailed later.

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Executing field initializers
+Execution of struct instance field initializers matches execution of [class field initializers](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#instance-variable-initializers):
+> When an instance constructor has no constructor initializer, ... that constructor implicitly performs the initializations specified by the _variable_initializers_ of the instance fields ... . This corresponds to a sequence of assignments that are executed immediately upon entry to the constructor and before the implicit invocation of the direct base class constructor. The variable initializers are executed in the textual order in which they appear in the ... declaration.
+
+### Definite assignment
+Instance fields must be definitely assigned in struct instance constructors that do not have a `this()` initializer (see [struct constructors](https://github.com/dotnet/csharplang/blob/master/spec/structs.md#constructors)).
+
+Definite assignment is required for explicit parameterless constructors as well.
```
jaredpar, line 37: Think we need to clean up the wording of this section a bit to make it clearer when we're talking about definite assignment within the `struct` constructor and when we're talking about uses of a `struct`

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,157 @@
...
+    S(object y) { this.y = y; } // ok
+}
+```
+
+_The synthesized parameterless constructor may need to emit `ldarg.0 initobj S` before any field initializers. Describe the specifics. How is this handled in Visual Basic?_
+
+### No `base()` initializer
+A `base()` initializer is disallowed in struct constructors.
+
+The compiler will not emit a call to the base `System.ValueType` constructor for any struct instance constructors including explicit and synthesized parameterless constructors.
```
jaredpar, line 68: Except in the cases where chain with `this(...)` right?

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Constructor use
+
+The parameterless constructor may be less accessible than the containing value type.
```
jaredpar, line 72: Can we make the accessibility of the `struct` themselves explicit here just to drive home the sample?

proposals/csharp-10.0/parameterless-struct-constructors.md
cston: From a constructor with a `this(...)` initializer, the compiler emits a call to that constructor from the same type, but I don't think the compiler emits a call to `System.ValueType::.ctor()` from any struct constructor.

---
In reply to: [583307707](https://github.com/dotnet/csharplang/pull/4459#discussion_r583307707) [](ancestors = 583307707)

proposals/csharp-10.0/parameterless-struct-constructors.md
cston: Added a comment that the same set of modifiers apply to parameterless constructors. Constructors cannot be declared `partial` though.

---
In reply to: [583304832](https://github.com/dotnet/csharplang/pull/4459#discussion_r583304832) [](ancestors = 583304832)

proposals/csharp-10.0/parameterless-struct-constructors.md
cston: > there's really no other way to new up an array of structs that have parameterless constructors

Since there's no way to avoid the warning, I've removed the warning comment and added an open issue.

---
In reply to: [581504602](https://github.com/dotnet/csharplang/pull/4459#discussion_r581504602) [](ancestors = 581504602)

proposals/csharp-10.0/parameterless-struct-constructors.md
cston: I've added an open issue for now so this can be discussed in LDM.

---
In reply to: [583120868](https://github.com/dotnet/csharplang/pull/4459#discussion_r583120868) [](ancestors = 583120868)

333fred (approved):
Initial spec with open questions LGTM

proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -0,0 +1,161 @@
...
+_ = CreateNew<PublicConstructor>();   // ok
+_ = CreateNew<InternalConstructor>(); // error: 'InternalConstructor..ctor()' is not public
+_ = CreateNew<PrivateConstructor>();  // error: 'PrivateConstructor..ctor()' is not public
+```
+
+`new T()` is emitted as a call to `System.Activator.CreateInstance<T>()`, and the compiler assumes the implementation of `CreateInstance<T>()` invokes the `public` parameterless constructor if defined.
+
+There is a gap in type parameter constraint checking because the `new()` constraint is satisfied by a type parameter with a `struct` constraint (see [satisfying constraints](https://github.com/dotnet/csharplang/blob/master/spec/types.md#satisfying-constraints)).
+
+As a result, the following will be allowed by the compiler but `Activator.CreateInstance<InternalConstructor>()` will fail at runtime.
```
jcouv, line 138: Sorry for being late to the party. Can we avoid this problem by requiring `public` accessibility on the parameterless constructor?

jaredpar: Unfortunately not really. The ecosystem today already contains `struct` with non-public ctors. That has to be accounted for in the design here. It would be possible for us to say we don't support them at a syntax level but we need to account for what happens when we see them in metadata. 

Given we have to account for them in metadata I would allow them in source otherwise it creates an odd friction point.

jcouv: Right, it makes sense that we still have to handle types in the wild. But are we helping users by making it easy to step into this gap?
If possible, let's bring this as an open question at LDM tomorrow.

PR #4493, opened 2021-03-03
Add initial required properties proposal

333fred:
This is the initial required properties proposal discussed today, modulo some clarifications and adding more open questions. I'd like to get this checked in, then we can start working through open questions in future LDMs.

### DIFF
--- a/proposals/required-members.md
+++ b/proposals/required-members.md
@@ -124,7 +124,7 @@ public class C
     public C() : init(Prop2)
     {
         Prop2 = 2;
-        Console.WriteLine($"Prop2 is {Prop1}")
+        Console.WriteLine($"Prop2 is {Prop2}")
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


RikkiGibson (approved):

proposals/required-members.md
```diff
@@ -0,0 +1,350 @@
...
+we again still have the second issue: properties in C# have been optional since C# 1.0. Nullable reference types, introduced in C# 8.0, addressed part of this issue: if a constructor
+does not initialize a non-nullable reference-type property, then the user is warned about it. However, this doesn't solve the problem: the user here wants to not repeat large parts
+of their type in the constructor, they want to pass the _requirement_ to set properties on to their consumers. It also doesn't provide any warnings about `ID` from `Student`, as that
+is a value type. These scenarios are extremely common in database model ORMs, such as EF Core, which need to have a public parameterless constructor but then drive nullability of the
+rows based on the nullability of the properties.
+
+This proposal seeks to address these concerns by introducing a new feature to C#: required members. Required members will be required to be initialized by consumers, rather than by
+the type author, with various customizations to allow flexibility for multiple constructors and other scenarios.
+
+## Detailed Design
```
RikkiGibson, line 74: It feels like a property without a `set` or `init` accessor can't be marked `required`. It might be good to specify this here.

333fred: I actually don't agree. This is one of those "silly" scenarios, but it would mean that the user _must_ initialize it in their constructor.

333fred: Though, perhaps we can say that a get-only non-auto-prop cannot be marked required.

RikkiGibson: do we have a notion of an "assignable" property? that would be a property where either:

1. it has a `set` or `init` accessor,
2. it is a getter-only auto property, or
3. it is ref-returning

I suppose the most straightforward thing would be to allow all 3

333fred: I'm not certain we have such a notion.

MadsTorgersen (approved):

PR #4503, opened 2021-03-09
Update ==/!= spec

RikkiGibson:
/cc @jcouv @333fred.

This spec change makes explicit that in the following scenario:
`a?.b(out x) == c.d(out y)`

- `y` is "definitely assigned" after the `==` expression
- `x` is "definitely assigned when true" after the `==` expression

Staring at this is making me wonder if there is a more reasonable way to generalize about how non-conditional state flows out of these expressions. e.g. `a.b(out x)?.c(out y)` unconditionally assigns `x` but not `y`.


jcouv:
> y is "definitely assigned" after the == expression

This makes sense, but I'd expect that to fall out of existing rules (both sides of `==` get evaluated), no?

> x is "definitely assigned when true" after the == expression

I didn't understand that. Couldn't `==` be true if `a` and `c.d(out y)` are both `null`?

RikkiGibson:
Ahh, I neglected the key detail which is that `c.d(out y)` is of a non-nullable value type.

I'll review the general rules for expressions again and try to determine if there is a better way to specify this.

Issue #1407, opened 2018-03-22
Proposal: Allow [AsyncMethodBuilder(...)] on methods


stephentoub:
EDIT: Proposal added 11/5/2020:
https://github.com/dotnet/csharplang/blob/master/proposals/csharp-10.0/async-method-builders.md

### Background

AsyncMethodBuilderAttribute can be put on a type to be used as the return type of an async method, e.g.
```C#
public struct AsyncCoolTypeMethodBuilder
{
    public static AsyncCoolTypeMethodBuilder Create();
    ...
}

[AsyncMethodBuilder(typeof(AsyncCoolTypeMethodBuilder))]
public struct CoolType { ... }

public async CoolType SomeMethodAsync() { ... } // will implicitly use AsyncCoolTypeMethodBuilder.Create()
```

This, however, means that:
- every use of a given return type requires the same builder.
- there's no context available to be passed to the builder.
- if you don't own the type and it's not already attributed, you can't use it as an async return type.

### Proposal

Two parts:

1. Allow AsyncMethodBuilderAttribute to be applied to methods.  When applied to an async method, it would be used as the builder for that method, overriding any AsyncMethodBuilderAttribute on the return type.
2. Allow the builder's Create method to have arguments.  Specifically, it can have arguments that match the parameters to the method to which it's applied, including the implicit `this` for instance methods.  The compiler will then forward the arguments to the method's invocation into the builder's Create.

#### Ammortized allocation-free async methods

There are four types in .NET with built-in builders: `Task`, `ValueTask`, `Task<T>`, and `ValueTask<T>`.  In .NET Core, significant work has gone in to optimizing async methods with these types, and thus in the majority of uses, each async method invocation will incur at most one allocation of overhead, e.g. a synchronously completing async method returning `ValueTask<T>` won’t have any additional allocations, and an asynchronously completing async method returning `ValueTask<T>` will incur one allocation for the underlying state machine object.

[...]

PR #4512, opened 2021-03-09
Spec determination of builder type from overrides

jcouv:
LDM recommend that we try and simplify the module-level usage of overrides. This proposal explains how we determine the applicable override/builder type when the attribute doesn't specify a "target return type".

@stephentoub @dotnet/roslyn-compiler for review.

Relates to async override proposal: https://github.com/dotnet/csharplang/issues/1407

### DIFF
--- a/proposals/async-method-builders.md
+++ b/proposals/async-method-builders.md
@@ -46,7 +46,7 @@ namespace System.Runtime.CompilerServices
     /// Indicates the type of the async method builder that should be used by a language compiler to
     /// build the attributed method.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface | AttributeTargets.Method | AttributeTargets.Module, Inherited = false, AllowMultiple = true)]
+    [System.AttributeUsage(System.AttributeTargets.Class | System.AttributeTargets.Struct | System.AttributeTargets.Interface | System.AttributeTargets.Method | System.AttributeTargets.Constructor | System.AttributeTargets.Event | System.AttributeTargets.Property | System.AttributeTargets.Module, Inherited = false, AllowMultiple = true)]
     public sealed class AsyncMethodBuilderOverrideAttribute : Attribute
     {
         /// <summary>Initializes the <see cref="AsyncMethodBuilderOverrideAttribute"/>.</summary>
@@ -59,7 +59,7 @@ namespace System.Runtime.CompilerServices
 }
 ```
 
-The attribute can be applied on methods, types and modules.
+The attribute can be applied on methods (or local function), constructors, events, properties, types and modules.
 
 Example of usage on a method:  
 ```C#
@@ -67,7 +67,7 @@ Example of usage on a method:
 static async ValueTask<int> ExampleAsync() { ... }
 ```
 
-It is an error to apply the attribute multiple times on a given method or local function.
+It is an error to apply the attribute multiple times on a given method (or local function).
 
 A developer who wants to use a specific custom builder for all of their methods can do so by putting the relevant attribute on each method.  
 Example of usage on module:  
@@ -83,7 +83,7 @@ class MyClass
 }
 ```
 
-### Determining the builder type for an async method
+### Determining the builder type for an async method (or local function or lambda)
 
 When compiling an async method (or local function or lambda), the builder type is determined by:
 1. looking in the containing scopes for an override attribute that specifies a builder type compatible with the method's return type.
@@ -178,8 +178,8 @@ Note that when the synthesized entry-point for top-level statements is async, it
 ## Unresolved questions
 [unresolved]: #unresolved-questions
 
-1. Confirm that the compiler should produce a diagnostic if the method-level attribute was found not compatible? (recommend yes)
-2. Confirm that the compiler should produce a diagnostic if multiple override attributes are specified on a method? (recommend yes)
+1. Confirm that the compiler should produce a diagnostic if the method-level attribute was found not compatible. (recommend yes)
+2. Confirm that the compiler should produce a diagnostic if multiple override attributes are specified on a method. (recommend yes)
 3. **Attribute.** Should we reuse `[AsyncMethodBuilder(typeof(...))]` or introduce yet another attribute? (answer: we need a new attribute)
 4. **Replace or also create.** All of the examples in this proposal are about replacing a buildable task-like's builder.  Should the feature be scoped to just that? Or should you be able to use this attribute on a method with a return type that doesn't already have a builder (e.g. some common interface)?  That could impact overload resolution.
 5. **Virtuals / Interfaces.** What is the behavior if the attribute is specified on an interface method?  I think it should either be a nop or a compiler warning/error, but it shouldn't impact implementations of the interface.  A similar question exists for base methods that are overridden, and there again I don't think the attribute on the base method should impact how an override implementation behaves. Note the current attribute has Inherited = false on its AttributeUsage.
@@ -187,3 +187,5 @@ Note that when the synthesized entry-point for top-level statements is async, it
 7. **Private Builders**. Should the compiler support non-public async method builders? This is not spec'd today, but experimentally we only support public ones.  That makes some sense when the attribute is applied to a type to control what builder is used with that type, since anyone writing an async method with that type as the return type would need access to the builder.  However, with this new feature, when that attribute is applied to a method, it only impacts the implementation of that method, and thus could reasonably reference a non-public builder.  Likely we will want to support library authors who have non-public ones they want to use.
 8. **Passthrough state to enable more efficient pooling**.  Consider a type like SslStream or WebSocket.  These expose read/write async operations, and allow for reading and writing to happen concurrently but at most 1 read operation at a time and at most 1 write operation at a time.  That makes these ideal for pooling, as each SslStream or WebSocket instance would need at most one pooled object for reads and one pooled object for writes.  Further, a centralized pool is overkill: rather than paying the costs of having a central pool that needs to be managed and access synchronized, every SslStream/WebSocket could just maintain a field to store the singleton for the reader and a singleton for the writer, eliminating all contention for pooling and eliminating all management associated with pool limits.  The problem is, how to connect an arbitrary field on the instance with the pooling mechanism employed by the builder.  We could at least make it possible if we passed through all arguments to the async method into a corresponding signature on the builder's Create method (or maybe a separate Bind method, or some such thing), but then the builder would need to be specialized for that specific type, knowing about its fields.  The Create method could potentially take a rent delegate and a return delegate, and the async method could be specially crafted to accept such arguments (along with an object state to be passed in).  It would be great to come up with a good solution here, as it would make the mechanism significantly more powerful and valuable.
 9. Should we allow **Assembly** target for the override attribute? (recommend no)
+10. Confirm that we should we allow **Property**, **Event**, **Constructor** targets for the override attribute. (recommend yes)
+11. Confirm that we should allow using an open generic type as override even when the attribute is used directly on a method. (recommend yes)
--- a/proposals/async-method-builders.md
+++ b/proposals/async-method-builders.md
@@ -36,7 +36,7 @@ We need a way to have an individual async method opt-in to a specific builder.
 ## Detailed design
 [design]: #detailed-design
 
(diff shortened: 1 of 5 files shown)


proposals/async-method-builders.md
```diff
@@ -46,23 +46,69 @@ namespace System.Runtime.CompilerServices
...
+Example of usage on a method:  
+```C#
+[AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder<int>))] // new, referring to some custom builder type
+static async ValueTask<int> ExampleAsync() { ... }
+```
+
+A developer that wants to using a specific custom builder for all of their methods can do so by putting the relevant attribute on each method.  
+Example of usage on module:  
+```C#
+[module: AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder), typeof(ValueTask)]
```
cston, line 73: >, typeof(ValueTask) [](start = 78, length = 19)

Is this argument needed?

cston, line 73: >] [](start = 97, length = 1)

Missing `)`.

proposals/async-method-builders.md
```diff
@@ -46,23 +46,69 @@ namespace System.Runtime.CompilerServices
...
+}
+```
+
+### Determining the builder type for an async methd
+
+When compiling an async method, the builder type is determined by:
+1. looking in the containing scopes for an override attribute that specifies a builder type compatible with the method's return type.
+2. otherwise, falling back to the builder type determined by previous approach. (see [spec for task-like types](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.0/task-types.md)) 
+
+The containing scope are considered starting from the method, then its containing types and finally the module. Each scope is considered
```
cston, line 90: >scope [](start = 15, length = 5)

"scopes"?

proposals/async-method-builders.md
```diff
@@ -46,23 +46,69 @@ namespace System.Runtime.CompilerServices
...
+in turn.  
+If a scope has one or more override attributes, the following process is used to determine if it is compatible with the current async method:
+1. take the override type specified by the attribute and construct it if necessary.  
+  If the override type is an open generic type, take the single type argument of the async method's return type and substitute it into the override type. If
```
cston, line 94: >If the override type is an open generic type, take the single type argument of the async method's return type and substitute it into the override type. [](start = 2, length = 151)

Does this mean an open type can be used for the attribute even when applied to a method directly?
```C#
[AsyncMethodBuilderOverride(typeof(MyBuilder<int>))] async Task<int> F1();
[AsyncMethodBuilderOverride(typeof(MyBuilder<>))] async Task<int> F2();
```

proposals/async-method-builders.md
```diff
@@ -46,23 +46,69 @@ namespace System.Runtime.CompilerServices
...
+  the async method's return type does not have a single type argument, then that override is not compatible.
+2. look for the public `Create` method with no type parameters and no parameters on the constructed override type:  
+  If the method is not found, then that override is not compatible.
+3. consider the return type of that `Create` method (a builder type) and look for the public `Task` property.  
+  If the property is not found, then that override is not compatible.
+4. consider the type of that `Task` property (a task-like type):  
+  If the task-like type matches the return type of the async method, then the override is compatible. Otherwise, it is not compatible.
+
+If no compatible override is found on a given scope, then look at the next scope.
```
cston, line 103: >If no compatible override is found on a given scope, then look at the next scope. [](start = 0, length = 81)

If the attribute is applied to a method directly but is incompatible, should we report an error instead of silently ignoring that attribute?

proposals/async-method-builders.md
jcouv: Yes, it would. Treating attributes on methods just like attributes on other scopes is a double-edged sword: it's simpler (more regular) but it does open the door to some strange usages...

proposals/async-method-builders.md
jcouv: Added special rule when the attribute is applied to method. Thanks

proposals/async-method-builders.md
```diff
@@ -46,23 +46,70 @@ namespace System.Runtime.CompilerServices
...
-In the C# compiler, prefer the attribute on the method when determining what builder to use over the one defined on the type.  For example, today if a method is defined as:
+The attribute can be applied on methods, types and modules.
+
+Example of usage on a method:  
+```C#
+[AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder<int>))] // new, referring to some custom builder type
+static async ValueTask<int> ExampleAsync() { ... }
+```
+
+A developer that wants to using a specific custom builder for all of their methods can do so by putting the relevant attribute on each method.
```
cston, line 70: >using [](start = 26, length = 5)

"use"?

proposals/async-method-builders.md
```diff
@@ -46,23 +46,70 @@ namespace System.Runtime.CompilerServices
     /// Indicates the type of the async method builder that should be used by a language compiler to
     /// build the attributed method.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface | AttributeTargets.Method | AttributeTargets.Module, Inherited = false, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface | AttributeTargets.Method | AttributeTargets.Module, Inherited = false, AllowMultiple = true)]
```
jaredpar, line 49: Think we should include `AttributeTargets.Assembly` here. Yes the compiler team is typically focused on modules because that is the unit of emit that we have to deal with. 99% of users though don't understand the difference between assembly and module and they are all used to assembly level attributes. Think we should support that and ask users to write it.

proposals/async-method-builders.md
```diff
@@ -46,23 +46,70 @@ namespace System.Runtime.CompilerServices
...
+Example of usage on module:  
+```C#
+[module: AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder))]
+[module: AsyncMethodBuilderOverride(typeof(PoolingAsyncValueTaskMethodBuilder<>))]
+
+class MyClass
+{
+    public async ValueTask Method1Async() { ... } // would use PoolingAsyncValueTaskMethodBuilder
+    public async ValueTask<int> Method2Async() { ... } // would use PoolingAsyncValueTaskMethodBuilder<int>
+    public async ValueTask<string> Method3Async() { ... } // would use PoolingAsyncValueTaskMethodBuilder<string>
```
jaredpar, line 80: Consider adding a counter example here using a task-like type that does not hit the custom builder. Essentially demonstrating the fallback / default case.

proposals/async-method-builders.md
```diff
@@ -46,23 +46,70 @@ namespace System.Runtime.CompilerServices
...
+}
+```
+
+### Determining the builder type for an async method
```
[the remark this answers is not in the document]
jaredpar: I would not expect that to be the case however I do think we need to spec out whether or not this should apply to `async` lambda as well once we have the ability to add attributes.

proposals/async-method-builders.md
```diff
@@ -46,23 +46,70 @@ namespace System.Runtime.CompilerServices
...
+
+When compiling an async method, the builder type is determined by:
+1. looking in the containing scopes for an override attribute that specifies a builder type compatible with the method's return type.
+2. otherwise, falling back to the builder type determined by previous approach. (see [spec for task-like types](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.0/task-types.md)) 
+
+The containing scopes are considered starting from the method, then its containing types and finally the module. Each scope is considered
```
jaredpar, line 90: Couple of items that I think we should add clarity on: 

- How does this work for local functions? Does a local function consider the attributes on the function that contains it or is it containing scope the type?
- By containing types here we mean we consider the inner most nested type and walk outwards to the containing type correct?

proposals/async-method-builders.md
```diff
@@ -46,23 +46,70 @@ namespace System.Runtime.CompilerServices
...
+  If the override type is an open generic type, take the single type argument of the async method's return type and substitute it into the override type. If
+  the async method's return type does not have a single type argument, then that override is not compatible.
+2. look for the public `Create` method with no type parameters and no parameters on the constructed override type:  
+  If the method is not found, then that override is not compatible.
+3. consider the return type of that `Create` method (a builder type) and look for the public `Task` property.  
+  If the property is not found, then that override is not compatible.
+4. consider the type of that `Task` property (a task-like type):  
+  If the task-like type matches the return type of the async method, then the override is compatible. Otherwise, it is not compatible.
+
+If the attribute was specified on a method, but it turned out to be incompatible, an error is produced.
```
jaredpar, line 103: What do we do in the case multiple attributes are defined on the method? 

My recommendation is we should emit an error. The `AllowMultiple = true` is necessary for the assembly / type case but doesn't make much sense at the method level.

proposals/async-method-builders.md
```diff
@@ -268,13 +175,12 @@ This would not only make it more convenient than putting the attribute on every
...
-
-## Design meetings
+1. Should the compiler produce a diagnostic if the method-level attribute was found not compatible?
```
jaredpar, line 178: Strongly believe yes. Putting an attribute on a method is about as strong of a declaration the customer can make that they intend for the builder to be used there. If we instead silently fell back to the default or one on the containing type it seems a virtual certainty that this would not be the behavior the user expected and hence we should produce a diagnostic.

proposals/async-method-builders.md
```diff
@@ -268,13 +175,12 @@ This would not only make it more convenient than putting the attribute on every
...
+2. Should the compiler produce a diagnostic if multiple override attributes are specified on a method?
```
jaredpar, line 179: Yes. Similar logic as above. It's a bit of a non-sensical declaration. Hard to even justify it from say a source generator perspective.

proposals/async-method-builders.md
jcouv: For skiplocalsinit, the [attribute ](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.skiplocalsinitattribute?view=net-5.0) is allowed on modules, but not assemblies. 
I think our other attribute-based features did the same. We should probably stick to that approach.

On the other hand, should we allow `Property` (accessors could use async lambdas or local functions), `Event` (ditto) and `Constructor`?

proposals/async-method-builders.md
jcouv: - Good catch for local functions and lambdas. I'm proposing that the attribute also apply to those and those scopes contribute to the lookup chain.

- Yes, we'll look at the containing types in turn, walking outwards.

proposals/async-method-builders.md
jcouv: Added an open question. 
Do you think we should disallow open generics when the attribute is applied to a method directly? Seems unnecessarily verbose.

proposals/async-method-builders.md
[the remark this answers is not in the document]
jcouv: I'm proposing that async local functions and lambdas be subject to the same rules. Also tracking this in open questions section.

proposals/async-method-builders.md
jcouv: I'm proposing that we should allow `Property`, `Event` and `Constructor`, but not `Assembly`. Also added a tracking issue in open questions.

proposals/async-method-builders.md
jcouv: Added example of picking inner scope and also falling back to non-override behavior.

proposals/async-method-builders.md
```diff
@@ -46,23 +46,76 @@ namespace System.Runtime.CompilerServices
...
-        public AsyncMethodBuilderOverrideAttribute(Type builderType, Type targetReturnType) => ...
-
         /// <summary>Gets the <see cref="Type"/> of the associated builder.</summary>
         public Type BuilderType { get; }
     }
 }
 ```
 
-In the C# compiler, prefer the attribute on the method when determining what builder to use over the one defined on the type.  For example, today if a method is defined as:
+The attribute can be applied on methods (or local function), constructors, events, properties, types and modules.
```
cston, line 62: >function [](start = 50, length = 8)

"functions"?

cston (approved):

PR #4521, opened 2021-03-11
Update parameterless struct constructor proposal

cston:
Updates based on recent discussions.

### DIFF
--- a/proposals/parameterless-struct-constructors.md
+++ b/proposals/parameterless-struct-constructors.md
@@ -86,7 +86,6 @@ struct S1
 {
     int x = 1;
     object y;  // error: field 'y' must be assigned from synthesized constructor
-    S(object y) { this.y = y; } // ok
 }
 ```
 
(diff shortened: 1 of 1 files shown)


333fred (approved):

proposals/parameterless-struct-constructors.md
```diff
@@ -116,10 +95,33 @@ A `base()` initializer is disallowed in struct constructors.
...
+record struct R5(int F)
+{
+    public R5() { } // error: must call 'this(string S)'
+    public int F =  F;
+}
+```
+
 ### Fields
-The synthesized parameterless constructor will zero fields rather than calling any parameterless constructors for the field types.
+The synthesized parameterless constructor will zero fields rather than calling any parameterless constructors for the field types. No warnings are reported that field constructors are ignored.
```
RikkiGibson, line 122: > The synthesized parameterless constructor

Does this refer to the same thing as the "default value type constructor"?

cston: Yes. I've changed this to "implicitly-defined parameterless constructor" to match the "implicitly has a parameterless instance constructor" earlier (which I believe is the only other reference.

PR #4524, opened 2021-03-11
Fix inactive and rejected proposal links

Youssef1313:


Youssef1313:
Cherry-picked and added with #4525

PR #4548, opened 2021-03-17
Update structs.md

johnnygerard:
Move "partial modifier" link and add missing link


333fred:
@BillWagner, is there anything you need to do with the ecma side to integrate this?

PR #4549, opened 2021-03-17
Spec change for auto-props

CyrusNajmabadi:


spec/classes.md
```diff
@@ -2873,9 +2873,16 @@ the underlying `TextWriter` for the output device is created. But if the applica
...
+An automatically implemented property (or ***auto-property*** for short), is a non-abstract non-extern property with at least:
 
-When a property is specified as an automatically implemented property, a hidden backing field is automatically available for the property, and the accessors are implemented to read from and write to that backing field. If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
+1. one semicolon-only body
+2. usage of the `field` contextual keyword ([Keywords](lexical-structure.md#keywords)) within the `property_body` of the `property`. The `field` identifier is only considered the `field` contextual when there is no symbol named `field` in scope at that location. For the purposes of simplicity, from this point on any usages of `field` in the specification description will assume there is no `field` symbol in scope at that location.
```
jcouv:
```suggestion
2. usage of the `field` contextual keyword ([Keywords](lexical-structure.md#keywords)) within the `property_body` of the `property`. The `field` identifier is only considered the `field` contextual keyword when there is no symbol named `field` in scope at that location. For the purposes of simplicity, from this point on any usages of `field` in the specification description will assume there is no `field` symbol in scope at that location.
```

spec/classes.md
```diff
@@ -2873,9 +2873,16 @@ the underlying `TextWriter` for the output device is created. But if the applica
...
+
+Auto-properties must have a get accessor and can optionally have a set accessor. Because an expression-bodied property defines a get accessor, if the expression-bodied property refers to `field`, it satisfies the requirements and is considered an auto-property.
+
+When a property is specified as an auto-property, a hidden, unnamed, backing field is automatically available for the property, and any semicolon-only accessors are implemented to read from and write to that backing field. The hidden backing field is in scope in all accessors and can be referenced directly using `field`.  Because the field is unnamed, it cannot be used in a `nameof` expression.
+
+ If the auto-property has no set accessor, and does not use `field`, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
```
jcouv: That doesn't seem obvious. Can't I have an auto-incrementing get-only property?

spec/classes.md
```diff
@@ -2920,6 +2927,56 @@ public class ReadOnlyPoint
 Notice that the assignments to the readonly field are legal, because they occur within the constructor.
 
 
+The following example:
+```csharp
+// No 'field' symbol in scope.
+public class Point
+{
+	public int X { get; set; }
```
jcouv: nit: indent in code snippets look unusual

spec/classes.md
```diff
@@ -2920,6 +2927,56 @@ public class ReadOnlyPoint
...
+    public int Y { get { return __y; } set { __y = value; } }
+}
+```
+
+The following example:
+```csharp
+// No 'field' symbol in scope.
+public class LazyInit
+{
+	public string Value => field ??= ComputeValue();
```
jcouv: nit: this indent is an eye sore. Also below (line 2975)

jcouv (commented):
Done with review pass.
You can look at Aleksey's "global usings" for an example of making a separate spec document that shows a diff from official spec:
https://github.com/dotnet/csharplang/blob/8bddff3b2f6dea3e8647af81cdd808e7e86f1119/proposals/GlobalUsingDirective.md

spec/classes.md
```diff
@@ -2873,9 +2873,18 @@ the underlying `TextWriter` for the output device is created. But if the applica
...
+1. an accessor with semicolon-only body
+2. usage of the `field` contextual keyword ([Keywords](lexical-structure.md#keywords)) within the `property_body` of the `property`. The `field` identifier is only considered the `field` keyword when there is no existing symbol named `field` in scope at that location.
+
+Auto-properties must have a get accessor and can optionally have a set accessor. Because an expression-bodied property defines a get accessor, if the expression-bodied property refers to the `field` keyword, it satisfies the requirements and is considered an auto-property.
+
+When a property is specified as an auto-property, a hidden, unnamed, backing field is automatically available for the property, and any semicolon-only accessors are implemented to read from and write to that backing field. The backing field is in scope in all accessors, the `property_initializer`, the property expression body, and can be referenced directly using the `field` keyword. Because the field is unnamed, it cannot be used in a `nameof` expression.
```
jcouv, line 2883: Should probably state what is the type of the backing field ;-)

CyrusNajmabadi: interesting.  we never seemed to need that before :)

spec/classes.md
CyrusNajmabadi: yup.  all whitespace was fixed.

spec/classes.md
CyrusNajmabadi: this was tweaked to separate out the concepts of 'readonly' (which only occurs for a `get;`-only autoprop) and 'constructor can assign directly to field'.  Can you recheck?

spec/classes.md
```diff
@@ -2900,26 +2909,76 @@ The following example:
...
+    public int Y { get { return __y; } set { __y = value; } }
+}
+```
+
+The following example:
+```csharp
+// No 'field' symbol in scope.
+public class LazyInit
+{
+    public string Value => field ??= ComputeValue();
```
alrz, line 2967: Happy to see we don't need `auto` modifier :)

Is there any special meaning for `@field`?

CyrusNajmabadi: yes.  it would no longer be a keyword (contextual or otherwise) so would not count.  this falls out from our definitions around keywords/identifiers and whatnot.

CyrusNajmabadi: > Happy to see we don't need auto modifier :)

LDM still to decide on this.  it will come back to if we think it is critical to be able to tell if somethign is an autoprop without having to do examination of the bodies.  My perspective is that it is not critical and it's acceptable to have no top-level sigil indicating this information.

spec/classes.md
```diff
@@ -2873,9 +2873,18 @@ the underlying `TextWriter` for the output device is created. But if the applica
...
+1. an accessor with semicolon-only body
+2. usage of the `field` contextual keyword ([Keywords](lexical-structure.md#keywords)) within the accessors or expression body of the property. The `field` identifier is only considered the `field` keyword when there is no existing symbol named `field` in scope at that location.
+
+Auto-properties must have a get accessor and can optionally have a set accessor. Because an expression-bodied property defines a get accessor, if the expression-bodied property refers to the `field` keyword, it satisfies the requirements and is considered an auto-property.
+
+When a property is specified as an auto-property, a hidden, unnamed, backing field is automatically available for the property, and any semicolon-only accessors are implemented to read from and write to that backing field. The backing field can be referenced directly using the `field` keyword within all accessors and within the property expression body. Because the field is unnamed, it cannot be used in a `nameof` expression.
```
333fred, line 2883:
```suggestion
When a property is specified as an auto-property, a hidden, unnamed backing field is automatically available for the property, and any semicolon-only accessors are implemented to read from (if a `get` accessor) or write to (if a `set` accessor) that backing field. The backing field can be referenced directly using the `field` keyword within all accessors and within the property expression body. Because the field is unnamed, it cannot be used in a `nameof` expression.
```

333fred: (I know those parens are not in the original, but what the original says is kinda ambiguous at the moment).

spec/classes.md
```diff
@@ -2873,9 +2873,18 @@ the underlying `TextWriter` for the output device is created. But if the applica
...
+
+If the auto-property does not have a set accessor, the backing field can still be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
```
333fred, line 2885:
```suggestion
Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
```

I much prefer the original wording here.

spec/classes.md
CyrusNajmabadi: Yeah, i'm ok with keeping as is given the original.  We can see what LDM thinks on this.   but i think it's fairly clear contextually :)

spec/classes.md
CyrusNajmabadi: the problem here is what does 'getter only' mean.  It doesn't mean it has an explicit 'get accessor', as `=>` auto-props would count here.  So really, all we're saying is that even if there is no set-accessor, the auto-prop is still assignable in the constructor.

PR #4574, opened 2021-03-22
Update partial method specification

maxkoshevoi:
Updated `partial` method specification according to changes made in C# 9

Fixes: https://github.com/dotnet/docs/issues/23084

### DIFF
--- a/spec/classes.md
+++ b/spec/classes.md
@@ -712,8 +712,7 @@ class Customer
 }
 ```
 
-Partial methods can also be useful in combination with source generators. For example a regex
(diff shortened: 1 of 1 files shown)


333fred:
Thanks for this change! However, this spec covers an older version of C#, not the C# 9 changes. You can find the proposed C# 9 specification here: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/extending-partial-methods.md. When those changes are integrated back into the main specification, this document will be updated.

Issue #3435, opened 2020-05-07
Proposal: list pattern


gafter:
See https://github.com/dotnet/csharplang/pull/3245

HaloFour:
Awesome.  Is this going to build on the linked proposal (with your additional feedback) or is this to track the work of designing list/collection/dictionary/etc patterns in general?

gafter:
@HaloFour This is to track answering questions like those.

alrz:
>list/collection/dictionary/etc patterns

I wonder how you'd imagine dictionary/indexer patterns work and what use cases you have in mind.

Syntax-wise it could be `{[constant]: pattern}` but that's useless because any common type that exposes an indexer, throws if it's out of range.

gafter:
Maybe we could use a `TryGet` method.

alrz:
To me, that's too specific to have a dedicated syntax for. I think we should enable patterns for all Try-like methods which is essentially what user-defined positional patterns are.
```cs
map is TryGetValue("constant key", pattern) // maybe with another name
```
This could made to work only on "some" methods as mentioned in https://github.com/dotnet/csharplang/issues/1047#issuecomment-440254430

Syntactical symmetry with indexer initializers makes it attractive though, however, the fact that it would not call the indexer might be confusing.

HurricanKai:
Would this allow something like
```cs
Array: { Length: 1, [1] }
```
checking whether there is only one element and whether that element is `1`?
Also, looking at the PR it's unclear to me whether this would allow checking for just the 5th element for example? Looks like that's not possible?

alrz:
@HurricanKai 

The pattern `[1]` already includes the check for length=1 so you don't need to add anything else.

To match Nth element you *could* prepend N-1 discards - it tends to get long but it is definitely possible.

The team is interested in supporting indexers, so using range indexer pattern, we could support `{ [5]: var elem }` and generate `e.Length >= 5 && e[5] is var elem` (we'd only accept constant indexer args so we can calculate the minimum size).

I'm not sure if that's common enough for a dedicated support, plus it won't work with generic dictionaries which is probably a deal breaker.

HurricanKai:
I'd generally much prefer `[idx]: { ... }` because it looks much more natural inside of more pattern matching, in fact I tried typing it out like that today cause I thought it would work 😄 ie `[2]: { [5]: { Length: 5 } }` on a `string[][]` would be quite amazing if possible.

omikhailov:
[...]

PR #4575, opened 2021-03-23
Spec list-pattern on enumerable collections

jcouv:
Tagging @dotnet/roslyn-compiler @alrz @MadsTorgersen for review.

Relates to proposal https://github.com/dotnet/csharplang/issues/3435

### DIFF
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -220,6 +220,11 @@ class ListPatternHelper
     return count;
   }
 
+  public bool Last()
+  {
+    return !enumerator.MoveNext();
+  }
+
   public int Count()
   {
     return MoveNextIfNeeded(-1);
@@ -266,7 +271,7 @@ class ListPatternHelper
 
   helper.TryGetStartElement(index: 0, out var element0) && element0 is 0 &&
   helper.TryGetStartElement(1, out var element1) && element1 is 1 &&
-  helper.Count() == 2
+  helper.Last()
 }
 ```
 
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -324,4 +324,4 @@ All multi-dimensional arrays can be non-zero-based. We can either:
 2. Call `GetLowerBound` and add it to each indexer access to pass the *correct* index.
 3. Assume all arrays are zero-based since that's the default for arrays created by `new` expressions.
 4. Should we try and optimize list-patterns like `{ 1, _, _ }` on a countable enumerable type? We could just check the first enumerated element then check `Length`/`Count`...
-5. Should we try to enumeration short for length-patterns on enumerables in some cases? (computing max acceptable count and checking partial count against that)
+5. Should we try to cut the enumeration short for length-patterns on enumerables in some cases? (computing min/max acceptable count and checking partial count against that)
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -93,18 +93,18 @@ This set of rules is derived from the [***range indexer pattern***](https://gith
 
 #### Semantics on enumerable type
 
-If the input type is *enumerable* but not *countable*, then the *length-pattern* is checked on the number of elements obtained from enumerating the collection.
+If the input type is *enumerable* but not *countable*, then the *length_pattern* is checked on the number of elements obtained from enumerating the collection.
 
-If the input type is *enumerable* but not *indexable*, then the *list-pattern* enumerates elements from the collection and checks them against the listed patterns:  
-Patterns at the start of the *list-pattern* — that are before the `..` *slice-pattern* if one is present, or all otherwise — are matched against the elements produced at the start of the enumeration.  
+If the input type is *enumerable* but not *indexable*, then the *list_pattern* enumerates elements from the collection and checks them against the listed patterns:  
+Patterns at the start of the *list_pattern* — that are before the `..` *slice_pattern* if one is present, or all otherwise — are matched against the elements produced at the start of the enumeration.  
 If the collection does not produce enough elements to get a value corresponding to a starting pattern, the match fails. So the *constant-pattern* `3` in `{ 1, 2, 3, .. }` doesn't match when the collection has fewer than 3 elements.  
-Patterns at the end of the *list-pattern* (that are following the `..` *slice-pattern* if one is present) are matched against the elements produced at the end of the enumeration.  
+Patterns at the end of the *list_pattern* (that are following the `..` *slice_pattern* if one is present) are matched against the elements produced at the end of the enumeration.  
 If the collection does not produce enough elements to get a value corresponding to an ending pattern, the match fails. So the *constant-pattern* `3` in `{ 1, .., 3 }` doesn't match when the collection has fewer than 2 elements.  
-A *list-pattern* without a *splice-pattern* only matches if the number of elements produced by complete enumeration and the number of patterns are equals. So `{ _, _, _ }` only matches when the collection produces exactly 3 elements.
+A *list_pattern* without a *splice-pattern* only matches if the number of elements produced by complete enumeration and the number of patterns are equals. So `{ _, _, _ }` only matches when the collection produces exactly 3 elements.
 
 Note that those implicit checks for number of elements in the collection are unaffected by the collection type being *countable*. So `{ _, _, _ }` will not make use of `Length` or `Count` even if one is available.
 
-When multiple *list-patterns* are applied to one input value the collection will be enumerated once at most:  
+When multiple *list_patterns* are applied to one input value the collection will be enumerated once at most:  
 ```
 _ = collection switch
 {
@@ -121,9 +121,9 @@ _ = collectionContainer switch
 };
 ```
 
-It is possible that the collection will not be completely enumerated. For example, if one of the patterns in the *list-pattern* doesn't match or when there are no ending patterns in a *list-pattern* (e.g. `collection is { 1, 2, .. }`).
+It is possible that the collection will not be completely enumerated. For example, if one of the patterns in the *list_pattern* doesn't match or when there are no ending patterns in a *list_pattern* (e.g. `collection is { 1, 2, .. }`).
 
-If an enumerator is produced when a *list-pattern* is applied to an enumerable type and that enumerator is disposable it will be disposed when a top-level pattern containing the *list-pattern* successfully matches, or when none of the patterns matches (in the case of a `switch` statement or expression). It is possible for an enumerator to be disposed more than once and the enumerator must ignore all calls to `Dispose` after the first one.
+If an enumerator is produced when a *list_pattern* is applied to an enumerable type and that enumerator is disposable it will be disposed when a top-level pattern containing the *list_pattern* successfully matches, or when none of the patterns match (in the case of a `switch` statement or expression). It is possible for an enumerator to be disposed more than once and the enumerator must ignore all calls to `Dispose` after the first one.
 ```
 // any enumerator used to evaluate this switch statement is disposed at the indicated locations
 _ = collection switch
@@ -226,8 +226,8 @@ class ListPatternHelper
   // fulfills the role of `[index]` for start elements when enough elements are available
   public bool TryGetStartElement(int index, out ElementType value)
   {
-    Debug.Assert(index < startBuffer.Length);
-    PullStart(index);
+    Debug.Assert(index >= 0 && index < startBuffer.Length);
+    MoveNextIfNeeded(index);
     if (count > index)
     {
       value = startBuffer[index];
@@ -240,7 +240,7 @@ class ListPatternHelper
   // fulfills the role of `[^hatIndex]` for end elements when enough elements are available
   public bool TryGetEndElement(int hatIndex, out ElementType value)
   {
-    Debug.Assert(hatIndex < endBuffer.Length);
+    Debug.Assert(hatIndex > 0 && hatIndex <= endBuffer.Length);
     Count();
     if (count < startBuffer.Length + endBuffer.Length)
     {
@@ -306,7 +306,7 @@ class ListPatternHelper
 }
 ```
 
-The same way that a `Type { name: pattern }` *property-pattern* checks that the input has the expected type and isn't null before using that as receiver for the property checks, so can we have the `{ ..., ... }` *list-pattern* initialize a helper and use that as the pseudo-receiver for element accesses.  
+The same way that a `Type { name: pattern }` *property-pattern* checks that the input has the expected type and isn't null before using that as receiver for the property checks, so can we have the `{ ..., ... }` *list_pattern* initialize a helper and use that as the pseudo-receiver for element accesses.  
 This should allow merging branches of the patterns DAG, thus avoiding creating multiple enumerators.
 
 ### Additional types
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -325,3 +325,4 @@ All multi-dimensional arrays can be non-zero-based. We can either:
 3. Assume all arrays are zero-based since that's the default for arrays created by `new` expressions.
 4. Should we try and optimize list-patterns like `{ 1, _, _ }` on a countable enumerable type? We could just check the first enumerated element then check `Length`/`Count`...
 5. Should we try to cut the enumeration short for length-patterns on enumerables in some cases? (computing min/max acceptable count and checking partial count against that)
+  What if the enumerable type has some sort of `TryGetNonEnumeratedCount` API?
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -80,7 +80,7 @@ A *list_pattern* is compatible with any type that is *countable* as well as *ind
 A *list_pattern* is also compatible with any type that is *enumerable*.
 
 A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.  
-A *slice_pattern* without a sub-pattern is also compatible with any type that is *enumerable*.
+A *slice_pattern* without a sub_pattern is also compatible with any type that is *enumerable*.
 
 ```
 enumerable is { 1, 2, .. } // okay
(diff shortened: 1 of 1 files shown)


jcouv:
I need to tweak this a bit more, to be able to share the data structure between two branches where the staring patterns in one may overlap with the ending patterns in another:
```
collection switch
{
   { 1, 2, 3, .., 5 } => ... // causes start buffer to hold 3 elements
   {1,  .., 3, 4, 5 } => ... // causes end buffer to hold 3 elements
}

**Comment by alrz** [comment]
We could possibly ditch the separate start buffer if we disallow element match after a slice pattern.
```cs
case { .., 1, 2 }:
case { 1, 2, .. }: // subsumption error
```
```cs
case { 1, 2, .. }:
case { .., 1, 2 }: // ok
```
Note that we need to start buffering from the beginning anyways to facilitate forthcoming matches but we only need to maintain a single buffer for that.

This might be too restrictive but I'm not sure how this is common to worth the added complexity.

alrz:
That said, I'd prefer we have a separate helper for skipping elements,
```cs
@{
  // bufferSize: computed from the max number of trailing patterns in all arms
  var helper = new ListPatternHelper(collection, bufferSize); 

  // leading patterns
  helper.TryGetNext(0, out var element0) && ...

  // slice pattern; either:
  helper.SkipToEnd() // always succeeds
  helper.TrySkipToEnd(maxLength)

  // trailing patterns
  helper.TryGetLast(2, out var hatElement2) && ..

  // length pattern; either:
  helper.TryGetCount(out var count)
  helper.TryGetCount(minLength, out var count)
  helper.TryGetCount(maxLength, out var count)
  helper.TryGetCount(minLength, maxLength, out var count)
}
```
For `TryGetCount`, we could possibly optimize a length pattern without binding e.g. `[<N]` to not iterate past the required length.

jcouv:
> For TryGetCount, we could possibly optimize a length pattern without binding e.g. [<N] to not iterate past the required length.

I think that's going to be difficult. Patterns currently rely on having a ready value as input.

> // slice pattern; ...
> helper.SkipToEnd() // always succeeds

I've made a small change along those lines. A slice pattern with some following ending patterns can call and check `Count()` once. That way we don't need to do redundant count checks on each ending patterns.

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+> **Open question**: Confirm that async enumerables are out-of-scope.  
+> **Open question**: Confirm that slice patterns with a sub_pattern (such as `..var x`) are out-of-scope.  
+
+Although a helper type is not necessary, it helps simplify and illustrate the logic.
+
+```
+class ListPatternHelper
+{
+  // Notes: 
+  // We could inline this logic to avoid creating a new type and to handle the pattern-based enumeration scenarios.
```
alrz, line 186: I think this is a pretty serious issue. So far, I tried to not depend on the buffer type for iteration exactly because of it.
This would be useful to reduce code size, but we should design around pattern-based case if we want to support it.

proposals/list-patterns.md
```diff
@@ -73,16 +73,67 @@ Notes:
...
+Patterns at the start of the *list_pattern* — that are before the `..` *slice_pattern* if one is present, or all otherwise — are matched against the elements produced at the start of the enumeration.  
+If the collection does not produce enough elements to get a value corresponding to a starting pattern, the match fails. So the *constant_pattern* `3` in `{ 1, 2, 3, .. }` doesn't match when the collection has fewer than 3 elements.  
+Patterns at the end of the *list_pattern* (that are following the `..` *slice_pattern* if one is present) are matched against the elements produced at the end of the enumeration.  
+If the collection does not produce enough elements to get values corresponding to the ending patterns, the *splice_pattern* does not match. So the *splice_pattern* in `{ 1, .., 3 }` doesn't match when the collection has fewer than 2 elements.  
+A *list_pattern* without a *splice_pattern* only matches if the number of elements produced by complete enumeration and the number of patterns are equals. So `{ _, _, _ }` only matches when the collection produces exactly 3 elements.
```
alrz, line 103: typo: *splice_pattern* -> *slice_pattern* (a few of these throughout the document)

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
 ```
 The *input type* for the *slice_pattern* is the return type of the underlying `this[Range]` or `Slice` method with two exceptions: For `string` and arrays, `string.Substring` and `RuntimeHelpers.GetSubArray` will be used, respectively.
 
+#### Lowering on enumerable type
+
+> **Open question**: Confirm that async enumerables are out-of-scope.  
+> **Open question**: Confirm that slice patterns with a sub_pattern (such as `..var x`) are out-of-scope.  
+
+Although a helper type is not necessary, it helps simplify and illustrate the logic.
```
alrz, line 180: I gather the following is not the rigorous codegen spec and we may make changes as we go through the impl?

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+  helper.TryGetStartElement(1, out var element1) && element1 is 1
+}
+```
+
+`collection is { .., 3, 4 }` is lowered to
+```
+@{
+  var helper = new ListPatternHelper(collection, 0, 2);
+
+  helper.Count() > 2 && // `..` with 2 ending patterns
```
alrz, line 288: `Count() >= 2`

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+```
+
+`collection is { 1, 2, .., 3, 4 }` is lowered to
+```
+@{
+  var helper = new ListPatternHelper(collection, 2, 2);
+
+  helper.TryGetStartElement(index: 0, out var element0) && element0 is 1 &&
+  helper.TryGetStartElement(1, out var element1) && element1 is 2 &&
+  helper.Count() > 4 && // `..` with 2 starting patterns and 2 ending patterns
```
alrz, line 301: `Count() >= 4`

proposals/list-patterns.md
```diff
@@ -73,16 +73,67 @@ Notes:
...
+If the collection does not produce enough elements to get a value corresponding to a starting pattern, the match fails. So the *constant_pattern* `3` in `{ 1, 2, 3, .. }` doesn't match when the collection has fewer than 3 elements.  
+Patterns at the end of the *list_pattern* (that are following the `..` *slice_pattern* if one is present) are matched against the elements produced at the end of the enumeration.  
+If the collection does not produce enough elements to get values corresponding to the ending patterns, the *splice_pattern* does not match. So the *splice_pattern* in `{ 1, .., 3 }` doesn't match when the collection has fewer than 2 elements.  
+A *list_pattern* without a *splice_pattern* only matches if the number of elements produced by complete enumeration and the number of patterns are equals. So `{ _, _, _ }` only matches when the collection produces exactly 3 elements.
+
+Note that those implicit checks for number of elements in the collection are unaffected by the collection type being *countable*. So `{ _, _, _ }` will not make use of `Length` or `Count` even if one is available.
```
alrz, line 105: I think if the type is countable we can just use it e.g. if we have Length but not an indexer, we match  Length while enumerating for elements.

proposals/list-patterns.md
```diff
@@ -73,16 +73,67 @@ Notes:
...
+
+If an enumerator is produced when a *list_pattern* is applied to an enumerable type and that enumerator is disposable it will be disposed when a top-level pattern containing the *list_pattern* successfully matches, or when none of the patterns match (in the case of a `switch` statement or expression). It is possible for an enumerator to be disposed more than once and the enumerator must ignore all calls to `Dispose` after the first one.
+```
+// any enumerator used to evaluate this switch statement is disposed at the indicated locations
+_ = collection switch
+{
+  { 1 } => /* here */  ...,
+  _ => /* here */ ...,
+};
+/* here too, with a spilled try/finally around the switch expression */
```
alrz, line 134: This isn't quite clear to me. DAG lowring happens in it's own block, so we only need one try/finally per enumerator, (provided we have a single node per each after simplification), from there we emit jumps to the target code section.

alrz:
To facilitate early failures, we calculate max/min length based on the pattern itself and the length pattern so that we can check **as we iterate**. Therefore, we require this set to be contiguous (e.g. `[1 or 3]` is an error on enumerables). I think it helps to mention that in the spec.

proposals/list-patterns.md
jcouv: Right.

This rough codegen was mostly to convince myself that it would fit in the DAG binding design (evaluation steps, test steps, ...). Overall I expect it will fit okay.

The parts I'm not sure yet:
- how do we account for evaluation steps returning a value or not? (in contrast, property accesses always succeed)
- who is responsible for the end-of-enumeration count check (when exiting the list-pattern, like `{ 1, 2 }`)?
- can we offer enough guarantees in the DAG binding that we don't need to cache starting values at all? (those values would get cached into temps if they are enumeratoed in the right order)
- is there enough helper logic that we could extract to a BCL type?
- how do we keep track of enumerators to dispose?

proposals/list-patterns.md
jcouv: My thinking was that if we're enumerating to the end anyways, we might as well use the enumerated count that we've accumulated for the check at the end of the list-pattern.
But it's true that we could omit tailing discard patterns. So `{ 1, _, _ }` could be "check first element, check Count == 3".

On the other hand, imagine that the `Count` property needs to enumerate again from the start.

I'll add an open issue.

proposals/list-patterns.md
jcouv: You're right. I'm not sure how to best represent this. What I'm trying to illustrate is that we dispose the enumerators as early as possible and in every case (even if an exception is thrown somewhere in a pattern evaluation).

jcouv:
> we require this set to be contiguous (e.g. [1 or 3] is an error on enumerables

Two concerns:
1. The evaluation model for patterns is "get a value" followed by "check whether the value matches the pattern". But what you're proposing is that we have special understanding of some patterns inside of length-patterns so that we can check before we have the (final) value.
2. From a language perspective, why place such a restriction? The `[1 or 3]` pattern seems fine to me.

I'll add that as an open question.

proposals/list-patterns.md
alrz: >how do we account for evaluation steps returning a value or not?

We either use an evaluation followed by a standard test, or a dedicated test node. As long as these won't affect subsumption checking, we can do away with simple eval nodes. In the prototype, I tried the first approach initially, but with trailing patterns, had to introduce a test node to disallow MoveNext after a slice which can happen in multi-arm matches.

>who is responsible for the end-of-enumeration count check 

If the enumerator has a length (e.g. via TryGetNonEnumeratedCount), we check that first for an early failure, otherwise we only count if there's a length pattern. This happens right after we enumerate the sequence to the end. While we're doing it, we don't want to exceed the maxLength (if specified) and after that, we want to ensure we've reached the minLength (either inferred or specified). This cover the "early failure" that was discussed in the last LDM.

Note that a simple `helper.Count() is P` always enumerate the sequence to the end which is not what we want, therefore we need to compute the actual value set that we're testing for. Since we operate on a range of values, something like `[1 or 3]` causes incorrect results and should be gated.

>can we offer enough guarantees in the DAG binding that we don't need to cache starting values at all? (those values would get cached into temps if they are enumerated in the right order)

That is correct. Starting patterns won't need any kind of buffer since we have a temp per each.

>is there enough helper logic that we could extract to a BCL type?

I think if we use a standard buffer type we can indeed propose to it BCL. Currently I used a generic fixed-size stack with no additional logic.


>how do we keep track of enumerators to dispose?

Since we can't "join" leaf nodes in a DAG, we just wrap the "rest" of the lowering in a try/finally, starting at each GetEnumerator. As a consequence enumerators might stack up. For instance,  in `{ enum1: {0}, enum2: {0} }`, both enumerators won't get disposed until after the whole pattern is executed.

proposals/list-patterns.md
alrz: > "check first element, check Count == 3".

We always check Count first if it's available without enumerating, subsequently `MoveNext() is 1 && MoveNext() && MoveNext() && !MoveNext()` will be emitted to match elements which also tests for the length in itself (in case it wasn't available)

proposals/list-patterns.md
alrz: >What I'm trying to illustrate is that we dispose the enumerators as early as possible and in every case

As I explained above, the earliest we could do that is at the end of the DAG lowering. At least that's what I could think of.

proposals/list-patterns.md
```diff
@@ -73,16 +73,67 @@ Notes:
...
 
-A *list_pattern* is compatible with any type that is *countable* as well as *indexable* - it has an accessible indexer that takes an `Index` or `int` argument. If both indexers are present, the former is preferred.
+A *list_pattern* is compatible with any type that is *countable* as well as *indexable* — it has an accessible indexer that takes an `Index` or `int` argument. If both indexers are present, the former is preferred.  
+A *list_pattern* is also compatible with any type that is *enumerable*.
 
-A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* - it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.
+A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.  
+A *slice_pattern* without a sub_pattern is also compatible with any type that is *enumerable*.
```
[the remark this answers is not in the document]
alrz: > `sub-pattern`

"sub" is not a word. If I read that correctly, it means "substitute" by itself.

proposals/list-patterns.md
jcouv: If we're going to emit 4 `MoveNext` evaluations and enumerate completely, then there is no need to also call `Count`/`Length`. We don't need to check the count twice.

alrz: We dont *need* to do it but if we can get count without enumerating we'll check it first to fail early ifit doesn't match. This is from last LDM.

jcouv: What I'm saying is that when there is a count check for the closing brace `}` of the list-pattern on an enumerable, that should use the enumerated count, since we'll have enumerated anyways. Similarly, the check that occurs to confirm that we had at least 3 elements before applying the `3` would use the enumeration.

I think that in your framing, all I'm saying is that before we check whether pattern `3` matches, we must have had 3 successful `MoveNext()` calls and that at the closing brace `}` we check that `!MoveNext()`. Those checks just rely on the enumerator, there was no need to call `Count` on a countable type.

> If the enumerable is actually backed by a concrete list type, we don't need to do any enumeration of the enumerable to check the length pattern.

Yes, for the length-pattern (such as `[<6]`) it is fine to introduce a third concept of count, a "try-count", or extend the concept of the `Count` API, but that's orthogonal to what I'm saying about the list-pattern. 
Note that the `{ 1, 2, 3, .., 10 }` does not have an implicit end-of-list-pattern count, since it has a `..`. But it does have implicit checks that we could get at least 4 elements before we apply the pattern `10`.

I hope that makes sense.
By the way, if you're already down the implementation path and feel there is a more natural way to spec this, feel free to update this doc. I'll merge later today so you can stack a PR.

jcouv:
@dotnet/roslyn-compiler for review.

proposals/list-patterns.md
jcouv: Sorry, one more point of clarification.
This (`Note that those implicit checks for number of elements in the collection are unaffected by the collection type being *countable*. So `{ _, _, _ }` will not make use of `Length` or `Count` even if one is available.`) is merely calling out a consequence of what is stated in the preceding paragraph:

> If the collection does not produce enough elements to get a value corresponding to a starting pattern, the match fails. [...]

> If the collection does not produce enough elements to get values corresponding to the ending patterns, the *slice_pattern* does not match. [...]

> A *list_pattern* without a *slice_pattern* only matches if the number of elements produced by complete enumeration and the number of patterns are equals. [...]

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+  private EnumeratorType enumerator;
+  private int count;
+  private ElementType[] startBuffer;
+  private ElementType[] endCircularBuffer;
+
+  public ListPatternHelper(EnumerableType enumerable, int startPatternsCount, int endPatternsCount)
+  {
+    count = 0;
+    enumerator = enumerable.GetEnumerator();
+    startBuffer = startPatternsCount == 0 ? null : new ElementType[startPatternsCount];
```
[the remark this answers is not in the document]
333fred: And having an `IDisposable` implementation for returning them.

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+  }
+}
+```
+
+`collection is [3]` is lowered to
+```
+@{
+  var helper = new ListPatternHelper(collection, 0, 0);
+
+  helper.Count() == 3
```
333fred, line 258: This does not feel good. We probably want to take a parameter of some kind an avoid enumerating the whole enumerable if we pass an upper limit, 3 in this case.

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+```
+
+`collection is { 0, 1 }` is lowered to
+```
+@{
+  var helper = new ListPatternHelper(collection, 2, 0);
+
+  helper.TryGetStartElement(index: 0, out var element0) && element0 is 0 &&
+  helper.TryGetStartElement(1, out var element1) && element1 is 1 &&
+  helper.Count() == 2
```
333fred, line 269: Same comment about enumerable length.

proposals/list-patterns.md
```diff
@@ -121,6 +172,141 @@ expr.Length is >= 2
...
+  helper.TryGetStartElement(1, out var element1) && element1 is 1
+}
+```
+
+`collection is { .., 3, 4 }` is lowered to
+```
+@{
+  var helper = new ListPatternHelper(collection, 0, 2);
+
+  helper.Count() >= 2 && // `..` with 2 ending patterns
```
333fred, line 288: This one isn't so bad, since we need enumerate the whole thing anyway.

proposals/list-patterns.md
jcouv: In the current design of patterns in general, we get a value then we check the value. I've added an open question on length-pattern cutting enumerations short (checking non-final value in some way), following Ali's suggestion.

proposals/list-patterns.md
jcouv: Thanks. We only need to check we're at the end of the enumeration.

proposals/list-patterns.md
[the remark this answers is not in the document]
jcouv: Right. Since we'll need to track disposal of enumerator we have some options here.
Stackalloc might be an option too if we inline this logic (then the state is just locals to the expression).

---
In reply to: [601740384](https://github.com/dotnet/csharplang/pull/4575#discussion_r601740384) [](ancestors = 601740384)

333fred (approved):
LGTM (commit 11). We'll keep whacking on that enumeration question in the implementation.

proposals/list-patterns.md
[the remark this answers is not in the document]
333fred: I don't think we'll want to use stackalloc (unconditionally, anyway). This might be a place where we'd like a runtime helper like https://github.com/dotnet/runtime/issues/25423.

PR #4073, opened 2020-10-27
Initial spec-work for single-line-namespace

CyrusNajmabadi:
Initial spec changes for https://github.com/dotnet/csharplang/issues/137.  To be taken to LDM.

There is an open question on what to do here for top-level-statements.  We could consider either:

```
compilation_unit
    : extern_alias_directive* using_directive* global_attributes? (single_line_namespace_declaration | (statement* namespace_member_declaration*))
    ;
```

or

```
compilation_unit
    : extern_alias_directive* using_directive* global_attributes? statement* (single_line_namespace_declaration | namespace_member_declaration*))
    ;
```

The effective difference between the two is if you can have a file with top-level-statements and a single-line namespace. I personally feel like it's a bit weird to have both.  But i suppose no more weird than top level statements followed by a namespace-declaration.

--

Update: LDM decision is we will not allow top level statements with a single-line namespace.

PR #4581, opened 2021-03-24
Initial specification for file-scoped-namespaces

CyrusNajmabadi:
Supersedes https://github.com/dotnet/csharplang/pull/4073


proposals/csharp-10.0/file-scoped-namespaces.md
```diff
@@ -0,0 +1,129 @@
...
+    class C
+    {
+    }
+}
+```
+
+Specifically, a *single_line_namespace_declaration* is treated the same as a *namespace_declaration* at the same location in the *compilation_unit* with the same *qualified_identifier*.  The *using_directive*s and *type_declaration*s of that *single_line_namespace_declaration* act as if they were declared in the same order inside the *namespace_body* of that *namespace_declaration*.
+
+A source file cannot contain both a *single_line_namespace_declaration* and a *namespace_declaration*.  A source file cannot contain multiple *single_line_namespace_declaration*s. A *compilation_unit* cannot contain both a *single_line_namespace_declaration* and any *statement* at the top level.
+
```
CyrusNajmabadi, line 126: note: this last paragraph is just for convenience/clarity.  The grammar enforces this.  However, i'm not sure how much we like to help explain these sorts of restrictions.

proposals/csharp-10.0/file-scoped-namespaces.md
```diff
@@ -0,0 +1,129 @@
+## File Scoped Namespaces
```
jcouv, line 1: nit: some places say "file-scoped" but others say "single-line". Consider unifying

jcouv (approved):
LGTM Thanks (iteration 9)

proposals/csharp-10.0/file-scoped-namespaces.md
CyrusNajmabadi: indeed.

CyrusNajmabadi:
@cston Can you ptal?  Thanks! :)

333fred (approved):

proposals/csharp-10.0/file-scoped-namespaces.md
```diff
@@ -0,0 +1,129 @@
...
+
+namespace X.Y.Z
+{
+    // types
+}
+```
+
+However, both these forms force the user to indent the majority of their code and add a fair amount of ceremony for what is effectively a very basic concept.  This affects clarity, uses horizontal and vertical space, and is often unsatisfying for users both used to C# and coming from other languages (which commonly have less ceremony here).
+
+The primary goal of the feature therefor is to meet the needs of the majority of the ecosystem with less unnecessary boilerplate.
```
cston, line 50: >therefor [](start = 32, length = 8)

Typo?

proposals/csharp-10.0/file-scoped-namespaces.md
```diff
@@ -0,0 +1,129 @@
...
+A *namespace_declaration* consists of the keyword `namespace`, followed by a namespace name and body, optionally followed by a semicolon.
+A *file_scoped_namespace_declaration* consists of the keyword `namespace`, followed by a namespace name, a semicolon and an optional list of *extern_alias_directive*s, *using_directive*s and *type_declaration*s.
+
+```antlr
+namespace_declaration
+    : 'namespace' qualified_identifier namespace_body ';'?
+    ;
+    
+file_scoped_namespace_declaration
+    : 'namespace' qualified_identifier ';' extern_alias_directive* using_directive* type_declaration*
```
cston, line 89: >extern_alias_directive* using_directive* type_declaration* [](start = 43, length = 58)

I'm surprised the `file_scoped_namespace_declaration` contains the directives and type declarations that follow. Is that necessary, and does that mean the corresponding syntax nodes would have the same nesting?

proposals/csharp-10.0/file-scoped-namespaces.md
```diff
@@ -0,0 +1,129 @@
...
+    ;
+
+... unchanged ...
+```
+
+... unchanged ...
+
+the two namespace declarations above contribute to the same declaration space, in this case declaring two classes with the fully qualified names `N1.N2.A` and `N1.N2.B`. Because the two declarations contribute to the same declaration space, it would have been an error if each contained a declaration of a member with the same name.
```
cston, line 97: >the [](start = 0, length = 3)

"The"

jaredpar:
@chsienki

proposals/csharp-10.0/file-scoped-namespaces.md
CyrusNajmabadi: the purpose of the grammar here is to just define strictly what people can and cannot write.  In this case, i found this to be hte simplest and clearest way to write the grammar to achieve the design we wanted.

> and does that mean the corresponding syntax nodes would have the same nesting?

As it turns out, yes.  My approach here (which is similar towhat we did in typescript) is that there is now a BaseNamespaceDeclaration, which has hte name, and members. And then there are two subclasses representing the two different syntactic forms.  This approach makes the implementation (for both the IDE and compiler) extremely trivial as the same 'tree shape' (esp. wrt to parent/children nodes) is present for both trees.

cston: > My approach here (which is similar towhat we did in typescript) is that there is now a BaseNamespaceDeclaration, which has hte name, and members. And then there are two subclasses representing the two different syntactic forms.

But it doesn't feel like there is a syntactic relationship between the `namespace identifier;` and the subsequent directives and types.

The only analogy I can think of to `namespace identifier;` in terms of how it applies to subsequent code is a `using var identifier = expr;`, and the statements following the using statement are not children of using statement syntax node.

---
In reply to: [601267615](https://github.com/dotnet/csharplang/pull/4581#discussion_r601267615) [](ancestors = 601267615)

CyrusNajmabadi: > But it doesn't feel like there is a syntactic relationship between the namespace identifier; and the subsequent directives and types.

It's going to come down to perspective. Either approaches are valid, however one is vastly more costly. For example, every piece of code that walks upwards to determine what namespace it is in must now be revised to not only do the standard walk, but also then got the compilation unit *and* then check if there is one of these directives. 

As mentioned earlier, we took this approach in ts with great success. We could take this other approach but it will mean a higher cost for both codebases.

proposals/csharp-10.0/file-scoped-namespaces.md
CyrusNajmabadi: Actually how it is in the spec:

image

https://github.com/dotnet/csharplang/blob/main/spec/namespaces.md#namespace-declarations

CyrusNajmabadi:
@chsienki ?

chsienki (approved):

chsienki:
@CyrusNajmabadi Weird, I thought I already approved this a couple days ago, but it doesn't show. 

My only comment was about the nesting / not nesting child nodes. Personally I prefer the approach you're taking, but agree with @cston's concerns and want to be absolutely sure that's the route we want to take as we don't have any precedent for doing it previously.

CyrusNajmabadi:
> , but agree with @cston's concerns and want to be absolutely sure that's the route we want to take as we don't have any precedent for doing it previously.

I'm not sure we don't have precedent.  In many cases we've reused nodes, but used different kinds, to allow for easier migration.  This is very similar.  Conceptually (and practically) this is literally what the node means (as per the spec).  i.e. in teh spec we say: this node is effectively *identical* to this other form.

We also follow all teh syntactic invariant rules here.  i.e. these are not AST nodes.  THey still remain concrete syntax nodes with every bit of the semantics around that we expect.  The only difference here is simply that we have some nodes as children that one could decide they wanted to be siblings.  However, IMO, the parent/child relationship here *is* appropriate given what tehse nodes represent.  if i have `namespace N; class C { }`.  the class *is* a child of that namespace :)

Issue #4394, opened 2021-02-05
Proposal: Extended property patterns


alrz:
# Extended property patterns

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

## Summary
[summary]: #summary

Allow property subpatterns to reference nested members.

## Motivation
[motivation]: #motivation

As discussion in https://github.com/dotnet/csharplang/discussions/4114 when you want to match a child property, nesting another recursive pattern adds too much noise which will hurt readability with no real advantage.

## Detailed design
[design]: #detailed-design

### Semantics

A pattern of the form `{ Property1.Property2: pattern }` is exactly equivalent to `{ Property1: { Property2: pattern } }`.

This will include the null check for `Nullable<T>` values as it is the case for the expanded form, so we only see the underlying type's members when we dot off of a property pattern.

Repeated member paths are allowed. Under the hood, such member accesses are simplified to be evaluated once.

### Syntax

Currently a `SubpatternSyntax` uses `NameColonSyntax` which is not able to hold a chain of identifiers.

There are several avenues we could take:

1. **Accept a generic `ExpressoinSyntax` in place of the name**
This would particularly help with parsing and avoid lookaheads, since patterns and expressions have a common parsing path, we can start with the pattern and if we get to a colon, we just adjust it as the name and continue to parse the actual pattern.
2. **Accept a generic `NameSyntax` in place of the name**
This seems to be the "correct" node to use, but is more expensive to parse. Also. if and when we introduce *indexer patterns* we probably want to relax property patterns to enable nesting those as well e.g. `{ Property1[0].Property2[1]: pattern }`. In that case we'll need to go with option (1) for forward compatibility.
3. **Accept a new `QualifiedNameColonSyntax` in addition to `NameColonSyntax`**
This option doesn't quite help with anything other than keeping the API consistent.

In the first two options, we're able to hold a simple identifier so we might want to consider deprecating `NameColon` API.

### Alternatives

We could use `{ P1?.P2: p }` syntax to make the implicit null check more apparent. But then we still want to consider `P1.P2` in case `?.` doesn't apply e.g. for structs. Since patterns shouldn't throw in regular usage and we should emit the null-check anyways, that distinction would seem to be unnecessary.

## Design meetings

None.

alrz:
cc @CyrusNajmabadi

canton7:
[...]

PR #4585, opened 2021-03-25
Add extended-property-patterns.md

alrz:
Championed issue: https://github.com/dotnet/csharplang/issues/4394
Prototype: https://github.com/dotnet/roslyn/pull/52139
Proposal: [Rendered](https://github.com/alrz/csharplang/blob/patch-3/proposals/extended-property-patterns.md)

### DIFF
--- a/proposals/extended-property-patterns.md
+++ b/proposals/extended-property-patterns.md
@@ -3,7 +3,14 @@
 ## Summary
 [summary]: #summary
 
-Allow property subpatterns to reference nested members.
+Allow property subpatterns to reference nested members, for instance:
+```cs
+if (e is MethodCallExpression { Method.Name: "MethodName" })
+``` 
+Instead of:
+```cs
+if (e is MethodCallExpression { Method: { Name: "MethodName" } })
+```
 
 ## Motivation
 [motivation]: #motivation
@@ -13,7 +20,7 @@ When you want to match a child property, nesting another recursive pattern adds 
 ## Detailed design
 [design]: #detailed-design
 
-The pattern syntax is modified as follow:
+The [*property_pattern*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md#property-pattern) syntax is modified as follow:
 
 ```diff
 property_pattern
--- a/proposals/extended-property-patterns.md
+++ b/proposals/extended-property-patterns.md
@@ -1,0 +1,45 @@
+  
+
+# Extended property patterns
+
+## Summary
+[summary]: #summary
+
+Allow property subpatterns to reference nested members.
+
+## Motivation
+[motivation]: #motivation
+
+When you want to match a child property, nesting another recursive pattern adds too much noise which will hurt readability with no real advantage.
+
+## Detailed design
+[design]: #detailed-design
+
+The pattern syntax is modified as follow:
+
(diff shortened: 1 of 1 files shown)


proposals/extended-property-patterns.md
```diff
@@ -0,0 +1,44 @@
...
++ : subpattern_name ':' pattern
+  ;
+
++subpattern_name
++ : identifier 
++ | subpattern_name '.' identifier
++ ;
+```
+
+The receiver for each name lookup is the stripped type of the previous member starting from the *input type* of the *property_pattern*.
```
alrz, line 38: >stripped type

This is probably not the correct term to use. How do we refer to such type?

jcouv: Here's the type of language used elsewhere in the spec: `Determine the type T0. If T is a nullable type, T0 is its underlying type, otherwise T0 is equal to T.`

---
In reply to: [601701046](https://github.com/dotnet/csharplang/pull/4585#discussion_r601701046) [](ancestors = 601701046)

proposals/extended-property-patterns.md
```diff
@@ -0,0 +1,44 @@
+# Extended property patterns
+
+## Summary
+[summary]: #summary
+
+Allow property subpatterns to reference nested members.
```
jcouv, line 6: Let's add an example to illustrate up front.

Also, let's add a link to property patterns spec somewhere in this doc: [https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md#property-pattern](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md#property-pattern)
Take example on it to write this one.

alrz: Didn't find anything in the proposal to include. 

Added an example that I encountered first that was kinda annoying :)

jcouv (approved):
LGTM Thanks (iteration 7)

jcouv:
@333fred @dotnet/roslyn-compiler for review.

proposals/extended-property-patterns.md
```diff
@@ -0,0 +1,51 @@
...
++ : identifier 
++ | subpattern_name '.' identifier
++ ;
+```
+
+The receiver for each name lookup is the type of the previous member *T0*, starting from the *input type* of the *property_pattern*. if *T* is a nullable type, *T0* is its underlying type, otherwise *T0* is equal to *T*.
+
+For example, a pattern of the form `{ Prop1.Prop2: pattern }` is exactly equivalent to `{ Prop1: { Prop1: pattern } }`.
+
+Note that this will include the null check for `Nullable<T>` values as it is the case for the expanded form, so we only see the underlying type's members when we dot off of a name in a *property_pattern*.
```
333fred, line 49:
```suggestion
Note that this will include the null check when `T` is a nullable value type or a reference type. This null check means that the nested properties available will be the properties of `T0`, not of `T`.
```

333fred (approved):

333fred:
Thanks @alrz!

jcouv:
FYI following discussion with Mads, I added a proposal to mirror this in `with` and object creation (as a separate but related feature):
https://github.com/dotnet/csharplang/issues/4587

PR #4588, opened 2021-03-26
Update interpolated string proposal with feedback

333fred:
* Split the TryFormat methods into TryFormatBaseString and TryFormatInterpolationHole. Naming is definitely subject to review, but hopefully the concept is clear.
* Added new open questions for logging scenarios after review with the ILogger folks.
* Added leeway for async rules in using the interpolated string builder for string types.
* Added open question for disposal of builder types.

### DIFF
--- a/proposals/improved-interpolated-strings.md
+++ b/proposals/improved-interpolated-strings.md
@@ -44,16 +44,16 @@ picturing this:
 // The builder that will actually "build" the interpolated string"
 public ref struct TraceLoggerParamsBuilder
 {
-    public static bool Create(int baseLength, int formatHoleCount, Logger logger, out TraceLoggerParamsBuilder builder)
+    public static TraceLoggerParamsBuilder Create(int baseLength, int formatHoleCount, Logger logger, out bool builderIsValid)
     {
         if (!logger._logLevelEnabled)
         {
-            builder = default;
-            return false;
+            builderIsValid = false;
+            return default;
         }
 
-        builder = TraceLoggerParamsBuilder(baseLength, formatHoleCount, logger.EnabledLevel);
-        return true;
+        builderIsValid = true;
+        return TraceLoggerParamsBuilder(baseLength, formatHoleCount, logger.EnabledLevel);
     }
 
     // Storage for the built-up string
@@ -99,16 +99,18 @@ var name = "Fred Silberberg";
 logger.LogTrace($"{name} will never be printed because info is < trace!");
 
 // This is converted to:
+var name = "Fred Silberberg";
 var receiverTemp = logger;
-_ = TraceLoggerParamsBuilder.Create(baseLength: 47, formatHoleCount: 1, receiverTemp, out var builder) &&
-    builder.TryFormatInterpolationHole("Fred Silberberg") &&
+var builder = TraceLoggerParamsBuilder.Create(baseLength: 47, formatHoleCount: 1, receiverTemp, out var builderIsValid);
+_ = builderIsValid &&
+    builder.TryFormatInterpolationHole(name) &&
     builder.TryFormatBaseString(" will never be printed because info is < trace!");
 receiverTemp.LogTrace(builder);
 ```
 
-Here, because `TraceLoggerParamsBuilder` has static method called `Create` with the correct parameters, including an out param that is the type
-the `LogTrace` call was expecting, we say that the interpolated string has an implicit builder conversion to that parameter, and it lowers to the pattern shown above.
-The specese needed for this is a bit complicated, and is expanded below.
+Here, because `TraceLoggerParamsBuilder` has static method called `Create` with the correct parameters and returns the type the `LogTrace` call was expecting,
+we say that the interpolated string has an implicit builder conversion to that parameter, and it lowers to the pattern shown above. The specese needed for this
+is a bit complicated, and is expanded below.
 
 #### Builder type applicability
 
@@ -153,20 +155,21 @@ to be converted to an _applicable\_interpolated\_string\_builder\_type_. There a
 1. A method argument is converted as part of determining applicable function members (covered below), or
 2. Given an _interpolated\_string\_expression_ `S` being converted to type `T`, the following is true:
     * `T` is an _applicable\_interpolated\_string\_builder\_type_, and
-    * `T` has an accessible static bool-returning method `Create` that takes 2 int parameters and 1 out parameter of type `T`, in that order.
+    * One of the following is true:
+        * `T` has an accessible static T-returning method `Create` that takes 2 int parameters and 1 out parameter of type `bool`, in that order.
+        * `T` has an accessible static T-returning method `Create` that takes 2 int parameters, in that order.
 
-We want to make `Create` a static method with an `out` parameter for 2 reasons:
+We want to make `Create` a static method because by making it a `static` method instead of a constructor, we allow the implementation to pool builders if it so decides to.
+If we limited the pattern to constructors, then the implementation would be required to always return new instances.
 
-1. By making it a `static` method instead of a constructor, we allow the implementation to pool builders if it so decides to. If we limited the pattern to constructors,
-then the implementation would be required to always return new instances.
-2. By making the builder an `out` parameter we allow the `Create` method to return a bool indicating whether to continue formatting, which is useful
-for scenarios like the logger above that may want to skip any argument evaluation at all for cases when the log level isn't enabled.
+Additionally, by returning the builder type instead of requiring it to go in an out parameter, we marginally improve the codegen and significantly simplify the rules around
+ref struct lifetimes vs the traditional .NET `TryX` pattern, and we expect a number of these builder types to be ref structs.
+
+This can and will be impacted by abstract statics in interfaces for generic contexts. We will need to make sure the interactions are considered and tested.
 
 **Open Question**
 
-Should we swap the out parameter and return? IE, make the return the builder type, and have an optional bool out parameter for whether the builder was successfully created? This
-might be easier for authoring and have better perf characteristics. While it does differ from the standard `TryX` pattern in C#, this method isn't intended for direct usage, so
-diverging may be fine.
+Need to confirm this pattern.
 
 #### Applicable function member adjustments
 
@@ -176,13 +179,17 @@ as follows (a new sub-bullet is added at the front of each section, in bold):
 A function member is said to be an ***applicable function member*** with respect to an argument list `A` when all of the following are true:
 *  Each argument in `A` corresponds to a parameter in the function member declaration as described in [Corresponding parameters](expressions.md#corresponding-parameters), and any parameter to which no argument corresponds is an optional parameter.
 *  For each argument in `A`, the parameter passing mode of the argument (i.e., value, `ref`, or `out`) is identical to the parameter passing mode of the corresponding parameter, and
-   *  **for an interpolated string argument to a value parameter when `A` is an instance method or static extension method invoked in reduced from, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `Ai` succeeds with 1 invocable member. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
+   *  **for an interpolated string argument to a value parameter when `A` is an instance method or static extension method invoked in reduced from, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and one of the following overload resolutions succeeds. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
+      * **Overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `bool` succeeds with 1 invocable member, and the return type of that member is `Ai`, or**
+      * **Overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters and the receiver type of `A` succeeds with 1 invocable member, and the return type of that member is `Ai`.**
    *  for a value parameter or a parameter array, an implicit conversion ([Implicit conversions](conversions.md#implicit-conversions)) exists from the argument to the type of the corresponding parameter, or
    *  for a `ref` or `out` parameter, the type of the argument is identical to the type of the corresponding parameter. After all, a `ref` or `out` parameter is an alias for the argument passed.
 For a function member that includes a parameter array, if the function member is applicable by the above rules, it is said to be applicable in its ***normal form***. If a function member that includes a parameter array is not applicable in its normal form, the function member may instead be applicable in its ***expanded form***:
 *  The expanded form is constructed by replacing the parameter array in the function member declaration with zero or more value parameters of the element type of the parameter array such that the number of arguments in the argument list `A` matches the total number of parameters. If `A` has fewer arguments than the number of fixed parameters in the function member declaration, the expanded form of the function member cannot be constructed and is thus not applicable.
 *  Otherwise, the expanded form is applicable if for each argument in `A` the parameter passing mode of the argument is identical to the parameter passing mode of the corresponding parameter, and
-   *  **for an interpolated string argument to a fixed value parameter or a value parameter created by the expansion when `A` is an instance method or static extension method invoked in reduced form, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `Ai` succeeds with 1 invocable member. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
+   *  **for an interpolated string argument to a fixed value parameter or a value parameter created by the expansion when `A` is an instance method or static extension method invoked in reduced form, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and one of the following overload resolutions succeeds. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
+      * **Overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `bool` succeeds with 1 invocable member, and the return type of that member is `Ai`, or**
+      * **Overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters and the receiver type of `A` succeeds with 1 invocable member, and the return type of that member is `Ai`.**
    *  for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([Implicit conversions](conversions.md#implicit-conversions)) exists from the type of the argument to the type of the corresponding parameter, or
    *  for a `ref` or `out` parameter, the type of the argument is identical to the type of the corresponding parameter.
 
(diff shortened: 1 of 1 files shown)


proposals/improved-interpolated-strings.md
```diff
@@ -35,16 +35,16 @@ convenient interpolation syntax.
...
+and has an out parameter of the type of original method's parameter and that type has instance `TryFormatX` methods can be invoked for every part of the interpolated
 string, then we lower the interpolation using that, instead of into a traditional call to `string.Format(formatStr, args)`. A more concrete example is helpful for
 picturing this:
 
 ```cs
 // The builder that will actually "build" the interpolated string"
 public ref struct TraceLoggerParamsBuilder
 {
-    public static bool GetInterpolatedStringBuilder(int baseLength, int formatHoleCount, Logger logger, out TraceLoggerParamsBuilder builder)
+    public static bool Create(int baseLength, int formatHoleCount, Logger logger, out TraceLoggerParamsBuilder builder)
```
jaredpar, line 47: One of the suggestions from the previous PR was to make these methods return the builder type and have the `bool` as the `out` parameter. Is there a reason we didn't change to use that form? 

The current form is less efficient performance wise (returning complex types is faster than initializing them by `ref / out`) and also has implications with `ref struct` life time rules. A non-trivial amount of our interpolated builders are going to be `ref struct` and once they are passed as  `ref / out` then we need to do a lot of due dilligence around the life time rules here. 

The details here are mundane but important and the `ref` field proposal probably does the best job of explaining them 

https://github.com/dotnet/csharplang/blob/main/proposals/low-level-struct-improvements.md#provide-parameter-does-not-escape-annotations

In short though we would need to make guarantees about the lifetimes of the temporaries that we pass as the `out` here. No such guarantees need to be done if we simple return them here.

proposals/improved-interpolated-strings.md
```diff
@@ -138,15 +153,21 @@ to be converted to an _applicable\_interpolated\_string\_builder\_type_. There a
...
 then the implementation would be required to always return new instances.
-2. By making the builder an `out` parameter we allow the `GetInterpolatedStringBuilder` method to return a bool indicating whether to continue formatting, which is useful
+2. By making the builder an `out` parameter we allow the `Create` method to return a bool indicating whether to continue formatting, which is useful
 for scenarios like the logger above that may want to skip any argument evaluation at all for cases when the log level isn't enabled.
 
+**Open Question**
+
+Should we swap the out parameter and return? IE, make the return the builder type, and have an optional bool out parameter for whether the builder was successfully created? This
+might be easier for authoring and have better perf characteristics. While it does differ from the standard `TryX` pattern in C#, this method isn't intended for direct usage, so
+diverging may be fine.
```
jaredpar, line 169: The current form also means you need to spend more time specing out `ref struct` lifetime issues. I'm happy to sit down and work with you on this but I'd prefer to pick the form that doesn't require us to do this :smile:

proposals/improved-interpolated-strings.md
```diff
@@ -35,16 +35,16 @@ convenient interpolation syntax.
 We introduce a new builder pattern that can represent an interpolated string passed as an argument to a method. The simple English of the pattern is as follows:
 
 When an _interpolated\_string\_expression_ is passed as an argument to a method, we look at the type of the parameter. If the parameter type has a static method
-`GetInterpolatedStringBuilder` that can invoked with 2 int parameters, `baseLength` and `formatHoleCount`, optionally takes a parameter the receiver is convertible to,
-and has an out parameter of the type of original method's parameter and that type has instance `TryFormat` methods can be invoked for every part of the interpolated
+`Create` that can invoked with 2 int parameters, `baseLength` and `formatHoleCount`, optionally takes a parameter the receiver is convertible to,
+and has an out parameter of the type of original method's parameter and that type has instance `TryFormatX` methods can be invoked for every part of the interpolated
```
jaredpar, line 39: >  If the parameter type has a static method `Create` that can invoked with 2 int parameters

This will impact the static virtual interfaces proposal as type parameters will become eligable in certain cases to be interpolated builders. Can we add a tracking issue in the proposal that we need to handle this at some point in the future.

Don't want to design it, just want to make sure we are tracking that these features overlap in this space.

proposals/improved-interpolated-strings.md
```diff
@@ -100,36 +100,51 @@ logger.LogTrace($"{name} will never be printed because info is < trace!");
 
 // This is converted to:
 var receiverTemp = logger;
-_ = TraceLoggerParamsBuilder.GetInterpolatedStringBuilder(baseLength: 47, formatHoleCount: 1, receiverTemp, out var builder) &&
-    builder.TryFormat("Fred Silberberg") &&
-    builder.TryFormat(" will never be printed because info is < trace!");
+_ = TraceLoggerParamsBuilder.Create(baseLength: 47, formatHoleCount: 1, receiverTemp, out var builder) &&
+    builder.TryFormatInterpolationHole("Fred Silberberg") &&
```
jaredpar, line 104:
```suggestion
    builder.TryFormatInterpolationHole(name) &&
```

Looks like a typo from the previous iteration of the proposal

proposals/improved-interpolated-strings.md
333fred: Discussed this offline. While LDM hasn't approved that change currently, I'll go ahead and make the change, and leave an open question to confirm it.

333fred:
@jaredpar addressed your feedback. Anything else?

jaredpar (approved):

333fred:
@stephentoub @chsienki please take a look.

stephentoub:
I will try to look tonight.  Sorry for the delay.

proposals/improved-interpolated-strings.md
```diff
@@ -99,37 +99,54 @@ var name = "Fred Silberberg";
...
+with an identifier of `TryFormatInterpolationHole` and parameter of the type of `Si` and name  succeeds, and contains a single instance method that returns a `bool`.
 * For every _regular\_balanced\_text_ component of `S` (`Si`) with an _interpolation\_format_ component and no _constant\_expression_ (alignment) component, overload resolution
-with an identifier of `TryFormat` and parameter types of `Si` and `string`(in that order) succeeds, and contains a single instance method that returns a `bool`.
+with an identifier of `TryFormatInterpolationHole` and parameter types of `Si` and `string`(in that order) succeeds, and contains a single instance method that returns a `bool`.
 * For every _regular\_balanced\_text_ component of `S` (`Si`) with a _constant\_expression_ (alignment) component and no _interpolation\_format_ component, overload resolution
-with an identifier of `TryFormat` and parameter types of `Si` and `int` (in that order) succeeds, and contains a single instance method that returns a `bool`.
+with an identifier of `TryFormatInterpolationHole and parameter types of `Si` and `int` (in that order) succeeds, and contains a single instance method that returns a `bool`.
```
chsienki, line 125:
```suggestion
with an identifier of `TryFormatInterpolationHole` and parameter types of `Si` and `int` (in that order) succeeds, and contains a single instance method that returns a `bool`.
```

proposals/improved-interpolated-strings.md
```diff
@@ -99,37 +99,54 @@ var name = "Fred Silberberg";
...
 * For every _regular\_balanced\_text_ component of `S` (`Si`) with an _interpolation\_format_ component and a _constant\_expression_ (alignment) component, overload resolution
-with an identifier of `TryFormat` and parameter types of `Si`, `int`, and `string` (in that order) succeeds, and contains a single instance method that returns a `bool`.
+with an identifier of `TryFormatInterpolationHole` and parameter types of `Si`, `int`, and `string` (in that order) succeeds, and contains a single instance method that returns a `bool`.
 
-Note that these rules do not permit extension methods for the `TryFormat` calls. We could consider enabling that if we choose, but this is analogous to the enumerator
+The rest of this proposal will use `TryFormatX` to refer to either of `TryFormatBaseString` or `TryFormatInterpolationHoles` in cases when both are applicable.
```
chsienki, line 129: I found this confusing that you used `TryFormatX` above in the 'simple english' section, but didn't define it until now. I would also suggest using `TryFormat...` as (IMO) its clearer what you mean, and isn't a valid identifier.

proposals/improved-interpolated-strings.md
```diff
@@ -99,37 +99,54 @@ var name = "Fred Silberberg";
...
+
+Note that these rules do not permit extension methods for the `TryFormatX` calls. We could consider enabling that if we choose, but this is analogous to the enumerator
 pattern, where we allow `GetEnumerator` to be an extension method, but not `Current` or `MoveNext()`.
 
-These rules _do_ permit default parameters for the `TryFormat` calls, which will work with things like `CallerLineNumber` or `CallerArgumentExpression` (when supported by
+These rules _do_ permit default parameters for the `TryFormatX` calls, which will work with things like `CallerLineNumber` or `CallerArgumentExpression` (when supported by
 the language).
 
+We have separate overload lookup rules for base elements vs interpolation holes because some builders will want to be able to understand the difference between the components
```
chsienki, line 137: I totally get the use case for this, but it seems complicated, especially for simpler cases that don't care. Could we alternatively have a single method that takes a `struct` with all the info in it?

```csharp
struct InteroplatedStringBuilderHoleInfo
{
    string content;
    bool isBaseString;
    string? interpolationFormat;
    int? alignment;
}
```

That way, the given builder can just look at whatever info is useful to it, we get a place to add extra info (like the caller expression if we go that route) and reduces duplication in the builder when not needed?

proposals/improved-interpolated-strings.md
```diff
@@ -138,14 +155,21 @@ to be converted to an _applicable\_interpolated\_string\_builder\_type_. There a
 1. A method argument is converted as part of determining applicable function members (covered below), or
 2. Given an _interpolated\_string\_expression_ `S` being converted to type `T`, the following is true:
     * `T` is an _applicable\_interpolated\_string\_builder\_type_, and
-    * `T` has an accessible static bool-returning method `GetInterpolatedStringBuilder` that takes 2 int parameters and 1 out parameter of type `T`, in that order.
+    * One of the following is true:
+        * `T` has an accessible static T-returning method `Create` that takes 2 int parameters and 1 out parameter of type `bool`, in that order.
+        * `T` has an accessible static T-returning method `Create` that takes 2 int parameters, in that order.
+
+We want to make `Create` a static method because by making it a `static` method instead of a constructor, we allow the implementation to pool builders if it so decides to.
```
chsienki, line 162:
```suggestion
By requiring `Create` to be a `static` method instead of a constructor, we allow the implementation to pool builders if it so decides to.
```

proposals/improved-interpolated-strings.md
```diff
@@ -155,13 +179,17 @@ as follows (a new sub-bullet is added at the front of each section, in bold):
...
-   *  **for an interpolated string argument to a value parameter when `A` is an instance method or static extension method invoked in reduced from, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and overload resolution on `Ai` with the identifier `GetInterpolatedStringBuilder` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `Ai` succeeds with 1 invocable member. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
+   *  **for an interpolated string argument to a value parameter when `A` is an instance method or static extension method invoked in reduced from, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and one of the following overload resolutions succeeds. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
```
chsienki, line 182:
```suggestion
   *  **for an interpolated string argument to a value parameter when `A` is an instance method or static extension method invoked in reduced form, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and one of the following overload resolutions succeeds. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
```

proposals/improved-interpolated-strings.md
```diff
@@ -155,13 +179,17 @@ as follows (a new sub-bullet is added at the front of each section, in bold):
...
+      * **Overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `bool` succeeds with 1 invocable member, and the return type of that member is `Ai`, or**
```
chsienki, line 183: Move these sub-bullets to `An interpolated string argument applicable in this way...`

stephentoub:
I'm still not clear on what we're choosing to incorporate into the doc in these PR updates and what we're holding off on.  I'd love it if the proposal doc outlined what we currently believe the end result should be, even if it hasn't been agreed on by LDM and even if alternatives are spelled out in open issues.  For example, we know we need to be able to pass in additional arguments, but that / the attribute for that aren't part of the proposal, just called out as an open issue.  And we know (or at least I've strongly argued) that we should have the format methods be able to optionally return void vs bool, but that's still just called out as an open issue.  Can we do one update to get the proposal into the state you / I believe it should be, and then just iterate based on differing decisions from LDM?  Otherwise, I don't know what it means to provide feedback on the doc, don't know how to interpret what you're pushing for vs what you're not, etc.

proposals/improved-interpolated-strings.md
```diff
@@ -155,13 +179,17 @@ as follows (a new sub-bullet is added at the front of each section, in bold):
...
+   *  **for an interpolated string argument to a fixed value parameter or a value parameter created by the expansion when `A` is an instance method or static extension method invoked in reduced form, the type of the corresponding parameter is an _applicable\_interpolated\_string\_builder\_type_ `Ai`, and one of the following overload resolutions succeeds. An interpolated string argument applicable in this way is said to be immediately converted to the corresponding parameter type with an _implicit\_string\_builder\_conversion_. Or,**
+      * **Overload resolution on `Ai` with the identifier `Create` and a parameter list of 2 int parameters, the receiver type of `A`, and an out parameter of type `bool` succeeds with 1 invocable member, and the return type of that member is `Ai`, or**
```
chsienki, line 191: Same comment as above

proposals/improved-interpolated-strings.md
stephentoub: That will make every call more expensive, in order to set up the struct just to be able to pass in the relevant info.  That struct would also need to be either emitted for every different T or be generic on T, and the latter would prevent ref structs (e.g. spans) from being used in holes, which is important to be able to support.

proposals/improved-interpolated-strings.md
```diff
@@ -99,37 +99,54 @@ var name = "Fred Silberberg";
...
+with an identifier of `TryFormatInterpolationHole` and parameter types of `Si`, `int`, and `string` (in that order) succeeds, and contains a single instance method that returns a `bool`.
 
-Note that these rules do not permit extension methods for the `TryFormat` calls. We could consider enabling that if we choose, but this is analogous to the enumerator
+The rest of this proposal will use `TryFormatX` to refer to either of `TryFormatBaseString` or `TryFormatInterpolationHoles` in cases when both are applicable.
+
+Note that these rules do not permit extension methods for the `TryFormatX` calls. We could consider enabling that if we choose, but this is analogous to the enumerator
 pattern, where we allow `GetEnumerator` to be an extension method, but not `Current` or `MoveNext()`.
 
-These rules _do_ permit default parameters for the `TryFormat` calls, which will work with things like `CallerLineNumber` or `CallerArgumentExpression` (when supported by
+These rules _do_ permit default parameters for the `TryFormatX` calls, which will work with things like `CallerLineNumber` or `CallerArgumentExpression` (when supported by
```
stephentoub, line 134: I understand this for TryFormatInterpolationHole.  What does it mean to support CallerArgumentExpression with TryFormatBaseString?

proposals/improved-interpolated-strings.md
```diff
@@ -99,37 +99,54 @@ var name = "Fred Silberberg";
...
 
+We have separate overload lookup rules for base elements vs interpolation holes because some builders will want to be able to understand the difference between the components
+that were interpolated and the components that were part of the base string.
+
+**Open Question**
+
+Some scenarios, like structured logging, want to be able to provide names for interpolation elements. For example, today a logging call might look like
+`Log("{name} bought {itemCount} items", name, items.Count);`. The names inside the `{}` provide important structure information for loggers that help with ensuring output
+is consistent and uniform. Some cases might be able to reuse the `:format` component of an interpolation hole for this, but many loggers already understand format specifiers
+and have existing behavior for output formatting based on this info. Is there some syntax we can use to enable putting these named specifiers in?
```
stephentoub, line 145: If such a syntax existed, how would the data be passed to TryFormatInterpolationHole?

proposals/improved-interpolated-strings.md
```diff
@@ -138,14 +155,21 @@ to be converted to an _applicable\_interpolated\_string\_builder\_type_. There a
 1. A method argument is converted as part of determining applicable function members (covered below), or
 2. Given an _interpolated\_string\_expression_ `S` being converted to type `T`, the following is true:
     * `T` is an _applicable\_interpolated\_string\_builder\_type_, and
-    * `T` has an accessible static bool-returning method `GetInterpolatedStringBuilder` that takes 2 int parameters and 1 out parameter of type `T`, in that order.
+    * One of the following is true:
+        * `T` has an accessible static T-returning method `Create` that takes 2 int parameters and 1 out parameter of type `bool`, in that order.
+        * `T` has an accessible static T-returning method `Create` that takes 2 int parameters, in that order.
```
stephentoub, line 160: What about the other state that's passed in?  I know you have attributed arguments in open issues (even though I think it's really important), but your examples allow passing in the receiver, and that doesn't seem to be factored in here...?

proposals/improved-interpolated-strings.md
```diff
@@ -224,15 +251,15 @@ public ref struct InterpolatedStringBuilder
         ArrayPool<char>.Shared.Return(_array);
         Return result;
     }
-    public bool TryFormat(string s) => TryFormat((ReadOnlySpan<char>)s);
-    public bool TryFormat(ReadOnlySpan<char> s)
+    public bool TryFormatBaseString(string s) => TryFormatInterpolationHole((ReadOnlySpan<char>)s);
```
stephentoub, line 254: Nit: the cast is no longer needed

stephentoub (approved):

333fred:
> I'm still not clear on what we're choosing to incorporate into the doc in these PR updates and what we're holding off on.

The intent of this PR is to have enough specified that we can move forward with `InterpolatedStringBuilder` (or whatever else we decide to name it). The attribute changes are going to require more spec work, and I'd like to actually make progress on the implementation of strings at this point. There are still a number of open questions related to more complete support (including out-of-order evaluation and argument resolution), and what we have should be fine for the initial bit.

proposals/improved-interpolated-strings.md
333fred: That's what I would expect. Not that I expect anyone to _do_ that, but we would support it.

proposals/improved-interpolated-strings.md
333fred: Presumably, there would need to be an overload that took it.

proposals/improved-interpolated-strings.md
333fred: Neither of these is applicable there. Those are bullet one, covered in applicable function member.

proposals/improved-interpolated-strings.md
333fred: I specifically didn't do that because you can't have sub-bullets in the middle of a bullet.

proposals/improved-interpolated-strings.md
333fred: Same answer as above.

proposals/improved-interpolated-strings.md
333fred: Not if we have a generic. `TryFormatInterpolationHole<T>(T)` will be preferred over `TryFormatInterpolationHole(ReadOnlySpan<char>)`

stephentoub: That's ok.  It's just an optimization at that point.  The generic handed a string will work just fine.

proposals/improved-interpolated-strings.md
stephentoub: I don't follow, but you think it's fine, so ok.

333fred: Basically: these bullets don't allow for state to come in. The version that allows state to come in is covered in the applicable function member section.

chsienki (approved):

Issue #4018, opened 2020-10-16
Proposal: Permit variable declarations under disjunctive patterns


alrz:
# Permit variable declarations under disjunctive patterns

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

## Summary
[summary]: #summary

Allow pattern variables to be declared in different mutually exclusive patterns. This is the part of [pattern-matching proposal](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-9.0/patterns3.md) that did not make it into C# 9.0.

## Motivation
[motivation]: #motivation

The feature would reduce code duplication where switch expression arms or switch section bodies could be shared for a number of patterns.
```cs
switch (e)
{
    case (int i, 0):
        Use(i); break;
    case (0, int i):
        Use(i); break;
}

switch (e)
{
    case (int i, 0):
    case (0, int i):
        Use(i); break;
}
```
```cs
e switch
{
    (int i, 0) => i,
    (0, int i) => i,
}

e switch
{
    (int i, 0) or (0, int i) => i,
}
```

Today we can workaround this via breaking each pattern to separate arms or sections. 

For `is` expressions that is even worse as we need to use a different name for each branch.
```cs
if (e is (int i, 0) or (0, int i))
  Use(i);

if (e is (int i, 0))
  Use(i);
else if (e is (0, int j))
  Use(j);
```

## Detailed design
[design]: #detailed-design

For this to work, we basically require each mutually exclusive pattern to define the same set of variables with the same types, otherwise they would not be definitely assigned in that particular code path. This includes `or` patterns and each case of a switch section.

As a consequence, a pattern variable may be definitely assigned in a `when` clause but not in the switch section body, for example.

```cs
  case (int x, 0) a when Use(x, a): // ok
  case (0, int x) b when Use(x, b): // ok
     Use(x); // ok
     Use(a); // error; not definitely assigned
     Use(b); // error; not definitely assigned
     break;
```

Pattern variables may be recursively multiply defined on each side:

```cs
if (e is { A: (int x, 0) or (0, int x) } or
         { B: (int x, 0) or (0, int x) })
{
    Use(x);
}  
```

Under the hood, it is the same local being assigned in each pattern.

Since we require an identical type for each declaration, the narrowed type would not be affected by this change.

## Further considerations

[...]

PR #4592, opened 2021-03-28
Add pattern-variables.md

alrz:
Championed issue: https://github.com/dotnet/csharplang/issues/4018
Prototype: https://github.com/dotnet/roslyn/compare/main...alrz:pattern-variables
Proposal: [Rendered](https://github.com/alrz/csharplang/blob/patch-4/proposals/pattern-variables.md)

### DIFF
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -104,7 +104,7 @@ For a *disjunctive_pattern* of the form `left_pattern or right_pattern`:
 - The definite assignment state of *v* after *disjunctive_pattern* is determined by:
     - If the state of *v* after *left_pattern* is definitely assigned, and the state of *v* after *right_pattern* is conditionally or unconditionally definitely assigned, then the state of *v* after *disjunctive_pattern* is definitely assigned.
     - Otherwise, if the state of *v* after *right_pattern* is definitely assigned, and the state of *v* after *left_pattern* is "definitely assigned when true", then the state of *v* after *disjunctive_pattern* is definitely assigned.
-    - Otherwise, if the state of *v* after *right_pattern* is definitely assigned or "definitely assigned when false", and the state of *v* after *right_pattern* is conditionally or unconditionally definitely assigned,  then the state of *v* after *disjunctive_pattern* is "definitely assigned when false".
+    - Otherwise, if the state of *v* after *right_pattern* is definitely assigned or "definitely assigned when false", and the state of *v* after *left_pattern* is conditionally or unconditionally definitely assigned, then the state of *v* after *disjunctive_pattern* is "definitely assigned when false".
     - Otherwise, if the state of *v* after *left_pattern* is "definitely assigned when true", and the state of *v* after *right_pattern* is "definitely assigned when true", then the state of *v* after *disjunctive_pattern* is "definitely assigned when true".
     - Otherwise, the state of *v* after *disjunctive_pattern* is not definitely assigned.
 
@@ -113,7 +113,7 @@ For a *conjunctive_pattern* of the form `left_pattern and right_pattern`:
 - The definite assignment state of *v* after *conjunctive_pattern* is determined by:
     - If the state of *v* after *left_pattern* is definitely assigned, and the state of *v* after *right_pattern* is conditionally or unconditionally definitely assigned, then the state of *v* after *conjunctive_pattern* is definitely assigned.
     - Otherwise, if the state of *v* after *right_pattern* is definitely assigned, and the state of *v* after *left_pattern* is "definitely assigned when false", then the state of *v* after *conjunctive_pattern* is definitely assigned.
-    - Otherwise, if the state of *v* after *right_pattern* is definitely assigned or "definitely assigned when true", and the state of *v* after *right_pattern* is conditionally or unconditionally definitely assigned, then the state of *v* after *conjunctive_pattern* is "definitely assigned when true".
+    - Otherwise, if the state of *v* after *right_pattern* is definitely assigned or "definitely assigned when true", and the state of *v* after *left_pattern* is conditionally or unconditionally definitely assigned, then the state of *v* after *conjunctive_pattern* is "definitely assigned when true".
     - Otherwise, if the state of *v* after *left_pattern* is "definitely assigned when false", and the state of *v* after *right_pattern* is "definitely assigned when false", then the state of *v* after *conjunctive_pattern* is "definitely assigned when false".
     - Otherwise, the state of *v* after *conjunctive_pattern* is not definitely assigned.
 
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -87,8 +87,8 @@ For a *var_pattern* of the form `var variable_designation`:
 
 - The definite assignment state of *v* after *var_pattern* is determined by:
 
-    - The state of *v* after *var-pattern* is definitely assigned if *variable_designation* is a *single_variable_designation*.
-    - Otherwise, the state of *v* after *var-pattern* is "definitely assigned when true".
+    - The state of *v* after *var_pattern* is definitely assigned if *variable_designation* is a *single_variable_designation*.
+    - Otherwise, the state of *v* after *var_pattern* is "definitely assigned when true".
 
 #### General rules for pattern variables in simple patterns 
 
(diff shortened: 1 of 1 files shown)


alrz:
cc @333fred @RikkiGibson

I haven't started the impl yet particularly because DAG construction eliminates bindings after a false branch. 
See https://github.com/dotnet/roslyn/pull/49369#issuecomment-808706659

Opening up this early due to possible interaction with improved def assignment proposal.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,139 @@
...
+
+- The state of *v* after *is_pattern_expression* is the same as the state of *v* after *pattern*.
+
+#### General rules for pattern variables in simple patterns 
+
+Note: This section is unchanged and included for the sake of completeness.
+
+The following rules applies to any *primary_pattern* that declares a variable:
+
+- The state of *v* is definitely assigned after *primary_pattern* if the pattern is irrefutable; permitting use after:
```
alrz, line 84: This rule should probably be moved to the previous section under *is_pattern_expression*. Because only at the top-level we can determine if a pattern is irrefutable (at least that's how it works today).

proposals/pattern-variables.md
```diff
@@ -0,0 +1,139 @@
...
+
+- Otherwise, the state of *v* is "definitely assigned when true" after *primary_pattern*; permitting use after:
+
+    ```cs
+    if (o is int x)
+    ```
+
+#### Definite assignment rules for pattern variables in pattern combinators
+
+Note: This section is derived by definite assignment rules for expressions with an exception: for patterns we strictly require both operands to assign the same set of variables. For example, even though `e is var x or 1` is logically valid, the variable `x` is not definitely assigned. That said, some of cases might still result in impossible or irrefutable patterns by other measures.
```
alrz, line 100: I'm not sure if we should actually do this. There's no harm in allowing such cases to maintain symmetry between expressions and pattern variables. In any nonsensical case, you will have unassigned variables or an invalid pattern.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,139 @@
...
+
+For a *disjunctive_pattern* of the form `left_pattern or right_pattern`:
+
+- The definite assignment state of *v* after *disjunctive_pattern* is determined by:
+    - If the state of *v* after *left_pattern* is definitely assigned, and the state of *v* after *right_pattern* is conditionally or unconditionally definitely assigned, then the state of *v* after *disjunctive_pattern* is definitely assigned.
```
RikkiGibson, line 105: I didn't understand why we specified "conditionally or unconditionally definitely assigned" here. That is, I don't know why `right_pattern` can conditionally assign a variable, but the variable can be fully definitely assigned after the containing `disjunctive_pattern`.

It might also be useful to specify what "conditionally definitely assigned" means. I think it just means "definitely assigned when true, or definitely assigned when false". But the spec doesn't mention the term, so it would be useful to make that completely clear.

RikkiGibson:
I made a fairly quick pass over the proposal and didn't notice anywhere that it would conflict with "improved definite assignment". However, it will probably be a little while before I can review this proposal in detail, as our feature load for C# 10 is getting a little bit high.

proposals/pattern-variables.md
alrz: See https://github.com/dotnet/csharplang/pull/4592#discussion_r602993190

This portion can be removed to make it fairly identical to logical and/or expressions.

Other than that there's nothing new here. In practice, roughly speaking, we just apply binary operator rules to binary patterns.

alrz:
Added a link to the (very early) prototype. Will finish it up once this got checked in.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,150 @@
...
+
+    ```cs
+    if (o is int x)
+    ```
+
+#### Definite assignment rules for pattern variables in pattern combinators
+
+Note: This section is derived by definite assignment rules for expressions.
+
+> **Open question**: We should confirm if we want to allow patterns like `var x or 1`.
```
alrz, line 111: Ok. the correctness of such pattern would depend directly on the order of evaluation in patterns which is undefined.

So I think we should require both sides to define the same set of variables but that could be spec'ed under redeclaration rules instead of changing def assignment section.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,158 @@
+# Variable declarations under disjunctive patterns
+
+## Summary  
+
+## Motivation
```
jcouv, line 5: nit: Let's remove the unused sections

proposals/pattern-variables.md
```diff
@@ -0,0 +1,158 @@
...
+### Scope
+
+#### Pattern variable redeclaration
+
+Introducing a new concept similar to *name hiding*, but these names can possibly reference either of variables based on the result of the pattern-matching at runtime.
+
+There's two degrees of enforcement as to whether variables may be redeclared:
+
+- Pattern variables within pattern boundaries *must* be redeclared because assignment of such variables depend on the order of evaluation which is undefined for patterns.
+- Pattern variables outside pattern boundaries *can* be redeclared, definitive assignment rules for the containing expression specify where such variables are usable.
```
jcouv, line 18: typo: definite

jcouv:
Tagging @333fred since he's championing the proposal.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,158 @@
...
+For a *logical_or_expression* of the form `left_expr || right_expr`:
+- Pattern variables declared in *left_expr* **can be redeclared** in the *right_expr*; permitting:
+
+  ```cs
+  e is (0, var x) || e is (var x, 0)
+  ```
+  Note: This is the expression variation of the previous case.
+
+For a *negated_pattern* of the form `not pattern_operand`:
+- Pattern variables declared in the *pattern_operand* **must be redeclared** inside any other *negated_pattern* in the containing pattern; permitting:
```
jcouv, line 36: "... inside any other *negated_pattern*..."

What if there's more than one nested negated_pattern? `e is not (0, var x) and not not (0, 0)`

proposals/pattern-variables.md
jcouv: Consider linking to those original/unmodified sections instead of replicating them here.

jcouv (approved):
Done with review pass (iteration 13). Only minor comments

proposals/pattern-variables.md
alrz: There's no one place for this in previous proposals. We have three documents each of which mentioned definite assignment in prose.

In any case, since we don't want to add expressions to the mix just yet, I think we no longer need to specify definite assignment in this format.

proposals/pattern-variables.md
alrz: Haven't thought this through yet. To keep the proposal simple, we can add this at some later point if there's enough support to allow such patterns.

proposals/pattern-variables.md
```diff
@@ -2,157 +2,69 @@
...
+	Use(x);
+	break;
+}
+```
+Instead of:
+
+```cs
+if (e is (int x, 0))
+    Use(x);
+else
```
333fred, line 27: Missing the condition.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,73 @@
...
+        Use(x);
+        break;
+}
+```
+
+## Detailed design
+
+Variables *must* be redeclared under all disjuncitve patterns because assignment of such variables depend on the order of evaluation which is undefined in the context of pattern-matching.
+
+- In a *disjunctive_pattern*, pattern variables declared under the left-hand-side must be redeclared under the right-hand-side.
```
333fred, line 45: The way this is worded, technically I could declare variables under just the right-hand side. Perhaps "variables declared on one side must be redeclared on the other side"?

333fred (approved):
@jcouv please take a look.

proposals/pattern-variables.md
```diff
@@ -0,0 +1,73 @@
+# Variable declarations under disjunctive patterns
+
+## Summary  
+
+Allow variable declarations under `or` patterns and across `case` labels in a `switch` section.
+
+## Motivation
```
jcouv, line 7: I'd add some more practical (less abstract) motivating examples. 
Looking that the roslyn codebase, we often have a `switch` expression just to get some value out of different kinds of bound nodes. Many can be found just by searching for ["node switch"](https://grep.app/search?q=%20node%20switch&case=true&filter[repo][0]=dotnet/roslyn).
But I have to say looking at that search result, most of them don't look like they would benefit from this feature (either they use `when` clauses for different cases, or they call different property/method overloads in the different cases).

PR #4611, opened 2021-03-31
Add equality and conversion operators to static abstract members in interfaces

MadsTorgersen:
Proposed expansion of static abstract members in interfaces to support equality operators and user-defined conversions, as well ad default implementations

### DIFF
--- a/proposals/statics-in-interfaces.md
+++ b/proposals/statics-in-interfaces.md
@@ -196,9 +196,13 @@ An alternative approach would be to have "structural constraints" directly and e
     
 ## Default implementations
 
-An *additional* feature to this proposal is to allow static virtual members in interfaces to have default implementations, just as instance virtual members do. We're investigating this, but the semantics get very complicated: default implementations will want to call other static virtual members, but what syntax, semantics and implementation strategies should we use to ensure that those calls can in turn be virtual?
+An *additional* feature to this proposal is to allow static virtual members in interfaces to have default implementations, just as instance virtual members do. 
 
-This seems like a further improvement that can be done independently later, if the need and the solutions arise.
+One complication here is that default implementations would want to call other static virtual members "virtually". Allowing static virtual members to be called directly on the interface would require flowing a hidden type parameter representing the "self" type that the current static method really got invoked on. This seems complicated, expensive and potentially confusing.
+
+We discussed a simpler version which maintains the limitations of the current proposal that static virtual members can *only* be invoked on type parameters. Since interfaces with static virtual members will often have an explicit type parameter representing a "self" type, this wouldn't be a big loss: other static virtual members could just be called on that self type. This version is a lot simpler, and seems quite doable.
+
+However, it seems rare in practice that a default implementation would be beneficial, at least in our main driving scenario of numeric abstraction. Default implementations are *explicitly* implemented on implementing classes and structs, so they wouldn't result in a public member. Why would you want e.g. an operator implementation that only surfaces in a generic context, but is hidden on the concrete type? The main reason would be if we want to evolve some of the interfaces in a later release to e.g. expose more operators. In that case, we can add the language feature at that time.
 
 ## Virtual static members in classes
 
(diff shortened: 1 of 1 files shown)


proposals/statics-in-interfaces.md
```diff
@@ -56,20 +56,59 @@ Static interface members today are implicitly non-virtual, and do not allow `abs
...
+    
+    static abstract T P1 { get; set; }
+    static virtual T P2 { get => T.P1; set => T.P1 = value; }
+    
+    static abstract event Action E1;
+    static virtual event Action E2 { add => T.E1 += value; remove => T.E1 -= value; }
+    
     static abstract T operator +(T l, T r);
+    static virtual T operator -(T l, T r) => l;
+    static abstract bool operator ==(T l, T r);
```
AlekseyTs, line 80: At some point (not necessarily in his PR) it would be good to specify changes for requirements around signatures of operators, if any.

AlekseyTs:
> 

Should this section be removed now?

---
Refers to: proposals/statics-in-interfaces.md:235 in d80953e. [](commit_id = d80953e14d735f83acd374c5e4a3dc58ec9f787d, deletion_comment = False)

proposals/statics-in-interfaces.md
AlekseyTs: That is if "Operator restrictions" section is insufficient.

proposals/statics-in-interfaces.md
```diff
@@ -127,38 +174,38 @@ These requirements need to be relaxed so that a restricted operand is allowed to
...
 M<C>();  // Allowed: C is not an interface
-M<I<C>>(); // Disallowed: I is an interface
+M<I2<C>>(); // Disallowed: I is an interface
 ```
 
-## Accessing static abstract interface members
+## Accessing static virtual interface members
 
-A static abstract interface member `M` may be accessed on a type parameter `T` using the expression `T.M` when `T` is constrained by an interface `I` and `M` is an accessible static abstract member of `I`.
+A static virtual interface member `M` may be accessed on a type parameter `T` using the expression `T.M` when `T` is constrained (directly or indirectly) by an interface `I` and `M` is an accessible static virtual member of `I`.
```
AlekseyTs, line 200: >A static virtual interface member `M` may be accessed on a type parameter `T` using the expression `T.M` [](start = 0, length = 104)

Can they be accessed without qualification inside default implementation?

AlekseyTs:
> These requirements need to be relaxed so that a restricted operand is allowed to be of a type parameter that is constrained to `T`.

Is this relaxation applied to all operators, or only to virtual ones?

---
Refers to: proposals/statics-in-interfaces.md:173 in d80953e. [](commit_id = d80953e14d735f83acd374c5e4a3dc58ec9f787d, deletion_comment = False)

proposals/statics-in-interfaces.md
```diff
@@ -56,20 +56,59 @@ Static interface members today are implicitly non-virtual, and do not allow `abs
...
+    static virtual T operator -(T l, T r) => l;
+    static abstract bool operator ==(T l, T r);
+    static virtual bool operator !=(T l, T r) => !(l == r);
+    static abstract implicit operator T(string s);
+    static virtual explicit operator string(T t) => t.ToString();
 }
 ```
 
-***Open question:** Operators `==` and `!=` as well as the implicit and explicit conversion operators are disallowed in interfaces today. Should they be allowed?*
+Default implementations cannot be in the form of auto-properties or field-like events.
```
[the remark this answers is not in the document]
MadsTorgersen: The section went away again. :-D

proposals/statics-in-interfaces.md
MadsTorgersen: No. Only via a type parameter.

proposals/statics-in-interfaces.md
MadsTorgersen: I agree we should take another look at this.

PR #4640, opened 2021-04-11
Fix readonly property samples

meziantou:
The original code doesn't compile. `readonly` can be set on a getter or setter when only one of them is readonly. If all setters are readonly, `readonly` must be set on the property.

As this file is available in the documentation, I think this is better to keep it up to date.
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly#c-language-specification
- https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/readonly-instance-members

### DIFF
--- a/proposals/csharp-8.0/readonly-instance-members.md
+++ b/proposals/csharp-8.0/readonly-instance-members.md
@@ -51,9 +51,9 @@ public struct Vector2
         return tmp;
     }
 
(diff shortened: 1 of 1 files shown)


333fred (approved):

333fred:
Thanks @meziantou!

PR #4642, opened 2021-04-11
Update lambda improvements proposal

cston:
Update proposal based on recent discussions.

### DIFF
--- a/proposals/lambda-improvements.md
+++ b/proposals/lambda-improvements.md
@@ -156,9 +156,9 @@ int zero = ((int x) => x)(0); // ok
 ```
 
 ### Implicit conversions
-A consequence of inferring a natural type is that lambda expressions and method groups with natural type are implicitly convertible to `System.MulticastDelegate` and to base classes and interfaces implemented by `System.MulticastDelegate` (including `System.Delegate`, `System.Object` and `System.ICloneable`).
+A consequence of inferring a natural type is that lambda expressions and method groups with natural type are implicitly convertible to `System.MulticastDelegate` and to base classes and interfaces implemented by `System.MulticastDelegate` (including `System.Delegate`, `System.Object`, and `System.ICloneable`).
 
-If a natural type cannot be inferred, there is no implicit conversion to `System.Delegate` or base classes or interfaces.
+If a natural type cannot be inferred, there is no implicit conversion to `System.MulticastDelegate` or base classes or interfaces.
 ```csharp
 Delegate d1 = 1.GetHashCode; // ok
 Delegate d2 = 2.ToString;    // error: cannot convert to 'System.Delegate'; multiple 'ToString' methods
@@ -166,7 +166,7 @@ object o1 = (int x) => x;    // ok
 object o2 = x => x;          // error: cannot convert to 'System.Object'; no natural type for 'x => x'
 ```
 
-The compiler will also treat lambda expressions with natural type as implicitly convertible to `System.Linq.Expressions.Expression` as an expression tree. Base classes or interfaces implemented by `System.Linq.Expressions.Expression` are ignored when calculating conversions to expression trees.
+The compiler will also treat lambda expressions with natural type as implicitly convertible to `System.Linq.Expressions.Expression` as an expression tree. Base classes and interfaces implemented by `System.Linq.Expressions.Expression` are ignored when calculating conversions to expression trees.
 
 ### Overload resolution
 Overload resolution already prefers binding to a strongly-typed delegate over `System.Delegate`, and prefers binding a lambda expression to a strongly-typed `System.Linq.Expressions.Expression<TDelegate>` over the corresponding strongly-typed delegate `TDelegate`.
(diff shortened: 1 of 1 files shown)


proposals/lambda-improvements.md
```diff
@@ -41,26 +41,36 @@ app.MapAction([HttpPost("/")] ([FromBody] Todo todo) => todo);
...
+Parameter types are not required.
 ```csharp
-f = [MyAttribute] x => x;          // [MyAttribute]lambda
-f = [MyAttribute] (int x) => x;    // [MyAttribute]lambda
-f = [MyAttribute] static x => x;   // [MyAttribute]lambda
-f = [return: MyAttribute] () => 1; // [return: MyAttribute]lambda
+f = [A] () => { };        // [A]lambda
+f = [return:A] x => x;    // syntax error at '=>'
+f = [return:A] (x) => x;  // [A]lambda
+f = [A] static x => x;    // syntax error at '=>'
```
alrz, line 51: What is the correct way to do this? `[A] static (x) = x` or `static [A] (x) = x`? 

In C++ *capture list* always comes first (analogous to *attribute list*) but specifiers comes after params. So that probably wouldn't be useful. See https://en.cppreference.com/w/cpp/language/lambda

proposals/lambda-improvements.md
```diff
@@ -72,18 +82,21 @@ It should be valid to inspect the `MethodInfo` associated with a `Delegate` inst
...
-An explicit return type may be specified after the parameter list.
+An explicit return type may be specified before the parenthesized parameter list.
 ```csharp
-f = () : T => default;              // () : T
-f = x : short => 1;                 // <unknown> : short
-f = (ref int x) : ref int => ref x; // ref int : ref int
-f = static _ : void => { };         // <unknown> : void
+f = T () => default;                    // ok
+f = short x => 1;                       // syntax error at '=>'
+f = ref int (ref int x) => ref x;       // ok
```
alrz, line 89: IIRC trailing return types were also chosen for function pointers because this is hard to read.

Couldn't we stick to `->` just like C++ with some restriction as to when you can specify the return type? e.g. only when parameters are also explicitly typed. See https://en.cppreference.com/w/cpp/language/lambda

CyrusNajmabadi: We discussed this and decided we really wanted to keep things consistent with all other function-forms in the language.  Methods are 'return type first'.  Local functions a 'return type first'.  So we would like to keep consistent with that approach.  

> e.g. only when parameters are also explicitly typed

This is a good question.  I don't think we covered this particular aspect @cston .  Should we say that you are only allowed to supply the return type if you also supply types for all parameters?  On hte one hand, i like the symettry of either having it be explicit or inferred.  On hte other hand though, i can see cases where users would want the param types inferred, but want to supply the explicit return type because it can't be effectively inferred.

alrz: >Methods are 'return type first'. Local functions a 'return type first'.

I gave function pointers as an example where that's actually not true. Lambdas appear in the middle of an expression so if anything you'd expect the symmetry with the type it's being assigned to.

```cs
Func<A, B> M() { return (A p) -> B => e; }
Func<A, B> M() { return B(A p) => e; }
```
Even if I have to specify all the parameter types I'd def go with the former. That's just one way to disambiguate though.

bernd5: In my eyes both ways are reasonable - for short or built in types the "return type first syntax" looks quite good, too.
But on the other hand it looks quite similar to an invocation expression which decreases readability.

CyrusNajmabadi: > I gave function pointers as an example where that's actually not true.

Function pointers are a form of Type.  And it's true that in types we put return-last.  But lambdas are not really like types.  They're like an inline declaration of a method/local-function.  As such, we felt it should align most closely with that.  Same reasoning for making local functions look like methods.  

> But on the other hand it looks quite similar to an invocation expression which decreases readability.

Enh.  Only in the same way that a parenthesized lambda looks like a parenthesized expr.  You need to look forward to see if there are multiple params, or if you are followed by a `=>`.  Same as before.  So i think it's fine.  

Note that having the return type should be pretty rare.  only for cases where things break down and you really have no palatable alternative for the type to be understood.

proposals/lambda-improvements.md
```diff
@@ -72,18 +81,21 @@ It should be valid to inspect the `MethodInfo` associated with a `Delegate` inst
...
+f = static void (_) => { };             // ok
+f = async async (async async) => async; // ok
```
jcouv, line 90: Did LDM today say this should be an error, or could be an error?

proposals/lambda-improvements.md
```diff
@@ -127,60 +149,51 @@ var f7 = "".F1; // System.Action
...
-
-var fB = (IEnumerable<object> e, ref int i) => { }; // void DB$(IEnumerable<object>, ref int);
-fB = (IEnumerable<string> e, ref int i) => { };     // error: parameter type mismatch
+int zero = ((int x) => x)(0); // ok
 ```
 
-### Implicit conversion to `System.Delegate`
-A consequence of inferring a natural type is that lambda expressions and method groups with natural type are implicitly convertible to `System.Delegate`.
+### Implicit conversions
+A consequence of inferring a natural type is that lambda expressions and method groups with natural type are implicitly convertible to `System.Delegate` and to base classes and interfaces implemented by `System.Delegate` (such as `System.Object` and `System.ICloneable`).
```
jcouv, line 159: MulticastDelegate instead of Delegate?

jcouv (approved):
LGTM Thanks (iteration 6)

PR #4669, opened 2021-04-19
Natural language improvement

wuyuansushen:


YairHalberstadt:
Thanks for opening the PR! Efforts to fix spelling/grammar are always appreciated!

I think the original sentence was actually correct and intended:

> "c" and "result" have element names "f1" and "f2"

vs

> "c" and "result" have elements named "f1" and "f2"

The first is consistent with language throughout C# spec and proposals. E.g https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples#tuple-field-names which talks about "tuple field names".

PR #4675, opened 2021-04-20
Update ranges.md: length -> Length

Szune:
Extremely small fix to get the example code under "Implicit Range support" to run


333fred (approved):

333fred:
Thanks @Szune!

PR #4686, opened 2021-04-24
async is neccessary in Task Main entrypoint

wuyuansushen:
We need to explicitly use `return <Task>` statement to return a Task object without async modifier. It violates Task design principle (for asynchronous programing).

### DIFF
--- a/proposals/csharp-7.1/async-main.md
+++ b/proposals/csharp-7.1/async-main.md
@@ -48,14 +48,14 @@ static int Main(string[])
 We extend the list of allowed entrypoints to include:
 
 ```csharp
-static Task Main()
-static Task<int> Main()
-static Task Main(string[])
-static Task<int> Main(string[])
+static async Task Main()
+static async Task<int> Main()
+static async Task Main(string[])
+static async Task<int> Main(string[])
 ```
 
 To avoid compatibility risks, these new signatures will only be considered as valid entrypoints if no overloads of the previous set are present.
-The language / compiler will not require that the entrypoint be marked as `async`, though we expect the vast majority of uses will be marked as such.
+The language / compiler will require that the entrypoint be marked as `async`.
 
 When one of these is identified as the entrypoint, the compiler will synthesize an actual entrypoint method that calls one of these coded methods:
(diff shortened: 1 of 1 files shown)


yaakov-h:
The spec as described prior to this PR is accurate. The following code compiles and runs:

```csharp
using System.Threading.Tasks;

class Program
{
    static Task Main()
    {
        return Task.CompletedTask;
    }
}
```

And, as per the spec, the compiler emits a special `private static void <Main> => Main().GetAwaiter().GetResult()` method which is marked in IL with `.entrypoint`.

As such, the `async` modifier is not required in order to have a valid C# program as per the specification for the `async Main` feature. 

However, if you wish to actually have an `async Main` method, then you must declare it as `async`, otherwise it is just a `Main` method, not an `async Main` method.

wuyuansushen:
I used `ildasm.exe` to inspect `Mian` method without `async` keyword.
```
call       class [System.Runtime]System.Threading.Tasks.Task ConfigurationHost.Program::Main(string[])
callvirt   instance valuetype [System.Runtime]System.Runtime.CompilerServices.TaskAwaiter [System.Runtime]System.Threading.Tasks.Task::GetAwaiter()
call       instance void [System.Runtime]System.Runtime.CompilerServices.TaskAwaiter::GetResult()
```

The preceding IL proved that you mentioned `.GetAwaiter().GetResult()`.

But this private Main method's IL will be no different after added `async` to it.

What's most, `return Task` statement returns `Task` directly without `await` (`await` can't be added because `async` isn't set in method), It will cause some exceptions, such as [dispose services ahead of time intentionally](https://github.com/dotnet/docs/pull/23901#event-4637727413) I had reported in another dotnet/docs PR.

So, I conclude `async` is necessary for `async Main` method.

CyrusNajmabadi:
This change is not accurate or appropriate. `async` is not required here, and it's entirely reasonable for Main to just return a task directly.

CyrusNajmabadi:
> It will cause some exceptions, such as dispose services ahead of time intentionally I had reported in another dotnet/docs PR.

> So, I conclude async is necessary for async Main method.

The conclusion is incorrect. In that example you found, the code was incorrect. But it would have been incorrect even outside of Main. That doesn't mean then that having Main without async is then always incorrect.

PR #4703, opened 2021-04-27
Rework improved interpolated string builder proposal with LDM feedback.

333fred:
This updates the interpolated string proposal with the latest ldm feedback, and gets it ready for the remaining open questions to be asked tomorrow.

### DIFF
--- a/proposals/improved-interpolated-strings.md
+++ b/proposals/improved-interpolated-strings.md
@@ -42,6 +42,7 @@ can be invoked for every part of the interpolated string, then we lower the inte
 
 ```cs
 // The builder that will actually "build" the interpolated string"
+[InterpolatedStringBuilder(UseLazyComputation = true)]
 public ref struct TraceLoggerParamsBuilder
 {
     public static TraceLoggerParamsBuilder Create(int literalLength, int formattedCount, Logger logger, out bool builderIsValid)
@@ -86,7 +87,7 @@ public class Logger
     // Initialization code omitted
     public LogLevel EnabledLevel;
 
-    public void LogTrace(TraceLoggerParamsBuilder builder)
+    public void LogTrace([InterpolatedStringBuilderArguments("")]TraceLoggerParamsBuilder builder)
     {
         // Impl of logging
     }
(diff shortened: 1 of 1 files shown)


proposals/improved-interpolated-strings.md
```diff
@@ -112,100 +113,83 @@ Here, because `TraceLoggerParamsBuilder` has static method called `Create` with
...
+    }
+}
+```
 
-#### Interpolated string builder conversion
+This attribute is used by the compiler to determine if a type is a valid interpolated string builder type.
 
-We add a new implicit conversion type: The _implicit\_string\_builder\_conversion_. An _implicit\_string\_builder\_conversion_ permits an _interpolated\_string\_expression_
-to be converted to an _applicable\_interpolated\_string\_builder\_type_. There are 2 ways that this conversion can occur:
+The compiler also recognizes the `System.Runtime.CompilerServices.InterpolatedStringBuilderAttribute`:
```
AlekseyTs, line 140: It looks like we already said this at the beginning of the "New attributes section".

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+    then for every name `Argx` in the `Arguments` array of that attribute the compiler matches it to a parameter `px` that has the same name. The empty string is matched to the receiver
+    of `M1`.
+        * If any `Argx` is not able to be matched to a parameter of `M1`, or an `Argx` requests the receiver of `M1` and `M1` is a static method, an error is produced and no further
+        steps are taken.
+        * Otherwise, the type of every resolved `px` is added to the argument list, in the order specified by the `Arguments` array. These are all passed by value.
+    3. The final argument is a `bool`, passed as an `out` parameter.
+3. Traditional method invocation resolution is performed with method group `M` and argument list `A`.
+    * If a single-best method `F` was found and errors were reported during final validation, those errors are reported and no further steps are taken.
+    * If a single-best method `F` was found and no errors were reported during final validation, the result of overload resolution is `F`.
```
AlekseyTs, line 294: I think that this bullet point should be merged with the previous one. If we really want to say something about final validation, which I prefer we don't, we could say that final validation might still produce errors.

AlekseyTs: If the "final validation" mentioned here is the one specified as step 4, then it appears that we are referring to steps out of order

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+    * If no applicable methods were found, step 3 is retried, removing the final `bool` parameter from `A`. If this retry also finds no applicable members, an error is produced and
+    no further steps are taken.
+    * If no single-best method was found, the result of overload resolution is ambiguous, an error is produced, and no further steps are taken.
+4. Final validation on `F` is performed.
+    * If `F`'s return type is not a `T`, an error is produced and no further steps are taken.
+    * If any element of `A` occurred lexically after `i`, an error is produced and no further steps are taken.
+
+By requiring `Create` to be a `static` method instead of a constructor, we allow the implementation to pool builders if it so decides to.
```
AlekseyTs, line 302: It looks like the requirement for the method to be static isn't specified anywhere, but this sentence tries to explain the motivation behind the requirement.

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+#### `Append...` method overload resolution
+
+Given an _applicable\_interpolated\_string\_builder\_type_ `T` and an _interpolated\_string\_expression_ `i`, overload resolution for a set of valid `Append...` methods on `T` is
+performed as follows:
+
+1. If there are any _interpolated\_regular\_string\_character_ components in `i`:
+    1. Member lookup on `T` with the name `AppendLiteral` is performed. The resulting method group is called `Ml`.
+    2. The argument list `Al` is constructed with one value parameter of type `string`.
+    3. Traditional method invocation resolution is performed with method group `Ml` and argument list `Al`.
+        * If a single-best method `Fi` is found and no errors were found in final validation, the result of lookup is `Fi`.
```
AlekseyTs, line 319: Similar comment for the "final validation"

AlekseyTs: Also, is Fi a result of the lookup or result of the overload resolution?

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+        * Otherwise, an error is reported.
+2. For every _interpolation_ `ix` component of `i`:
+    1. Member lookup on `T` with the name `AppendFormatted` is performed. The resulting method group is called `Mf`.
+    2. The argument list `Af` is constructed:
+        1. The first parameter is the `expression` of `ix`, passed by value.
+        2. If `ix` contains a _constant\_expression_ component, then an integer parameter value parameter is added, with the name `alignment` specified.
+        3. If `ix` is followed by an _interpolation\_format_, then a string value parameter is added, with the name `format` specified.
+    3. Traditional method invocation resolution is performed with method group `Mf` and argument list `Af`.
+        * If a single-best method `Fi` is found and no errors were found in final validation, the result of lookup is `Fi`.
```
AlekseyTs, line 328: Similar comment as for the previous section.

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+        * Otherwise, an error is reported.
+3. Finally, for every `Fi` discovered in steps 1 and 2, final validation is performed:
+    1. If any `Fi` does not return `bool` by value or `void`, and error is reported.
```
AlekseyTs, line 331: Typo ? "and error"

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+
+Some scenarios, like structured logging, want to be able to provide names for interpolation elements. For example, today a logging call might look like
+`Log("{name} bought {itemCount} items", name, items.Count);`. The names inside the `{}` provide important structure information for loggers that help with ensuring output
+is consistent and uniform. Some cases might be able to reuse the `:format` component of an interpolation hole for this, but many loggers already understand format specifiers
+and have existing behavior for output formatting based on this info. Is there some syntax we can use to enable putting these named specifiers in?
+
+Some cases may be able to get away with `CallerArgumentExpression`, provided that support does land in C# 10. But for cases that invoke a method/property, that may not be
+sufficient.
+
+#### Lowering
```
AlekseyTs, line 353: Do we want to use term lowering? Is this just a way the conversion is performed? Like user-defined conversion is performed by calling the conversion operator, etc.

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+lowering for `i` is performed as follows:
+
+1. If the `InterpolatedStringBuilderAttribute` on `T` specifies false for `UseLazyComputation`, the following steps are taken:
+    1. Any arguments to `Fc` that occur lexically before `i` are evaluated and stored into temporary variables in lexical order.
+    2. All _interpolation_ components of `i` are evaluated in lexical order and stored into temporary variables.
+    3. `Fc` is called with the length of the interpolated string literal components, the number of _interpolation_ holes, any previously evaluated arguments, and an a `bool` out argument
+    (if `Fc` was resolved with one as the last parameter). The result is stored into a temporary value `ib`.
+    4. If `Fc` ended with a `bool` out argument, a check on that `bool` value is generated. If true, `Fa` will be called. Otherwise, they will not be called.
+    5. For every `Fax` in `Fa`, `Fax` is called on `ib` with either the current literal component or the saved temporary variable, as appropriate. If `Fax` returns a `bool`, the result is
+    logically anded with all preceeding `Fax` calls.
```
AlekseyTs, line 365: It feels like this conflicts with "specifies false for `UseLazyComputation`" stated above

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +272,106 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+    6. The result of the expression is `ib`.
```
AlekseyTs, line 366: Result of the conversion?

AlekseyTs:

proposals/improved-interpolated-strings.md
333fred: > I think that this bullet point should be merged with the previous one. If we really want to say something about final validation, which I prefer we don't, we could say that final validation might still produce errors.

I was trying to word it in the same manner that I plan on implementing it, which would likely suppress cascaded diagnostics if the single-best method failed, say, constraint checking.

> If the "final validation" mentioned here is the one specified as step 4, then it appears that we are referring to steps out of order

It's not the same, I'll reword.

proposals/improved-interpolated-strings.md
333fred: > Also, is Fi a result of the lookup or result of the overload resolution?

Neither: it's the result of traditional method invocation resolution, which includes overload resolution, and also its own steps of final validation (such as type constraint checking).

proposals/improved-interpolated-strings.md
333fred: Traditional method invocation resolution includes making sure that the method is being called from the right context, static in this case.

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +270,107 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
 
 _Answer_: No.
 
-### Lowering
+### Builder pattern codegen
+
+In this section, method invocation resolution refers to the steps listed [here](https://github.com/dotnet/csharplang/blob/main/spec/expressions.md#method-invocations).
+
+#### `Create` method invocation resolution
```
jaredpar, line 277: Have we got the `Create` vs. constructor discussin on the LDM agenda?

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +270,107 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+
+Given an _applicable\_interpolated\_string\_builder\_type_ `T` and an _interpolated\_string\_expression_ `i`, overload resolution for a set of valid `Append...` methods on `T` is
+performed as follows:
+
+1. If there are any _interpolated\_regular\_string\_character_ components in `i`:
+    1. Member lookup on `T` with the name `AppendLiteral` is performed. The resulting method group is called `Ml`.
+    2. The argument list `Al` is constructed with one value parameter of type `string`.
+    3. Traditional method invocation resolution is performed with method group `Ml` and argument list `Al`. For the purposes of method invocation final validation, the context of `Ml`
+    is treated as a _member\_access_ through an instance of `T`.
+        * If a single-best method `Fi` is found and no errors were produced, the result of lookup is `Fi`.
```
AlekseyTs, line 320: >the result of lookup is `Fi` [](http://example.com/codeflow?start=77&length=28)

"the result of overload resolution is `Fi`"? To match the previous section about the Create method.

proposals/improved-interpolated-strings.md
```diff
@@ -287,12 +270,107 @@ Do we want to have builders for `System.IFormattable` and `System.FormattableStr
...
+        * Otherwise, an error is reported.
+2. For every _interpolation_ `ix` component of `i`:
+    1. Member lookup on `T` with the name `AppendFormatted` is performed. The resulting method group is called `Mf`.
+    2. The argument list `Af` is constructed:
+        1. The first parameter is the `expression` of `ix`, passed by value.
+        2. If `ix` directly contains a _constant\_expression_ component, then an integer value parameter is added, with the name `alignment` specified.
+        3. If `ix` is directly followed by an _interpolation\_format_, then a string value parameter is added, with the name `format` specified.
+    3. Traditional method invocation resolution is performed with method group `Mf` and argument list `Af`. For the purposes of method invocation final validation, the context of `Mf`
+    is treated as a _member\_access_ through an instance of `T`.
+        * If a single-best method `Fi` is found, the result of lookup is `Fi`.
```
AlekseyTs, line 330: >the result of lookup is `Fi`. [](http://example.com/codeflow?start=49&length=29)

Same comment as for AppendLiteral section.

AlekseyTs (approved):
LGTM (commit 3) with a couple of suggestions.

proposals/improved-interpolated-strings.md
333fred: That's one of the 2 or 3 questions left, which I hope to fully cover in the next session.

PR #4761, opened 2021-05-18
Typo and rendering fixes

jnm2:
@333fred


meetings/2021/LDM-2021-05-17.md
```diff
@@ -78,7 +78,7 @@ We're mixed on language fences, leaning against supporting them for now. More de
 
 ### Interpolation
 
-This proposal didn't originally have interpolation, but after a large pushback from the community interpolation was added. Because the
+This proposal didn't originally have interpolation, but after a large pushback from the community, interpolation was added. Because the
```
333fred, line 81:
```suggestion
This proposal didn't originally have interpolation, but after a large pushback from the community interpolation was added. Because the
```

I don't believe the comma addition is correct.

jnm2: I don't think the lack is correct. This example sentence doesn't sound right:

> After a large pushback from the community interpolation was added.

333fred (approved):

PR #4778, opened 2021-05-26
Clarify struct field initialization with default constructor initializer

cston:
If a `struct` constructor calls `: this()` and the parameterless constructor is the default zero-init constructor, any fields initializers are executed in the calling constructor rather than the parameterless constructor.

Compare `S1(int)` and `S2(int)` below. Both call `: this()`. In `S1`, the parameterless constructor is the explicit, so `S1.X` is initialized in `S1()`. In `S2`, the parameterless constructor is the default zero-init constructor so `S2.X` is initialized in `S2(int)` instead.
```csharp
struct S1
{
    int X = 1;
    int Y;
    public S1() { } // X = 1;
    public S1(int y) : this() { Y = y; }
}

struct S2
{
    int X = 2;
    int Y;
    public S2(int y) : this() { Y = y; } // X = 2;
}
```

### DIFF
--- a/proposals/parameterless-struct-constructors.md
+++ b/proposals/parameterless-struct-constructors.md
@@ -58,8 +58,8 @@ Constructors can be declared `extern` or `unsafe`.
 Constructors cannot be `partial`.
 
 ### Executing field initializers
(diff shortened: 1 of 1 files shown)


jcouv (approved):
LGTM Thanks (iteration 1)

PR #4788, opened 2021-05-28
Fix typo

YairHalberstadt:


333fred (approved):
Don't really know if this is a "typo" so much as a mistake, but thanks :).

PR #4790, opened 2021-05-28
Update list-patterns.md

alrz:
Record recent LDM decisions and sync with the current impl.

Linking to test plan: https://github.com/dotnet/roslyn/issues/51289

### DIFF
--- a/proposals/list-patterns.md
+++ b/proposals/list-patterns.md
@@ -43,7 +43,7 @@ property_pattern
   ;
 
 slice_pattern
-  : '..' negated_pattern?
+  : '..' pattern?
   ;
 
 primary_pattern
@@ -59,7 +59,7 @@ There are three new patterns:
 - The *length_pattern* is used to match the length.
 - A *slice_pattern* is only permitted once and only directly in a *list_pattern_clause* and discards _**zero or more**_ elements.
 
-> **Open question**: Should we accept a general *pattern* following `..` in a *slice_pattern*?
+> **Open question**: Should we accept a general *pattern* following `..` in a *slice_pattern*? (answer [LDM 2021-05-26]: Yes, any pattern is allowed after a slice.)
 
 Notes:
 
@@ -69,28 +69,23 @@ Notes:
 - If the *type* is an *array_type*, the *length_pattern_clause* is disambiguated so that `int[] [0]` would match an empty integer array.
 - All other combinations are valid, for instance `T (p0, p1) [p2] { name: p3 } v` or `T (p0, p1) [p2] { p3 } v` where each clause can be omitted.
 
-> **Open question**: Should we support all these combinations?
-
 #### Pattern compatibility
 
 A *length_pattern* is compatible with any type that is *countable* — it has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.  
 A *length_pattern* is also compatible with any type that is *enumerable* — it can be used in `foreach`.
 
-A *list_pattern* is compatible with any type that is *countable* as well as *indexable* — it has an accessible indexer that takes an `Index` or `int` argument. If both indexers are present, the former is preferred.  
+A *list_pattern* is compatible with any type that is *countable* as well as *indexable* — it has an accessible indexer that takes an `Index` as the argument or otherwise an accessible indexer with a single `int` parameter. If both indexers are present, the former is preferred.  
 A *list_pattern* is also compatible with any type that is *enumerable*.
 
-A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.  
-A *slice_pattern* without a subpattern is also compatible with any type that is *enumerable*.
+A *slice_pattern* with a subpattern is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method with two `int` parameters. If both are present, the former is preferred.
+A *slice_pattern* without a subpattern is compatible with any type that is compatible with a *list_pattern*.
 
 ```
 enumerable is { 1, 2, .. } // okay
 enumerable is { 1, 2, ..var x } // error
 ```
 
-This set of rules is derived from the [***range indexer pattern***](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md#implicit-index-support) but relaxed to ignore optional or `params` parameters, if any.
-
-> **Open question**: We should define the exact binding rules for any of these members and decide if we want to diverge from the range spec.  
-> **Open question**: Should extension methods play a role in sliceability?  
+This set of rules is derived from the [***range indexer pattern***](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/ranges.md#implicit-index-support).
 
 #### Semantics on enumerable type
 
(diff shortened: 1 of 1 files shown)


proposals/list-patterns.md
```diff
@@ -153,7 +148,7 @@ case {.., 1, _}: // expr.Length is >= 2 && expr[^2] is 1
 
 The order in which subpatterns are matched at runtime is unspecified, and a failed match may not attempt to match all subpatterns.
 
-> **Open question**: The pattern `{..}` tests for  `expr.Length >= 0`. Should we omit such test (assuming `Length` is always non-negative)?
+> **Open question**: By this definition, the pattern `{..}` tests for `expr.Length >= 0`. Should we omit such test, assuming `Length` is always non-negative? (answer [LDM 2021-05-26]: `{ .. }` will not emit a Length check)
```
alrz, line 151: :memo: Notes were clear that this is about semantics and not an optimization.

alrz:
@333fred @jcouv for review. thanks.

(I didn't touch enumerable sections but I think we should move it to its own document eventually)

proposals/list-patterns.md
```diff
@@ -69,28 +69,23 @@ Notes:
...
+A *list_pattern* is compatible with any type that is *countable* as well as *indexable* — it has an accessible indexer that takes an `Index` as the argument or otherwise an accessible indexer with a single `int` parameter. If both indexers are present, the former is preferred.  
 A *list_pattern* is also compatible with any type that is *enumerable*.
 
-A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.  
-A *slice_pattern* without a subpattern is also compatible with any type that is *enumerable*.
+A *slice_pattern* with a subpattern is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method with two `int` parameters. If both are present, the former is preferred.
+A *slice_pattern* without a subpattern is compatible with any type that is compatible with a *list_pattern*.
```
alrz, line 81: :memo: We don't require Slice method if there is no subpattern. This is now consistent with the decision about `{..}`.

proposals/list-patterns.md
```diff
@@ -43,7 +43,7 @@ property_pattern
   ;
 
 slice_pattern
-  : '..' negated_pattern?
+  : '..' pattern?
```
alrz, line 46: Wasn't mentioned in LDM. But I think "all patterns" includes combinators.

333fred: It does. I'm not familiar enough with the various levels of the pattern grammar, does `negated_pattern` not include combinators?

proposals/list-patterns.md
```diff
@@ -69,28 +69,23 @@ Notes:
...
+A *list_pattern* is compatible with any type that is *countable* as well as *indexable* — it has an accessible indexer that takes an `Index` as an argument or otherwise an accessible indexer with a single `int` parameter. If both indexers are present, the former is preferred.  
 A *list_pattern* is also compatible with any type that is *enumerable*.
 
-A *slice_pattern* is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` argument or otherwise an accessible `Slice` method that takes two `int` arguments. If both are present, the former is preferred.  
-A *slice_pattern* without a subpattern is also compatible with any type that is *enumerable*.
+A *slice_pattern* with a subpattern is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` as an argument or otherwise an accessible `Slice` method with two `int` parameters. If both are present, the former is preferred.
+A *slice_pattern* without a subpattern is compatible with any type that is compatible with a *list_pattern*.
```
333fred, line 81: Might need to mention discard subpattern?

alrz: With this change, we have:

- `{..}` works even without `this[Range]` or a proper `Slice` method.
- `{.._}` requires `this[Range]`/`Slice`. The codegen, however, is permitted to omit the invocation (which we do). That is not limited to discards or the slice pattern, we skip an evaluation when the output is unused, e.g. with an unused variable etc.

alrz: Sounds like we currently don't include unused variables here, but it's something that could be done as an optimization.

alrz: Here's an example to demonstrate it's not limited to discard: `if (this is {IntProp: {}}){}` IntProp is not called here.

proposals/list-patterns.md
alrz: No it doesn't. I've started this as an unary operator, hence the precedence.

proposals/list-patterns.md
333fred: > The codegen, however, is permitted to omit the invocation (which we do).

This is a good distinction, but I do think it is good to make this not an optimization: it gets back to making sure that behavior of something like `default(ImmutableArray<int>) is { .._ }` is well-defined.

alrz: At the moment the behavior of `default(ImmutableArray<int>) is { .._ }` is an error. "unsupported type for slice pattern"

ImmutableArray doesn't expose a Slice method or range indexer.

333fred: Yeah, but additions are tracked by https://github.com/dotnet/runtime/issues/22928.

alrz: My main concern is that when a type is not sliceable, we should be able to do `{1, ..}` etc. Not every list should be required to be sliceable only to be able to skip elements in the list pattern.

jcouv: I'm fine with the behavior as spec'ed in this PR. I'd let `.._` and `..var unused` require a sliceable type. It's conceptually simpler.

For `{ ..var unused }`: during binding we don't know that `var unused` is unused.
For `{ .._ }`: @333fred , I can add a note to test plan to rediscuss if you think this should not require a sliceable type.

alrz:
To be clear, I don't think we should do anything specific for `ImmutableArray`, it's not a known type in the compiler and I don't believe it should influence the design. However, it's a use case that we definitely want to cover in this area.

To that end, I've proposed to use a hypothetical `bool Deconstruct` (already championed in some form) so that we can safely match an ImmutableArray whether or not it's initialized. I'm not sure if we want to go down that road though. But checking `IsDefault` is too type specific IMO, as far as I know, that doesn't have any special meaning anywhere in the language or BCL in general.

jcouv (approved):
LGTM Thanks (iteration 2)

jcouv:
> I don't think we should do anything specific for ImmutableArray,

Agreed.
FYI, based on email discussion, I suspect there will be a formal proposal soon to add a `null/default` operator. But that'll be orthogonal to the list-patterns feature.

proposals/list-patterns.md
jcouv: nit: Consider adding links to previous patterns specs:

https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.0/pattern-matching.md
https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md
https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/patterns3.md

proposals/list-patterns.md
alrz: >For `{ .._ }`

Just like we still need an indexable type for `{_}` that we're not going to use, `{.._}` should require a sliceable type that we may not use.  I think the part that we decide to not evaluate the method, is orthogonal to *binding rules*.

jcouv:
@333fred I started thread with LDM mentioning the `{ .._ }` question. Feel free to chime in. I think we can merge this PR if you're good with it (I am).

333fred (approved):

jcouv:
Merged/squashed. Thanks @alrz

PR #4829, opened 2021-06-14
fix: special type implementation for array

bernd5:
Spec refers to non existent type.


333fred (approved):
@BillWagner in case someone has already started on the csharpstandard side.

333fred:
Thanks @bernd5!

PR #4890, opened 2021-06-29
Specify when record struct synthesized members are 'readonly'

RikkiGibson:


jcouv:
I must have gotten confused. The note that I'd taken to test plan was this:
"LDM: Emitting GetHashCode and Equals as readonly (conditionally) (answer: let's not do anything special)"

Does that differ from LDM notes?

Youssef1313 (commented):
Can't `PrintMembers` and `Deconstruct` be always readonly? I don't get why there is a restriction. Those methods aren't supposed to write to fields/properties.
Anything I'm missing?

svick:
@Youssef1313 I believe the problematic case is when you have a property with a non-readonly getter.

proposals/record-structs.md
```diff
@@ -143,6 +143,8 @@ The record struct includes a synthesized method equivalent to a method declared
 public override string ToString();
 ```
 
+If the record's printable members do not include a readable property with a non-`readonly` `get` accessor, then the synthesized `PrintMembers` and `ToString` methods are `readonly`.
```
jcouv, line 146: The situation where a printable member is a non-`readonly` field should also be accounted for.

jcouv (commented):

proposals/record-structs.md
RikkiGibson:
```suggestion
If the record's printable members do not include a readable property with a non-`readonly` `get` accessor, then the synthesized `PrintMembers` and `ToString` methods are `readonly`. There is no requirement for the record's fields to be `readonly` for the `PrintMembers` and `ToString` methods to be `readonly`.
```

jcouv (approved):
LGTM Thanks (iteration 2)

jcouv:
Do we want to hold off this PR until we're ready to make the corresponding implementation change?

RikkiGibson:
> Do we want to hold off this PR until we're ready to make the corresponding implementation change?

Yeah, I think we should just merge it in sync with the implementation change.

RikkiGibson:
It was a bogus conflict (GitHub found a conflict, but local git reported no conflicts upon merge).

PR #4931, opened 2021-07-16
A meeting with lambda improvements mentioned

WhiteBlackGoose:
I was looking for the status of naturally typed lambdas and couldn't find. Then Fred gave me a link to a meeting which, I assume, should be mentioned in the lambda improvements proposal too, since it touches them.

### DIFF
--- a/proposals/lambda-improvements.md
+++ b/proposals/lambda-improvements.md
@@ -216,5 +216,6 @@ lambda_parameter
 
 ## Design meetings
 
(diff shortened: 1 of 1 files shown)


jcouv:
Thanks @WhiteBlackGoose

333fred (approved):
@cston there are probably others missing from this. In general, this is why I prefer to put them in the relevant issue, rather than in the proposal.

Issue #124, opened 2017-02-15
Champion Allow Generic Attributes


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

See
- https://github.com/dotnet/roslyn/issues/953
- https://roslyn.codeplex.com/discussions/542178
- https://github.com/dotnet/roslyn/pull/16814
- https://github.com/dotnet/coreclr/pull/9189

AlphaDriveLegal:
I've wanted this for a long time!

Mafii:
I would be happpy to have this, too!

weitzhandler:
Very compelling.

Here's my use case, one among many:

    public class ValidateIf<TValidationAttribute> : ValidationAttribute
    {
      public TValidationAttribute Attribute { get; private set; }
      //other properties to resemble condition
      public ValidateIf(Action<TValidationAttribute> configure)
      {
        Attribute = configure();
      }
      protected override ValidationResult IsValid(object value, ValidationContext validationContext)
      {
        var shouldValidate = this.ShouldValidate();
        if(!shouldValidate) return ValidationResult.Success;
        else return Attribute.GetValidationResult(value, validationContext);
      }
    }

Joe4evr:
@weitzhandler Your example is pretty flawed since lambdas can't be passed as an attribute parameter; only primitives, `string`, `typeof()` expressions, and `DateTime` literals (VB).

A more sensible example of where a generic attribute would be useful:
```cs
// This concerns a bot command system that can parse strings into typed arguments,
// but sometimes the "default" reader for something may be too narrow,
// so this on a parameter would override which class does the reading for that command.
[AttributeUsage(AttributeTargets.Parameter)]
public sealed class OverrideTypeReaderAttribute<TReader> : Attribute
    where TReader : TypeReader
{
    // We're only interested in the Type info at this point,
    // but we would really like the type-safety of the generic constraint
    public Type TypeReaderType { get; }

    public OverrideTypeReaderAttribute()
    {
        TypeReaderType = typeof(TReader);
    }
}
// Usage:
public Task Foo([OverrideTypeReader<AlternativeBoolReader>] bool switch)
{
    //.....
}
```
[...]

PR #4936, opened 2021-07-17
Add generic attributes proposal

RikkiGibson:
Related to #124

### DIFF
--- a/proposals/generic-attributes.md
+++ b/proposals/generic-attributes.md
@@ -1,0 +1,66 @@
+# Generic Attributes
+
+* [x] Proposed
+* [x] Prototype: [N/A](#)
+* [x] Implementation: [In Progress](https://github.com/dotnet/roslyn/tree/features/generic-attributes)
+* [ ] Specification: [Not Started](#)
+
+## Summary
+[summary]: #summary
+
+When generics were introduced in C# 2.0, attribute classes were not allowed to participate. We can make the language more composable by removing (rather, loosening) this restriction. The .NET Core runtime has added support for generic attributes. Now, all that's missing is support for generic attributes in the compiler.
+
+## Motivation
+[motivation]: #motivation
+
+Currently attribute authors can take a `System.Type` as a parameter and have users pass a `typeof` expression to provide the attribute with types that it needs. However, outside of analyzers, there's no way for an attribute author to constrain what types are allowed to be passed to an attribute via `typeof`. If attributes could be generic, then attribute authors could use the existing system of type parameter constraints to express the requirements for the types they take as input.
+
+## Detailed design
+[design]: #detailed-design
+
+The following section is amended: https://github.com/dotnet/csharplang/blob/main/spec/classes.md#base-classes
+
+> The direct base class of a class type must not be any of the following types: System.Array, System.Delegate, System.MulticastDelegate, System.Enum, or System.ValueType. ~~Furthermore, a generic class declaration cannot use System.Attribute as a direct or indirect base class.~~
+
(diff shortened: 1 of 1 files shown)


YairHalberstadt:
Would a generic attribute parameter be allowed to be a type argument, so long as it was instantiated with a constant?

E.g.

```csharp
public class Attr<T> : Attribute
{
    public Attr(T t){}
}

[Attr<string>("hello world")]
```

RikkiGibson:
Yes, I expect that scenario to work. I'll try and make sure the feature branch has a test for it :)

proposals/generic-attributes.md
```diff
@@ -0,0 +1,66 @@
+# Generic Attributes
+
+* [x] Proposed
+* [x] Prototype: [N/A](#)
+* [x] Implementation: [In Progress](https://github.com/dotnet/roslyn/tree/features/generic-attributes)
+* [ ] Specification: [Not Started](#)
```
jcouv, line 6: This section should be in the issue, but not the spec file

proposals/generic-attributes.md
```diff
@@ -0,0 +1,66 @@
...
+
+The following section is amended: https://github.com/dotnet/csharplang/blob/main/spec/classes.md#base-classes
+
+> The direct base class of a class type must not be any of the following types: System.Array, System.Delegate, System.MulticastDelegate, System.Enum, or System.ValueType. ~~Furthermore, a generic class declaration cannot use System.Attribute as a direct or indirect base class.~~
+
+One important note is that the following section of the spec is *unaffected* when referencing the point of usage of an attribute, i.e. within an attribute list: https://github.com/dotnet/csharplang/blob/main/spec/types.md#type-parameters
+
+> A type parameter cannot be used anywhere within an attribute.
+
+This means that when a generic attribute is used, its construction needs to be fully "closed", i.e. not containing any type parameters, which means the following is still disallowed:
```
jcouv, line 29: Is "closed" the correct term? "Opened" for generic types could refer to `Type<>`. I didn't think that `Type<T>` would be considered "open".

jcouv (commented):

proposals/generic-attributes.md
RikkiGibson: I think so: https://github.com/dotnet/csharplang/blob/main/spec/types.md#open-and-closed-types

I think `Type<>` is an *unbound* generic type while `Type<T>` is an *open* generic type.

proposals/generic-attributes.md
RikkiGibson: Removed the section.

proposals/generic-attributes.md
```diff
@@ -0,0 +1,61 @@
...
+## Detailed design
+[design]: #detailed-design
+
+The following section is amended: https://github.com/dotnet/csharplang/blob/main/spec/classes.md#base-classes
+
+> The direct base class of a class type must not be any of the following types: System.Array, System.Delegate, System.MulticastDelegate, System.Enum, or System.ValueType. ~~Furthermore, a generic class declaration cannot use System.Attribute as a direct or indirect base class.~~
+
+One important note is that the following section of the spec is *unaffected* when referencing the point of usage of an attribute, i.e. within an attribute list: https://github.com/dotnet/csharplang/blob/main/spec/types.md#type-parameters
+
+> A type parameter cannot be used anywhere within an attribute.
```
Joe4evr, line 22: Something about this spec line had me confused for a bit (as well as once before ~a year ago?), so I'm wondering if it needs to be rephrased to emphasize it refers to the attribute placement and not its definition.

jcouv (approved):
LGTM Thanks (iteration 2)

proposals/generic-attributes.md
RikkiGibson: It does seem like it would be a little more clear to adjust the terminology to make attribute usages more distinct from their definitions.

Piusgodw:
Today

PR #4951, opened 2021-07-20
Call EnsureSufficientExecutionStack in PrintMembers

RikkiGibson:


proposals/csharp-9.0/records.md
```diff
@@ -232,8 +232,9 @@ bool PrintMembers(System.Text.StringBuilder builder);
 The method is `private` if the record type is `sealed`. Otherwise, the method is `virtual` and `protected`.
 
 The method:
-1. for each of the record's printable members (non-static public field and readable property members), appends that member's name followed by " = " followed by the member's value separated with ", ",
-2. return true if the record has printable members.
+1. if the record has printable members, calls `System.Runtime.CompilerServices.RuntimeHelpers.EnsureSufficientExecutionStack()`.
```
jaredpar, line 235: Did we measure the cost of this method?

RikkiGibson: No.

RikkiGibson: I did look at the implementation and it seems to be an internal call which reads the value of a precomputed field on the current thread.

The limit is computed here: https://github.com/dotnet/runtime/blob/663013d579d5eb3e29b2e0f88aeaed0694e74f39/src/coreclr/vm/threads.cpp#L6429-L6447

The method is an internal call which is implemented here: https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/coreclr/vm/reflectioninvocation.cpp#L1718

jaredpar: I created the following benchmark test for this scenario. Tried to somewhat emulate a deep stack where we don't actually hit the limit 

https://gist.github.com/jaredpar/019d3ed170c144b6f8a129eba7368969

The differences are there but not sure it's that significant 


|             Method |     Mean |    Error |   StdDev |
|------------------- |---------:|---------:|---------:|
|           ToString | 16.64 ms | 0.161 ms | 0.134 ms |
| ToStringWithEnsure | 18.59 ms | 0.361 ms | 0.469 ms |

RikkiGibson: Thanks for taking the time to check. It's useful to know the difference.

jaredpar: Why did you decide to unconditionally call the method vs. returning early? Essentially truncating the output at the point we hit the stack limit?

jaredpar: The benchmark I did initially was for a `try / catch / return` approach. The actual proposal, now that I looked at the PR, has less of a measurable difference

|             Method |     Mean |    Error |   StdDev |
|------------------- |---------:|---------:|---------:|
|           ToString | 14.86 ms | 0.284 ms | 0.339 ms |
| ToStringWithEnsure | 15.64 ms | 0.266 ms | 0.316 ms |

RikkiGibson: Could you please share the modified version of the benchmark?

jaredpar: Sure thing

https://gist.github.com/jaredpar/d20e3cb923345427d2bb1ee0136623c2

RikkiGibson: > Why did you decide to unconditionally call the method vs. returning early? Essentially truncating the output at the point we hit the stack limit?

If we return early it means we can't prevent the caller from continuing to print more "sibling" fields of the containing object, potentially hitting the stack limit again in other paths. Making it so we truly "short-circuit" without throwing an exception would probably require changing the signature of `PrintMembers()` which I don't think is on the table.

Basically, my view is that a record with cycles is essentially an error condition, at least when compiler-generated recursive methods are in use. Simply throwing an exception gives the user a stack trace that they should be able to fairly easily use to track down the cause of the cycle. On the other hand, giving truncated output will mean in practice producing a huge single-line string from `ToString()`, which I think might end up being harder to traverse in practice.

jaredpar (approved):

333fred (approved):

PR #4978, opened 2021-07-29
Rename GlobalUsingDirective.md to global-using-directive.md

WhiteBlackGoose:
For consistency. Although if there are links to the old file, then the change is irrelevant


WhiteBlackGoose:
Actually it is probably useless change

PR #4982, opened 2021-07-30
Update list-patterns.md

alrz:


proposals/list-patterns.md
```diff
@@ -34,12 +34,18 @@ There are two new patterns:
...
 
-A *list_pattern* is compatible with any type that is *countable* as well as *indexable* — it has an accessible indexer that takes an `Index` as an argument or otherwise an accessible indexer with a single `int` parameter. If both indexers are present, the former is preferred.  
+A *list_pattern* is compatible with any type that is *countable* and *indexable*.
 
-A *slice_pattern* with a subpattern is compatible with any type that is *countable* as well as *sliceable* — it has an accessible indexer that takes a `Range` as an argument or otherwise an accessible `Slice` method with two `int` parameters. If both are present, the former is preferred.
+A *slice_pattern* with a subpattern is compatible with any type that is *countable* and *sliceable*.
 
 A *slice_pattern* without a subpattern is compatible with any type that is compatible with a *list_pattern*.
 
+A type is *countable* if it has an accessible property getter that returns an `int` and has the name `Length` or `Count`. If both properties are present, the former is preferred.
```
alrz, line 43: :memo: This definition was removed in a previous PR.

alrz:
A couple of questions:

- Should `iTuple switch { (1,2,3) => 1, [1,2,3] => 2, _ => 0 }` produce an error? ITuple is both countable and indexable.
- In `[..[1]]` we only check length on the outer list (length=1) but we still check indexer values off of slice (if not subsumed). Is that ok?

proposals/list-patterns.md
333fred: Why do we need to add it back? Both concepts already exist via the C# features around range and index, I'm not sure we should duplicate here.

333fred (commented):
> Should iTuple switch { (1,2,3) => 1, [1,2,3] => 2, _ => 0 } produce an error? ITuple is both countable and indexable.

I don't see how it could, unless we wanted to make assumptions about general correspondence between positional pattern indexes and list pattern indexes. This doesn't seem like a necessarily-safe assumption to me.

> In [..[1], 2] we only check length on the outer list (length=2) but we still check indexer values off of slice (if not subsumed by another test anywhere else). Is that ok?

Seems both find and desirable to me. That's what the pattern says should happen, so we should do it.

alrz:
>>Should iTuple switch { (1,2,3) => 1, [1,2,3] => 2, _ => 0 } produce an error? ITuple is both countable and indexable.

>I don't see how it could

Because ITuple can be matched with both a list pattern and a positional pattern, the codegen would be identical.

In fact, anything we specified for pattern compatibility here applies to ITuple, including the positive values for length.

jcouv:
> Should iTuple switch { (1,2,3) => 1, [1,2,3] => 2, _ => 0 } produce an error? ITuple is both countable and indexable.

I'm unsure about this one. Let's assume this shouldn't produce an error for now, treating the positional and the list patterns as orthogonal. 
We can bring up to LDM when back in session (in 3 weeks).

> In [..[1], 2] we only check length on the outer list (length=2) but we still check indexer values off of slice (if not subsumed by another test anywhere else). Is that ok?

Yes, that sounds good to me as well. This behaves the same as when the nested list-pattern has a designation.

jcouv (approved):

333fred:
Only thing I'm missing from signing off is the answer to why we're adding back the definition of indexable and countable here. I don't think they need to be in this document.

proposals/list-patterns.md
alrz: Reverted.

alrz:
>I'm unsure about this one. Let's assume this shouldn't produce an error for now, treating the positional and the list patterns as orthogonal.

Please make sure to record this in the test plan, I'm more than fifty percent sure that we should do this because ITuple is pattern-compatible with a list pattern, it just happens to be compatible with another pattern as well, retroactively.

333fred (approved):

jcouv:
FYI, added a note to test plan to follow up on ITuple question.

Issue #1881, opened 2018-09-23
Switching ReadOnlySpan<char> with constant strings


Happypig375:
With the recent addition of Spans, we can reduce a lot of allocations when dealing with strings. For example, substrings can be replaced with slices. However, there is one place where using strings are better:
```cs
string str = ...;
switch (str)
{
    case "string 1":
        break;
    case "string 2":
        break;
    case "string 3":
        break;
    case "string 4":
        break;
}
```
With spans, you are forced to write boilerplate code:
```cs
ReadOnlySpan<char> span = ...;
switch (span)
{
    case var _ where span == "string 1":
        break;
    case var _ where span == "string 2":
        break;
    case var _ where span == "string 3":
        break;
    case var _ where span == "string 4":
        break;
}
```
Which is identical to the if-else mess that switches aim to solve.
```cs
ReadOnlySpan<char> span = ...;
if (span == "string 1")
else if (span == "string 2")
else if (span == "string 3")
else if (span == "string 4")
```
It would be better if spans can be switched with constant strings.
```cs
ReadOnlySpan<char> span = ...;
switch (span)
{
    case "string 1":
        break;
    case "string 2":
        break;
    case "string 3":
        break;
    case "string 4":
        break;
}
```
Special-casing spans as a compiler-known type already have precedent in stackalloc expressions, it can also be done here.
Possible extension to regular Span<char> can also be considered, but it is not as important as ReadOnlySpan<char>.

agocke:
This does seem really annoying. I'm not sure I have a language proposal that would fix this, but it would be something I'd like to address.

The obvious route would be something like an implicit conversion on `string` to `ReadOnlySpan<char>`. I'm not sure if that would work.

Happypig375:
>The obvious route would be something like an implicit conversion on string to ReadOnlySpan<char>. I'm not sure if that would work.

Yes it does exist in .NET Core 2.1 (NOT with the portable version!!)
However, it is not considered to be a _constant conversion_ (e.g. int -> byte). The compiler prohibits that inside switch statements.

If this was implemented, this should work with the portable version too.

CyrusNajmabadi:
Note: is there an idea on what the codegen here would actually be?  The normal string-switch works by building a dictionary and switching on that.  That likely wouldn't work for ReadOnlySpans.  Perhaps emitting code for an inline-trie might be appropriate?

scalablecory:
[...]

PR #5001, opened 2021-08-05
pattern-match-span-of-char-on-string.md is not part of csharp 10

bernd5:
Move `pattern-match-span-of-char-on-string.md` out of the csharp-10 folder because it is not part of chsharp-10.


jcouv:
Thanks. Do you know what championed issue tracks this feature, by any chance?

jcouv:
Found it. I think it's https://github.com/dotnet/csharplang/issues/1881 (and it's worked on in https://github.com/dotnet/roslyn/pull/44388)

PR #5018, opened 2021-08-09
fix build warnings in docs

BillWagner:
Most of the fixes are updating links to sections of the spec so that they work correctly.

Other fixes include semantic markdown:
- exactly 1 H1 heading in each file.
- ATX style headers.

Discovered while working on dotnet/docs#25504

### DIFF
--- a/proposals/csharp-10.0/GlobalUsingDirective.md
+++ b/proposals/csharp-10.0/GlobalUsingDirective.md
@@ -34,22 +34,22 @@ The scope of a *global_using_directive* specifically does not include *using_dir
 
 The effect of adding a *global_using_directive* to a program can be thought of as the effect of adding a similar *using_directive* that resolves to the same target namespace or type to every compilation unit of the program. However, the target of a *global_using_directive* is resolved in context of the compilation unit that contains it. 
 
-# Scopes 
+## Scopes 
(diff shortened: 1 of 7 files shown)


jcouv (approved):
LGTM Thanks (iteration 1)

RikkiGibson:
Is there any chance we could get a CI job spun up which would report these warnings on csharplang PRs?

BillWagner:
> Is there any chance we could get a CI job spun up which would report these warnings on csharplang PRs?

We will catch them on the `main` branch when we do our regular build on docs. That's less than optimal because we only catch them after being merged. I'll see what tools we may have that would work.

(Meanwhile making a PR for a few of the warnings I missed in this PR).

BillWagner:
> Is there any chance we could get a CI job spun up which would report these warnings on csharplang PRs?

Ping @Youssef1313 

Would the link verifier you wrote work well in this repo?

Youssef1313:
@BillWagner I don't recall what would the behavior be if some docs-specific files are not found (e.g, openpublishing redirection file).

In case it doesn't work, it shouldn't be hard to tweak. As a first step, I think we'd need to enable GitHub Actions for this repo (it can be restricted as in dotnet/docs)

PR #5025, opened 2021-08-11
fix missing links

BillWagner:
A few links to the standard did not include the folder path.

These were missed in #5018


jcouv (approved):
LGTM Thanks (iteration 1)

PR #5046, opened 2021-08-14
Fix typo

geoff-wasilwa:


CyrusNajmabadi:
`E` was correct.

geoff-wasilwa:
Fixed

333fred:
Spec PRs should not go against this repository. Instead, the specification is being developed at https://github.com/dotnet/csharpstandard, and the spec here will be removed when that is done.

PR #5083, opened 2021-08-18
Clarify span safety rules for readonly receivers

jaredpar:
The compiler excludes receivers from the "Must Arguments Must Match"
rules when the type is a `readonly struct`. This exclusion was not
covered in the rules.

### DIFF
--- a/proposals/csharp-7.2/span-safety.md
+++ b/proposals/csharp-7.2/span-safety.md
@@ -279,7 +279,7 @@ We wish to ensure that no `ref` local variable, and no variable of `ref struct` 
 
 - For an assignment `e1 = e2`, if the type of `e1` is a `ref struct` type, then the *safe-to-escape* of `e2` must be at least as wide a scope as the *safe-to-escape* of `e1`.
 
-- For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver), with *safe-to-escape* E1, then no argument (including the receiver) may have a narrower *safe-to-escape* than E1. [Sample](#method-arguments-must-match)
+- For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with *safe-to-escape* E1, then no argument (including the receiver) may have a narrower *safe-to-escape* than E1. [Sample](#method-arguments-must-match)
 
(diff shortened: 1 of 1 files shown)


jaredpar:
@333fred because you are the one who nerd sniped me into looking at this.

proposals/csharp-7.2/span-safety.md
```diff
@@ -292,9 +292,9 @@ We wish to ensure that no `ref` local variable, and no variable of `ref struct`
 These explanations and samples help explain why many of the safety rules above exist
 
 ### Method Arguments Must Match
-When invoking a method where there is an `out`, `ref` parameter that is a `ref struct` including the receiver then all of the `ref struct` need to have the same lifetime. This is necessary because C# must make all of its decisions around lifetime safety based on the information available in the signature of the method and the lifetime of the values at the call site. 
+When invoking a method where there is an `out`, `ref` parameter that is a `ref struct` then all of the `ref struct` need to have the same lifetime. This is necessary because C# must make all of its decisions around lifetime safety based on the information available in the signature of the method and the lifetime of the values at the call site.
```
333fred, line 295:
```suggestion
When invoking a method where there is an `out` or `ref` parameter that is a `ref struct` then all of the `ref struct` parameters need to have the same lifetime. This is necessary because C# must make all of its decisions around lifetime safety based on the information available in the signature of the method and the lifetime of the values at the call site. 
```

proposals/csharp-7.2/span-safety.md
```diff
@@ -334,6 +334,8 @@ void Broken(ref S s)
 }
 ```
 
+For the purpose of this analysis the receiver is considered an `in`, not a `ref`, if the type is a `readonly struct`. In that case the receiver cannot be used to store values from other parameters. It is effectively an `in` parameter for analysis purposes. Hence the same example above is legal when `S` is `readonly` because the `span` cannot be stored anywhere.
```
333fred, line 337:
```suggestion
For the purpose of this analysis the receiver is considered an `in`, not a `ref`, if the type is a `readonly struct`. In that case the receiver cannot be used to store values from other parameters, it is effectively an `in` parameter for analysis purposes. Hence the same example above is legal when `S` is `readonly` because the `span` cannot be stored anywhere.
```

jaredpar:
@333fred accepted your feedback

333fred (approved):

PR #5100, opened 2021-08-21
Add "private protected" to introduction.md

KorneiDontsov:
Add "private protected" to the list of accesses in spec.


333fred:
Spec changes should go on dotnet/csharpstandard, not on csharplang. The spec part of this repository is no longer being changed, all the active work is in that repository.

PR #5102, opened 2021-08-23
Address docs build warnings

BillWagner:
There are three different types of changes in this PR:

- H1s should use ATX headers (`#`)
- tables should have a header row that is not empty.
- ~~Replace absolute links to docs.microsoft.com with links to the source on GitHub.~~

~~The third is less optimal, and I'm looking for a better solution.  The reason for the build warning is that the docs build prefers relative links, so those links work in isolated environments. The problem is that a relative links assumes the reader is on the docs.microsoft.com site, not on GitHub. Instead, I changed the destination to the GitHub source. That suppresses the warning, but doesn't actually fix the issue: those links still won't work in an isolated environment, if disconnected from the internet.~~

The better solution for the links is to use an escape hatch for the build warning in our docs system.

### DIFF
--- a/proposals/csharp-7.2/private-protected.md
+++ b/proposals/csharp-7.2/private-protected.md
@@ -22,7 +22,7 @@ Directly providing support for this access level in C# enables these circumstanc
 
 ### `private protected` access modifier
 
-We propose to add a new access modifier combination `private protected` (which can appear in any order among the modifiers). This maps to the CLR notion of protectedAndInternal, and borrows the same syntax currently used in [C++/CLI](https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/dotnet/how-to-define-and-consume-classes-and-structs-cpp-cli.md#-member-visibility).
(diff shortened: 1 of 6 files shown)


RikkiGibson:
@333fred what is the plan for where the spec lives? Are we expecting to reach a point where we remove the content of the specs folder here and just link to its new home?

BillWagner:
@RikkiGibson 

> what is the plan for where the spec lives? Are we expecting to reach a point where we remove the content of the specs folder here and just link to its new home?

Yes.  Once the remaining [C# 6.0 PRs](https://github.com/dotnet/csharpstandard/pulls?q=is%3Aopen+is%3Apr+milestone%3A%22C%23+6%22) are merged, the version of the spec in dotnet/csharplang shouldn't be needed. The standard will have caught up. :tada:.   However, the `proposals/<version>` folders will live longer until the committee catches up.  Rex and I are on an upcoming agenda at LDM to discuss and formalize the plan.

RikkiGibson (approved):

PR #5174, opened 2021-09-10
Create using-alias-types.md

CyrusNajmabadi:


proposals/using-alias-types.md
```diff
@@ -0,0 +1,52 @@
...
+## Motivation
+For ages, C# has had the ability to introduce aliases for namespaces and named types (classes, delegated, interfaces, records and structs).  This worked acceptably well as it provided a means to introduce non-conflicting names in cases where a normal named pulled in from `using_directive`s might be ambiguous, and it allowed a way to provide a simpler name when dealing with complex generic types.  However, the rise of additional complex type symbols in the language has caused more use to arise where aliases would be valuable but are currently not allowed.  For example, both tuples and function-pointers often can have large and complex regular textual forms that can be painful to continually write out, and a burden to try to read.  Aliases would help in these cases by giving a short, developer-provided, name that can then be used in place of those full structural forms.
+
+## Detailed design
+We will change the grammar of `using_alias_directive` thusly:
+
+```
+using_alias_directive
+-    : 'using' identifier '=' namespace_or_type_name ';'
++    : 'using' identifier '=' (namespace_name | type) ';'
```
jcouv, line 19: Does `type` include top-level nullability annotation?
I assume this allows built-in types `using Id = int;`. Is that right?

cston:
See also https://github.com/dotnet/csharplang/pull/4452.

proposals/using-alias-types.md
CyrusNajmabadi: > Does type include top-level nullability annotation?

Yes.  It would.

> I assume this allows built-in types using Id = int;. Is that right?

yes, it would.

CyrusNajmabadi:
> See also #4452.

This is very intentionally *not* trying to support the case of `using X<T> = ...`  While i think that is valuable for us to look into, it's orthogonal and much larger in scope (esp. on the impl side).  Whereas the above proposal is actually quite trivial to support and hits a lot of pain cases we have today.

proposals/using-alias-types.md
jcouv: > Yes. It would [include top-level nullability annotation].

From our offline discussion, this raises a question for how the three annotations (`?`, `!` and `~`/oblivious) interact in the declaration versus the use-site. For example, what happens if you put `?` in the declaration and the use-site?

proposals/using-alias-types.md
```diff
@@ -0,0 +1,57 @@
...
+    ;
+```
+
+Interestingly, most of the spec language in [using_alias_directive](https://github.com/dotnet/csharplang/blob/main/spec/namespaces.md#using-alias-directives) needs to change.  Most language in it already refers to 'namespace or type', for example:
```
333fred, line 23: Does _not_ need to change?

CyrusNajmabadi: yup.  fixed.

333fred (approved):
Spec looks good to me. Do we have a championed issue?

tannergooding:
@CyrusNajmabadi, would this also work for other aliases?

That is, could I do the following?
```csharp
using HANDLE = void*;
using HWND = HANDLE;
```

I was wanting to add a feature to the ClangSharp P/Invoke generator that generates `global using directives` for `typedefs` and I've found the following three issues that are blocking:
* Can't use keywords, which is particularly relevant since `nint` and `IntPtr` behave differently
* Can't use pointers (important for typedefs to pointers)
* Can't use other defined aliases (important for typedefs to typedefs)

CyrusNajmabadi:
> That is, could I do the following?
> 
> using HANDLE = void*;
> using HWND = HANDLE;

The first is completely allowed.  The second would not be allowed as a using directive cannot reference another using directive in the same block.  However, this is ok:

```c#
using HANDLE = void*;
namespace X
{
  using HWND = HANDLE;
}
```

or

```c#
using HANDLE = void*;
using HWND = void*;
```

> Can't use keywords, which is particularly relevant since nint and IntPtr behave differently

Keywords would be fine.

> Can't use pointers (important for typedefs to pointers)

You will be able to use pointers.

> Can't use other defined aliases (important for typedefs to typedefs)

If in the same scope this will continue to not work and you'll still have to explicitly write out the entire type signature again.

tannergooding:
> Keywords would be fine.
> You will be able to use pointers.

Awesome.

> If in the same scope this will continue to not work and you'll still have to explicitly write out the entire type signature again.

This is a bit unfortunate but also not overly complex to handle, thankfully.

CyrusNajmabadi:
> This is a bit unfortunate but also not overly complex to handle, thankfully.

Understood. Note that i'm not opposed to work here.  But i would consider it more of a *separate* feature.

tannergooding:
Yep definitely understand.

Its basically just traversing the typedefs to the "concrete" type for ClangSharp, I expect its a bit more complex for Roslyn to handle 😄

PR #5175, opened 2021-09-10
Fix only three " instead of four " in raw-strings proposal

petrroll:
The example (IMHO) is supposed to show that when the raw string includes """, it's supposed to be marked by """".

PS: It looks like such a useful feature :)


petrroll:
I actually cannot read properly. It's all correct as it should be.

CyrusNajmabadi:
:)

PR #5192, opened 2021-09-15
Remove broken link

gewarren:
Contributes to https://github.com/dotnet/docs/issues/25539.

### DIFF
--- a/proposals/csharp-10.0/improved-definite-assignment.md
+++ b/proposals/csharp-10.0/improved-definite-assignment.md
@@ -205,7 +205,7 @@ if (b ? x != null && Set(out y) : x != null && Set(out y))
 bool Set(out int x) { x = 0; return true; }
 ```
 
-This is an admittedly niche scenario, that compiles without error in the native compiler, but was broken in Roslyn in order to match the specification at the time. See [internal issue](http://vstfdevdiv:8080/DevDiv2/DevDiv/_workitems/edit/529603).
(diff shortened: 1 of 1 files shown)


proposals/csharp-10.0/improved-definite-assignment.md
```diff
@@ -205,7 +205,7 @@ if (b ? x != null && Set(out y) : x != null && Set(out y))
 bool Set(out int x) { x = 0; return true; }
 ```
 
-This is an admittedly niche scenario, that compiles without error in the native compiler, but was broken in Roslyn in order to match the specification at the time. See [internal issue](http://vstfdevdiv:8080/DevDiv2/DevDiv/_workitems/edit/529603).
```
333fred, line 208: @RikkiGibson @gewarren where would be a good place to leave this note? I get that we don't want to publish a page to docs that links to an internal defunct bug tracker, but we still likely want to keep the link somewhere.

RikkiGibson: What about the proposal issue #4465? Maybe we could link to this paragraph and as well as the internal issue?

333fred: Possibly in here? https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/Compiler%20Breaking%20Changes%20-%20VS2015.md

Issue #398, opened 2017-04-01
Proposal: Extend null-conditional (?., ?[]) and null-coalescing (??) operators to pointers


YaakovDavis:
I'd like to be able to write:
```
Data* p = GetUnmanagedData();
p?->Act();
var a = p?->X ?? 42;
var b = p?[3] ?? new Data();
var p2 = p ?? AnotherDataPointer();
```

The above should compile to the equivalent of the following:
```
Data* p = GetUnmanagedData();
if(p != null)
   p->Act();

var a = p != null ? p->X : 42;
var b = p != null ? p[3] : new Data();
var p2 = p != null ? p : AnotherDataPointer();
```

svick:
This is half of a duplicate of https://github.com/dotnet/csharplang/issues/323.

YaakovDavis:
@svick 
I thought I had encountered something similar, but couldn't find it.
In case, this proposal is more encompassing.

jacekbe:
When I proposed #323 I completely forgot about null-conditional operators. Obviously it would be nice if both null-conditional and null-coalescing operator worked also for pointer types.

Note that null-coalescing operator is already supported by VS2017 quick actions. VS2017 hints that it could be simplified to use null-coalescing operator even though resulting code does not compile.
```c#
 unsafe byte* SomeFunc(byte* p, byte* defaultVal)
 {
      return p != null ? p : defaultVal;            
 }
```

gafter:
I've queued this to be considered by the LDM at https://github.com/dotnet/csharplang/issues/418.

lachbaer:
Wouldn't it rather be `?->` than `?.`?

gafter:
@lachbaer The LDM discussed this today and we decided that while it isn't something we would devote resources to implementing, we'd probably accept a contributed implementation along with #418.

Rekkonnect:
@jaredpar this is only half a duplicate of #418, as noted [previously](https://github.com/dotnet/csharplang/issues/398#issuecomment-290953479). Please consider either updating the discussion created from #418, reopening this and converting it into a discussion, or creating a new parent issue tracking all those parts of the feature.

jaredpar:
I understand that these are not a perfect duplicate of each other, but at the same time having two championed issues / discussions for this topic isn't helping either. I duped them so that people would see the relationship between the discussions. Happy for the discussion on this to include the broader topics.

Issue #3619, opened 2020-06-27
Allow null-conditional access operator when accessing a pointer type


john-h-k:

Joe4evr:
Duplicate of #398.

john-h-k:
That did not cover this operator explicitly 🙂 , it appeared they were suggesting `?.` on pointers as `?->`, rather than using `?.` where the resulting type is a pointer

yaakov-h:
This does seem like an interesting case that isn't quite covered by #323 or #398.

To be clear, is the idea here that `notAPointerReference?.APointerReference` would become legal, and return a null pointer if `notAPointerReference` evalutes to null?

john-h-k:
Yes, exactly @yaakov-h

PR #5198, opened 2021-09-16
Create proposal spec file for null-conditional and null-coalescing operators on pointers

AlFasGD:
Adds a spec file for dotnet/csharplang#398

# Discussion
There already exists the spec for [null-coalescing on pointers](proposals/pointer-null-coalescing.md). Should it remain as an individual proposal that will be possibly handled from a separate PR, or have it merged into this one?


333fred:
Again, #3619 is not championed at the moment. You can have an open question about whether we should consider it as well at the same time, but it is not currently a championed proposal.

PR #5215, opened 2021-09-23
Fix code example in notes

alrz:


meetings/2021/LDM-2021-09-22.md
```diff
@@ -101,8 +101,8 @@ of the LDM misinterpreted it as modifying the nested instance, rather than doing
 if multiple properties on a nested instance need to be set. We have a couple of initial ideas:
 
 ```cs
-methodCallExpression { Method with { Name = "MethodName", BaseType = "C" } }
-methodCallExpression { Method = { Name = "MethodName", BaseType = "C" } } // more similar to nested object initializer?
+methodCallExpression with { Method with { Name = "MethodName", BaseType = "C" } }
+methodCallExpression with { Method = { Name = "MethodName", BaseType = "C" } } // more similar to nested object initializer?
```
alrz, line 105: When written correctly, the second form doesn't look as odd as the first. ;)

333fred (approved):

PR #5232, opened 2021-09-28
Deleted confusing task lists from proposals

svick:
Fixes https://github.com/dotnet/docs/issues/26288.

These task lists were often outdated, I don't think they are tracking anything useful and they didn't render quite right on docs.MS, so I think it makes sense to delete them.

In a couple of cases, this also deletes links to documentation in `roslyn/docs/features`. If these should be preserved, I can add them back.

Also note that this PR only changes proposal that are shown on docs.MS (i.e. those under `proposals/csharp-x.y`). Those directly under `proposals` and those under `proposals/rejected` are left unchanged.

### DIFF
--- a/proposals/csharp-7.1/async-main.md
+++ b/proposals/csharp-7.1/async-main.md
@@ -1,10 +1,5 @@
 # Async Main
 
-* [x] Proposed
-* [ ] Prototype
-* [ ] Implementation
-* [ ] Specification
-
 ## Summary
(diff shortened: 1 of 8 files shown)


jcouv (approved):

jcouv:
I'm glad for this clean up. Thanks!

BillWagner:
Thanks @svick I'm also quite happy with this.

PR #5382, opened 2021-11-03
Add proposal for `params Span<T>`

cston:
Champion issue: https://github.com/dotnet/csharplang/issues/1757

### DIFF
--- a/proposals/params-span.md
+++ b/proposals/params-span.md
@@ -1,7 +1,7 @@
 # `params Span<T>` and implicit allocation of arrays
 
 ## Summary
-Avoid heap allocation for implicit allocation of arrays in specific scenarios such as `params` arguments.
+Avoid heap allocation for implicit allocation of arrays in specific scenarios with `params` arguments.
 
 ## Motivation
 `params` array parameters provide a convenient way to call a method that takes an arbitrary length list of arguments.
@@ -11,7 +11,7 @@ If we extend `params` types to include the `ref struct` types `Span<T>` and `Rea
 
 And if we're extending `params` to other types, we  could also allow `params IEnumerable<T>` to avoid allocating and copying collections at call sites that have an `IEnumerable<T>` rather than `T[]`.
 
-The benefits of `params ReadOnlySpan<T>` and `params Span<T>` are primarily for new APIs. Existing commonly used APIs such as `Console.WriteLine()` and `StringBuilder.AppendFormat()` already have overloads that avoid array allocations for common cases.
+The benefits of `params ReadOnlySpan<T>` and `params Span<T>` are primarily for new APIs. Existing commonly used APIs such as `Console.WriteLine()` and `StringBuilder.AppendFormat()` already have overloads that avoid array allocations for common cases and those overloads would be needed for backward compatibility.
 ```csharp
 public static class Console
 {
@@ -28,9 +28,9 @@ public static class Console
 ### Extending `params`
 `params` parameters will be supported with types `Span<T>`, `ReadOnlySpan<T>`, and `IEnumerable<T>`.
 
-A call in _expanded_ form to a method with a `params T[]` or `params IEnumerable<T>` parameter will result in an array `T[]` allocated on the heap.
+A call in _expanded form_ to a method with a `params T[]` or `params IEnumerable<T>` parameter will result in an array `T[]` allocated on the heap.
 
-A call in _expanded_ form to a method with a `params ReadOnlySpan<T>` or `params Span<T>` parameter will result in an array `T[]` created on the stack _if the `params` array is within specific limits set by the compiler_.
+A call in _expanded form_ to a method with a `params ReadOnlySpan<T>` or `params Span<T>` parameter will result in an array `T[]` created on the stack _if the `params` array is within specific limits set by the compiler_.
 Otherwise the array will be allocated on the heap.
 
 ```csharp
@@ -47,9 +47,9 @@ Two overloads cannot differ by `params` modifier alone.
 `params` parameters will be marked in metadata with a `System.ParamArrayAttribute` regardless of type.
 
 ### Overload resolution
-Overload resolution will continue to prefer overloads that are applicable in _normal_ form rather than _expanded_ form.
+Overload resolution will continue to prefer overloads that are applicable in [_normal form_](../../spec/expressions.md#applicable-function-member) rather than [_expanded form_](../../spec/expressions.md#applicable-function-member).
 
-For overloads that are applicable in _expanded_ form, [Better function member](https://github.com/dotnet/csharplang/blob/main/spec/expressions.md#better-function-member) will be updated to prefer `params` types in a specific order:
+For overloads that are applicable in _expanded form_, [Better function member](../../spec/expressions.md#better-function-member) will be updated to prefer `params` types in a specific order:
 
 > When performing this evaluation, if `Mp` or `Mq` is applicable in its expanded form, then `Px` or `Qx` refers to a parameter in the expanded form of the parameter list.
 > 
@@ -101,7 +101,7 @@ Are scenarios for `params IEnumerable<T>` sufficiently compelling to justify tha
 
 ### Array limits
 The compiler may use heuristics to determine when to fallback to heap allocation for the underlying data for spans.
-Experimentation should establish the limits we agree on.
+If heuristics are necessary, experimentation should establish the limits we agree on.
 
 ### Lowering approach
 Determine the particular approach used to lower `params` and array creation expressions to avoid heap allocation.
--- a/proposals/params-span.md
+++ b/proposals/params-span.md
@@ -1,39 +1,77 @@
-# Stack allocation of `params Span<T>`
+# Stack allocation of arrays of managed types
 
 ## Summary
-Allow calling a method with a `params` parameter without allocating an array on the heap.
+Allow explicit stack allocation of short-lived arrays regardless of type, similar to the current support for arrays of managed types using `stackalloc` and `Span<T>`.
+And allow implicit stack allocation of arrays, including `params` arguments, in cases where the arrays are only used as `ReadOnlySpan<T>` or `Span<T>`.
 
-## Motivation
-A `params` array parameter allows callers to pass an arbitrary length list of arguments to the method without explicitly creating an array. Instead, the compiler is responsible for allocating and initializing an array for the trailing arguments.
+## Explicit stack allocation
+Explicit stack allocation will be supported for arrays of any type with `stackalloc`.
 
-For instance, a call to `Console.WriteLine(fmt, x, y, z, w);` is emitted as:
+The expression `stackalloc T[length]` will allocate `T[]` on the call stack unconditionally, regardless of `length`.
+The type of the expression is `Span<T>` unless `T` is an _unmanaged type_ target-typed to `T*`.
+
+`length` must be a non-negative `int` but does not need to be compile-time constant.
+
 ```csharp
-// public static void WriteLine(string format, params object?[]? arg)
-Console.WriteLine(fmt, new object[4] { x, y, z, w });
+public static ImmutableArray<TResult> Select<TSource, TResult>(this ImmutableArray<TSource> source, Func<TSource, TResult> map)
+{
+    int n = source.Length;
+    Span<TResult> result = n <= 16 ? stackalloc TResult[n] : new TResult[n];
+    for (int i = 0; i < n; i++)
+        result[i] = map(source[i]);
+    return ImmutableArray.Create(result); // requires ImmutableArray.Create<T>(ReadOnlySpan<T>)
+}
 ```
-However, the implicit array is allocated on the heap.
 
-To avoid the array allocation, some commonly used methods such as `Console.WriteLine()` include overloads with fixed numbers of arguments in addition to the `params` array overload.
+### Lowering `stackalloc`
+The runtime will provide a `StackAlloc<T>(int length)` method that unconditionally allocates an array of `T[length]` on the call stack.
 ```csharp
-public static void WriteLine(string value);
-public static void WriteLine(string format, object? arg0);
-public static void WriteLine(string format, object? arg0, object? arg1);
-public static void WriteLine(string format, params object?[]? arg);
+namespace System.Runtime.CompilerServices
+{
+    public static class RuntimeHelpers
+    {
+        public static Span<T> StackAlloc<T>(int length);
+    }
+}
 ```
 
-Ideally, the compiler should allocate the `params` array on the callstack rather than the heap when the compiler can ensure the value does not escape the method being called.
+A call to `Console.WriteLine(fmt, stackalloc object[] { x, y, z });` could be emitted as:
+```csharp
+var span = RuntimeHelpers.StackAlloc<object>(3);
+span[0] = x;
+span[1] = y;
+span[2] = z;
+Console.WriteLine(fmt, (ReadOnlySpan<object>)span); // WriteLine(string format, params ReadOnlySpan<object?> arg)
+```
 
-## Detailed design
-### Additional `params` types
+_Can we use `localloc` for managed types?_
+
+## Implicit stack allocation
+Stack allocation of arrays is extended to implicit allocations as well.
+In particular, array creation expressions that are target-typed to `ReadOnlySpan<T>` or `Span<T>` will be allocated on the stack if:
+- The length of the array is a constant integer <= 8 (the actual limit is arbitrary), and
+- The call-site has not explicitly _opted-out_ of implicit stack allocation.
+
+```csharp
+Span<int> s = new[] { i, j, k }; // stack allocation of int[]
+WriteLine(fmt, new[] { x, y });  // stack allocation of object[] for WriteLine(string fmt, ReadOnlySpan<object> args);
+```
+
+### `params Span<T>`
 The C# compiler will support parameters declared as `params ReadOnlySpan<T>` and `params Span<T>`.
 
-_List the restrictions from `params T[]`._ 
+_Include the restrictions from `params T[]`._
 
-The compiler will report an error if the `ReadOnlySpan<T>` or `Span<T>` parameter value is returned from the method or assigned to an `out` parameter.
-That ensures call-sites can allocate the underlying _array_ for the span on the callstack and reuse the underlying _array_ without concern for aliases.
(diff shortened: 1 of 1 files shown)


proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+An implicitly allocated array may be reused regardless of whether the array was created on the stack or the heap.
+
+### Lowering implicit allocation
+For an array `T[]` of constant length `N`, the compiler will synthesize a `struct` with `N` fields of type `T` where the layout and alignment of the fields matches the alignment of elements in `T[]`.
+The array will be represented with a value of the synthesized `struct`, and a `Span<T>` created from a `ref` to the first field.
+
+For example, a call to `Console.WriteLine(fmt, x, y, z);` would be emitted as:
+```csharp
+[StructLayout(LayoutKind.Sequential)]
+internal struct __ValueArray3<T> { public T Item1, Item2, Item3; };
```
jcouv, line 93: May be worth considering the kind of struct we already generate for [fixed size buffers](https://github.com/dotnet/csharplang/blob/main/spec/unsafe-code.md#fixed-size-buffers). Such struct has a single field with the type of an item, but the layout is stretched out. [sharplab](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQBsCWyaoA6wwQIB2yANACYgDUAPgAIBMAjALABQDAzAARxkokXjgRwAxjF4BhXgG9OvJbzCYAHhEq9MJKWoDa3ALoBuTgF8gA===)

proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+## Detailed design
+
+### Extending `params`
+`params` parameters will be supported with types `Span<T>`, `ReadOnlySpan<T>`, and `IEnumerable<T>`.
+
+A call in _expanded_ form to a method with a `params T[]` or `params IEnumerable<T>` parameter will result in an array `T[]` allocated on the heap.
+
+A call in _expanded_ form to a method with a `params ReadOnlySpan<T>` or `params Span<T>` parameter will result in an array `T[]` created on the stack if the `params` array has no more than 8 arguments (an arbitrary limit).
```
CyrusNajmabadi, line 33: do we want the actual language spec to state the limit?  or do we want to leave it to be a compiler/runtime/user possible heuristic?

proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+    public static void WriteLine(string format, object arg0, object arg1);
+    public static void WriteLine(string format, object arg0, object arg1, object arg2);
+    public static void WriteLine(string format, params object[] arg);
+}
+```
+
+## Detailed design
+
+### Extending `params`
+`params` parameters will be supported with types `Span<T>`, `ReadOnlySpan<T>`, and `IEnumerable<T>`.
```
CyrusNajmabadi, line 29: i think we could trivially expand this out to support any interface that Array implements (so IList as well).  This can be nice on the consumption side as it allows params, but also requires anything passed in have a known fixed length (which IEnumerable doesn't provide).

proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+
+var values = new __ValueArray3<object>() { Item1 = x, Item2 = y, Item3 = z };
+var span = MemoryMarshal.CreateSpan(ref values.Item1, 3);
+Console.WriteLine(fmt, (ReadOnlySpan<object>)span); // WriteLine(string format, params ReadOnlySpan<object?> arg)
+```
+
+The synthesized `struct` type will be `internal`, with an unspeakable name.
+
+Since `T` is a valid type argument (and used in `Span<T>`), the synthesized type will be a constructed generic type where the underlying generic type is shared across other call sites in the compilation with the same number of arguments.
+
```
CyrusNajmabadi, line 103: it's possible today to have a `params int*[]`.  Whatever we do, we have to make sure that that still works. i bring this up as pointers cannot be type parameters.  so while we may not allow a params span of pointers, we still need to make sure anything we do doesn't break the standard array case.

proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+## Open issues
+### Is `params Span<T>` necessary?
+Is there a reason to support `params` parameters of type `Span<T>` in addition to `ReadOnlySpan<T>`? Is allowing mutation within the `params` method useful?
```
CyrusNajmabadi, line 106: yes.  i think we would want this.  we have params arrays today, and those are mutable.  it would be unfortuante for those apis that depend on this to not be able to use params span and get the benefits.

proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Is `params IEnumerable<T>` necessary?
+If the compiler allows `params ReadOnlySpan<T>`, then new APIs that require `params` could use `params ReadOnlySpan<T>` instead of `params T[]` because `T[]` is implicitly convertible to `ReadOnlySpan<T>`. And even legacy APIs could add a `params ReadOnlySpan<T>` overload where the existing `params T[]` simply delegates to the new overload.
+
+There is no conversion from `IEnumerable<T>` to `ReadOnlySpan<T>` however, so allowing `params IEnumerable<T>` is essentially asking APIs to provide two overloads for `params` methods: `params ReadOnlySpan<T>` and `params IEnumerable<T>`. 
+
+Are scenarios for `params IEnumerable<T>` sufficiently compelling to justify that?
```
CyrusNajmabadi, line 113: yes, i believe so.  This feature is not jsut about perf, but about ergonomics.  The requirement to take in an array has been very limiting to many apis, requiring an array overload (for the ease of passing values), and an IEnumerable/IList/etc. overload (to allow other collections to be passed in).   It's been a very longstanding request from teh community that this sort of common pattern not require multiple overloads when one would be clear and sufficient.

proposals/params-span.md
jcouv: From LDM, turns out the fixed-size buffer approach doesn't work for this scenario (GC tracking wouldn't work).
Fixed-size buffers avoid this today by being limited to simple struct types.
```
public unsafe struct C {
    fixed object x[3]; // error CS1663: Fixed size buffer type must be one of the following: bool, byte, short, int, long, char, sbyte, ushort, uint, ulong, float or double
}
```

proposals/params-span.md
```diff
@@ -0,0 +1,138 @@
...
+### Explicit `stackalloc`
+Should we allow explicit stack allocation of arrays of managed types with `stackalloc` as well?
+```csharp
+public static ImmutableArray<TResult> Select<TSource, TResult>(this ImmutableArray<TSource> source, Func<TSource, TResult> map)
+{
+    int n = source.Length;
+    Span<TResult> result = n <= 16 ? stackalloc TResult[n] : new TResult[n];
+    for (int i = 0; i < n; i++)
+        result[i] = map(source[i]);
+    return ImmutableArray.Create(result); // requires ImmutableArray.Create<T>(ReadOnlySpan<T>)
```
RikkiGibson, line 124: I think this requires the 'DoesNotEscape' annotation on ImmutableArray.Create.

proposals/params-span.md
RikkiGibson: I am wary of arbitrary boundaries that users don't know when they've crossed. Perhaps this is fine for the "simple expanded call" scenario, where getting the stack allocation is just a bonus, but other users will want to be able to make a call that is "stack allocate this or error", i.e. it won't accidentally cross the threshold as the code evolves. I think this is an important reason to have an "explicit" equivalent to the expanded call.

That said, it occurs to me that this was probably discussed in an LDM that I missed, so this might have all been touched on already..

proposals/params-span.md
YairHalberstadt: Is there any need for a struct at all? Why not just use locals directly?

proposals/params-span.md
YairHalberstadt: Could the compiler generate a static instance of an ImmutableArray<T> and share it across all calls to `params IEnumerable<T>`?

proposals/params-span.md
cston: Created an open issue for heuristics. To explicitly create the span on the stack, we should consider allowing `stackalloc [] { x, y, z }`.

proposals/params-span.md
bernd5: Locals are not required to be stored in a sequentual order - usually they are not even stored in memory but in CPU register.
I think we should not force a lowering strategy (e.g. stackalloc, custom type with seq. fields or fixed size buffer).
We should just mention those as possible implementations.

proposals/params-span.md
bernd5: stackalloc is quite dangerous because: 

> stackalloc expression allocates memory from the current stack frame, and that memory may not be released until the current method call returns. If stackalloc is used in a loop, it can lead to stack overflows due to exhausting the stack memory.

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2014

proposals/params-span.md
cston: `params` parameters for types that are not valid type arguments (such as `int*`) will need to continue using `params T[]`.

proposals/params-span.md
cston: When might it be useful to modify the `params` array or `Span<T>` within the `params` method? If the parameter is marked `params`, presumably many callers will use _expanded_ form (with an implicit array or span), so the modifications to the array will be ignored.

proposals/params-span.md
```diff
@@ -0,0 +1,139 @@
...
+    Span<TResult> result = n <= 16 ? stackalloc TResult[n] : new TResult[n];
+    for (int i = 0; i < n; i++)
+        result[i] = map(source[i]);
+    return ImmutableArray.Create(result); // requires ImmutableArray.Create<T>([DoesNotEscape] ReadOnlySpan<T> items)
+}
+```
+
+This would require runtime support for stack allocation of arrays of non-constant length and any type, and GC tracking of the elements.
+
+Direct runtime support for stack allocation of arrays of managed types might be useful for lowering implicit allocation as well.
```
jaredpar, line 131: The other downside we should discuss here is that the JIT not track the lifetime of  `stackalloc` locals the same way it tracks normal locals. That is because we will immediately grab a `ref` to their contents and that breaks JIT ability to track them. It essentially gives up and considers them always alive. 

That means in order to not introduce inadvertent memory leaks the compiler has to consider manual management strategies. For example when we exit a scope where we did a `stackalloc` we need to consider calling `.Clear()` on the `Span<T>`. 

These problems, and the likely solutions, mirror the problems we have with lifted locals in state machines. Unless we take compensating actions the lifetime of locals won't match the intuition of developers.

proposals/params-span.md
```diff
@@ -0,0 +1,139 @@
...
+### Is `params IEnumerable<T>` necessary?
+If the compiler allows `params ReadOnlySpan<T>`, then new APIs that require `params` could use `params ReadOnlySpan<T>` instead of `params T[]` because `T[]` is implicitly convertible to `ReadOnlySpan<T>`. And even legacy APIs could add a `params ReadOnlySpan<T>` overload where the existing `params T[]` simply delegates to the new overload.
+
+There is no conversion from `IEnumerable<T>` to `ReadOnlySpan<T>` however, so allowing `params IEnumerable<T>` is essentially asking APIs to provide two overloads for `params` methods: `params ReadOnlySpan<T>` and `params IEnumerable<T>`. 
+
+Are scenarios for `params IEnumerable<T>` sufficiently compelling to justify that?
+
+### Heuristics
+The proposed approach falls back to heap allocation if the array length exceeds an arbitrary limit.
+Do we need a limit, and if so, what are the heuristics?
```
jaredpar, line 114: I would generalize this a bit. Rather than recommending specific heuristics I would say that the language, using only the tools we have today, is likely to have to resort to heuristics. Experimentation should establish the limits we agree on.

proposals/params-span.md
```diff
@@ -0,0 +1,139 @@
...
+### Array re-use
+The compiler _may_ reuse an implicitly allocated array across multiple uses within a single thread executing a method:
+- At the same call-site (within a loop) or
+- At distinct call-sites if the lifetime of the spans do not overlap, and the array length is sufficient, and
+  - the element types are managed types that are considered identical by the runtime, or
+  - the element types are unmanaged types of the same size.
+
+An implicitly allocated array may be reused regardless of whether the array was created on the stack or the heap.
+
+### Lowering implicit allocation
```
jaredpar, line 86: I would consider moving this to an open issues section. Essentially here is one potential way we can lower. 

For the current design I would say that the compiler is going to lower the calls to the most efficient way possible. Specifically avoiding heap allocations when possible. The exact details are still to be determined and may differ based on the target framework and hence the tools available.

proposals/params-span.md
jaredpar: No because in the case of anything but `Span<T>` type we must assume the implementation smuggles away the argument. Hence re-use is not safe.

YairHalberstadt: But if the backing type is an ImmutableArray that shouldn't matter? I'm taking about a case where the params are constants.

jaredpar: In that specific case yes we could put it in an `ImmutableArray<T>`. I'm not sure it's worth it. 

- Forces the compiler to depend on a new type and assign specific semantics to it. Generally we only do that for special types. 
- It's not a total win, there is still allocations at the call site (boxing the `ImmutableArray<T>`) and at the callee (allocation of `IEnumerator` during enumeration). 

Don't think we'll save enough for it to be worth the complexity.

YairHalberstadt: You called store the ImmutableArray already boxed as an IEnunerable to save on the first allocation. But agreed, less than perfect.

jaredpar (approved):

proposals/params-span.md
```diff
@@ -0,0 +1,146 @@
+# `params Span<T>` and implicit allocation of arrays
+
+## Summary
+Avoid heap allocation for implicit allocation of arrays in specific scenarios such as `params` arguments.
```
jcouv, line 4:
```suggestion
Avoid heap allocation for implicit allocation of arrays in specific scenarios with `params` arguments.
```

jcouv: The motivation for suggestion is that some `params Span<T>` scenarios are likely to still allocate.

proposals/params-span.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Motivation
+`params` array parameters provide a convenient way to call a method that takes an arbitrary length list of arguments.
+However, using an array type for the parameter means the compiler must implicitly allocate an array on the heap at each call site.
+
+If we extend `params` types to include the `ref struct` types `Span<T>` and `ReadOnlySpan<T>`, where values of those types cannot escape the call stack, the array at the call site may be created on the stack instead.
+
+And if we're extending `params` to other types, we  could also allow `params IEnumerable<T>` to avoid allocating and copying collections at call sites that have an `IEnumerable<T>` rather than `T[]`.
+
+The benefits of `params ReadOnlySpan<T>` and `params Span<T>` are primarily for new APIs. Existing commonly used APIs such as `Console.WriteLine()` and `StringBuilder.AppendFormat()` already have overloads that avoid array allocations for common cases.
```
jcouv, line 14: >The benefits [](http://example.com/codeflow?start=0&length=12)

There's also a benefit for existing APIs: they can be re-written to call a single overload.

proposals/params-span.md
```diff
@@ -0,0 +1,146 @@
...
+    public static void WriteLine(string format, params object[] arg);
+}
+```
+
+## Detailed design
+
+### Extending `params`
+`params` parameters will be supported with types `Span<T>`, `ReadOnlySpan<T>`, and `IEnumerable<T>`.
+
+A call in _expanded_ form to a method with a `params T[]` or `params IEnumerable<T>` parameter will result in an array `T[]` allocated on the heap.
```
jcouv, line 31: >_expanded_ form [](http://example.com/codeflow?start=10&length=15)

1. Consider including a link to spec section that defines "expanded form" [https://github.com/dotnet/csharplang/blob/main/spec/expressions.md#applicable-function-member](https://github.com/dotnet/csharplang/blob/main/spec/expressions.md#applicable-function-member)
2. Is allocating a `T[]` array on the heap the only option for `params IEnumerable<T>` in expanded form?

proposals/params-span.md
```diff
@@ -0,0 +1,146 @@
...
+### Is `params IEnumerable<T>` necessary?
+If the compiler allows `params ReadOnlySpan<T>`, then new APIs that require `params` could use `params ReadOnlySpan<T>` instead of `params T[]` because `T[]` is implicitly convertible to `ReadOnlySpan<T>`. And even legacy APIs could add a `params ReadOnlySpan<T>` overload where the existing `params T[]` simply delegates to the new overload.
+
+There is no conversion from `IEnumerable<T>` to `ReadOnlySpan<T>` however, so allowing `params IEnumerable<T>` is essentially asking APIs to provide two overloads for `params` methods: `params ReadOnlySpan<T>` and `params IEnumerable<T>`. 
+
+Are scenarios for `params IEnumerable<T>` sufficiently compelling to justify that?
+
+### Array limits
+The compiler may use heuristics to determine when to fallback to heap allocation for the underlying data for spans.
+Experimentation should establish the limits we agree on.
```
jcouv, line 104: As I mentioned in LDM, I'm not convinced that the compiler can determine what is right/preferrable. Maybe we should pick a default (for example some stack allocation) and have an escape for users who prefer the other mode (use explicit array allocation).

jcouv (approved):

proposals/params-span.md
cston: Existing overloads would need to be retained for backward compatibility however, so a `params ReadOnlySpan<T>` overload could be added, but would not replace existing overloads.

proposals/params-span.md
cston: I've added a link for _expanded form_, thanks.

For `params IEnumerable<T>`, the collection will need to be allocated on the heap since the method being called may hold onto a reference to the collection. Presumably it could be an `IEnumerable<T>` other than `T[]` though.

proposals/params-span.md
cston: Changed the comment to "If heuristics are necessary, ..."

proposals/params-span.md
jnm2: Maybe:

```cs
for (var i = 0; i < someParams.Length; i++)
    someParams[i] ??= string.Empty; // Or something
```

PR #5397, opened 2021-11-09
Update param-nullchecking proposal

RikkiGibson:
Uses the `!!` syntax and allows wiggle room in the implementation strategy.


proposals/param-nullchecking.md
```diff
@@ -37,8 +37,10 @@ void M(string name) {
 }
 ```
 
+Parameter null checks are used frequently in libraries with tight performance and size constraints. For this reason, the compiler implementation is free to define exactly how the above semantics are achieved. For example, to optimize code size, inlining, etc., the compiler may use helper methods to perform the null check a la the [ArgumentNullException.ThrowIfNull](https://github.com/dotnet/runtime/blob/1d08e154b942a41e72cbe044e01fff8b13c74496/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs#L56-L69) methods, or any other implementation strategy that achieves the required semantics.
```
333fred, line 40: The concern I have with the way that this is worded is that a helper method would _not_ work. The semantics of this code include where the exception is thrown, ie "what is the top frame of the stack trace." Unless we're collaborating with the runtime on an intrinsic that is erased from a stack trace such that it really has identical semantics, I think we need to clarify that variance here.

RikkiGibson: This is true. I'll try and clarify.

333fred (approved):

PR #5438, opened 2021-11-16
Clarify conversion rules for CallerArgumentExpression

333fred:
Relates to https://github.com/dotnet/roslyn/issues/52745
@Youssef1313 @jcouv ptal.

### DIFF
--- a/proposals/csharp-10.0/caller-argument-expression.md
+++ b/proposals/csharp-10.0/caller-argument-expression.md
@@ -148,7 +148,7 @@ There should always be an expression corresponding to the `this` parameter. Even
 - Multiple parameters marked with `CallerArgumentExpression` are permitted, as shown above.
 - The attribute's namespace will be `System.Runtime.CompilerServices`.
 - If `null` or a string that is not a parameter name (e.g. `"notAParameterName"`) is provided, the compiler will pass in an empty string.
-- The type of the parameter `CallerArgumentExpressionAttribute` is applied to must have a standard conversion from `string`. This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string` or `object`.
+- The type of the parameter `CallerArgumentExpressionAttribute` is applied to must have a standard implicit conversion from `string`. This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string`, `object`, or an interface implemented by `string`.
 
 ## Drawbacks
 [drawbacks]: #drawbacks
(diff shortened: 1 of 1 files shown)


proposals/csharp-10.0/caller-argument-expression.md
```diff
@@ -148,6 +148,7 @@ There should always be an expression corresponding to the `this` parameter. Even
 - Multiple parameters marked with `CallerArgumentExpression` are permitted, as shown above.
 - The attribute's namespace will be `System.Runtime.CompilerServices`.
 - If `null` or a string that is not a parameter name (e.g. `"notAParameterName"`) is provided, the compiler will pass in an empty string.
+- The type of the parameter `CallerArgumentExpressionAttribute` is applied to must have a standard conversion from `string`. This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string` or `object`.
```
Youssef1313, line 151: Any interface implemented by `string` will work, also `dynamic` will work. So it's not only `string` and `object`.

Youssef1313: Side note (out of scope for this PR): The other attributes error messages (and also the spec) mention "standard implicit conversion" instead of "standard conversion", which is wrong. This was pointed out by Aleksey in the code review IIRC.

yaakov-h: Does "user-defined conversions" mean `ReadOnlySpan<char>` is also out of the question?

333fred: Thanks for the clarification @Youssef1313.

> Does "user-defined conversions" mean ReadOnlySpan<char> is also out of the question?

Yes, it does mean that.

proposals/csharp-10.0/caller-argument-expression.md
```diff
@@ -148,6 +148,7 @@ There should always be an expression corresponding to the `this` parameter. Even
 - Multiple parameters marked with `CallerArgumentExpression` are permitted, as shown above.
 - The attribute's namespace will be `System.Runtime.CompilerServices`.
 - If `null` or a string that is not a parameter name (e.g. `"notAParameterName"`) is provided, the compiler will pass in an empty string.
+- The type of the parameter `CallerArgumentExpressionAttribute` is applied to must have a standard implicit conversion from `string`. This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string`, `object`, or an interface implemented by `string`.
```
Youssef1313, line 151: I think "standard conversion" is the correct term (as it originally was). My comment was saying that specs for other caller info mention "standard implicit conversion" incorrectly. I also added "dynamic"

```suggestion
- The type of the parameter `CallerArgumentExpressionAttribute` is applied to must have a standard conversion from `string`. This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string`, `object`, `dynamic`, or an interface implemented by `string`.
```

333fred: Dynamic is just object with a different name, I don't think it needs to be called out.

jcouv:
Do we have corresponding tests on the roslyn side?

jcouv (approved):
LGTM Thanks (iteration 3)

333fred:
> Do we have corresponding tests on the roslyn side?

Yes.

PR #5441, opened 2021-11-17
Update file-scoped-namespaces.md

Rick-Anderson:


proposals/csharp-10.0/file-scoped-namespaces.md
```diff
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Allow a simpler format for the common case of file containing only one namespace in it.  This format is `namespace X.Y.Z;` (note the semicolon and lack of braces).  This allows for files like so:
+File scoped namespaces use a less verbose format for the typical case of file containing only one namespace in it.  The file scoped namespace format is `namespace X.Y.Z;` (note the semicolon and lack of braces).  This allows for files like the following:
```
Rick-Anderson, line 5:
```suggestion
File scoped namespaces use a less verbose format for the typical case of files containing only one namespace.  The file scoped namespace format is `namespace X.Y.Z;` (note the semicolon and lack of braces).  This allows for files like the following:
```

Rick-Anderson: Why is it that I can write directly to main, but I can't commit my own suggestion?

Rick-Anderson: Never mind, I did an edit in browser, which is not convenient.

CyrusNajmabadi:
Seems reasonable to me :-)

PR #5539, opened 2021-12-14
Update param-nullchecking proposal

RikkiGibson:


333fred (approved):

jaredpar (approved):

proposals/param-nullchecking.md
```diff
@@ -37,6 +37,32 @@ void M(string name) {
 }
 ```
 
+There are a few guidelines limiting where `!!` can be used:
+
+1. Only a parameter of something with an implementation can use it. For example, an abstract method parameter cannot use `!!`. Further examples include:
+   - extern method parameters
+   - delegate parameters
+   - interface method parameters when the method is not a DIM
+2. It must be possible to include an equivalent "check then throw" of the given parameter at the beginning of the method, ignoring syntactic limitations such as the need to replace an expression body with a block body.
```
RikkiGibson, line 46: Record primary constructors seem to violate this. Since we think record primary constructors will eventually have explicit implementations, I'm choosing to allow them in the implementation. I'm not currently intending to say anything in the spec about it, assuming the problem will work itself out when record explicit implementations come about.

PR #5550, opened 2021-12-15
Added notes about interpolated strings in nullable flow analysis

333fred:
Followup from https://github.com/dotnet/roslyn/pull/57780.
Relates to test plan https://github.com/dotnet/roslyn/issues/51499

### DIFF
--- a/proposals/csharp-10.0/improved-interpolated-strings.md
+++ b/proposals/csharp-10.0/improved-interpolated-strings.md
@@ -452,6 +452,46 @@ rented array inside it, and if one of the interpolation holes throws an exceptio
 _Answer_: No. handlers can be assigned to locals (such as `MyHandler handler = $"{MyCode()};`), and the lifetime of such handlers is unclear. Unlike foreach enumerators, where the lifetime
 is obvious and no user-defined local is created for the enumerator.
 
+### Impact on nullable reference types
+
+To minimize complexity of the implementation, we have a few limitations on how we perform nullable analysis on interpolated string handler constructors used as arguments to a method or indexer.
+In particular, we do not flow information from the constructor back through to the original slots of parameters or arguments from the original context, and we do not use constructor parameter
+types to inform generic type inference for type parameters in the containing method. An example of where this can have an impact is:
+
+```cs
+string s = """";
+C c = new C();
+c.M(s, $"""", c.ToString(), s.ToString()); // No warnings on c.ToString() or s.ToString(), as the `MaybeNull` does not flow back.
+
+public class C
+{
+    public void M(string s1, [InterpolatedStringHandlerArgument("""", ""s1"")] CustomHandler c1, string s2, string s3) { }
+}
+
+[InterpolatedStringHandler]
+public partial struct CustomHandler
+{
+    public CustomHandler(int literalLength, int formattedCount, [MaybeNull] C c, [MaybeNull] string s) : this()
+    {
+    }
+}
+```
+
+```cs
+string? s = null;
+M(s, $""""); // Infers `string` for `T` because of the `T?` parameter, not `string?`, as flow analysis does not consider the unannotated `T` parameter of the constructor
+
+void M<T>(T? t, [InterpolatedStringHandlerArgument(""s1"")] CustomHandler<T> c) { }
+
+[InterpolatedStringHandler]
+public partial struct CustomHandler<T>
+{
+    public CustomHandler(int literalLength, int formattedCount, T t) : this()
+    {
+    }
+}
+```
+
 ## Other considerations
 
 ### Allow `string` types to be convertible to handlers as well
--- a/proposals/csharp-10.0/improved-interpolated-strings.md
+++ b/proposals/csharp-10.0/improved-interpolated-strings.md
@@ -459,13 +459,13 @@ In particular, we do not flow information from the constructor back through to t
 types to inform generic type inference for type parameters in the containing method. An example of where this can have an impact is:
 
 ```cs
-string s = """";
+string s = "";
 C c = new C();
-c.M(s, $"""", c.ToString(), s.ToString()); // No warnings on c.ToString() or s.ToString(), as the `MaybeNull` does not flow back.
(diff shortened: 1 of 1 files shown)


333fred:
@AlekseyTs @RikkiGibson for a look. This is documenting the limitations around nullable flow analysis from the roslyn PR.

proposals/csharp-10.0/improved-interpolated-strings.md
```diff
@@ -452,6 +452,46 @@ rented array inside it, and if one of the interpolation holes throws an exceptio
...
 is obvious and no user-defined local is created for the enumerator.
 
+### Impact on nullable reference types
+
+To minimize complexity of the implementation, we have a few limitations on how we perform nullable analysis on interpolated string handler constructors used as arguments to a method or indexer.
+In particular, we do not flow information from the constructor back through to the original slots of parameters or arguments from the original context, and we do not use constructor parameter
+types to inform generic type inference for type parameters in the containing method. An example of where this can have an impact is:
+
+```cs
+string s = """";
```
RikkiGibson, line 462: Was it intended to have doubled quote marks here and below? This looks like it might have been left over from pasting in a test.

proposals/csharp-10.0/improved-interpolated-strings.md
```diff
@@ -452,6 +452,46 @@ rented array inside it, and if one of the interpolation holes throws an exceptio
...
+C c = new C();
+c.M(s, $"""", c.ToString(), s.ToString()); // No warnings on c.ToString() or s.ToString(), as the `MaybeNull` does not flow back.
```
jaredpar, line 464: I wouldn't expect `[MaybeNull]` to have any impact here (interpolated strings are no). That attribute should only be impacting the input position, not the output position. What am I missing?

RikkiGibson: AllowNull is for inputs, MaybeNull is for outputs.

jaredpar: Huh, seems odd for that to have value in non-ref parameters. That isn't an output

333fred: Consider `Assert.NotNull`. That will use a `NotNull` attribute on a by-val parameter and throw/assert if it's `null`.

jaredpar: Correct. But it's reasonable for an API to strengthen our understanding of a piece of nullable state. The implementation of a method can do all manner of tricks to verify that a piece of input is not null. Effectively elevating the state from "maybe null" to "not null".

Having `[MaybeNull]` impact the state is saying that an API can downgrade the state. Taking something the compiler believes is "not null" and making it "maybe null". Having trouble seeing how an API can make such an assertion. Basically because you passed A to me you should assume A is now possible null.

333fred: I don't see a real reason to have it not impact things, and I don't think that this spec is the place to make the argument to change the feature in any case 😀

jaredpar: Don't want my debate to be about changing the feature. Want to make sure I understand `[MaybeNull]` so I can better evaluate this change.

333fred: Basically, in an ideal world, this feature would behave exactly as if you called all these methods in order manually, and that would mean any nullability post-condition attributes flowing back to the original expression (`MaybeNull` included). For implementation simplicity, we have avoided this.

RikkiGibson:
Looks like this is ready to go just pending minor fixes to the quotes in code examples.

RikkiGibson (approved):

proposals/csharp-10.0/improved-interpolated-strings.md
333fred, line 462:
```suggestion
string s = "";
```

proposals/csharp-10.0/improved-interpolated-strings.md
333fred, line 464:
```suggestion
c.M(s, $"", c.ToString(), s.ToString()); // No warnings on c.ToString() or s.ToString(), as the `MaybeNull` does not flow back.
```

proposals/csharp-10.0/improved-interpolated-strings.md
```diff
@@ -452,6 +452,46 @@ rented array inside it, and if one of the interpolation holes throws an exceptio
...
+
+public class C
+{
+    public void M(string s1, [InterpolatedStringHandlerArgument("""", ""s1"")] CustomHandler c1, string s2, string s3) { }
```
333fred, line 468:
```suggestion
    public void M(string s1, [InterpolatedStringHandlerArgument("", "s1")] CustomHandler c1, string s2, string s3) { }
```

proposals/csharp-10.0/improved-interpolated-strings.md
```diff
@@ -452,6 +452,46 @@ rented array inside it, and if one of the interpolation holes throws an exceptio
...
+{
+    public CustomHandler(int literalLength, int formattedCount, [MaybeNull] C c, [MaybeNull] string s) : this()
+    {
+    }
+}
+```
+
+```cs
+string? s = null;
+M(s, $""""); // Infers `string` for `T` because of the `T?` parameter, not `string?`, as flow analysis does not consider the unannotated `T` parameter of the constructor
```
333fred, line 482:
```suggestion
M(s, $""); // Infers `string` for `T` because of the `T?` parameter, not `string?`, as flow analysis does not consider the unannotated `T` parameter of the constructor
```

proposals/csharp-10.0/improved-interpolated-strings.md
```diff
@@ -452,6 +452,46 @@ rented array inside it, and if one of the interpolation holes throws an exceptio
...
+
+void M<T>(T? t, [InterpolatedStringHandlerArgument(""s1"")] CustomHandler<T> c) { }
```
333fred, line 484:
```suggestion
void M<T>(T? t, [InterpolatedStringHandlerArgument("s1")] CustomHandler<T> c) { }
```

PR #5564, opened 2021-12-21
Spec update to semi auto properties

Youssef1313:
cc @CyrusNajmabadi @333fred @RikkiGibson @AlekseyTs

### DIFF
--- a/proposals/semi-auto-properties.md
+++ b/proposals/semi-auto-properties.md
@@ -47,7 +47,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 +
 + If the auto-property has only a semicolon-only get accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)).
 +
-+ An auto-property is not allowed to only have a single semicolon-only `set` accessor without a `get` accessor.
++ An auto-property must have a `get` accessor, and can optionally have a `set` accessor.
 ```
 
 ...


CyrusNajmabadi:
@jnm2 do you remember why we phrased things this way in the spec instead of just keeping the original wording (which @Youssef1313 is returning us to)?

proposals/semi-auto-properties.md
```diff
@@ -47,7 +47,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 +
 + If the auto-property has only a semicolon-only get accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)).
 +
-+ An auto-property is not allowed to only have a single semicolon-only `set` accessor without a `get` accessor.
++ An auto-property must have a `get` accessor, and can optionally have a `set` accessor.
```
AlekseyTs, line 50: Either I am missing something, or the original wording looks right to me. I.e. The following code should be fine:
```
int P
{
    set { field+=value; if (field > 100) throw null;}
}
``` 

The original wording could be slightly adjusted without changing its meaning: "An auto-property is not allowed to only have a semicolon-only `set` accessor without a `get` accessor."

jnm2:
@CyrusNajmabadi I don't remember us talking about this line directly. It comes down to whether this is called an auto property since it does have an autogenerated backing field even though it has no getter:

```cs
public int Foo { set => Bar(field ??= value); }
```

If this is called an auto property, then the previous should probably be too:

```cs
public int Foo { get => field; set => field = value; }
```

proposals/semi-auto-properties.md
Youssef1313: @AlekseyTs That's exactly it. The discussion is about whether this piece of code will be allowed or not.

jnm2: Wouldn't it seem odd not to allow it? Is there a benefit in making it asymmetrical with `get` accessors this way?

CyrusNajmabadi: It's more just preserving what rules we had before. And keeping symmetry with the existing constructs.

AlekseyTs: > It's more just preserving what rules we had before. And keeping symmetry with the existing constructs.

The rules we had before made sense at that time because an auto-property property like that couldn't do anything useful. That is not the case any more. So the wording prior to this PR actually preserves the old rule in the new reality. I do not think there is a requirement to preserve the old rule verbatim.

Youssef1313:
Closing per https://github.com/dotnet/csharplang/pull/5564#discussion_r773582135

PR #5566, opened 2021-12-23
Update the required members spec

333fred:
Adds some additional details on various topics, including OHI and metadata. Also adds new questions for LDM to address in a full session in early Jan.

Relates to test plan https://github.com/dotnet/roslyn/issues/57046

### DIFF
--- a/proposals/required-members.md
+++ b/proposals/required-members.md
@@ -265,7 +265,7 @@ for the generic instantiation to ensure that the requirements are satisfied.
 
 ### Overriding, Hiding, and Inheriting
 
-It is an error to mark a member required if the member is not at least as visible as the containing member's type. This means the following cases are not allowed:
+It is an error to mark a member required if the member is less accessible than the member's containing type. This means the following cases are not allowed:
 
 ```cs
 interface I
@@ -294,7 +294,7 @@ When overriding a `required` member, the `required` keyword must be included on 
 unrequiring a property with an override in the future, we have design space to do so.
 
 Overrides are allowed to mark a member `required` where it was not `required` in the base type. A member so-marked is added to the required members
-list of derived type.
+list of the derived type.
 
 ### Metadata Representation
 
@@ -328,23 +328,23 @@ the string `"Types with required members are not supported in this version of yo
 any older compilers from using these constructors. We don't use a `modreq` here because it is a goal to maintain binary compat: if the last `required`
 property was removed from a type, the compiler would no longer synthesize this `modreq`, which is a binary-breaking change and all consumers would
 need to be recompiled. A compiler that understands `required` members will ignore this obsolete attribute. Note that members can come from base types
-as well: even there are no new `required` members in the current type, if any base type has `required` members, this `Obsolete` attribute will be
+as well: even if there are no new `required` members in the current type, if any base type has `required` members, this `Obsolete` attribute will be
 generated. If the constructor already has an `Obsolete` attribute, no additional attribute will be generated. If a by-ref-like type also has `required`
 members, the generated `Obsolete` attribute will reference the `required` error message, under the assumption that if a compiler is new enough to
-understand `required` properties, it will also understand by-ref-like-types.
+understand `required` properties, it will also understand by-ref-like types.
 
-To build the full list of all `required` members for a given type T, including all base types, all the `RequiredMembersAttribute`s from T and its
-base type Tb (recursively until reaching `System.Object`) are collected. At every Ti, if it has a `RequiredMembersAttribute` with members R1..Rn,
+To build the full list of `required` members for a given type `T`, including all base types, all the `RequiredMembersAttribute`s from `T` and its
+base type `Tb` (recursively until reaching `System.Object`) are collected. At every `Ti`, if it has a `RequiredMembersAttribute` with members `R1`..`Rn`,
 the following algorithm is performed:
 
-1. Perform standard member lookup on Ti for name Ri with 0 arguments.
-2. If this lookup was ambiguous or did not produce a single field or property result, an error occurs and the required member list of T cannot be
+1. Perform standard member lookup on `Ti` for name `Ri` with 0 arguments.
+2. If this lookup was ambiguous or did not produce a single field or property result, an error occurs and the required member list of `T` cannot be
 determined. No further steps are taken, and calling any constructor on `T` not marked with a `NoRequiredMembersAttribute` issues an error.
 3. Otherwise, the result of the member lookup is added to `T`'s list of required members.
 
 ## Open Questions
 
-### Requiring less-public members
+### Accessibility requirements and `init`
 
 In versions of this proposal with the `init` clause, we talked about being able to have the following scenario:
 
@@ -381,13 +381,23 @@ previously.
 The current spec says that the `required` keyword needs to be copied over and that overrides can make a member _more_ required, but not less. Is that what we want to do?
 Allowing removal of requirements needs more contract modification abilities than we are currently proposing.
 
+### Alternative metadata representation
+
+We could also take a different approach to metadata representation, taking a page from extension methods. We could put a `RequiredMemberAttribute` on the type to indicate
+that the type contains required members, and then put a `RequiredMemberAttribute` on each member that is required. This would simplify the lookup sequence (no need to do
+member lookup, just look for members with the attribute).
+
 ### Metadata Representation
 
 The [Metadata Representation](#metadata-representation) needs to be approved. We additionally need to decide whether these attributes should be included in the BCL.
 
 1. For `RequiredMembersAttribute`, this attribute is more akin to the general embedded attributes we use for nullable/nint/tuple member names, and will not be manually
-applied by the user.
+applied by the user in C#. It's possible that other languages might want to manually apply this attribute, however.
 2. `NoRequiredMembersAttribute`, on the other hand, is directly used by consumers, and thus should likely be in the BCL.
+
+If we go with the alternative representation in the previous section, that might change the calculus on `RequiredMemberAttribute`: instead of being similar to the general
+embedded attributes for `nint`/nullable/tuple member names, it's closer to `System.Runtime.CompilerServices.ExtensionAttribute`, which has been in the framework since
+extension methods shipped.
 
 ### Warning vs Error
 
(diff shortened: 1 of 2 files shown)


333fred:
@jcouv @RikkiGibson for review.

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
 A type with a parameterless constructor that advertises a _contract_ is not allowed to be substituted for a type parameter constrained to `new()`, as there is no way
 for the generic instantiation to ensure that the requirements are satisfied.
 
+### Overriding, Hiding, and Inheriting
+
+It is an error to mark a member required if the member is not at least as visible as the containing member's type. This means the following cases are not allowed:
```
RikkiGibson, line 268: Consider using the phrase "less accessible than" instead of "not at least as visible as".

Consider using "containing type" instead of "containing member's type". If the thing contains required members, it is always a type, right?

jcouv, line 268: >containing member's type [](http://example.com/codeflow?start=89&length=24)

as the member's containing type?

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+
+When overriding a `required` member, the `required` keyword must be included on the method signature. This is done so that if we ever want to allow
+unrequiring a property with an override in the future, we have design space to do so.
+
+Overrides are allowed to mark a member `required` where it was not `required` in the base type. A member so-marked is added to the required members
+list of derived type.
+
+### Metadata Representation
+
+The following 2 attributes are known to the C# compiler and required for this feature to function:
```
RikkiGibson, line 301: > required for this feature to function

Is this to say that the compiler will just error in the absence of these? It won't synthesize them?

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+when an override adds `required` to a member that was not `required` in the base type. As these members can only be fields or properties, duplicate
+names cannot exist.
+
+Any constructor in a type with `required` members that does not have `NoRequiredMembers` applied to it is marked with an `Obsolete` attribute with
+the string `"Types with required members are not supported in this version of your compiler"`, and the attribute is marked as an error, to prevent
+any older compilers from using these constructors. We don't use a `modreq` here because it is a goal to maintain binary compat: if the last `required`
+property was removed from a type, the compiler would no longer synthesize this `modreq`, which is a binary-breaking change and all consumers would
+need to be recompiled. A compiler that understands `required` members will ignore this obsolete attribute. Note that members can come from base types
+as well: even there are no new `required` members in the current type, if any base type has `required` members, this `Obsolete` attribute will be
```
RikkiGibson, line 331: "even if there are"?

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+}
+```
+
+It is an error to hide a `required` member, as that member can no longer be set by a consumer.
+
+When overriding a `required` member, the `required` keyword must be included on the method signature. This is done so that if we ever want to allow
+unrequiring a property with an override in the future, we have design space to do so.
+
+Overrides are allowed to mark a member `required` where it was not `required` in the base type. A member so-marked is added to the required members
+list of derived type.
```
RikkiGibson, line 297: "of the derived type"?

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+The following 2 attributes are known to the C# compiler and required for this feature to function:
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = false)]
+public sealed class RequiredMembersAttribute : Attribute
+{
+    public RequiredMembersAttribute(string[] members) => Members = members;
+    public string[] Members { get; }
```
RikkiGibson, line 310: It might be worthwhile to discuss in an "alternatives" section why this encoding is preferable to putting a `[Required]` attribute straight on the members which were `required` in source, and perhaps putting some attribute on the containing type to indicate "this type contains required members", similar to types that contain extension methods.

RikkiGibson (approved):

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+public class Derived : Base, I
+{
+    required int I.Prop1 { get; } // Error: explicit interface implementions cannot be required as they cannot be set in an object initializer
+
+    public required override int Prop2 { get; } // Error: this property is hidden by Derived.Prop2 and cannot be set in an object initializer
+    public new int Prop2 { get; }
+}
+```
+
+It is an error to hide a `required` member, as that member can no longer be set by a consumer.
```
jcouv, line 291: nit: I don't know if Mads or someone else has created a proposal for addressing members of specific types (something like `x.(Type.Method)(argument)`). If there's one, let's add a note there about this scenario.

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+[AttributeUsage(AttributeTargets.Constructor)]
+public sealed class NoRequiredMembersAttribute : Attribute
+{
+    public NoRequiredMembersAttribute() {}
+}
+```
+
+It is an error to manually apply `RequiredMembersAttribute` to a type.
+
+All members that are required by a type are listed in a `RequiredMembersAttribute` on the type. Members from any base type are not included, except
```
jcouv, line 322: >Members from any base type are not included [](http://example.com/codeflow?start=96&length=43)

Do we have any precedent in terms of encoding incrementally at each layer vs. encoding the full/accumulated set?

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+names cannot exist.
+
+Any constructor in a type with `required` members that does not have `NoRequiredMembers` applied to it is marked with an `Obsolete` attribute with
+the string `"Types with required members are not supported in this version of your compiler"`, and the attribute is marked as an error, to prevent
+any older compilers from using these constructors. We don't use a `modreq` here because it is a goal to maintain binary compat: if the last `required`
+property was removed from a type, the compiler would no longer synthesize this `modreq`, which is a binary-breaking change and all consumers would
+need to be recompiled. A compiler that understands `required` members will ignore this obsolete attribute. Note that members can come from base types
+as well: even there are no new `required` members in the current type, if any base type has `required` members, this `Obsolete` attribute will be
+generated. If the constructor already has an `Obsolete` attribute, no additional attribute will be generated. If a by-ref-like type also has `required`
```
jcouv, line 332: >If the constructor already has an `Obsolete` attribute, no additional attribute will be generated. [](http://example.com/codeflow?start=11&length=98)

Is that what we do for other "not supported in this version of your compiler" Obsolete attributes? What if the explicit Obsolete attribute is only marked as "warning"?

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+members, the generated `Obsolete` attribute will reference the `required` error message, under the assumption that if a compiler is new enough to
+understand `required` properties, it will also understand by-ref-like-types.
+
+To build the full list of all `required` members for a given type T, including all base types, all the `RequiredMembersAttribute`s from T and its
```
jcouv, line 336: >T [](http://example.com/codeflow?start=66&length=1)

nit: may want to add back-ticks to the names in this sentence (`T`, `Tb`, `Ti`, `R1`, etc)
nit: "all" feels redundant in "the full list of all `required` members"

This is not a spec problem, but we'll want to test with invalid attributes on base type (`RequiredMembers("missing")` for instance). Will add a note to test plan.

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+
+1. Perform standard member lookup on Ti for name Ri with 0 arguments.
+2. If this lookup was ambiguous or did not produce a single field or property result, an error occurs and the required member list of T cannot be
+determined. No further steps are taken, and calling any constructor on `T` not marked with a `NoRequiredMembersAttribute` issues an error.
+3. Otherwise, the result of the member lookup is added to `T`'s list of required members.
+
 ## Open Questions
 
-### Syntax questions
+### Requiring less-public members
```
jcouv, line 347: >Requiring less-public members [](http://example.com/codeflow?start=4&length=29)

nit: Sounds awkward. Consider "Accessibility requirements and `init`"

proposals/required-members.md
```diff
@@ -263,42 +263,131 @@ init_argument_initializer
...
+public class Base
+{
+    public virtual int Prop2 { get; }
+
+    protected required int _field; // Error: _field is not at least as visible as Base. Open question below about the protected constructor scenario
+    protected Base() { }
+}
+public class Derived : Base, I
+{
+    required int I.Prop1 { get; } // Error: explicit interface implementions cannot be required as they cannot be set in an object initializer
```
jcouv, line 284: >object initializer [](http://example.com/codeflow?start=124&length=18)

Not blocking this PR: I feel like the spec is missing a section on the caller side of the story, ie object initializers ;-) 
We should talk about nested initializers (difference between `{ x = new Point { ... } }` and `{ x = { ... } }`).

jcouv (commented):

proposals/required-members.md
333fred: I don't know of one. But the whole point of this feature is to encode by layer, not the full set.

proposals/required-members.md
333fred: Yes, it is. If a type has an obsolete marker, it takes precedence, no matter the warning level. This is even true for `ref struct` types: https://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Symbols/Source/SourceNamedTypeSymbol.cs,1544

proposals/required-members.md
333fred: That's an open question below.

proposals/required-members.md
333fred: Added a question for this.

proposals/required-members.md
333fred: @MadsTorgersen?

jcouv (approved):
LGTM Thanks (iteration 2)

RikkiGibson (approved):

PR #5573, opened 2021-12-24
Open questions for semi auto properties

Youssef1313:


CyrusNajmabadi (approved):

proposals/semi-auto-properties.md
```diff
@@ -108,6 +108,24 @@ Open LDM questions:
 
 1. If a type does have an existing accessible `field` symbol in scope (like a field called `field`) should there be any way for an auto-prop to still use `field` internally to both create and refer to an auto-prop field.  Under the current rules there is no way to do that.  This is certainly unfortunate for those users, however this is ideally not a significant enough issue to warrant extra dispensation.  The user, after all, can always still write out their properties like they do today, they just lose out from the convenience here in that small case.
 
+2. Should initializers use the backing field or the property setter? If the latter, what about `public int P { get => field; } = 5;`?
+
+3. Can the field be written into in a property without setter?
```
jnm2, line 113: I definitely hope so: `public string Foo => field ??= GetFoo();`

CyrusNajmabadi: Yes. This is a core scenario. It has to work.

CyrusNajmabadi: For '2' I think it should use the setter.

Youssef1313: Should I remove `3` from the PR if the answer is clear and doesn't need to go in LDM?

> For '2' I think it should use the setter.

And if there is no setter? Should be error or use the field directly?

jnm2: Thus far, constructor property assignments fall back to using the field directly if there is no setter.

I'm feeling skeptical that property initializers calling the setter will be a nice surprise.

CyrusNajmabadi: > Should I remove 3 from the PR if the answer is clear and doesn't need to go in LDM?

Yup

CyrusNajmabadi: > And if there is no setter? Should be error or use the field directly?

The latter.  Just like a normal auto prop without a setter.

CyrusNajmabadi: > And if there is no setter? Should be error or use the field directly?

The latter.  Just like a normal auto prop without a setter.

CyrusNajmabadi: > A principle in the proposal is that get; and get => field; have identical semantics.

Yup yup

jnm2: With `public int Foo { get; set => DoStuff(ref field, value); } = 1;`, I think we'll want a way to initialize the field without executing `DoStuff`, without losing the ability to use an autoprop/`field`.

There's already a way to go the other way if you *do* want to explicitly execute the setter: turn the property initializer into a constructor assignment.

Youssef1313: For `2`, I agree with @jnm2 that it's more flexible if the initializer uses the backing field.

For `3`, I removed it from this PR and opened a spec change: https://github.com/dotnet/csharplang/pull/5575

proposals/semi-auto-properties.md
```diff
@@ -110,22 +110,6 @@ Open LDM questions:
 
 2. Should initializers use the backing field or the property setter? If the latter, what about `public int P { get => field; } = 5;`?
 
-3. Can the field be written into in a property without setter?
```
AlekseyTs: I am not sure why this issue got removed. I think the current specification disallows that because it says that the backing field should be readonly when there is no setter. Is there a plan to ament the specification to enable that. Until that is done, I would keep this on the list.

CyrusNajmabadi: > Is there a plan to amend the specification to enable that? Until that is done, I would keep this on the list.

Yes.  There is a plan to amend the specification.  This was simply trying to maintain the rule that if you had `T P { get; }` that it was readonly.  

Trying to think of every single case and subtlety was just too difficult when doing the initial spec.  So we wrote it up as best as we could, but the purpose of this prototype was to help uncover those 'a ha!' cases that would help us revise.

proposals/semi-auto-properties.md
```diff
@@ -108,6 +108,8 @@ Open LDM questions:
 
 1. If a type does have an existing accessible `field` symbol in scope (like a field called `field`) should there be any way for an auto-prop to still use `field` internally to both create and refer to an auto-prop field.  Under the current rules there is no way to do that.  This is certainly unfortunate for those users, however this is ideally not a significant enough issue to warrant extra dispensation.  The user, after all, can always still write out their properties like they do today, they just lose out from the convenience here in that small case.
 
+2. Should initializers use the backing field or the property setter? If the latter, what about `public int P { get => field; } = 5;`?
```
AlekseyTs, line 111: I think we should capture the following information:
- Calling a setter for an initializer is not an option because initializers are processed before calling base constructor and it is illegal to call any instance method before the base constructor is called.
- If we say that initializer assigns directly to the backing field.  If there is a setter, then we are getting into a situation when initializer does one thing and an assignment to the property within constructor does a different thing (calls the setter). A behavior like that can be a trap for a user. Today, there is no semantical difference between an initializer and an assignment in constructor when both are allowed. This invariant will be broken. However, people are changing between the forms often.

CyrusNajmabadi: Ah, that's a great point.  Will def revise things.

proposals/semi-auto-properties.md
AlekseyTs: That is good. I suggest keeping this item as a record of our intent to amend the specification. It can be removed at the same time the specification is adjusted. I assume this is going to happen in this document as well. This will remind us to do the work and will inform other readers that we are aware of the issue and tracking it.

Youssef1313: @AlekseyTs The spec is now updated. Should I remove `3` again?

AlekseyTs: > The spec is now updated. Should I remove `3` again?

If that change got merged before this one, it might make sense to remove it.

AlekseyTs:
Another design question related to initialization came to mind. How compiler is supposed to know when all structure fields are initialized?

For example, the following scenario doesn't compile with a regular property. Is it supposed to compile with a ```field```-like property? 
```
struct S
{
    S(int x)
    {
        P = 1;
    }
    
    //private int field;
    int P
    {
        get => field;
        set => field = value;
    }
}
```

Youssef1313:
@AlekseyTs I think this is scheduled for a LDM? https://github.com/dotnet/csharplang/commit/2cf6389e2632dd2200865da3d35483c506a2e77b

AlekseyTs:
>  I think this is scheduled for a LDM?

Do we want to capture this as an open question here?

proposals/semi-auto-properties.md
AlekseyTs: @Youssef1313 Do you plan to act on this suggestion?

proposals/semi-auto-properties.md
CyrusNajmabadi: I'm coming around to this position.  Though i think it's def got a downside (as well as a virtue) that field-assign and constructor-assign mean separate things.

For now, i think we shoudl make it so that:

1. property-init means field-assign (it legit can't mean 'call setter' unless we do some interesting emit stuff).
2. constructor-assign means 'call setter'.
3. we bring this to LDM to verify this is the behavior we want.

jnm2: Additionally, if the property initializer does mean calling the setter, then people who actually can't invoke the setter are pushed over the cliff and have to refactor to declaring their own field. This pain could be mitigated by implementing property-scoped fields, making this syntax possible:

```cs
public int Foo { int field = 1; get; set => Set(ref field, value); }
```

Even with that ability, I would personally still prefer this syntax to not call the setter:
```cs
public int Foo { get; set => Set(ref field, value); } = 1;
```

proposals/semi-auto-properties.md
333fred: I pushed a commit adding these points.

AlekseyTs (approved):
LGTM

proposals/semi-auto-properties.md
jnm2: Adding this info in https://github.com/dotnet/csharplang/pull/5837

PR #5575, opened 2021-12-25
Spec change around backing field being readonly

Youssef1313:
cc @AlekseyTs @CyrusNajmabadi

### DIFF
--- a/proposals/semi-auto-properties.md
+++ b/proposals/semi-auto-properties.md
@@ -54,7 +54,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 
 ```diff
 - If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
-+ If the auto-property has semicolon-only get accessor without a set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a semicolon-getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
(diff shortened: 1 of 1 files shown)


proposals/semi-auto-properties.md
```diff
@@ -52,6 +52,13 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 
 ...
 
+```diff
+- If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
++ If the auto-property has semicolon-only get accessor without a set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a semicolon-getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
```
Youssef1313, line 57: I'm not sure if "semicolon-getter-only" is clear enough.

jnm2: It is to me, or it could be `an auto property which only has a semicolon getter`. I'd nitpick:

- has *a* semicolon-only get accessor
- `auto property` not `auto-property` (I know, this was original)

jnm2: I wonder if the spec should say something about getter-only properties being assignable in the body of a constructor even when the getter is not semicolon-only and therefore the backing field is not readonly.

Youssef1313: Yup I just noticed an overlap and a possibility to divide this into two separate sentences depending on what the initializer behavior would be. Let me mark as a draft for now as it doesn't seem to be accurate.

Youssef1313: I think the wording is more accurate now?

333fred (approved):

CyrusNajmabadi (approved):

proposals/semi-auto-properties.md
jnm2: Lgtm!

CyrusNajmabadi:
thanks! :)

RikkiGibson:
It feels like if you had a `get; init => field = Validate(value);` property, you might expect that:

1. Backing field is readonly
2. Init accessor will be called when property is assigned in constructor

Does the spec account for this as written?

Youssef1313:
The spec doesn't account for that currently. Let me open a new PR for that.

PR #5582, opened 2021-12-27
Update semi auto props spec around readonly backing field for init-only

Youssef1313:
Addresses https://github.com/dotnet/csharplang/pull/5575#issuecomment-1001714937

cc @RikkiGibson @AlekseyTs @333fred @CyrusNajmabadi  @jnm2

### DIFF
--- a/proposals/semi-auto-properties.md
+++ b/proposals/semi-auto-properties.md
@@ -54,7 +54,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 
 ```diff
 - If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
-+ If the auto-property has semicolon-only get accessor without a set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
++ If the auto-property has semicolon-only get accessor (without a set accessor or with an init-only accessor), the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
 ```
 
 ...
--- a/proposals/semi-auto-properties.md
+++ b/proposals/semi-auto-properties.md
@@ -54,7 +54,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 
 ```diff
 - If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
-+ If the auto-property has semicolon-only get accessor (without a set accessor or with an init-only accessor), the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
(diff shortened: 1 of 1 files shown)


proposals/semi-auto-properties.md
```diff
@@ -54,7 +54,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 
 ```diff
 - If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
-+ If the auto-property has semicolon-only get accessor without a set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
++ If the auto-property has semicolon-only get accessor (without a set accessor or with an init-only accessor), the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
 ```
```
CyrusNajmabadi, line 58: lol.  i got so confused fro a second.  i was going to say these two seem almost identical and should be merged.  then i realized i was looking at a diff 🗡️

proposals/semi-auto-properties.md
```diff
@@ -54,7 +54,7 @@ The following changes are to be made to [classes.md](https://github.com/dotnet/c
 
 ```diff
 - If the auto-property has no set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto-property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
-+ If the auto-property has semicolon-only get accessor without a set accessor, the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
++ If the auto-property has semicolon-only get accessor (without a set accessor or with an init-only accessor), the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
```
CyrusNajmabadi, line 57: `init-only accessor` or just `init accessor`

CyrusNajmabadi (approved):

proposals/semi-auto-properties.md
Youssef1313:
```suggestion
+ If the auto-property has semicolon-only get accessor (without a set accessor or with an init accessor), the backing field is considered `readonly` ([Readonly fields](classes.md#readonly-fields)). Just like a `readonly` field, a getter-only auto property can also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the backing field of the property.
```

Youssef1313: Yup it's just `init accessor`.

proposals/semi-auto-properties.md
Youssef1313: Yeah reviewing a diff in a diff hurts eyes 😄

333fred (approved):

proposals/semi-auto-properties.md
jnm2: Not sure the added parens are right for this, since they aren't expanding into a more detailed version of the thing right before the parens.

jnm2: Maybe similar to this?

> If the auto-property has semicolon-only get accessor and has no set accessor (or has an init accessor),

proposals/semi-auto-properties.md
CyrusNajmabadi: hrmm... does teh constructor actually assign to the field?  it doesn't go through the init-setter?

Youssef1313: @CyrusNajmabadi I think the spec is now wrong regarding this point. I focused more on the first sentence. We need to adjust.

CyrusNajmabadi: yeah, i think the second sentence needs to be clear that it's only covering the `{get;}` case and not the `init` case.

Youssef1313: Actually it seems like it's accurate already? It mentions `getter-only auto property`, so it doesn't apply for the "with an init-only accessor" part of the previous sentence.

CyrusNajmabadi: right.  i think the language is correct, but potentially confusing in context.  i wasn't sure if `getter only` meant: only `{ get; }`, or also somehow applied to get+init (since hte prior sentence covers that).  I feel like it could use with jsut a little clarity, only to avoid confusion.

PR #5602, opened 2022-01-03
Low level struct updates

jaredpar:
Updates to the low level struct improvements document. Changes:

1. Tried to put the compat issue up front. Want readers to understand that as a separate concept as it comes up many times in the feature design
2. Solved the issue of how non-constructors could return `ref struct` which had `ref` fields
3. Have firm rules now about which life time annotations we will support and how they interact with the existing rules

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -7,7 +7,7 @@ This proposal is an aggregation of several different proposals for `struct` perf
 ## Motivation
 Earlier versions of C# added a number of low level performance features to the language: `ref` returns, `ref struct`, function pointers, etc. ... These enabled .NET developers to create write highly performant code while continuing to leverage the C# language rules for type and memory safety.  It also allowed the creation of fundamental performance types in the .NET libraries like `Span<T>`.
 
-As these features have gained traction in the .NET ecosystem developers, both internal and external, have been providing us with information on remaining friction points in the ecosystem. Places where they still need to drop to `unsafe` code to get their work, or require the runtime to special case types like `Span<T>`. 
+As these features have gained traction in the .NET ecosystem developers, both internal and external, have been providing us with information on remaining friction points in the ecosystem. Places where they still need to drop to `unsafe` code to get their work done, or require the runtime to special case types like `Span<T>`. 
 
 Today `Span<T>` is accomplished by using the `internal` type `ByReference<T>` which the runtime effectively treats as a `ref` field. This provides the benefit of `ref` fields but with the downside that the language provides no safety verification for it, as it does for other uses of `ref`. Further only dotnet/runtime can use this type as it's `internal`, so 3rd parties can not design their own primitives based on `ref` fields. Part of the [motivation for this work](https://github.com/dotnet/runtime/issues/32060) is to remove `ByReference<T>` and use proper `ref` fields in all code bases. 
 
@@ -506,7 +506,7 @@ One of the most notable friction points is the inability to return fields by `re
 ```c#
 struct S
 {
-     _field;
+     int _field;
 
     // Error: this, and hence _field, can't return by ref
     public ref int Prop => ref _field;
@@ -530,7 +530,7 @@ struct S
 
 // Option 2 
 // The ref-safe-to-escape for this is now the calling method on all members
-[ThisRefEcapes]
+[RefThisEscapes]
 struct S
 {
     int _field;
@@ -816,7 +816,7 @@ ref struct S<T> {
 ```
 
 ### Why isn't assignability to ref field a scope?
-It may seem attractive to think of the ability to assign to a `ref` field as just another scope concept instead of an orthogonal concept. Essentially *assignable to field* as an escape scope between *heap* and *calling method*. This falls a part though when looking at practical examples. Consider for example the following: 
+It may seem attractive to think of the ability to assign to a `ref` field as just another scope concept instead of an orthogonal concept. Essentially *assignable to field* as an escape scope between *heap* and *calling method*. This falls apart though when looking at practical examples. Consider for example the following: 
 
 ```c#
 void M(ref int parameter)
@@ -833,9 +833,9 @@ The only practical way to represent whether a location can assign to a `ref` fie
 ## Open Issues
 
 ### Take the breaking change
-Consider for a minute a design where the compat problem was approached from the other direction. Effectively make it such that every method was implicitly `[RefFieldsEscape]` and then have an attribute that restores the span safety rules in place today. Say `[RefFieldDoesNotEsacpe]`. 
+Consider for a minute a design where the compat problem was approached from the other direction. Effectively make it such that every method was implicitly `[RefFieldsEscape]` and then have an attribute that restores the span safety rules in place today. Say `[RefFieldDoesNotEscape]`. 
 
-This would be a breaking change and it's easy to construct code samples that trigger an error when upgrading to a new version of C# (as demonstrated [here](#new-span-challenges)). It's hard to determine though how prevelant these types of patterns are. Essentially methods which have both the following attributes:
+This would be a breaking change and it's easy to construct code samples that trigger an error when upgrading to a new version of C# (as demonstrated [here](#new-span-challenges)). It's hard to determine though how prevalent these types of patterns are. Essentially methods which have both the following attributes:
 
 - Have a `Span<T>` or `ref struct` which has a `Span<T>` as
     - The return type 
@@ -858,7 +858,7 @@ Span<int> CreateSpan(ref int i)
 Span<int> Method()
 {
     int local = 42;
-    var span = new Span<int>(ref local);
+    var span = CreateSpan(ref local);
     return span;
 }
 ```
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -1,38 +1,6 @@
 Low Level Struct Improvements
 =====
 
-***
-TODO
-types of lifetime capture we need to solve with annotations
-- methods that capture a `ref` into a `ref` field (true for `ref` parameter or `ref`
-    field of an argument) and that is returned to the caller. Do we annotate the value
-    or the method
-- `[RefEscapes]` `this` of a `struct` being ref-safe-to-escape (maybe this is a generalized `ref`
-    escape that could be applied to existing `ref` parameter)
-- `[DoesNotRefEscape]` valuable once we have `ref` fields and we need to distinguish
-    which of the parameters is the captured ones 
-- `[DoesNotEscape]` used to indicate that a `ref struct` can't escape from a method 
-    and that helps infer lifetime issues
-
-types of lifetime annotations that are not needed
-- `[Escapes]` this is always true for non `ref struct` values and for a `ref struct`
-    it is the natural default. Can't think of a case where it would be needed
-- `[DoesNotInput]` today `out` is treated no differently than `ref` and that can cause
-    confusion in places. This could be fixed by an annotation where we don't consider
-    the input of a `out` in lifetime calculation. This is narrow though and don't 
-    see the need to do it.
-
-annotations need to be subject to verification. Consider marking a normal parameter
-as `[RefEscapes]` is non-sense and must be a compilation error. Likely need an allow
-list for when these can be added
-
-what is the default escape rules of `ref` fields?
-
-intersection of annotations and the `[InlineArray]` attribute. Maybe it implies and 
-/ or requires `[RefEscapes]`
-
-***
-
 ## Summary
 This proposal is an aggregation of several different proposals for `struct` performance improvements: ref fields and attributes to override lifetime defaults.  The goal being a design which takes into account the various proposals to create a single overarching feature set for `struct` improvements.
 
@@ -51,7 +19,7 @@ remove special case type like `ByReference<T>`
 in `struct`
 
 ## Detailed Design 
-The rules for `ref struct` safety are defined in the [span-safety document](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md).  This document will describe the required changes to this document as a result of this proposal. Once accepted as an approved feature these changes will be incorporated into that document.
+The rules for `ref struct` safety are defined in the [span safety document](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md).  This document will describe the required changes to this document as a result of this proposal. Once accepted as an approved feature these changes will be incorporated into that document.
 
 ### Provide ref fields
 The language will allow developers to declare `ref` fields inside of a `ref struct`. This can be useful for example when encapsulating large mutable `struct` instances or defining high performance types like `Span<T>` in libraries besides the runtime.
@@ -62,7 +30,7 @@ Before diving into the problems here it should be noted that `ref` fields only r
 
 To understand the challenges here let's first consider how `Span<T>` will look once `ref` fields are supported.
 
-```cs
+```c#
 // This is the eventual definition of Span<T> once we add ref fields into the
 // language
 readonly ref struct Span<T>
@@ -84,7 +52,7 @@ readonly ref struct Span<T>
 
 The constructor defined here presents a problem because its return values must necessarily have restricted lifetimes for many inputs. Consider that if a local parameter is passed by `ref` into this constructor that the returned `Span<T>` must have a *safe-to-escape* scope of the local's declaration scope.
 
-```cs
+```c#
 Span<int> CreatingAndReturningSpan()
 {
     int i = 42;
@@ -101,7 +69,7 @@ Span<int> CreatingAndReturningSpan()
 
 At the same time it is legal to have methods today which take a `ref` parameter and return a `Span<T>`.  These methods bear a lot of similarity to the newly added `Span<T>` constructor: take a `ref`, return a `Span<T>`. However the lifetime of the return value of these methods is never restricted by the inputs.  The existing span safety rules consider such values as effectively always *safe-to-escape* outside the enclosing method.
 
-```cs
+```c#
 class ExistingScenarios
 {
     Span<T> CreateSpan<T>(ref T p)
@@ -128,7 +96,7 @@ class ExistingScenarios
 
 The reason that all of the above samples are legal is because in the existing design there is no way for the return `Span<T>` to store a reference to the input state of the method call. This is because the span safety rules explicitly depend on `Span<T>` not having a constructor which takes a `ref` parameter and stores it as a field. 
 
-```cs
+```c#
 class ExistingAssumptions
 {
     Span<T> CreateSpan<T>(ref T p)
@@ -144,7 +112,7 @@ class ExistingAssumptions
 
 The rules we define for `ref` fields must ensure the `Span<T>` constructor properly restricts the *safe-to-escape* scope of constructed objects in the cases it captures `ref` state. At the same time it must ensure that we don't break the existing consumption rules for methods like `CreateSpan<T>`. 
 
-```cs
+```c#
 class GoalOfRefFields
 {
     Span<T> CreateSpan<T>(ref T p)
@@ -181,7 +149,7 @@ To do this we will change the escape rules for a constructor invocation, which t
 
 Let's examine these rules in the context of samples to better understand their impact.
 
-```cs
+```c#
 ref struct RS
 {
     ref int _field;
@@ -233,7 +201,7 @@ ref struct RS
 
 It is important to note that for the purposes of the rule above any use of constructor chaining via `this` is considered a constructor invocation. The result of the chained constructor call is considered to be returning to the original constructor hence *safe-to-escape* rules come into play. That is important in avoiding unsafe examples like the following:
 
-```cs
+```c#
 ref struct RS1
 {
     ref int _field;
@@ -275,7 +243,7 @@ The limiting of the constructor rules to just `ref struct` that directly contain
 
 Example of where this comes into play.
 
-```cs
+```c#
 ref struct Container
 {
     LargeStruct _largeStruct;
@@ -291,7 +259,7 @@ ref struct Container
 
 Much like the `CreateSpan<T>` example before the *safe-to-escape* return of the `Container` constructor is not impacted by the `largeStruct` parameter. If the new constructor rules were applied to this type then it would break compatibility with existing code. The existing rules are also sufficient for existing constructors to prevent them from simulating `ref` fields by storing them into `Span<T>` fields.
 
-```cs
+```c#
 ref struct RS4
 {
     Span<int> _span;
@@ -346,7 +314,7 @@ The fourth item provides the critical safety point around a `ref struct` exposin
 
 Consider as an example the indexer on `Span<T>` which is returning `ref` fields by `ref` today. The fourth item here is what provides the safety here:
 
-```cs
+```c#
 ref int Examples()
 {
     Span<int> s1 = stackalloc int[5];
@@ -372,7 +340,7 @@ To account for `ref` fields the *ref-safe-to-escape* rules for fields will be ad
 
 This explicitly allows for `ref` fields being returned as `ref` from a `ref struct` but not normal fields (that will be covered later). 
 
-```cs
+```c#
 ref struct RS
 {
     ref int _refField;
@@ -436,7 +404,7 @@ A `ref` field can only be assigned outside a constructor when the value is known
 
 This design does not allow for general `ref` field assignment outside object construction due to existing limitations on lifetimes. Specifically it poses challenges for scenarios like the following:
 
-```cs
+```c#
 ref struct SmallSpan
 {
     public ref int _field;
@@ -489,7 +457,7 @@ A `ref` field will be emitted into metadata using the `ELEMENT_TYPE_BYREF` signa
 
 Developers can continue to initialize a `ref struct` with a `ref` field using the `default` expression in which case all declared `ref` fields will have the value `null`. Any attempt to use such fields will result in a `NullReferenceException` being thrown.
 
-```cs
+```c#
 struct S1 
 {
     public ref int Value;
@@ -501,7 +469,7 @@ local.Value.ToString(); // throws NullReferenceException
 
 While the C# language pretends that a `ref` cannot be `null` this is legal at the runtime level and has well defined semantics. Developers who introduce `ref` fields into their types need to be aware of this possibility and should be **strongly** discouraged from leaking this detail into consuming code. Instead `ref` fields should be validated as non-null using the [runtime helpers](https://github.com/dotnet/runtime/pull/40008) and throwing when an uninitialized `struct` is used incorrectly.
 
-```cs
+```c#
 struct S1 
 {
     private ref int Value;
@@ -521,31 +489,34 @@ struct S1
 Misc Notes:
 - A `ref` field can only be declared inside of a `ref struct` 
 - A `ref` field cannot be declared `static`
-- A `ref` field can only be `ref` assigned in the constructor of the declaring
-type.
-- The reference assembly generation process must preserve the presence of a 
-`ref` field inside a `ref struct` 
+- A `ref` field can only be `ref` assigned in the constructor of the declaring type.
+- The reference assembly generation process must preserve the presence of a `ref` field inside a `ref struct` 
 - A `readonly ref struct` must declare its `ref` fields as `readonly ref`
-- The span safety rules for constructors, fields and assignment must be updated
-as outlined in this document.
-- The span safety rules need to include the definition of `ref` values that 
-"refer to the heap". 
+- The span safety rules for constructors, fields and assignment must be updated as outlined in this document.
+- The span safety rules need to include the definition of `ref` values that "refer to the heap". 
 
-### Provide struct this escape annotation
-The rules for the scope of `this` in a `struct` limit the *ref-safe-to-escape* scope to the current method. That means neither `this`, nor any of its fields can return by reference to the caller.
+### Provide lifetime annotations
+The [span safety document](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md) assigns escape scopes to locations based on their declaration: parameters are *ref-safe-to-escape* to calling method, `this` in a `struct` is *ref-safe-to-escape* within the current method, etc ... These defaults were chosen to make `Span<T>` and `ref` returns work with the predominant coding patterns in .NET.
 
-```cs
+The lifetime decisions for these locations is not fundamental to correctness. For example parameters could default to *ref-safe-to-escape* to within the current method or `this` in a `struct` could be *ref-safe-to-escape* to the calling method in our span safety rules and it would not make `ref struct` usage unsafe. It would just make it significantly less usable.
+
+While the defaults have allowed broad adoption of `ref struct` within .NET they do create a number of friction points in low level code. These friction points often force developers to resort to using `unsafe` which de-values our `ref struct` efforts. 
+
+One of the most notable friction points in the inability to use `ref` returns within a `struct`. This means developers can't create `ref` returning methods / properties and have to resort to exposing fields directly.
+
+```c#
 struct S
 {
     int _field;
+
     // Error: this, and hence _field, can't return by ref
     public ref int Prop => ref _field;
 }
 ```
 
-There is nothing inherently wrong with a `struct` escaping `this` by reference.  Instead the justification for this rule is that it strikes a balance between the usability of `struct` and `interfaces`. If a `struct` could escape `this` by reference then it would significantly reduce the use of `ref` returns in interfaces.
+Again there is nothing inherently wrong with a `struct` escaping `this` by reference, it is simply the default chosen by the span safety rules. The justification for this rule is that it strikes a balance between the usability of `struct` and `interfaces`. If a `struct` could escape `this` by reference then it would significantly reduce the use of `ref` returns in interfaces.
 
-```cs
+```c#
 interface I1
 {
     ref int Prop { get; }
@@ -564,9 +535,57 @@ struct S1 : I1
 
 The justification here is reasonable but it also introduces unnecessary friction on `struct` members that don't participate in interface invocations. 
 
+Readers may have noticed that the existing defaults limit the usefulness of `ref` fields. The above section only provides for a `ref` value to be captured and returned from a constructor. No other method can do this because of our existing rules around `ref` capture. 
+
+```c#
+ref struct MySpan<T> {
+    ref T _field;o
+
+    // Legal and returns a MySpan<T> as outlined in this document
+    public MySpan(ref T field) {
+        _field = ref field
+    }
+
+    // Error: this is illegal because of compat rules that assume the `ref` cannot be 
+    // captured inside of the returned value
+    public static MySpan<T> Create(ref T field)
+    {
+        return new MySpan<T>(ref field);
+    }
+}
+```
+
+This, and several other friction points, can be removed if the language provides developers a way to invert the defaults by applying attributes to specific locations. 
+
+The `System.Runtime.CompilerServices.ThisRefEscapesAttribute` will be used to specify that `this` is *ref-safe-to-escape* to the calling method. It can be applied to individual methods or properties in `struct`. When applied to `struct` itself it is treated as if it were applied to all methods / properties in the type. 
+
+
+```c#
+// Option 1 
+struct S
+{
+    int _field;
+
+    // The ref-safe-to-escape for this is now the calling method hence this is legal
+    [ThisRefEscapes]
+    public ref int Prop => ref _field;
+}
+
+// Option 2 
+// The ref-safe-to-escape for this is now the calling method on all members
+[ThisRefEcapes]
+struct S
+{
+    int _field;
+
+    public ref int Prop => ref _field;
+}
+```
+
+**BELOW WILL EVENTUALLY BE DELETED**
 One key compatibility scenario that we have to keep in mind when approaching changes here is the following:
 
-```cs
+```c#
 struct S1
 {
     ref int GetValue() => ...
@@ -589,7 +608,7 @@ To remove this friction the language will provide the attribute `[ThisRefEscapes
 
 This allows for greater flexibility in `struct` definitions as they can begin returning `ref` to their fields. That allows for types like `FrugalList<T>`:
 
-```cs
+```c#
 struct FrugalList<T>
 {
     private T _item0;
@@ -617,7 +636,7 @@ struct FrugalList<T>
 
 This will naturally, by the existing rules in the span safety spec, allow for returning transitive fields in addition to direct fields.
 
-```cs
+```c#
 struct ListWithDefault<T>
 {
     private FrugalList<T> _list;
@@ -653,7 +672,7 @@ Misc Notes:
 ### Safe fixed size buffers
 The language will relax the restrictions on fixed sized arrays such that they can be declared in safe code and the element type can be managed or unmanaged.  This will make types like the following legal:
 
-```cs
+```c#
 internal struct CharBuffer
 {
     internal fixed char Data[128];
@@ -672,7 +691,7 @@ For each `fixed` declaration in a type where the element type is `T` the languag
 
 For example, the signature of the indexer for `CharBuffer.Data` will be the following:
 
-```cs
+```c#
 [ThisRefEscapes]
 internal ref char <>DataIndexer(int index) => ...;
 ```
@@ -700,7 +719,7 @@ This rule most commonly comes into play with instance methods on `ref struct` wh
 
 This rule exists to prevent scenarios like the following:
 
-```cs
+```c#
 ref struct RS
 {
     Span<int> _field;
@@ -724,7 +743,7 @@ Essentially this rule exists because the language must assume that all inputs to
 
 In practice though there are many such methods which never escape the parameter.  It is just a value that is used within the implementation. 
 
-```cs
+```c#
 ref struct JsonReader
 {
     Span<char> _buffer;
@@ -761,7 +780,7 @@ In order to work around this low level code will resort to `unsafe` tricks to li
 
 The other place where parameter default escape scope causes friction is when they are re-assigned within a method body. For instance if a method body decides to conditionally apply escaping to input by using stack allocated values. Once again this runs into some friction.
 
-```cs
+```c#
 void WriteData(ReadOnlySpan<char> data)
 {
     if (data.Contains(':'))
@@ -783,7 +802,7 @@ This pattern is fairly common across .NET code and it works just fine when a `re
 
 To remove this friction the language will provide the attribute `[DoesNotEscape]`. This can be applied to parameters of any type or instance members defined on `ref struct`. When applied to parameters the *safe-to-escape* and *ref-safe-to-escape* scope will be the current method scope. When applied to instance members of `ref struct` the same limitation will apply to the `this` parameter.
 
-```cs
+```c#
 class C
 {
     static Span<int> M1(Span<int> p1, [DoesNotEscape] Span<int> p2)
@@ -811,7 +830,7 @@ To account for this change the "Parameters" section of the span safety document 
 
 It's important to note that this will naturally block the ability for such parameters to escape by being stored as fields. Receivers that are passed by `ref`, or `this` on `ref struct`, have a *safe-to-escape* scope outside the current method. Hence assignment from a `[DoesNotEscape]` parameter to a field on such a value fails by existing field assignment rules: the scope of the receiver is greater than the value being assigned.
 
-```cs
+```c#
 ref struct S
 {
     Span<int> _field;
@@ -855,7 +874,7 @@ This does mean that program correctness will be defined in terms of attributes t
 ### Allow fixed buffer locals
 This design allows for safe `fixed` buffers that can support any type. One possible extension here is allowing such `fixed` buffers to be declared as local variables. This would allow a number of existing `stackalloc` operations to be replaced with a `fixed` buffer. It would also expand the set of scenarios we could have stack style allocations as `stackalloc` is limited to unmanaged element types while `fixed` buffers are not. 
 
-```cs
+```c#
 class FixedBufferLocals
 {
     void Example()
@@ -873,7 +892,7 @@ Example of where this would be beneficial: https://github.com/dotnet/runtime/pul
 ### Allow multi-dimensional fixed buffers
 Should the design for `fixed` buffers be extended to include multi-dimensional style arrays? Essentially allowing for declarations like the following:
 
-```cs
+```c#
 struct Dimensions
 {
     int array[42, 13];
@@ -885,7 +904,7 @@ struct Dimensions
 ### Allowing attributes on locals
 Another friction point for developers using `ref struct` is local variables can suffer from the same issues as parameters with respect to their lifetimes being decided at declaration. Than can make it difficult to work with `ref struct` that are assigned on multiple paths where at least one of the paths is a limited *safe-to-escape* scope. 
 
-```cs
+```c#
 int length = ...;
 Span<byte> span;
 if (length > StackAllocLimit)
@@ -902,7 +921,7 @@ else
 
 For `Span<T>` specifically developers can work around this by initializing the local with a `stackalloc` of size zero. This changes the *safe-to-escape* scope to be the current method and is optimized away by the compiler. It's effectively a syntax for making a `[DoesNotEscape]` local.
 
-```cs
+```c#
 int length = ...;
 Span<byte> span = stackalloc byte[0];
 if (length > StackAllocLimit)
@@ -947,7 +966,7 @@ This snippet wants to mutate a parameter by escaping elements of the data. The e
 
 ### Fun Samples
 
-```cs
+```c#
 ref struct StackLinkedListNode<T>
 {
     T _value;
@@ -983,3 +1002,41 @@ ref struct StackLinkedListNode<T>
     }
 }
 ```
+
+# TODO
+types of lifetime capture we need to solve with annotations
+- methods that capture a `ref` into a `ref` field (true for `ref` parameter or `ref`
+    field of an argument) and that is returned to the caller. Do we annotate the value
+    or the method
+- `[ThisEscapes]` `this` of a `struct` being ref-safe-to-escape (maybe this is a generalized `ref`
+- `[RefFieldEscapes]`: parameters can escape as a ref field
+    escape that could be applied to existing `ref` parameter)
+- `[DoesNotRefFieldEscape]` valuable once we have `ref` fields and we need to distinguish
+    which of the parameters is the captured ones 
+- `[DoesNotEscape]` used to indicate that a `ref struct` can't escape from a method 
+    and that helps infer lifetime issues
+
+types of lifetime annotations that are not needed
+- `[Escapes]` this is always true for non `ref struct` values and for a `ref struct`
+    it is the natural default. Can't think of a case where it would be needed
+- `[DoesNotInput]` today `out` is treated no differently than `ref` and that can cause
+    confusion in places. This could be fixed by an annotation where we don't consider
+    the input of a `out` in lifetime calculation. This is narrow though and don't 
+    see the need to do it.
+
+annotations need to be subject to verification. Consider marking a normal parameter
+as `[RefEscapes]` is non-sense and must be a compilation error. Likely need an allow
+list for when these can be added
+
+what is the default escape rules of `ref` fields?
+
+intersection of annotations and the `[InlineArray]` attribute. Maybe it implies and 
+/ or requires `[RefEscapes]`
+https://github.com/dotnet/runtime/issues/61135
+
+- convert cs to c# in the code blocks
+- ensure the language is consistent for *to the calling method* and *within the current method*
+- consider naming the compat problem
+- how do we mark the return of a `ref` field ctor as not returnable in the compat case. Maybe the attributes save us here because 
+
+***
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -37,6 +37,7 @@ readonly ref struct Span<T>
     // of changing Span<T> to have ref fields. It is a convenient, and
     // safe, way to create a length one span over a stack value that today 
     // requires unsafe code.
+    [RefFieldsEscape]
     public Span(ref T value)
     {
         _field = ref value;
@@ -130,7 +131,9 @@ Next the rules for method invocation will change as follows when the target meth
 - If the current method is not annotated with `[RefFieldsEscape]` then
     - If all arguments have a *ref-safe-to-escape* scope of *heap* then the *safe-to-escape* scope of the return is *calling method*
     - Else the *safe-to-escape* scope is *current method*
-- Else the *safe-to-escape* scope of the return is calculated normally
+- Else the *safe-to-escape* scope of the return is the minimum of 
+    - The existing *safe-to-escape* calculation 
+    - The minimum of the *ref-safe-to-escape* values of all `ref` and `in` arguments
 
 **TODO: think through out parameters**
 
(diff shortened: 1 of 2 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -2,46 +2,41 @@ Low Level Struct Improvements
...
 
 As these features have gained traction in the .NET ecosystem developers, both internal and external, have been providing us with information on remaining friction points in the ecosystem. Places where they still need to drop to `unsafe` code to get their work, or require the runtime to special case types like `Span<T>`. 
 
-This proposal aims to address many of these concerns by building on top of our existing low level features. Specifically it aims to:
+Today `Span<T>` is accomplished by using the `internal` type `ByReferenec<T>` which the runtime effectively treats as a `ref` field. This provides the benefit of `ref` fields but with the downside that the language provides no safety verification for it, as it does for other uses of `ref`. Further only dotnet/runtime can use this type as it's `internal`, so 3rd parties can not design their own primitives based on `ref` fields. Part of the [motivation for this work](https://github.com/dotnet/runtime/issues/32060) is to remove `ByReference<T>` and use proper `ref` fields in all code bases.
```
CyrusNajmabadi, line 12:
```suggestion
Today `Span<T>` is accomplished by using the `internal` type `ByReference<T>` which the runtime effectively treats as a `ref` field. This provides the benefit of `ref` fields but with the downside that the language provides no safety verification for it, as it does for other uses of `ref`. Further only dotnet/runtime can use this type as it's `internal`, so 3rd parties can not design their own primitives based on `ref` fields. Part of the [motivation for this work](https://github.com/dotnet/runtime/issues/32060) is to remove `ByReference<T>` and use proper `ref` fields in all code bases. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -2,46 +2,41 @@ Low Level Struct Improvements
...
 
-This proposal aims to address many of these concerns by building on top of our existing low level features. Specifically it aims to:
+Today `Span<T>` is accomplished by using the `internal` type `ByReferenec<T>` which the runtime effectively treats as a `ref` field. This provides the benefit of `ref` fields but with the downside that the language provides no safety verification for it, as it does for other uses of `ref`. Further only dotnet/runtime can use this type as it's `internal`, so 3rd parties can not design their own primitives based on `ref` fields. Part of the [motivation for this work](https://github.com/dotnet/runtime/issues/32060) is to remove `ByReference<T>` and use proper `ref` fields in all code bases. 
+
+This proposal plans to address these issues by building on top of our existing low level features. Specifically it aims to:
 
 - Allow `ref struct` types to declare `ref` fields.
-- Allow the runtime to fully define `Span<T>` using the C# type system and 
-remove special case type like `ByReference<T>`
+- Allow the runtime to fully define `Span<T>` using the C# type system and remove special case type like `ByReference<T>`
```
CyrusNajmabadi, line 17: are tehre any other speical types the runtime has that they could get rid of as well now? (just curious)

jaredpar: No other special types but it's likely that once we have these features we can essentially eliminate the concept of restricted types. They are expressable now in terms of a `ref struct` with a `ref` field.

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
+    // Okay
+    int local = 42;
+    return CreateSpan<int>(ref local);
 
-        // Legal today, must remain legal
-        T local = default;
-        return CreateSpan(ref local);
+    // Okay
+    Span<T> span = stackalloc int[1];
+    return CreateSpan<int>(ref span[0]);
```
CyrusNajmabadi, line 72: wait... why is this ok? :)   furthermore... if it was legal before... woudl it be bad to become illegal?

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
-        return CreateSpan(...);
-    }
+    // Okay
+    int[] array = new int[42];
+    return CreatSpan(ref array[0]);
 }
 ```
 
-The reason that all of the above samples are legal is because in the existing design there is no way for the return `Span<T>` to store a reference to the input state of the method call. This is because the span safety rules explicitly depend on `Span<T>` not having a constructor which takes a `ref` parameter and stores it as a field. 
+All of the above `return` statements are legal by our existing rules because the return of `CreateSpan` is always *safe-to-escape* to the *calling method*. The challenge in this proposal, which comes up many times, is they **must** remain legal in the version of the language which implements these features and / or when `Span<T>` moves to using `ref` fields. The above patterns can legally exist today and cannot become errors when moving to a new language version.
```
CyrusNajmabadi, line 80: > and cannot become errors when moving to a new language version. 

Would it be ok for them to become errors if there was true memory unsafety before?

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
-    {
-        // !!! Cannot happen today !!!
-        // The existing span safety rules specifically call out that this method
-        // cannot exist hence they can assume all returns from CreateSpan are
-        // safe to return.
-        return new Span<T>(ref p);
-    }
+    // Error: if this works it breaks the above compat requirements because the implementation
+    // is now capturing `ref` state.
+    return Span<T>(ref parameter);
```
CyrusNajmabadi, line 89: i'm not sure if this is a great example.  since the ref is captured, but only in somethign passed back to the caller, this is safe right (and i'm not sure how it breaks a compat requirement "if this works").

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
+    return new Span<int>(ref local);
 
-    Span<int> ConsumptionCompatibility()
-    {
-        // Okay: this is legal today and must remain legal.
-        int local = 42;
-        return CreateSpan(ref local);
+    // This must be an error because of the compat issues of the existing span safety 
+    // rules.
+    return new Span<int>(ref parameter);
```
CyrusNajmabadi, line 118: i'm not really getting this statement (i feel like it hasn't been explained well enough).  Why is it necessary for something safe to be an error.  compat issues tell me that prior working code should continue to work.  but it doesn't tell my why something must be an error.

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
+These rules ensure that the caller can understand when constructors will or will not capture parameters by `ref`. 
 
-Let's examine these rules in the context of samples to better understand their impact.
+Next the rules for method invocation will change as follows when the target method is annotated with `[RefFieldEscapes]`:
+- If the current method is not annotated with `[RefFieldEscapes]` then
+    - If all arguments have a *ref-safe-to-escape* scope of *heap* then the *safe-to-escape* scope of the return is to the *calling method*
+    - Else the *safe-to-escape* scope is to the *current method*
+- Else the *safe-to-escape* scope of the return is the minimum of 
+    - The existing *safe-to-escape* calculation for method invocation
+    - All of the *ref-safe-to-escape* values of `ref` and `in` arguments
```
CyrusNajmabadi, line 144: whooboy.  this last part went entirely over my head :)

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
 
-```cs
+The design of `[RefFieldEscape]` will be discussed in detail [later in the propsal](#reffieldescapes).
```
CyrusNajmabadi, line 146:
```suggestion
The design of `[RefFieldEscape]` will be discussed in detail [later in the proposal](#reffieldescapes).
```

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,226 @@ readonly ref struct Span<T>
...
     }
 }
 ```
 
-The limiting of the constructor rules to just `ref struct` that directly contain `ref` field is another important compatibility concern. Consider that the majority of `ref struct` defined today indirectly contain `Span<T>` references.  That mean by extension they will indirectly contain `ref` fields once `Span<T>` adopts `ref` fields. Hence it's important to ensure the *safe-to-return* rules of constructors on these types do not change. That is why the restrictions must only apply to types that directly contain a `ref` field.
+The samples here have the same patterns as the [compat considerations](#compat-considerations) above. This means it will allow the introduction of `ref` fields without breaking existing code.
 
-Example of where this comes into play.
+Constructor chaining needs to consider these new rules as well. When the original constructor calls a chained constructor via `:this(...)` the chained constructor effectively escapes from the original. That means a chained constructor call is olny legal if the *safe-to-escape* value is not smaller than the original constructor one. This will be accomplished with the following rules:
```
CyrusNajmabadi, line 239:
```suggestion
Constructor chaining needs to consider these new rules as well. When the original constructor calls a chained constructor via `:this(...)` the chained constructor effectively escapes from the original. That means a chained constructor call is only legal if the *safe-to-escape* value is not smaller than the original constructor one. This will be accomplished with the following rules:
```

proposals/low-level-struct-improvements.md
tannergooding: Warning wave maybe?

There is, on the BCL side, the interesting case of `MemoryMarshal.CreateSpan()` which functionally works like the proposed constructor (and via an internal `Span(ref T ptr, int length)` constructor):
```csharp
[RefFieldEscapes]
public Span(ref T value) { }
```

The BCL will need to decide how it works moving forward, including whether it starts erroring just like the proposed constructor, or continues allowing users to do "unsafe" things (potentially with an analyzer, etc).

jaredpar: > Would it be ok for them to become errors if there was true memory unsafety before?

This pattern is completely safe though. The only way this could be unsafe is if `Span<T>` had a constructor in the form of `Span(ref T value)` which is explicitly disallowed by the span safety rules. Yes `Span<T>` does have this CTOR but it's `internal` and known to be `unsafe`. 

So there is nothing wrong with these patterns. They are safe, legal and honestly expected. The issue is that we must maintain those properties going forward.

jaredpar: > Warning wave maybe?

How would a warning wave help though? Consider that by adding `[RefFieldEscapes]` to a method you change the semantics from the perspective of the caller. Or consider it this way:

```
Span<int> MyMethod() {
  int local = 42;
  Span<int> span = CreateSpan(ref local); 
  return span;
}
``` 

This is 100% legal today. If we suddenly warn customers into using `[RefFieldsEscape]` on such methods then it suddenly breaks callers of that method.

CyrusNajmabadi: Just so i understand with that example.  You're sayign that that code coudl exist and have legit uses (for example, not captuing 'local' within the returned span).  And as such, because of hte legal reasons, we couldn't jsut always start warning/error'ing here.

*however*, what we can do is have CreateSpan announce it's intentions about what happens with the passed in 'ref'.  If it ignored it (as an example) it would just do nothing, and the above would continue to be fine.  However, if it ends up pushing that parameter into the span it returns, then it shoudl have this attribute (virally spread hopefully from `Span<T>` inside it) so that hte caller can now error *in that case*.

--

In other words, the attribute allows cooperation between the caller and callee here on legal cases that should continue working, vs bad cases that ideally should break because they're entire memory unsafe.  Is that right?

tannergooding: > This is 100% legal today.

Right, and in the case where `CreateSpan` isn't updated to use some method/constructor marked `RefFieldsEscape` to create the returned span, it continues being "safe".
- If `CreateSpan` were updated to use `[RefFieldEscapes] public Span(ref T value) { }`, then I would assume the compiler errors as it expects `CreateSpan` to then be equivalently annotated, etc. This would be a breaking change for a library to make and therefore not recommended.

There are, however, scenarios like with `MemoryMarshal.CreateSpan` where its known to be "unsafe" and where the compiler (or an analyzer) could surface a warning that it functionally works like `[RefFieldEscapes] public Span(ref T value) { }` which would help catch problematic cases (particularly for users transitioning from "old unsafe logic" to "new safe logic").

You could also consider cases like `return new Span(&local, 1)` where its unsafe and legal, but its also a detectable problem and where a warning being surfaced is probably a "good idea" (maybe these are better suited to a library analyzer though)

proposals/low-level-struct-improvements.md
jaredpar: This was legal because there was no mechanism by which `CreateSpan(ref  int local)` could every capture and return the `ref`. Basically there is no way given the rules of C# and the safe APIs of dotnet/runtime to make the returned `Span<T>` refer to the passed in value. 

Yes you could do this using `MemoryMarshal` or `Unsafe` but those are APIs that are considered `unsafe`. They are essentially outside the bounds of span safety guidelines (and are documented as such)

proposals/low-level-struct-improvements.md
jaredpar: > Just so i understand with that example. You're sayign that that code coudl exist and have legit uses (for example, not captuing 'local' within the returned span). And as such, because of hte legal reasons, we couldn't jsut always start warning/error'ing here.

Exactly. It's perfectly okay to create such an API and give it legitamate uses. That exact pattern exists a few places in dotnet/runtime today (`SecureString` and some parts of `DateTime` parsing I was able to find). 

> In other words, the attribute allows cooperation between the caller and callee here on legal cases that should continue working, vs bad cases that ideally should break because they're entire memory unsafe. Is that right?

yep.

jaredpar: > There are, however, scenarios like with MemoryMarshal.CreateSpan where its known to be "unsafe" and where the compiler (or an analyzer) could surface a warning ... 

I'd actually vote in `net6.0` that we obsolete this API. Basically say "use `new Span<T>`" instead please. That would force the caller to use `unsafe` in the cases where actual `unsafe` behavior was still needed. 

I'm not sure how feasible that is though.

proposals/low-level-struct-improvements.md
jaredpar: Let me try expanding that sample out a bit. Essentially getting through the parts that will work (calling the ctor) and the parts that won't (returning it).

proposals/low-level-struct-improvements.md
jaredpar: I put a lot of the details into the earlier section but didn't have the full set of examples. I'm just going to expand the earlier section to include these, and link to that from here.

proposals/low-level-struct-improvements.md
jaredpar: That language takes some getting used to but it's how the original span safety doc was written. I want this doc to provide a map for how these features will fit into our existing spec. Essentially what changes do we need to make to extend our existing rules.

proposals/low-level-struct-improvements.md
CyrusNajmabadi: > yep.

Gracias!  you probably did write all that out in the PR.  but it helped to get that clarity here :)

proposals/low-level-struct-improvements.md
```diff
@@ -486,189 +480,120 @@ struct S1
...
-- The span safety rules for constructors, fields and assignment must be updated
-as outlined in this document.
-- The span safety rules need to include the definition of `ref` values that 
-"refer to the heap". 
+- The span safety rules for constructors, fields and assignment must be updated as outlined in this document.
+- The span safety rules need to include the definition of `ref` values that "refer to the heap". 
 
-### Provide struct this escape annotation
-The rules for the scope of `this` in a `struct` limit the *ref-safe-to-escape* scope to the current method. That means neither `this`, nor any of its fields can return by reference to the caller.
+### Provide lifetime annotations
```
davidwrighton, line 494: I wonder if we should have another annotation for extremely specialized cases where the method implementor wants to prohibit escape, but its not a capturing escape. For instance, the StackAlloc api I proposed in a hackathon a few months ago.
```
class RuntimeHelpers
{
    static Span<T> StackAlloc<T>(int count) {... }
}
```

Ideally there would be an annotation that allows specifying that the Span<T> returned from StackAlloc is not safe to escape.

OTOH, this may be so incredibly niche that we should just special case the function and move on.

jaredpar: I thought about this too. I think it would come down to how many of these functions we ended up creating. If it's only going ot be a small number, say < 5, and they're all effectively runtime intrinsics then possibly we just special case them. If it's more than that though then I'd favor adding an annotation for them.

proposals/low-level-struct-improvements.md
```diff
@@ -50,248 +45,233 @@ readonly ref struct Span<T>
...
+        // constructor is not annotated with [RefFieldEscapes] hence only known heap values are 
+        // allowed
+        _field = ref tuple.Item1;
     }
 
-    static RS RuleExamples(ref int i, int[] array)
+    static RS CreateRS(ref int parameter)
     {
-        var rs1 = new RS(ref i);
+        // Error: The RS(ref int) ctor is annotated as [RefFieldEscapes] however CreateRS is
```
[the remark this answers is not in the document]
jaredpar: Generics should not add any particular complication here. This problem only comes into effect when: 

- A `ref struct` with a `ref` field is returned from a method. Either by the method return or via a `out` / `ref` parameter
- There is an additional `ref`, `out` or `in` parameter of any type

Hence generics don't impact this much at all because `ref struct` can't be generic. Even if we add support for generic `ref struct` then it's an opt in scenario, like `where T : ref struct`, hence it should still be very limited. 

Note: in searching through the BCL today I've only found a very limited number of these methods and most are non-public. If you are aware of more I'd like to hear about them.

proposals/low-level-struct-improvements.md
```diff
@@ -486,189 +480,120 @@ struct S1
...
-```cs
-interface I1
-{
-    ref int Prop { get; }
-}
+This, and several other friction points, can be removed if the language provides developers a way to invert the defaults by applying attributes to specific locations. The language can recognize these attributes and simply adjust the lifetime calculation for locations when evaluating span safety.
 
-struct S1 : I1
+#### RefThisEscapes
+One of the most notable friction points in the inability to return fields by `ref` in instance members of a `struct`. This means developers can't create `ref` returning methods / properties and have to resort to exposing fields directly. This reduces the usefulness of `ref` returns in `struct` where it is often the most desired.
```
AaronRobinsonMSFT, line 504:
```suggestion
One of the most notable friction points is the inability to return fields by `ref` in instance members of a `struct`. This means developers can't create `ref` returning methods / properties and have to resort to exposing fields directly. This reduces the usefulness of `ref` returns in `struct` where it is often the most desired. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -486,189 +480,120 @@ struct S1
...
+- The method invocation rules will include the *ref-safe-to-escape* value of the receiver when it is marked by `[RefThisEscapes]`
+- The parameter lifetime rules will change to note that the *ref-safe-to-escape* of `this` is dependent on `[RefThisEscapes]`
 
-This also has the added benefit that it will make `fixed` buffers easier to consume from other languages. Named indexers is a feature that has existed since the 1.0 release of .NET. Even languages which cannot directly emit a named indexer can generally consume them (C# is actually a good example of this).
+Detailed Notes:
+- An instance method or property annotated with `[RefThisEscapes]` has *ref-safe-to-escape* of `this` set to the *calling method*
+- A `struct` annotated with `[RefThisEscapes]` has the same effect of annotating every instance method and property with `[RefThisEscapes]`
```
AaronRobinsonMSFT, line 571: I think it would be helpful for an example that demonstrates why one may want to annotate a single method/property as opposed to the entire type.

proposals/low-level-struct-improvements.md
```diff
@@ -729,15 +654,15 @@ In order to work around this low level code will resort to `unsafe` tricks to li
...
 void WriteData(ReadOnlySpan<char> data)
 {
     if (data.Contains(':'))
     {
         Span<char> buffer = stackalloc char[256];
         Escape(data, buffer, out var length);
 
-        // Error: Cannot assign `buffer` to `data` here as the *safe-to-escape*
+        // Error: Cannot assign `buffer` to `data` here as the safe-to-escape
         // scope of `buffer` is to the current method scope while `buffer` is
```
AaronRobinsonMSFT, line 666:
```suggestion
        // scope of `buffer` is to the current method scope while `data` is
```

proposals/low-level-struct-improvements.md
```diff
@@ -749,23 +674,23 @@ void WriteData(ReadOnlySpan<char> data)
...
+        // Error: the [DoesNotEscape] attribute changes the safe-to-escape*
         // to be limited to the current method scope. Hence it cannot be 
         // returned
         return p2; 
 
-        // ERROR: `local` has the same *safe-to-escape* as `p2` hence it cannot
+        // Error: `local` has the same safe-to-escape as `p2` hence it cannot
         // be returned.
         Span<int> local = p2;
         return p2;
```
AaronRobinsonMSFT, line 696:
```suggestion
        return local; 
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,7 +715,7 @@ ref struct S
         // of the current method and hence the same as the receiver.
         _field = p1;
 
-        // ERROR: the [DoesNotEscape] attribute changes the *safe-to-escape* 
+        // Error: the [DoesNotEscape] attribute changes the *safe-to-escape* 
         // to be limited to the current method scope. Hence it cannot be 
         // assigned to a receiver than has a *safe-to-escape* scope outside the
```
AaronRobinsonMSFT, line 720:
```suggestion
        // assigned to a receiver that has a *safe-to-escape* scope outside the 
```

proposals/low-level-struct-improvements.md
```diff
@@ -805,25 +730,119 @@ For example the last line of calculating *safe-to-escape* of returns will change
 
 > the safe-to-escape of all argument expressions including the receiver. **This will exclude all arguments that line up with parameters marked as [DoesNotEscape]**
 
-Misc Notes:
-- The  `DoesNotEscapeAttribute` will be defined in the `System.Runtime.CompilerServices` namespace.
-- The `DoesNotEscapeAttribute` cannot be combined with the `[ThisRefEscapes]` attribute, doing so results in an error.
-- The `DoesNotEscapeAttribute` will be emitted as a `modreq`
+Detailed Notes:
+- The `[DoesNotEscape]` and `[RefThisEscapes]` cannot be combined on the same method
```
AaronRobinsonMSFT, line 734: Isn't `DoesNotEscape` the default for member functions of a `ref struct`? Trying to confirm my understanding here. If so, can we should have an example where one would want to have `DoesNotEscape` applied to a member function.

proposals/low-level-struct-improvements.md
```diff
@@ -805,25 +730,119 @@ For example the last line of calculating *safe-to-escape* of returns will change
...
+{
+    internal char Data[128];
+}
+```
+
+These declarations, much like their `unsafe` counter parts, will define a sequence of `N` elements in the containing type. These members can be accessed with an indexer and can also be converted to `Span<T>` and `ReadOnlySpan<T>` instances.
+
+When indexing into a `fixed` buffer of type `T` the `readonly` state of the container must be taken into account.  If the container is `readonly` then the indexer returns `ref readonly T` else it returns `ref T`. 
+
+Accessing a `fixed` buffer without an indexer has no natural type however it is convertible to `Span<T>` types. In the case the container is `readonly` the buffer is implicitly convertible to `ReadOnlySpan<T>`, else it can implicitly convert to `Span<T>` or `ReadOnlySpan<T>` (the `Span<T>` conversion is considered *better*).
```
AaronRobinsonMSFT, line 750: > the `Span<T>` conversion is considered *better*

Why is this true?

proposals/low-level-struct-improvements.md
```diff
@@ -805,25 +730,119 @@ For example the last line of calculating *safe-to-escape* of returns will change
...
+
+The resulting `Span<T>` instance will have a length equal to the size declared on the `fixed` buffer. The *safe-to-escape* scope of the returned value will be equal to the *safe-to-escape* scope of the container, just as it would if the backing data was accessed as a field.
+
+For each `fixed` declaration in a type where the element type is `T` the language will generate a corresponding `get` only indexer method whose return type is `ref T`. The indexer will be annotated with the `[RefThisEscapes]` attribute as the implementation will be returning fields of the declaring type. The accessibility of the member will match the accessibility on the `fixed` field.
+
+For example, the signature of the indexer for `CharBuffer.Data` will be the following:
+
+```c#
+[RefThisEscapes]
+internal ref char <>DataIndexer(int index) => ...;
```
AaronRobinsonMSFT, line 760: Is the `<>` correct in the above signature?

proposals/low-level-struct-improvements.md
```diff
@@ -805,25 +730,119 @@ For example the last line of calculating *safe-to-escape* of returns will change
...
+
+There will also be a named accessor generated for each `fixed` buffer that provides by value `get` and `set` operations. Having this means that `fixed` buffers will more closely resemble existing array semantics by having a `ref` accessor as well as byval `get` and `set` operations. This means compilers will have the same flexibility when emitting code consuming `fixed` buffers as they do when consuming arrays. This should be operations like `await` over `fixed` buffers easier to emit. 
+
+This also has the added benefit that it will make `fixed` buffers easier to consume from other languages. Named indexers is a feature that has existed since the 1.0 release of .NET. Even languages which cannot directly emit a named indexer can generally consume them (C# is actually a good example of this).
+
+There will also be a by value `get` and `set` accessor generated for every
```
AaronRobinsonMSFT, line 769: Incomplete thought?

proposals/low-level-struct-improvements.md
```diff
@@ -805,25 +730,119 @@ For example the last line of calculating *safe-to-escape* of returns will change
...
+
+This also has the added benefit that it will make `fixed` buffers easier to consume from other languages. Named indexers is a feature that has existed since the 1.0 release of .NET. Even languages which cannot directly emit a named indexer can generally consume them (C# is actually a good example of this).
+
+There will also be a by value `get` and `set` accessor generated for every 
+
+The backing storage for the buffer will be generated using the `[InlineArray]` attribute. This is a mechanism discussed in [isuse 12320](https://github.com/dotnet/runtime/issues/12320) which allows specifically for the case of efficiently declaring sequence of fields of the same type. This particular issue is still under active discussion and the expectation is that the implementation of this feature will follow however that discussion goes.
```
AaronRobinsonMSFT, line 771:
```suggestion
The backing storage for the buffer will be generated using the `[InlineArray]` attribute. This is a mechanism discussed in [issue 12320](https://github.com/dotnet/runtime/issues/12320) which allows specifically for the case of efficiently declaring sequence of fields of the same type. This particular issue is still under active discussion and the expectation is that the implementation of this feature will follow however that discussion goes.
```

proposals/low-level-struct-improvements.md
```diff
@@ -805,25 +730,119 @@ For example the last line of calculating *safe-to-escape* of returns will change
...
+
 ## Considerations
 
 ### Keywords vs. attributes
-This design calls for using attributes to annotate the new lifetime rules for `struct` members. This also could've been done just as easily with contextual keywords. For instance: `scoped` and `escapes` could have been used instead of `DoesNotEscape` and `ThisRefEscapes`.
+This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance: `scoped` and `escapes` could have been used instead of `DoesNotEscape` and `RefThisEscapes`.
 
 Keywords, even the contextual ones, have a much heavier weight in the language than attributes. The use cases these features solve, while very valuable, impact a small number of developers. Consider that only a fraction of high end developers are defining `ref struct` instances and then consider that only a fraction of those developers will be using these new lifetime features.  That doesn't seem to justify adding a new contextual keyword to the language.
```
AaronRobinsonMSFT, line 778: Attributes though do seem to escape the language. In this case, will these attributes be meaningful in other .NET languages like VB.NET or F#? If they are applied will they be nops or having some compiler indicate they are unnecessary?

proposals/low-level-struct-improvements.md
tannergooding: "Anonymous structs/unions" used in interop.

There are many cases where `FieldOffset` is not viable because some other member in the struct is variable sized (`nint`, `nuint`, a `pointer`, etc). In those cases, the only solution is to have something like:
```csharp
public unsafe partial struct MyStruct
{
    public _Anonymous_e__Union Anonymous;
    
    public unsafe partial struct _Anonymous_e__Union
    {
        public float m11;
    }
} 
```

In such a scenario, there are a lot of "usability" issues with forcing users to do `value.Anonymous.m11` rather than just allowing `value.m11`. The "simplest" solution here is to expose a `ref property`:
```csharp
public unsafe partial struct MyStruct
{
    public _Anonymous_e__Union Anonymous;
    
    public ref float m11 => ref MemoryMarshal.GetReference(MemoryMarshal.CreateSpan(ref Anonymous.m11, 1));
    
    public unsafe partial struct _Anonymous_e__Union
    {
        public float m11;
    }
} 
```

However, this is unsafe and doesn't properly track the lifetime of the `ref`. With the introduction of `RefThisEscapes`, this becomes:
```csharp
public unsafe partial struct MyStruct
{
    public _Anonymous_e__Union Anonymous;
    
    [RefThisEscapes]
    public ref float m11 => ref Anonymous.m11;
    
    public unsafe partial struct _Anonymous_e__Union
    {
        public float m11;
    }
} 
```

You wouldn't want to annotate the entire type, because things like `operator ==(in MyStruct left, in MyStruct right)` take a ref, but don't return a reference to the `this` field.

proposals/low-level-struct-improvements.md
tannergooding: I would assume because the data is mutable so `Span<T>` is a better fit.

Plus, if you get a `Span<T>`, you can explicitly convert to `ROSpan<T>`. You cannot trivially do the inverse (and the inverse is generally considered "unsafe")

proposals/low-level-struct-improvements.md
AaronRobinsonMSFT: Thanks. Great example.

proposals/low-level-struct-improvements.md
jaredpar: For a instance member of a `struct` of any kind the `this` parameter is *safe-to-escape* to the calling method but *ref-safe-to-escape* within the current method. The `[DoesNotEscape]` annotation moves both to current method. 

An example where you would need it is the following: 

```c# 
ref struct MyType { 
  Span<int> _span;
  public MyType(Span<int> span) { }
  public void Method(ref ValueStringBuilder builder) { ... }
}
```

This method falls squarely into the [method arguments must match](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md#method-arguments-must-match) problem because both `this` and `builder` are *safe-to-escape* to the calling method. That means from the point of view of the compiler `builder` could end up smuggling out `ref` state from `this`. Yes realistically this can't happen do to no common data but compiler doesn't do field level analysis, it's just looking at the escape scopes of the locations.

This mean though practically that you run into problems like this: 

``` 
void Example(ref ValueStringBuilder builder) {
  var myType = new MyType(stackalloc int[42]); 
  
 // Error: builder and myType have different lifetimes
  myType.Method(ref builder); 
``` 

This can be fixed though by noting that `this` does not escape back into the parameters 

``` 
[DoesNotEscape]
public void Method(ref ValueStringBuilder builder) { ... }
```

proposals/low-level-struct-improvements.md
jaredpar: Indeed that is the reason.

proposals/low-level-struct-improvements.md
jaredpar: Yes it's meant to indicate that it's a compiler generated member that can't be directly accessed by consumer code. The `<>` is part of all synthesized non-acecssible members we generate (usually has a bit of text in between)

proposals/low-level-struct-improvements.md
jaredpar: The thought was complete, the text was lacking 😄

proposals/low-level-struct-improvements.md
jaredpar: By default these attributes won't be meaningful to other languages. But that's true today. Consider that if they put the `[Dynamic]` attribute on method parameters it has no meaning in F# but will when it's invoked from C#. 

There is a growing list of APIs for which they seem to only have meaning in C#. I'm not sure how or if we should dela with that.

AaronRobinsonMSFT: > I'm not sure how or if we should dela with that.

Okay. I'll think this does come with some concerns though. From the language it may not matter, but how the runtime may or may not respond to that attribute in the future seems to me to be a potential concern. In fact I could see a strong argument for having tests in those languages to ensure the runtime doesn't do things the generated IL doesn't expect. I don't have a concrete example but I could imagine method level attributes that let the runtime inform the JIT about some lifetime assumption that isn't there in a non-C# language and that ends up causing issues.

I'm in the middle of a GC Stress bug and it is consuming a fair amount of mental resources so it could be simple top of mind and I see potential issues everywhere.

proposals/low-level-struct-improvements.md
tannergooding: @jaredpar, I think the "issue" here is that the code is confusing to read.

If you were to think about this from the perspective of the framework design guidelines and how the BCL names things; the "logical assumption" of `CreateSpan<int>(ref T)` is that it creates a `span` over the `ref` input and so the readers initial "assumption" is that `CreateSpan<int>(stackalloc)` is returning a span that captures the stack allocation.

The language prevents this today (outside of unsafe code), but its still not how the code reads. This is somewhat made more "confusing" by the fact that part of this doc is meant to enable scenarios where `CreateSpan(ref T)` capturing the input is possible.

Perhaps it would be more clear if this was `CreateSpanThatDoesntCaptureTheInput` (e.g. a method without `RefThisEscapes`) and `CreateSpanThatCapturesTheInput` (e.g. a method with `RefThisEscapes`)?
-- or better names, because names are hard and my suggestions are terrible 😄

jaredpar: The goal of this section though is to really drive home the point that 

1. This method creates a `Span<T>` 
2. It can never, under any circumstances, capture that input value

If I name it say `CreateSpanThatDoesntCaptureTheInput` it might imply that this particular method does not capture the input but maybe some other method could. Where the problem is there is just no way this can happen, no matter the name or implementatoin.

proposals/low-level-struct-improvements.md
```diff
@@ -583,7 +583,7 @@ This attribute can be applied to methods, constructors and operators. Applying t
 The semantics of this attribute, and how it impacts span safety rules, are described [above](#ref-fields-escape)
 
 #### DoesNotEscape
-One source of repeated friction in low level code is the default escape for parameters is permissive. They are *safe-to-escape* to the *calling method* and in the case of `ref` parameters also *ref-safe-to-escape* to the *calling method*. These are sensible defaults because it lines up with the coding patterns of .NET as a whole. In low level code though there is a larger use of `ref` and `ref struct` and these defaults can cause friction with other parts of the span safet rules.
+One source of repeated friction in low level code is the default escape for parameters is permissive. They are *safe-to-escape* to the *calling method*. This is a sensible defaults because it lines up with the coding patterns of .NET as a whole. In low level code though there is a larger use of  and `ref struct` and this default can cause friction with other parts of the span safety rules.
```
cston, line 586:
```suggestion
One source of repeated friction in low level code is the default escape for parameters is permissive. They are *safe-to-escape* to the *calling method*. This is a sensible default because it lines up with the coding patterns of .NET as a whole. In low level code though there is a larger use of  `ref struct` and this default can cause friction with other parts of the span safety rules.
```

proposals/low-level-struct-improvements.md
```diff
@@ -698,9 +698,11 @@ class C
 }
 ```
 
+This attribute cannot be used on `ref` or `out` parameters. Those always implicitly escape to the calling method (that is how `ref` works). Instead `in` is likely a more appropriate designation for those scenarios.
+
 To account for this change the "Parameters" section of the span safety document will be updated to include the following bullet:
 
-- If the parameter is marked with `[DoesNotEscape]`it is *safe-to-escape* and *ref-safe-to-escape* are to the *current method*
+- If the parameter is marked with `[DoesNotEscape]`it is *safe-to-escape* is to the *current method*
```
cston, line 705:
```suggestion
- If the parameter is marked with `[DoesNotEscape]` it is *safe-to-escape* to the *current method*
```

proposals/low-level-struct-improvements.md
```diff
@@ -724,14 +726,15 @@ ref struct S
...
+Given that parameters are restricted in this way we will also update the "Method Invocation" section to relax its rules. In all cases where it is considering the *safe-to-escape* lifetimes of arguments the spec will change to ignore those arguments which line up to parameters which are marked as `[DoesNotEscape]`. Because these arguments cannot escape their lifetime does not need to be considered when considering the lifetime of returned values.
 
 For example the last line of calculating *safe-to-escape* of returns will change to 
 
 > the safe-to-escape of all argument expressions including the receiver. **This will exclude all arguments that line up with parameters marked as [DoesNotEscape]**
 
 Detailed Notes:
 - The `[DoesNotEscape]` and `[RefThisEscapes]` cannot be combined on the same method 
+- The `[DoesNotEsacpe]` attribute cannot be used on parameters that are `ref`, `out` or `in`.
```
cston, line 737:
```suggestion
- The `[DoesNotEscape]` attribute cannot be used on parameters that are `ref`, `out` or `in`.
```

proposals/low-level-struct-improvements.md
steveharter: FWIW an existing special\restricted types being discussed for supporting https://github.com/dotnet/runtime/issues/45152 (`ref struct` + reflection with related "fast invoke" feature) is `TypedReference` which contains a loosely-typed `ByReference` expressed as a [`ByReference<byte>`](https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/TypedReference.CoreCLR.cs#L16).

In order to support https://github.com/dotnet/runtime/issues/45152, a `static TypedReference FromRef<T>(ref T value)` [factory method](https://github.com/dotnet/runtime/compare/main...steveharter:FastInvoke2#diff-df560ce5bcf241f546b22cd124819fcf314e42ec480aaaf0c66fd653ce92e38cR72) is proposed which will create a `TypedReference` that references a `ref struct`. Today calling such a factory method is not possible with a `ref struct` since a `ref struct` type can't be used as a generic parameter, however that is a different issue; also note the method may be called "FromRefStruct" if necessary to distinguish between a `ref struct` and non-`ref-structs` so a new constraint like `where T : ref struct` can be added if necessary.

proposals/low-level-struct-improvements.md
tannergooding: > If I name it say CreateSpanThatDoesntCaptureTheInput it might imply that this particular method does not capture the input but maybe some other method could. Where the problem is there is just no way this can happen, no matter the name or implementatoin.

I thought that was the point of `RefThisEscapes`? `RefThisEscapes` means that you can have a span that captures the `ref input`. The language would then block it in the case of `stackalloc` because its unsafe.

That is, there are at least four basic scenarios to consider here, three of which are legal:
```csharp
// this is legal as `createdSpan` can't capture `ref span[0]` as it isn't attributed with `RefThisEscapes`
Span<T> stackSpan = stackalloc int[1];
var createdSpan = CreateSpanWithoutRefThisEscapes<int>(ref span[0]);

// this is legal as `createdSpan` can't capture `ref span[0]` as it isn't attributed with `RefThisEscapes`
Span<T> stackSpan = stackalloc int[1];
return CreateSpanWithoutRefThisEscapes<int>(ref span[0]);

// this is legal as even though `createdSpan` can capture `ref span[0]` since it is attributed with `RefThisEscapes`, it isn't escaping the scope of `stackSpan`
Span<T> stackSpan = stackalloc int[1];
var createdSpan = CreateSpanWithRefThisEscapes<int>(ref span[0]);

// this is ILLEGAL as `createdSpan` can capture `ref span[0]` and if it did, it would allow `stackSpan` to escape its scope
Span<T> stackSpan = stackalloc int[1];
return CreateSpanWithRefThisEscapes<int>(ref span[0]);
```

proposals/low-level-struct-improvements.md
jaredpar: > Today calling such a factory method is not possible with a ref struct since a ref struct type can't be used as a generic parameter, however that is a different issue; 

There is going to be a proposal soon-ish about how to make `ref struct` work with generic type arguments. I don't think we fully need that though to enable the `TypedReference` support. I think as a stop gap measure we could always just adjust the intrinsic `__makeref` to handle `ref struct`.

proposals/low-level-struct-improvements.md
jaredpar: > RefThisEscapes means that you can have a span that captures the ref input

No. The only thing that `RefThisEscapes` alters is that it makes it possible to return `this` by-ref from instance methods on `struct`. It does not change your ability to capture a `ref` as a field.

proposals/low-level-struct-improvements.md
```diff
@@ -2,46 +2,41 @@ Low Level Struct Improvements
...
 
 ## Motivation
-features to the language: `ref` returns, `ref struct`, function pointers, etc. ... These enabled .NET developers to create write highly performant code while continuing to leverage the C# language rules for type and memory safety.  It also allowed the creation of fundamental performance types in the .NET libraries like `Span<T>`.
+Earlier versions of C# added a number of low level performance features to the language: `ref` returns, `ref struct`, function pointers, etc. ... These enabled .NET developers to create write highly performant code while continuing to leverage the C# language rules for type and memory safety.  It also allowed the creation of fundamental performance types in the .NET libraries like `Span<T>`.
 
 As these features have gained traction in the .NET ecosystem developers, both internal and external, have been providing us with information on remaining friction points in the ecosystem. Places where they still need to drop to `unsafe` code to get their work, or require the runtime to special case types like `Span<T>`.
```
cston, line 10:
```suggestion
As these features have gained traction in the .NET ecosystem developers, both internal and external, have been providing us with information on remaining friction points in the ecosystem. Places where they still need to drop to `unsafe` code to get their work done, or require the runtime to special case types like `Span<T>`. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,133 @@ ref struct S
...
+```
+
+The design explicitly wants to ensure this is legal. Yet the *ref-safe-to-escape* scope of `local` is *current method* and it can't be anything greater. At the same time it must be assignable to a `ref` field. Further we need to ensure that `span` is itself assignable to a `ref` field of a different type and it also has *current method* as the *ref-safe-to-escape* scope.
+
+The only practical way to represent whether a location can assign to a `ref` field is as a separate concept from escape scopes.
+
 ## Open Issues
 
+### Take the breaking change
+Consider for a minute a design where the compat problem was approached from the other direction. Effectively make it such that every method was implicitly `[RefFieldsEscape]` and then have an attribute that restores the span safety rules in place today. Say `[RefFieldDoesNotEsacpe]`.
```
cston, line 811:
```suggestion
Consider for a minute a design where the compat problem was approached from the other direction. Effectively make it such that every method was implicitly `[RefFieldsEscape]` and then have an attribute that restores the span safety rules in place today. Say `[RefFieldDoesNotEscape]`. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -486,189 +480,120 @@ struct S1
...
-
-To remove this friction the language will provide the attribute `[ThisRefEscapes]`.  When this attribute is applied to an instance method, instance property or instance accessor of a `struct` or `ref struct` then the `this` parameter will be considered *ref-safe-to-escape* outside the enclosing method.
-
-This allows for greater flexibility in `struct` definitions as they can begin returning `ref` to their fields. That allows for types like `FrugalList<T>`:
-
-```cs
-struct FrugalList<T>
+// Option 2 
+// The ref-safe-to-escape for this is now the calling method on all members
+[ThisRefEcapes]
```
cston, line 533:
```suggestion
[ThisRefEscapes]
```

proposals/low-level-struct-improvements.md
```diff
@@ -394,17 +388,17 @@ ref struct RS
...
 
 - Constructors: The value must be *ref-safe-to-escape* outside the constructor
 - `init` accessors:  The value limited to values that are known to refer to the heap as accessors can't have `ref` parameters
 - object initializers: The value can have any *ref-safe-to-escape* value as this will feed into the calculation of the *safe-to-escape* of the constructed object by existing rules.
 
-A `ref` field can only be assigned outside a constructor when the value is known to refer to the heap. That is allowed because it is both safe at the assignment location (meets the field assignment rules for ensuring the value being assigned has a lifetime at least as large as the receiver) as well as requires no updates to the existing method invocation rules. 
+A `ref` field can only be `ref` assigned outside a constructor when the value is known to refer to the heap. That is allowed because it is both safe at the assignment location (meets the field assignment rules for ensuring the value being assigned has a lifetime at least as large as the receiver) as well as requires no updates to the existing method invocation rules.
```
cston, line 397: >when the value is known to refer to the heap [](http://example.com/codeflow?start=63&length=44)

What if the lifetime of the value is known to encompass the lifetime of the `ref struct` containing the field?

proposals/low-level-struct-improvements.md
```diff
@@ -486,189 +480,120 @@ struct S1
...
+#### RefThisEscapes
+One of the most notable friction points is the inability to return fields by `ref` in instance members of a `struct`. This means developers can't create `ref` returning methods / properties and have to resort to exposing fields directly. This reduces the usefulness of `ref` returns in `struct` where it is often the most desired. 
 
-struct S1 : I1
+```c#
+struct S
 {
-    int _field;
-    public ref int Prop => _ref field;
+     _field;
```
cston, line 509:
```suggestion
     int _field;
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,133 @@ ref struct S
...
+- Special casing methods that return `ref struct` that define a `ref`. There are no such methods today because `ref` fields do not exist hence we could say that methods which return `ref struct` that defined a `ref` field have opted-in to `[ReFFieldsEscape]` semantics. This works but it essentially prevents any existing `ref struct` from adding `ref` fields. Doing so would cause span safety rules to be interpreted differently in all methods that returned the type. 
+
+These implicit opt-in strategies all have significant holes while an explicit opt-in is fully generalizable and makes the span safety rule different explicit in the code.
+
+### Why isn't assignability to ref field a scope?
+It may seem attractive to think of the ability to assign to a `ref` field as just another scope concept instead of an orthogonal concept. Essentially *assignable to field* as an escape scope between *heap* and *calling method*. This falls a part though when looking at practical examples. Consider for example the following:
```
cston, line 794:
```suggestion
It may seem attractive to think of the ability to assign to a `ref` field as just another scope concept instead of an orthogonal concept. Essentially *assignable to field* as an escape scope between *heap* and *calling method*. This falls apart though when looking at practical examples. Consider for example the following: 
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,133 @@ ref struct S
...
+
+### Why isn't assignability to ref field a scope?
+It may seem attractive to think of the ability to assign to a `ref` field as just another scope concept instead of an orthogonal concept. Essentially *assignable to field* as an escape scope between *heap* and *calling method*. This falls a part though when looking at practical examples. Consider for example the following: 
+
+```c#
+void M(ref int parameter)
+{
+    int local = 0;
+    Span<T> span = new(ref local);
+}
```
cston, line 801: Assuming the constructor has `[RefFieldsEscape]`, doesn't this fit the escape scope model since `local` and `span` both have the same scope?

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,133 @@ ref struct S
...
+- Special casing methods that return `ref struct` that define a `ref`. There are no such methods today because `ref` fields do not exist hence we could say that methods which return `ref struct` that defined a `ref` field have opted-in to `[ReFFieldsEscape]` semantics. This works but it essentially prevents any existing `ref struct` from adding `ref` fields. Doing so would cause span safety rules to be interpreted differently in all methods that returned the type. 
+
+These implicit opt-in strategies all have significant holes while an explicit opt-in is fully generalizable and makes the span safety rule different explicit in the code.
+
+### Why isn't assignability to ref field a scope?
+It may seem attractive to think of the ability to assign to a `ref` field as just another scope concept instead of an orthogonal concept. Essentially *assignable to field* as an escape scope between *heap* and *calling method*. This falls a part though when looking at practical examples. Consider for example the following: 
+
+```c#
+void M(ref int parameter)
```
cston, line 797: >int parameter [](http://example.com/codeflow?start=11&length=13)

Should `parameter` be used below?

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,133 @@ ref struct S
...
+The design explicitly wants to ensure this is legal. Yet the *ref-safe-to-escape* scope of `local` is *current method* and it can't be anything greater. At the same time it must be assignable to a `ref` field. Further we need to ensure that `span` is itself assignable to a `ref` field of a different type and it also has *current method* as the *ref-safe-to-escape* scope.
+
+The only practical way to represent whether a location can assign to a `ref` field is as a separate concept from escape scopes.
+
 ## Open Issues
 
+### Take the breaking change
+Consider for a minute a design where the compat problem was approached from the other direction. Effectively make it such that every method was implicitly `[RefFieldsEscape]` and then have an attribute that restores the span safety rules in place today. Say `[RefFieldDoesNotEsacpe]`. 
+
+This would be a breaking change and it's easy to construct code samples that trigger an error when upgrading to a new version of C# (as demonstrated [here](#new-span-challenges)). It's hard to determine though how prevelant these types of patterns are. Essentially methods which have both the following attributes:
```
cston, line 813:
```suggestion
This would be a breaking change and it's easy to construct code samples that trigger an error when upgrading to a new version of C# (as demonstrated [here](#new-span-challenges)). It's hard to determine though how prevalant these types of patterns are. Essentially methods which have both the following attributes:
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,133 @@ ref struct S
...
+{
+    ...
+    return new Span<int>(new int[i]);
+}
+
+// App.exe
+Span<int> Method()
+{
+    int local = 42;
+    var span = new Span<int>(ref local);
```
cston, line 836:
```suggestion
    var span = CreateSpan(ref local);
```

cston:
>         Span<int> span = stakalloc int[42];

`stackalloc`

---
Refers to: proposals/low-level-struct-improvements.md:851 in d10b06d. [](commit_id = d10b06d1d90adc81714fe433a099c47a7bb65e39, deletion_comment = False)

proposals/low-level-struct-improvements.md
tannergooding: Maybe I'm getting the attributes mixed up then?

This allows ref fields, which means that you can functionally `new Span(ref param)` and the created `Span` will have the scope of the passed input.

So if the `CreateSpan` call uses that method and returns the span, its lifetime is the same as the input and the compiler does the right analysis.

proposals/low-level-struct-improvements.md
```diff
@@ -790,6 +790,31 @@ Several ideas for having implicit opt-in to `ref` capture were explored and disc
...
+
+```c#
+// Impl assembly 
+ref struct S<T> {
+    ref T _field;
+}
+
+// Ref assembly 
+ref struct S<T> {
+    object _o; // force unmanaged
```
tannergooding, line 813: Was this meant to be either `force not unmanaged` or `force managed`?

proposals/low-level-struct-improvements.md
jaredpar: True it could be expressed that way. A `ref struct` really only has two lifetime states: usable within the current method, returnable to the containing method. If we make it such that the field lifetime has to be at least the containing `ref struct` then that would also allow for a bit more flexibility. 

I need to think about this a bit though and make sure I'm not missing a subtle rule here.

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,158 @@ ref struct S
...
+
+These implicit opt-in strategies all have significant holes while an explicit opt-in is fully generalizable and makes the span safety rule different explicit in the code.
+
+### Reference Assemblies
+A reference assembly for a compilation using features described in this proposal must maintain the elements that convey span safety information. That means all lifetime annotation attributes and `[RefFieldEscapes]` must be preserved in their original position. Any attempt to replace or omit them can lead to invalid reference assemblies.
+
+Representing `ref` fields is more nuanaced. Ideally a `ref` field would appear in a reference assembly as would any other field. However a `ref` field represents a change to the metadata format and that can cause issues with tool chains that are not updated to understand this metadata change. A concrete example is C++/CLI which will likely error if it consumes a `ref` field. Hence it's advantageous if `ref` fields can be omitted from reference assemblies in our core libraries.
```
cston, line 796:
```suggestion
Representing `ref` fields is more nuanced. Ideally a `ref` field would appear in a reference assembly as would any other field. However a `ref` field represents a change to the metadata format and that can cause issues with tool chains that are not updated to understand this metadata change. A concrete example is C++/CLI which will likely error if it consumes a `ref` field. Hence it's advantageous if `ref` fields can be omitted from reference assemblies in our core libraries. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,158 @@ ref struct S
...
+
+### Reference Assemblies
+A reference assembly for a compilation using features described in this proposal must maintain the elements that convey span safety information. That means all lifetime annotation attributes and `[RefFieldEscapes]` must be preserved in their original position. Any attempt to replace or omit them can lead to invalid reference assemblies.
+
+Representing `ref` fields is more nuanaced. Ideally a `ref` field would appear in a reference assembly as would any other field. However a `ref` field represents a change to the metadata format and that can cause issues with tool chains that are not updated to understand this metadata change. A concrete example is C++/CLI which will likely error if it consumes a `ref` field. Hence it's advantageous if `ref` fields can be omitted from reference assemblies in our core libraries. 
+
+A `ref` field by itself has no impact on span safety rules. As a conecrete example consider that flipping the existing `Span<T>` defintion to use a `ref` field has no impact on consumption. Hence the `ref` itself can be omitted safely. However a `ref` field does have other impacts to consumption that must be preserved:
```
cston, line 798:
```suggestion
A `ref` field by itself has no impact on span safety rules. As a concrete example consider that flipping the existing `Span<T>` definition to use a `ref` field has no impact on consumption. Hence the `ref` itself can be omitted safely. However a `ref` field does have other impacts to consumption that must be preserved: 
```

proposals/low-level-struct-improvements.md
```diff
@@ -790,40 +717,158 @@ ref struct S
...
+
+```c#
+// Impl assembly 
+ref struct S<T> {
+    ref T _field;
+}
+
+// Ref assembly 
+ref struct S<T> {
+    object _o; // force unmanaged
```
cston, line 813: Doesn't the additional field change the size of the `ref struct` for the consumer?

proposals/low-level-struct-improvements.md
jaredpar: yep that is a typo

proposals/low-level-struct-improvements.md
jaredpar: Generally size is calculated at runtime based on the impl assembly, not the reference assembly. It's generally considered acceptable for reference assemblies to add / remove fields. Been a common practice in GenAPI for a number of years.

proposals/low-level-struct-improvements.md
jaredpar: It just shouldn't be there at all

proposals/low-level-struct-improvements.md
jaredpar: Correct. If `[RefFieldEscapes]` is on the ctor then both `local` and `span` have the same *safe-to-escape* scope

proposals/low-level-struct-improvements.md
```diff
@@ -394,7 +394,10 @@ The rules for `ref` assignment also need to be adjusted to account for `ref` fie
 - `init` accessors:  The value limited to values that are known to refer to the heap as accessors can't have `ref` parameters
 - object initializers: The value can have any *ref-safe-to-escape* value as this will feed into the calculation of the *safe-to-escape* of the constructed object by existing rules.
 
-A `ref` field can only be `ref` assigned outside a constructor when the value is known to refer to the heap. That is allowed because it is both safe at the assignment location (meets the field assignment rules for ensuring the value being assigned has a lifetime at least as large as the receiver) as well as requires no updates to the existing method invocation rules. 
+A `ref` field can only be `ref` assigned outside a constructor when the value is known to have a lifetime greater than the reciever. Specifically:
```
cston, line 397: Perhaps update to allow a lifetime that is the same span:
```suggestion
A `ref` field can only be `ref` assigned outside a constructor when the value is known to have a lifetime greater than or equal to the receiver. Specifically: 
```

jaredpar:
Thanks everyone for their feedback on the proposal. I'm going to proceed with the merge as more detailed planning around this feature is starting to kick off. Need a merged proposal to allow better linking into relevant sections of the doc. 

Please don't let that stop you from continuing to ask questions, file issues or suggest changes. More than happy to keep the convo going on this!

PR #5617, opened 2022-01-07
Fix code format

lindexi:
See: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

> We use PascalCasing for all method names, including local functions.

### DIFF
--- a/spec/documentation-comments.md
+++ b/spec/documentation-comments.md
@@ -32,8 +32,8 @@ __Example:__
 ///
 public class Point 
 {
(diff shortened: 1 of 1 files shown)


333fred:
We're not accepting new PRs to the spec folder at this time, as we're nearly ready to move the spec over to the https://github.com/dotnet/csharpstandard repo. It looks like their version of this file still has the lowercase function names, so I expect they'd accept this PR.

lindexi:
Thank you @333fred

PR #5676, opened 2022-01-24
Track runtime feature flag

jaredpar:


jcouv:
You may want to add ref fields to the feature list and open a test plan. That way we can link all the related work items from one place.

jaredpar:
Right now everything is being tracked here cause there is a lot of cross team dependencies on this 

https://github.com/dotnet/runtime/issues/63768

Yeah I do need to update lang feature status though.

PR #5689, opened 2022-01-26
Support for ref struct as generic arguments

jaredpar:
This proposal details how support can be added to the language to allow `ref struct` to be used as generic arguments. While the proposal is functional there are two areas that I am currently undecided on:

- The proposal uses the `where T : ref struct` syntax which is using a constraint to express what is really an anti-constraint. There is a section detailing an alternate syntax but I'm unsure if it meets the bar for new syntax.
- The proposal prefers a user consumable solution for how we deal with the problem APIs on `Span<T>`. Still unsure if we should prefer that or go for special casing as we do for `T[]` and generics.

### DIFF
--- a/proposals/generic-ref-struct.md
+++ b/proposals/generic-ref-struct.md
@@ -142,7 +142,51 @@ Note: the [alternative](#special-case) here is simply special casing `Span<T>` a
 ### ref struct interfaces
 <a name="interfaces"></a>
 
-**TODO** look up the definition of the constrained prefix for low level rules on where this is legal
+The language will allow `ref struct` to implement interfaces. The rules for implementation will be the same as for normal `struct`. This includes allowing the implementations to be implicit or explicit.
+
+**TODO: Need better example than animals but feeling lazy***
+
+```c#
+interface IAnimal
+{
+    void Speak(); 
+}
+
+ref struct Lab : IAnimal
+{
+    public void Speak() => Console.WriteLine("FEED ME");
+}
+```
+
+The one exception is when `ref struct` intersects with default implementation methods. When interfaces containing default implemented methods are used on `struct` types there is an implicit boxing operation for the invocation. That means effectively a DIM call when the implementing type was a `ref struct` would violate our rules on them appearing in the heap. As such this will be disallowed by both the language and runtime. 
+
+- A `ref struct` that implements an `interface` must provide implementations for all methods, even those with a default implementation
+- A DIM invocation on a `ref struct` will result in a runtime exception
+
+The ability to implement interfaces does not change the rules around conversions though. A `ref struct` cannot be boxed, conversion from a `struct` to an `interface` has an implicit box associated with it hence conversion from `ref struct` to its implemented interfaces remains an error: 
+
+```c#
+ref struct S : IUtil { ... }
+
+S local = ...;
+IUtil util = local; // Error: a ref struct cannot be boxed 
+```
+
+The interfaces implemented by a `ref struct` can be used to satisfy constraints on a generic type parameter. Invocation of the interface methods on the type parameters is allowed because such invocations use the `.constrained` prefix for execution which eliminates boxing when the underlying type is a `struct`. 
+
+```c# 
+void Write<T>(T value)
+    where T : ref struct, ISpanFormattable
+{
+    var buffer = stackalloc char[100];
+
+    // Constrained interface call which does not box
+    if (value.TryFormat(buffer, out var written, default, null))
+    {
+        this.Write(buffer);
+    }
+}
+```
 
 ### delegates
 
--- a/proposals/generic-ref-struct.md
+++ b/proposals/generic-ref-struct.md
@@ -37,7 +37,7 @@ Swap<byte>(ref span1[0], ref span1[1]);
 
 At the call site this relaxes the set of legal type arguments to inclued `ref struct` *in addition* to the types already allowed. Even though this appears in the constraint position it serves more of an anti-constraint because it expands the set of possible types. 
 
-As an expansion though it means that `T where T : ref struct` is not compatible with type parameters lacking the same annotation:
+As an expansion though it means that `where T : ref struct` is not compatible with type parameters lacking the same annotation:
 
 ```c#
 void M1<U>() { }
@@ -49,7 +49,7 @@ void M2<T>()
 }
 ```
 
-At the implementation all type parameters having the `where T : ref struct` annotation are treated as if the type is a `ref struct`. Most importantly it has all of the limitations that would be expected from a `ref struct`. For example: 
+At the implementation all type parameters having the `where T : ref struct` annotation are treated as if the type is a `ref struct`. This means the type parameter has all of the limitations that would be expected a `ref struct` type. For example:
 
 - Instances of `T` cannot be boxed or appear on the heap
 - Instances of `T` can only appear as fields inside a `ref struct`
@@ -144,7 +144,7 @@ Note: the [alternative](#special-case) here is simply special casing `Span<T>` a
 
 The language will allow `ref struct` to implement interfaces. The rules for implementation will be the same as for normal `struct`. This includes allowing the implementations to be implicit or explicit.
 
-**TODO: Need better example than animals but feeling lazy***
+**TODO: Need better example than animals but feeling lazy**
 
 ```c#
 interface IAnimal
@@ -191,9 +191,25 @@ void Write<T>(T value)
 ### delegates
 The features described thus far allow developers the flexibility to opt existing generic types into `where T : ref struct` where legal. This though requires developers modify existing types and deal with any incompatibilities that may occur. Essentially it is not automatic, it requires work. 
 
-There is one class of types for which no human inpsection is needed: delegates. Delegates represent types that are signature only, there is no implicit implementation. That means the compiler itself can infer whether a type parameter could be validly `where T : ref struct` or not. It simply has to examine the signature and determine if the type parameter is used in any type where the corresponding parameter is not `where T : ref struct`. 
+There is one class of types for which no human inpsection is needed: delegates. Delegates represent types that are signature only, there is no implementation to consider. That means the compiler itself can infer whether a type parameter could be validly `where T : ref struct` or not. It can do so by examining the places where type parameters are used and validating if the constraints are compatible with `where T : ref struct`. For example: 
 
-To expediate the adoption of this feature and make it easier to flow through th ecosystem the compiler will automatically mark type parameters on `delegate` instances as `where T : ref struct` when it detects it is legal. That means all existing `Action` and `Func` definition will be immediately eligable to take and return `Span<T>` instances.
+```c#
+// T is never used inside another generic hence implicitly supports ref struct
+delegate T Delegate1<T>(); 
+delegate void Delegate2<T>(T p);
+delegate void Deleaget3<T>(ref T p);
+
+// T is used in a generic param that does not support ref struct hence it does not 
+// support it
+delegate void Delegate4<T>(List<T> list);
+```
+
+The compiler will automatically mark type parameters on such `delegate` declarations as having the `where T : ref struct` annotation. This will eexpediate the adoption of this feature and not force a lot of extra annotations by developers. Consider that this will automatically make all known `Action` and `Func` delegates available for `ref struct` usage.
+
+## Open Issues
+
+- Special case `Span<T>` like array vs. using an attribute and making it end user consumable
+- Use constraint syntax vs. a newly defined allow style syntax
 
 ## Considerations
 
@@ -335,11 +351,16 @@ This also means that the use of `where T : any` on interfaces is not consequence
 It is not believed that the `interface` scenario is worth the extra syntax here. If more evidence arrives that demonstrates it to be worth it we should consider this in a future version of the language.
 
 ### Take delegates further
-**TODO: explain how delegates could use ref T as a type argument**
+The [low level hackathon](https://github.com/dotnet/runtime/tree/feature/lowlevelhackathon) demonstrated that at the `delegate` level, where there are only signatures, generic instantiation can go beyond even `ref struct`. It is possible to support generic arguments with all types not supported today, like pointers, as well as items like `ref T` which are more type modifiers than types in the language.
 
-## Open Issues
+```c# 
+int i = 0;
+Action<ref int> a = (ref int j) => j++;
+a(ref i);
+Console.WriteLine(i); // 1
+```
 
-## Future Considerations
+It is unlikely that this support will be added to the language in C# 11 or runtime in .NET 7. But support is a realistic possibility in future releases. It is another reason to consider leaving delegates us implicitly opt in. That would mean if the support is extended in the runtime in the future then delegates could continue to automatically get better at the same time.
 
 ## Examples 
 
@@ -399,3 +420,5 @@ sealed class RefList<T> where T : ref struct
 ### Related Work
 
 - https://github.com/dotnet/runtime/issues/13627
+- https://github.com/dotnet/csharplang/issues/1148
+-
--- a/proposals/generic-ref-struct.md
+++ b/proposals/generic-ref-struct.md
@@ -7,9 +7,11 @@ Unconstraining our generics
 
 ## Detailed Design 
 
-### Where T : ref struct
+### ref struct generic arguments
 
-### where T : any
+### ref struct interfaces
+
+### delegates
 
 ## Considerations
 
@@ -19,8 +21,87 @@ Why have an attribute when we could just special case it like arrays
 ### Why not attribute per type parameter
 Can't really encode type parameters in attributes
 
+### Generalize the delegate support
+The use of any type in `delegate` may cause readers to desire a more general feature that could be applied to any generic parameter. Essentially could we design a syntax for type parameters that could be any type (pointer, `ref`, etc ...) and allow it on any generic parameter? That would leave `delegate` instances as effectively having an implicit version of this syntax.
+
+It is possible to design such a syntax. Let's use `where T : any` for discussion purposes.
+
+```c#
+// Developer writes
+public delegate T Func<T>();
+
+// Developer effectively writes
+public delegate T Func<T>() where T : any;
+```
+
+The problem with this feature is that it's only useful in signature only scenarios. Those work becase the there is no action associated with the generic parameters. The compiler only needs to consider the standard constraint checking. Effectively the compiler needs to ensure that a `T` with `where T : any` isn't passed to an incompatible generic type parameter.
+
+```c#
+interface I1<U> { }
+interface I2<U> where U : IFormattable { }
+
+// Error: T is more permissive than U
+interface IImpl1<T> : I1<T> where T : any { }
+
+// Error: T doesn't implement IFormattable
+interface IImpl2<T> : I2<T> where T : any { }
+```
+
+Once the feature is applied to scenarios where code can execute on `T` it becomes clear that virtually nothing can be done with `T`. That is because the langauge, and runtime, have to consider all the different permutations that `T` can take. Lets look at a few simple examples: 
+
+```c#
+void M<T>(T parameter)
+    where T : any
+{
+    T local = parameter;
+    ...
+}
+```
+
+This, assigning a parameter to a local, seems simple enough but what are the semantics here? Consider that most C# developers would believe this to be simple value assignment but what happens when this method is instantiated as `ref int`? That effectively translates the assignment to `ref int local = parameter` which is an error. This is a hard error to reconcile because it violates basic `ref` local semantics. 
+
+Attempts to be clever here by saying assignment auto expands to taking the `ref` of the right hand side when the left is `ref` quickly fall part once you expand out the scenarios a bit more: 
+
+```c# 
+T GetValue<T>() where T : any { ... }
+
+void Use<T>(T parameter) where T : any, new()
+{
+    parameter = GetValue<T>(); // Okay: taking ref of RHS is legal when T is ref 
+    parameter = new T(); // Error: can't take ref of non-location
+    parameter = default; // Error: can't take ref of non-location
+}
+```
+
+Effectively rules meant to make assignment work just break it in other cases. It does not appear to be generally solvable. This means basic value manipulation is not possible.
+
+The problems go even deeper though. Consider the following:
+
+```c#
+void M2<T>(T parameter)
+    where T : any
+{
+    // Error: T doesn't necessarily inherit from object
+    Console.WriteLine(parameter.ToString());
+}
+```
+
+This example must be an error. The compiler must consider the case where `T` is instantiated as a pointer, say `int*`. That type does not derive from `object` and hence has no `ToString` to invoke.
+
+The other issue is that a syntax like `where T : any` needs to satisfy *any* type. That includes type combinations that are not allowed in C# today, or added to the CLI in the future. Consider as a concrete example that `ref ref` is not allowed in CIL or C# today but that it's reasonable for this to be added in [the future](https://github.com/dotnet/runtime/pull/63659#discussion_r784986222). It seems implausible that the semantics for these types of features could be correctly predicted in advance.
+
+Effectively the only item that can be taken with `T where T : any` is to instantiate other generics. No concrete actions can be taken on instances of the type itself. Doing so would require significant changes to the runtime and hard to explain C# semantics. There are some discussions around making runtime changs, [example](https://github.com/dotnet/runtime/issues/13627), but those aren't general enough for this to work.
+
+This also means that the use of `where T : any` on interfaces is not consequence free. Implicitly adding this to an interface will break any default implementated method on that interface. Hence unlike `delegate` it cannot be retroactively added to our existing types. It is possible as an opt-in for `interface` but doing so effectively removes the ability for an `interface` to participate in DIM. 
+
+It is not believed that the `interface` scenario is worth the extra syntax here. If more evidence arrives that demonstrates it to be worth it we should consider this in a future version of the language.
+
 ## Open Issues
 
 ## Future Considerations
 
 ### Issues
+
+### Related Work
+
+- https://github.com/dotnet/runtime/issues/13627
--- a/proposals/generic-ref-struct.md
+++ b/proposals/generic-ref-struct.md
@@ -4,14 +4,18 @@ Unconstraining our generics
 ## Summary
 This proposal is an aggregation of several different proposals for making `ref struct` more generally usable in our type system. The goal being to unify code paths today that have to specialize on `ref struct` and other types do to the inability of `ref struct` to participate in generics and implement interfaces. This proposal will allow us to generalize on `ref struct` as naturally as other types. 
 
-
 ## Motivation
+This feature will allow developers to remove `ref struct` specific code by instead letting them implement interfaces that are used in generic contexts to generalize algorithms. For example an explicit goal is to allow `ref struct` to implement `ISpanFormattable`. The `ISpanFormattable` interface and `ref struct` are used in low level perf sensitive areas but our current language limitations prevent them from being used together.
 
 ## Detailed Design 
 
 ### ref struct generic arguments
 
+This means type parameters with the `ref struct` constraint have all of the restrictions 
+
 ### ref struct interfaces
+
+**TODO** look up the definition of the constrained prefix for low level rules on where this is legal
 
 ### delegates
 
@@ -44,6 +48,8 @@ struct S<T, U>
 In this example the method `M` is only invalid when `T` is a `ref struct`. The type of `U` is immaterial to the legality of the method as it's not used in a manner incompatible with `ref struct`. In this case it would be desirable to have a more granular opt out. 
 
 At this time though there are no practical examples of this pattern. Adding this granularity will add cost to the runtime, as well as to the language as it's hard to reference type parameters in attributes. Going for simplicity at the moment but can reconsider in future versions if evidence for the scenario appears.
+
+**TODO** is IgnoreForRefStruct the right name (no)
 
 ### Constraints vs. anti constraints
 The syntax `where T : ref struct` appears in the space C# refers to as constraints. That is the syntax is meant to constraint or limit the set of types that are legal for the generic type parameter.
--- a/proposals/generic-ref-struct.md
+++ b/proposals/generic-ref-struct.md
@@ -137,6 +137,8 @@ void M<T, U>()
 
 The runtime will recognize this attribute and ensure that any use of these members at runtime with a type that is a `ref struct` will result in a runtime exception.
 
+Note: the [alternative](#special-case) here is simply special casing `Span<T>` as array is special cased today.
+
 ### ref struct interfaces
 <a name="interfaces"></a>
 
@@ -147,7 +149,7 @@ The runtime will recognize this attribute and ensure that any use of these membe
 ## Considerations
 
 ### Treat it like array
-**TODO** Use the name `SupportsOnlyNonByRefLike`
+<a name="special-case"></a>
 
 Rather than `[SupportsOnlyNonByRefLike]` attribute why not just special case `Span<T>` and other types as the language special cases arrays today. 
 
@@ -181,7 +183,32 @@ The syntax `where T : ref struct` appears in the space C# refers to as constrain
 
 This feature though expands the set of types available. The set of types applicable to `T where T : ref struct` is larger than simply `T`. This means the feature is an anti-constraint yet it occupies a constraint location. 
 
-It is reasonable to consider a new syntax here that cleanly differentiates our anti-constraints from actual constraints. At the time of this writing though there are only two proposed anti-constraints, one is lacking supporting evidence, and no obvious better syntax for them. Hence the proposal is written using the familiair `where T :` style of syntax. If an intuitive syntax for anti-constraints is proposed it is likely to get consideration in the proposal.
+It is reasonable to consider a new syntax here that cleanly differentiates our anti-constraints from actual constraints. At the time of this writing though there are only two proposed anti-constraints and one is lacking supporting evidence. This makes it questionable if it meets the bar for a new syntax or not.
+
+If a new syntax is desired to separate anti-constraints from constraints then `allow` is one to consider: 
+
+```c#
+readonly ref struct Span<T> 
+    allow T : ref struct
+{
+
+}
+```
+
+This makes the support for `[SupportsOnlyNonByRefLike]` cleaner as the implementation is to simply ignore the set of `allow` items. It may even lead to a better name such as `[RemoveAllows]`. 
+
+The only downside is that it means type parameter declarations get more complex because there are now both `where` and `allow` lists on them.
+
+```c#
+void Write<T>(T value)
+    where T : ISpanFormattable
+    allow T : ref struct
+{
+
+}
+```
+
+At the same time it's possible this will read clearer to developers. It also opens the door for future allows such as pointers. 
 
 ### Generalize the delegate support
 The use of any type in `delegate` may cause readers to desire a more general feature that could be applied to any generic parameter. Essentially could we design a syntax for type parameters that could be any type (pointer, `ref`, etc ...) and allow it on any generic parameter? That would leave `delegate` instances as effectively having an implicit version of this syntax.
--- a/proposals/generic-ref-struct.md
+++ b/proposals/generic-ref-struct.md
@@ -189,6 +189,11 @@ void Write<T>(T value)
 ```
 
 ### delegates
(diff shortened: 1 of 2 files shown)


MichalPetryka:
I feel like if the path with `where` is to be taken, dropping the `struct` part in the constraint (i.e. `where T : ref`) could look a bit cleaner.

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+Span<byte> span2 = default;
+
+// Can use ref struct type arguments 
+Swap<Span<byte>>(ref span1, ref span2);
+
+// Can also use normal type arguments
+Swap<byte>(ref span1[0], ref span1[1]);
+```
+
+At the call site this relaxes the set of legal type arguments to inclued `ref struct` *in addition* to the types already allowed. Even though this appears in the constraint position it serves more of an anti-constraint because it expands the set of possible types.
```
cston, line 38:
```suggestion
At the call site this relaxes the set of legal type arguments to included `ref struct` *in addition* to the types already allowed. Even though this appears in the constraint position it serves more of an anti-constraint because it expands the set of possible types. 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+As an expansion though it means that `where T : ref struct` is not compatible with type parameters lacking the same annotation:
+
+```c#
+void M1<U>() { }
+void M2<T>()
+    where T : ref struct
+{
+    M1<T>(); // Error: type parameter T is not compatible with U because it does not 
+             // allow `ref sturct`
```
cston, line 48:
```suggestion
             // allow `ref struct`
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+    {
+
+    }
+}
+```
+
+Positionally in syntax the `ref struct` annotation must come after the type kind, such as `struct`, but before the interface list or `new()` constraint (interfaces are covered [later in the proposal](#interfaces)). The combination with type kinds has the following effects:
+
+- `where T : struct, ref struct`: describes a type parameter that can be a `struct` or `ref struct` but not other types like `class` or `interface`.
+- `where T : unmanaged, ref struct`: similar to `struct, ref struct` except that the type also meets the `unmanaged` constraint. This is believed to have limited use though as `ref struct` containing any type of `ref` data will not meet the `unmanaged` contraint and hence are likely better defined as simply `struct`.
```
cston, line 80:
```suggestion
- `where T : unmanaged, ref struct`: similar to `struct, ref struct` except that the type also meets the `unmanaged` constraint. This is believed to have limited use though as `ref struct` containing any type of `ref` data will not meet the `unmanaged` constraint and hence are likely better defined as simply `struct`.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+- `where T : notnull, ref struct`: this is the same as the `notnull` constraint but also allows for `ref struct` types to satisfy it.
+- `where T : class, ref struct`: is an error as the `ref struct` portion is meaningless when combined with `class`. 
+- `where T : default, ref struct`: is an error as the declarations are contradictory.
+- `where T : ref struct, BaseType` where `BaseType` denotes a reference type. This is an error because the `ref sruct` portion is meaningless in this context.
```
cston, line 84:
```suggestion
- `where T : ref struct, BaseType` where `BaseType` denotes a reference type. This is an error because the `ref struct` portion is meaningless in this context.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+### delegates
+The features described thus far allow developers the flexibility to opt existing generic types into `where T : ref struct` where legal. This though requires developers modify existing types and deal with any incompatibilities that may occur. Essentially it is not automatic, it requires work. 
+
+There is one class of types for which no human inpsection is needed: delegates. Delegates represent types that are signature only, there is no implementation to consider. That means the compiler itself can infer whether a type parameter could be validly `where T : ref struct` or not. It can do so by examining the places where type parameters are used and validating if the constraints are compatible with `where T : ref struct`. For example: 
+
+```c#
+// T is never used inside another generic hence implicitly supports ref struct
+delegate T Delegate1<T>(); 
+delegate void Delegate2<T>(T p);
+delegate void Deleaget3<T>(ref T p);
```
cston, line 200:
```suggestion
delegate void Delegate3<T>(ref T p);
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+// T is used in a generic param that does not support ref struct hence it does not 
+// support it
+delegate void Delegate4<T>(List<T> list);
+```
+
+The compiler will automatically mark type parameters on such `delegate` declarations as having the `where T : ref struct` annotation. This will eexpediate the adoption of this feature and not force a lot of extra annotations by developers. Consider that this will automatically make all known `Action` and `Func` delegates available for `ref struct` usage.
```
cston, line 207:
```suggestion
The compiler will automatically mark type parameters on such `delegate` declarations as having the `where T : ref struct` annotation. This will expedite the adoption of this feature and not force a lot of extra annotations by developers. Consider that this will automatically make all known `Action` and `Func` delegates available for `ref struct` usage.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+```c#
+// Developer writes
+public delegate T Func<T>();
+
+// Developer effectively writes
+public delegate T Func<T>() where T : any;
+```
+
+The problem with this feature is that it's only useful in signature only scenarios. Those work becase the there is no action associated with the generic parameters. The compiler only needs to consider the standard constraint checking. Effectively the compiler needs to ensure that a `T` with `where T : any` isn't passed to an incompatible generic type parameter.
```
cston, line 291:
```suggestion
The problem with this feature is that it's only useful in signature only scenarios. Those work because the there is no action associated with the generic parameters. The compiler only needs to consider the standard constraint checking. Effectively the compiler needs to ensure that a `T` with `where T : any` isn't passed to an incompatible generic type parameter.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+    where T : any
+{
+    T local = parameter;
+    ...
+}
+```
+
+This, assigning a parameter to a local, seems simple enough but what are the semantics here? Consider that most C# developers would believe this to be simple value assignment but what happens when this method is instantiated as `ref int`? That effectively translates the assignment to `ref int local = parameter` which is an error. This is a hard error to reconcile because it violates basic `ref` local semantics. 
+
+Attempts to be clever here by saying assignment auto expands to taking the `ref` of the right hand side when the left is `ref` quickly fall part once you expand out the scenarios a bit more:
```
cston, line 317:
```suggestion
Attempts to be clever here by saying assignment auto expands to taking the `ref` of the right hand side when the left is `ref` quickly fall apart once you expand out the scenarios a bit more: 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+Effectively the only item that can be taken with `T where T : any` is to instantiate other generics. No concrete actions can be taken on instances of the type itself. Doing so would require significant changes to the runtime and hard to explain C# semantics. There are some discussions around making runtime changs, [example](https://github.com/dotnet/runtime/issues/13627), but those aren't general enough for this to work.
+
+This also means that the use of `where T : any` on interfaces is not consequence free. Implicitly adding this to an interface will break any default implementated method on that interface. Hence unlike `delegate` it cannot be retroactively added to our existing types. It is possible as an opt-in for `interface` but doing so effectively removes the ability for an `interface` to participate in DIM.
```
cston, line 349:
```suggestion
This also means that the use of `where T : any` on interfaces is not consequence free. Implicitly adding this to an interface will break any default implemented method on that interface. Hence unlike `delegate` it cannot be retroactively added to our existing types. It is possible as an opt-in for `interface` but doing so effectively removes the ability for an `interface` to participate in DIM. 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+ref struct Lab : IAnimal
+{
+    public void Speak() => Console.WriteLine("FEED ME");
+}
+```
+
+The one exception is when `ref struct` intersects with default implementation methods. When interfaces containing default implemented methods are used on `struct` types there is an implicit boxing operation for the invocation. That means effectively a DIM call when the implementing type was a `ref struct` would violate our rules on them appearing in the heap. As such this will be disallowed by both the language and runtime. 
+
+- A `ref struct` that implements an `interface` must provide implementations for all methods, even those with a default implementation
```
YairHalberstadt, line 163: An important motivation for DIMs is they allow developers to evolve interfaces without breaking implementors.

Forcing ref structs to implement DIMs means that adding a DIM will break any ref structs that implement the interface.

Are we ok saying that that's rare enough it shouldn't be a concern to interface authors?

Alternatively should interfaces have to opt into being implemented by ref structs, or have the ability to opt out?

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+    public Span(T[]? array, int index, int length) { ... }
+
+    public static implicit operator Span<T>(T[]? array) { ... }
+}
+```
+
+If the `Span<T>` type was being designed from scratch in the presence of `where T : ref struct` it's possible different choices would've been made here. Possibly leveraging `static` factories and pushing implicit conversions into the compiler. Unfortunately this is not being designed from scratch and in order to meet the desire to have `Span<Span<T>>` these existing APIs need to be rationalized.
+
+The compiler will introduce the type `System.Runtime.CompilerServices.SupportsOnlyNonByRefLikeAttribute` that can be applied to methods and operators. When applied to a member it has the effect of removing `where T : ref struct` from the constraints on all type parameters on the *immediate* containing type. All other constraints remain in place.
```
YairHalberstadt, line 105: Is there any room to consider a more general feature, where users can further constrain specific methods on types, so that a method only applies to the subset of generic instantiations that fulfill the increased constraints?

Currently it's common to do this by implementing extension methods, but this doesn't work for cases where you need access to private data, or for static methods.

If this was allowed you could add a `not ref struct` constraint which all type parameters have by default, and write:

```csharp
public Span(T[]? array) where T : not ref struct { ... }
```

I don't think this is incredibly useful, but I prefer general elegant solutions over specific hacky ones, and given the plethora of extension methods which do this there definitely will be plenty of use cases which would have preferred an instance method, but were forced to use an extension method.

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
+Unconstraining our generics
+=====
+
+## Summary
+This proposal is an aggregation of several different proposals for making `ref struct` more generally usable in our type system. The goal being to unify code paths today that have to specialize on `ref struct` and other types do to the inability of `ref struct` to participate in generics and implement interfaces. This proposal will allow us to generalize on `ref struct` as naturally as other types.
```
333fred, line 5: Well, I was going to try and read through this without correcting spelling and grammar, but ya got me good Jared.

```suggestion
This proposal is an aggregation of several different proposals for making `ref struct` more generally usable in our type system. The goal is to unify code paths today that have to specialize on `ref struct` and other types due to the inability of `ref struct` to participate in generics and implement interfaces. This proposal will allow us to generalize on `ref struct` as naturally as other types. 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
+Unconstraining our generics
+=====
+
+## Summary
+This proposal is an aggregation of several different proposals for making `ref struct` more generally usable in our type system. The goal being to unify code paths today that have to specialize on `ref struct` and other types do to the inability of `ref struct` to participate in generics and implement interfaces. This proposal will allow us to generalize on `ref struct` as naturally as other types. 
+
+## Motivation
+This feature will allow developers to remove `ref struct` specific code by instead letting them implement interfaces that are used in generic contexts to generalize algorithms. For example an explicit goal is to allow `ref struct` to implement `ISpanFormattable`. The `ISpanFormattable` interface and `ref struct` are used in low level perf sensitive areas but our current language limitations prevent them from being used together.
+
+This will also allow for combinations like `Span<Span<T>>`. This is a long desire of the runtime team who often want to represent collections of `Span<T>` but are unable to do so due to our current limitations.
```
333fred, line 10:
```suggestion
This will also allow for combinations like `Span<Span<T>>`. This has long been a desire of the runtime team, who often want to represent collections of `Span<T>` but are unable to do so due to our current limitations.
```

333fred: Or alternately
```suggestion
This will also allow for combinations like `Span<Span<T>>`. This is a long-held desire of the runtime team who often want to represent collections of `Span<T>` but are unable to do so due to our current limitations.
```

proposals/generic-ref-struct.md
333fred:
```suggestion
At the use site this relaxes the set of legal type arguments to include `ref struct` *in addition* to the types already allowed. Even though this appears in the constraint position it serves more of an anti-constraint because it expands the set of possible types. 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+// Can use ref struct type arguments 
+Swap<Span<byte>>(ref span1, ref span2);
+
+// Can also use normal type arguments
+Swap<byte>(ref span1[0], ref span1[1]);
+```
+
+At the call site this relaxes the set of legal type arguments to inclued `ref struct` *in addition* to the types already allowed. Even though this appears in the constraint position it serves more of an anti-constraint because it expands the set of possible types. 
+
+As an expansion though it means that `where T : ref struct` is not compatible with type parameters lacking the same annotation:
```
333fred, line 40:
```suggestion
As an expansion it means that `where T : ref struct` is not compatible with type parameters lacking the same annotation:
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+    M1<T>(); // Error: type parameter T is not compatible with U because it does not 
+             // allow `ref sturct`
+}
+```
+
+At the implementation all type parameters having the `where T : ref struct` annotation are treated as if the type is a `ref struct`. This means the type parameter has all of the limitations that would be expected a `ref struct` type. For example:
+
+- Instances of `T` cannot be boxed or appear on the heap
+- Instances of `T` can only appear as fields inside a `ref struct`
+- Instances of `T` are considered `ref struct` for the purposes of the [span safety rules](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md)
```
333fred, line 56:
```suggestion
- Instances of `T` are considered `ref struct`s for the purposes of the [span safety rules](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md) 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+- `where T : notnull, ref struct`: this is the same as the `notnull` constraint but also allows for `ref struct` types to satisfy it.
+- `where T : class, ref struct`: is an error as the `ref struct` portion is meaningless when combined with `class`. 
+- `where T : default, ref struct`: is an error as the declarations are contradictory.
+- `where T : ref struct, BaseType` where `BaseType` denotes a reference type. This is an error because the `ref sruct` portion is meaningless in this context.
+
+The `where T : ref struct` annotation will be encoded by attributing the type parameter with the attribute `System.Runtime.CompilerServices.SupportsByRefLike`. This will not result in a `modreq` on the containing member nor will `[Obsolete]` tricks be added to the containing type. The intent is for this to be a binary compatible change. The explicit intent is for us to allow types in libraries to relax their type parameters to accept `ref struct` without breaking consumers.
```
333fred, line 86: Did you mean `SupportsByRefLikeAttribute`? Or were you intending to leave the suffix off?

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Supporting Span<T>
+A key motivation of this proposal is to allow for instantiations like `Span<Span<T>>`. Effectively moving the `Span<T>` type to have the `where T : ref struct` annotation. This is a problem because a number of the APIs on `Span<T>` violate the rules of `where T : ref struct` as `T[]` is a common type in the APIs.
```
333fred, line 89:
```suggestion
A key motivation of this proposal is to allow for instantiations like `Span<Span<T>>`, effectively moving the `Span<T>` type to have the `where T : ref struct` annotation. This is a problem because a number of the APIs on `Span<T>` violate the rules of `where T : ref struct` as `T[]` is a common type in the APIs. 
```

proposals/generic-ref-struct.md
333fred: This is effectively inching into generic specialization (though really, it's anti-specialization?), and I wonder what the thoughts from the runtime team on generalized specialization are.

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+The runtime will recognize this attribute and ensure that any use of these members at runtime with a type that is a `ref struct` will result in a runtime exception.
+
+Note: the [alternative](#special-case) here is simply special casing `Span<T>` as array is special cased today.
+
+### ref struct interfaces
+<a name="interfaces"></a>
+
+The language will allow `ref struct` to implement interfaces. The rules for implementation will be the same as for normal `struct`. This includes allowing the implementations to be implicit or explicit.
```
333fred, line 145:
```suggestion
The language will allow `ref struct`s to implement interfaces. The rules for implementation will be the same as for normal `struct`s. This includes allowing the implementations to be implicit or explicit.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+```c#
+interface IAnimal
+{
+    void Speak(); 
+}
+
+ref struct Lab : IAnimal
+{
+    public void Speak() => Console.WriteLine("FEED ME");
```
333fred, line 157: Was Penny that hungry when you were writing this?

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+}
+```
+
+The one exception is when `ref struct` intersects with default implementation methods. When interfaces containing default implemented methods are used on `struct` types there is an implicit boxing operation for the invocation. That means effectively a DIM call when the implementing type was a `ref struct` would violate our rules on them appearing in the heap. As such this will be disallowed by both the language and runtime. 
+
+- A `ref struct` that implements an `interface` must provide implementations for all methods, even those with a default implementation
+- A DIM invocation on a `ref struct` will result in a runtime exception
+
+The ability to implement interfaces does not change the rules around conversions though. A `ref struct` cannot be boxed, conversion from a `struct` to an `interface` has an implicit box associated with it hence conversion from `ref struct` to its implemented interfaces remains an error:
```
333fred, line 166:
```suggestion
The ability to implement interfaces does not change the rules around conversions though. A `ref struct` cannot be boxed, and since conversion from a `struct` to an `interface` has an implicit box associated with it, conversion from a `ref struct` to its implemented interfaces remains an error: 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+    // Constrained interface call which does not box
+    if (value.TryFormat(buffer, out var written, default, null))
+    {
+        this.Write(buffer);
+    }
+}
+```
+
+### delegates
```
333fred, line 191:
```suggestion
### Delegates
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+The features described thus far allow developers the flexibility to opt existing generic types into `where T : ref struct` where legal. This though requires developers modify existing types and deal with any incompatibilities that may occur. Essentially it is not automatic, it requires work.
```
333fred, line 192:
```suggestion
The features described thus far allow developers the flexibility to opt existing generic types into `where T : ref struct` where legal. However, this requires developers modify existing types and deal with any incompatibilities that may occur. Essentially it is not automatic, it requires work. 
```

proposals/generic-ref-struct.md
333fred, line 207: It's unclear to me what you mean by mark here. My first read implied to me that you meant automatically applying the attribute to the type parameter, but I suspect that's not what you meant after re-reading it. Consider being more explicit.

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+<a name="special-case"></a>
+
+Rather than `[SupportsOnlyNonByRefLike]` attribute why not just special case `Span<T>` and other types as the language special cases arrays today. 
+
+Arrays today conditionally implement interfaces based on the element type of the array. For example a `T[]` implements `IEnumerable<T>` when `T` is a non-pointer type. That means `int*[]` does not but `int[]` does. The issues around generics and arrays is very good analogy for `where T : ref struct` and `Span<T>`. Effectively there is a set of APIs that need to be illegal based on the element type of `Span<T>` hence why not apply the same rules as array.
+
+In discussion with the runtime team there was agreement that the analogy is good. However under the hood the runtime was going to need some way to track the methods in question as invalid when the element type of `Span<T>` was a `ref struct`: an internal list, attribute, etc ... Given that some level of book keeping was needed and attributes were cheap it made more sense to expose this as a general feature. This way third parties can push `where T : ref struct` through types similar to `Span<T>` where they have APIs they can't walk away from.
```
333fred, line 223:
```suggestion
In discussion with the runtime team there was agreement that the analogy is good. However under the hood the runtime was going to need some way to track the methods in question as invalid when the element type of `Span<T>` was a `ref struct`: an internal list, attribute, etc ... Given that some level of book keeping was needed and attributes are cheap it made more sense to expose this as a general feature. This way third parties can push `where T : ref struct` through types similar to `Span<T>` where they have APIs they can't walk away from.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+In this example the method `M` is only invalid when `T` is a `ref struct`. The type of `U` is immaterial to the legality of the method as it's not used in a manner incompatible with `ref struct`. In this case it would be desirable to have a more granular opt out. 
+
+At this time though there are no practical examples of this pattern. Adding this granularity will add cost to the runtime, as well as to the language as it's hard to reference type parameters in attributes. Going for simplicity at the moment but can reconsider in future versions if evidence for the scenario appears.
+
+### Constraints vs. anti constraints
+The syntax `where T : ref struct` appears in the space C# refers to as constraints. That is the syntax is meant to constraint or limit the set of types that are legal for the generic type parameter.
+
+This feature though expands the set of types available. The set of types applicable to `T where T : ref struct` is larger than simply `T`. This means the feature is an anti-constraint yet it occupies a constraint location.
```
333fred, line 249:
```suggestion
However, this feature expands the set of types available. The set of types applicable to `T where T : ref struct` is larger than simply `T`. This means the feature is an anti-constraint yet it occupies a constraint location. 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+
+At this time though there are no practical examples of this pattern. Adding this granularity will add cost to the runtime, as well as to the language as it's hard to reference type parameters in attributes. Going for simplicity at the moment but can reconsider in future versions if evidence for the scenario appears.
+
+### Constraints vs. anti constraints
+The syntax `where T : ref struct` appears in the space C# refers to as constraints. That is the syntax is meant to constraint or limit the set of types that are legal for the generic type parameter.
+
+This feature though expands the set of types available. The set of types applicable to `T where T : ref struct` is larger than simply `T`. This means the feature is an anti-constraint yet it occupies a constraint location. 
+
+It is reasonable to consider a new syntax here that cleanly differentiates our anti-constraints from actual constraints. At the time of this writing though there are only two proposed anti-constraints and one is lacking supporting evidence. This makes it questionable if it meets the bar for a new syntax or not.
```
333fred, line 251: You got `differentiates` right but missed `struct` earlier? :P

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+```c#
+void Write<T>(T value)
+    where T : ISpanFormattable
+    allow T : ref struct
+{
+
+}
+```
+
+At the same time it's possible this will read clearer to developers. It also opens the door for future allows such as pointers.
```
333fred, line 276:
```suggestion
At the same time it's possible this will read clearer to developers. It also opens the door for future `allow`s such as pointers. 
```

proposals/generic-ref-struct.md
333fred, line 291:
```suggestion
The problem with this feature is that it's only useful in signature only scenarios. Those work because there is no action associated with the generic parameters. The compiler only needs to consider the standard constraint checking. Effectively the compiler needs to ensure that a `T` with `where T : any` isn't passed to an incompatible generic type parameter.
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+interface I2<U> where U : IFormattable { }
+
+// Error: T is more permissive than U
+interface IImpl1<T> : I1<T> where T : any { }
+
+// Error: T doesn't implement IFormattable
+interface IImpl2<T> : I2<T> where T : any { }
+```
+
+Once the feature is applied to scenarios where code can execute on `T` it becomes clear that virtually nothing can be done with `T`. That is because the langauge, and runtime, have to consider all the different permutations that `T` can take. Lets look at a few simple examples:
```
333fred, line 304:
```suggestion
Once the feature is applied to scenarios where code can execute on `T` it becomes clear that virtually nothing can be done with `T`. That is because the language, and runtime, have to consider all the different permutations that `T` can take. Lets look at a few simple examples: 
```

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+Console.WriteLine(i); // 1
+```
+
+It is unlikely that this support will be added to the language in C# 11 or runtime in .NET 7. But support is a realistic possibility in future releases. It is another reason to consider leaving delegates us implicitly opt in. That would mean if the support is extended in the runtime in the future then delegates could continue to automatically get better at the same time.
+
+## Examples 
+
+### RefList<T>
+<a name="ref-list"></a>
+This is an eample a `List<T>` style data structure with a hard capacity that can store `ref struct` in addition to normal types:
```
333fred, line 369:
```suggestion
This is an example `List<T>`-style data structure with a hard capacity that can store `ref struct`s in addition to normal types:
```

proposals/generic-ref-struct.md
jaredpar: Always

proposals/generic-ref-struct.md
jaredpar: Nah that is just a typo

proposals/generic-ref-struct.md
jaredpar: > I don't think this is incredibly useful, but I prefer general elegant solutions over specific hacky ones, and given the plethora of extension methods which do this there definitely will be plenty of use cases which would have preferred an instance method, but were forced to use an extension method.

Why do you feel this is a hacky solution vs a general one? This proposal is effectively relaxing a restriction that prevented an entire class of types from being used in generic arguments. I agree the syntax can seem off at first, which is why there are a few listed alternatives, but conceptually this is a general feature. 

> If this was allowed you could add a not ref struct constraint which all type parameters have by default, and write:

The default though is effectively not `ref struct` though aWhy do we need a new syntax here?

proposals/generic-ref-struct.md
jaredpar: > Are we ok saying that that's rare enough it shouldn't be a concern to interface authors?

I tend to think of it in the opposite direction. Today there is no way to implement an interface on a `ref struct` hence we are starting from scratch here (no compat concerns). To me the cost of implementing an interface on a `ref struct` becomes that your implementation will not benefit from DIM. Basically you must manually implement all interface members. 

 > Alternatively should interfaces have to opt into being implemented by ref structs, or have the ability to opt out?

I think forcing interfaces to opt-into `ref struct` would unnecessarily slow down the ecosystem. Every time a developer wanted to adapt an `ref struct` to an `interface`, even if it was for a limited use case, they'd have to get the developer to opt the interface in, go through an update cycle, etc ... Very slow. 

As for opting an interface out of `ref struct` implementations I'm not sure of the value. The only advantage I could see that providing is it gives developers the feeling that they could add DIM without breaking people. Problem is that is just not a reality. Due to diamond implementation problems there is always a risk to adding a DIM to an existing shipped interface. The risk depends on a number of factors but in many cases it's non-zero (it's not just a .NET thing here, Java has the same issues). 

This is a case where I'd say "probably better as an analyzer" if you really wanted it.

proposals/generic-ref-struct.md
jaredpar: I know how to spell most words, I just type them wrong for some reason. So every now and then I get lucky in strange ways like that.

proposals/generic-ref-struct.md
YairHalberstadt: @jaredpar I was referring to `System.Runtime.CompilerServices.SupportsOnlyNonByRefLikeAttribute` specifically

RikkiGibson: I think my main concern with the name is that it doesn't specify that it is changing the meaning of type parameters. It makes it sound a little like "also, you can't use spans and stuff in here"

RikkiGibson (approved):

proposals/generic-ref-struct.md
jaredpar: Okay. Yes that is definitely an icky part of this design. 

My original idea was to just special case the methods in .NET that needed this change. Essentially the way we special case some behaviors of arrays. After some discussions I was talked into it being a more general feature. As I wrote it up though I wasn't happy with the result. Been waffling on just making it back to a special case.

proposals/generic-ref-struct.md
333fred: @jaredpar this is still unaddressed.

proposals/generic-ref-struct.md
```diff
@@ -0,0 +1,424 @@
...
+{
+    M1<T>(); // Error: type parameter T is not compatible with U because it does not 
+             // allow `ref struct`
+}
+```
+
+At the implementation all type parameters having the `where T : ref struct` annotation are treated as if the type is a `ref struct`. This means the type parameter has all of the limitations that would be expected a `ref struct` type. For example:
+
+- Instances of `T` cannot be boxed or appear on the heap
+- Instances of `T` can only appear as fields inside a `ref struct`
```
teo-tsirpanis, line 55: We could loosen this restriction and make ordinary generic structs be seen as `ref struct`s if at least one of their generic parameters (and perhaps but not necessarily) that is stored as a field is either a `ref struct`, or a generic parameter that can be a `ref struct`.

For an example, we could redefine `ValueTuple<T1, T2>` like the following:

``` csharp
public struct ValueTuple<T1, T2> where T1 : ref struct, T2 : ref struct
{
    public T1 Item1;
    public T2 Item2;
}
```

and `(int, ReadOnlySpan<int>)` would be allowed and be restricted by the compiler to live only on the stack. All other value tuples would still be allowed to reside on the heap, including generic ones, unless their generic parameters are allowed to be `ref struct`s.

RikkiGibson: Some version of this might be possible but I was concerned about how such a struct could implement interfaces, for example. Would we need to go through a bunch of built-in generic types and add `where T : ref struct` to make it all work?

I haven't thought very deeply about this so I anticipate there could be more stumbling blocks here that make it more difficult than just "let ValueTuple contain ref structs".

teo-tsirpanis: Oh, that's right, we couldn't implement `ITuple` on value tuples for example.

GrabYourPitchforks:
Steve Toub pointed me at this after today's API review. In today's review, we considered an API:

```cs
ReadOnlySpan<T> MemoryExtensions.RemovePrefix<T>(this ReadOnlySpan<T> span, ReadOnlySpan<T> prefix);
```

But this would need three more overloads accounting for `Span<T>`, `ReadOnlyMemory<T>`, and `Memory<T>` so that the return type could be correct. If you expand the scope to include `string.RemovePrefix(...)`, your total overload count hits 5 or more.

Having the feature proposed here would _significantly_ reduce the number of overloads we'd need to achieve these scenarios, since we could use one genericized overload and satisfy the entire world. Consider:

```cs
public interface IContiguousBuffer<TBuffer, TElement> where TBuffer : ref struct, IContiguousBuffer<TBuffer, TElement>
{
    ReadOnlySpan<T> GetSpan();
    TBuffer Slice(int offset, int length);
}

public sealed class string : IContiguousBuffer<string, char> /* , ... */;
public ref struct Span<T> : IContiguousBuffer<Span<T>, T> /* , ... */;
public ref struct ReadOnlySpan<T> : IContiguousBuffer<ReadOnlySpan<T>, T> /* , ... */;
public struct Memory<T> : IContiguousBuffer<Memory<T>, T> /* , ... */;
public struct ReadOnlyMemory<T> : IContiguousBuffer<ReadOnlyMemory<T>, T> /* , ... */;

// one single implementation of RemovePrefix to rule them all
TBuffer MemoryExtensions.RemovePrefix<TBuffer, TElement>(this TBuffer buffer, ReadOnlySpan<TElement> prefix)
    where TBuffer : ref struct, IContiguousBuffer<TBuffer, TElement>
{
    ReadOnlySpan<T> span = buffer.GetSpan();
    if (span.StartsWith(prefix)) { return buffer.Slice(prefix.Length, span.Length - prefix.Length); }
    else { return buffer; }
}
```

This reduces overload explosion within our own APIs. And if some new contiguous buffer type is introduced in the future - let's say Utf8String rises from the ashes and is defined as `public sealed class Utf8String : IContiguousBuffer<Utf8String, byte>` - then the existing extension method will "just work" with no action required from the dev introducing the new buffer type.

IllidanS4:
`where T : ref struct` looks kinda weird. It would feel strange and a bit lengthy to use `where T : ref struct, struct` in constraints.

bernd5:
Why you would use `where T : ref struct, struct` - `ref struct` implies `struct`...?

IllidanS4:
Why should `where T : ref struct` imply `struct`? If this is an "anti-constraint" it should give you an option to use *all* types, classes and ref structs together, and `where T : ref struct` should just give you the possibility to use ref structs alongside what would be normally allowed. Or am I mistaken?

HaloFour:
@IllidanS4 

> Why should `where T : ref struct` imply `struct`? If this is an "anti-constraint" it should give you an option to use all types, classes and ref structs together, and `where T : ref struct` should just give you the possibility to use ref structs alongside what would be normally allowed. Or am I mistaken?

I would expect that such a generic type constraint would limit generic type arguments to only allow `ref struct`, which would be a subset of `struct` making a `struct` constraint redundant.

IllidanS4:
But why have such a limitation? There is nothing `struct`s cannot do that `ref struct`s can, it's all just a matter of that the actual generic type/method is allowed to do. What use-cases does actually restricting to only `ref struct`s have?

It'd make sense to me to have a `where T : ref` to *allow* a `ref struct` where previously one couldn't, but *in addition* to the existing constraints (that's what I understand by "anti-constraint"). This way you could actually add `where T : ref` to existing parameters without breaking the API, and possibly have things like `Span<Span<T>>`.

HaloFour:
@IllidanS4 

I guess that'd be an option.  The proposal does mean being unsatisfied by the syntax, and does explicitly mention the combination of `struct` and `ref struct`.  But if the constraint is to describe that the type/method only handles the generic type argument in a way that is safe for `ref` types that doesn't preclude other generic type arguments.

333fred:
@IllidanS4 brings up a key concern with this proposal, that `where T : ref struct` is very specifically _not_ a constraint. It is, in fact, a loosening, allowing more types to be substituted for T by the caller, and thereby restricting what the callee can do with that T. This is something LDM will need to think about before proceeding with any syntax here.

IllidanS4:
@HaloFour True; it remains to be seen whether there is some safety concern non-`ref` types would *not* provide to warrant restricting to them as well. After all, you can very easily circumvent that:
```cs
ref struct Wrapper<T>
{
    public T value;
}
```
This is something that cannot be done for other constraints, without changing the semantics.

jaredpar:
> where T : ref struct looks kinda weird. It would feel strange and a bit lengthy to use where T : ref struct, struct in constraints.

I agree. This proposal is a bit behind my current thinking on the topic. In discord and email I've been leaning more towards `allow T : ref struct`. That makes it a lot clearer that it's an anti-constraint vs. constraint. 

> Why should where T : ref struct imply struct? 

It specifically should not imply `struct`. This is yet another reason I now prefer `allow T : ref struct`.

PR #5781, opened 2022-02-15
Update pattern-variables.md

alrz:
Prototype: https://github.com/dotnet/roslyn/pull/59702
Rendered: [pattern-variables.md](https://github.com/alrz/csharplang/blob/patch-12/proposals/pattern-variables.md)

### DIFF
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -16,7 +16,6 @@ Expr Simplify(Expr e)
   switch (e) {
     case Mult(Const(1), var x):
     case Mult(var x, Const(1)): 
-        return Simplify(x);
     case Add(Const(0), var x):
     case Add(var x, Const(0)):
         return Simplify(x);
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -8,8 +8,8 @@ Allow variable declarations under `or` patterns and across `case` labels in a `s
 
 This feature would reduce code duplication where we could use the same piece of code if either of patterns is satisfied. For instance:
 ```cs
-if (e is Foo foo or Wrapper { P: Foo foo })
-    return foo;
+if (e is C c or Wrapper { Prop: C c })
+    return c;
   
 Expr Simplify(Expr e)
 {
@@ -26,10 +26,10 @@ Expr Simplify(Expr e)
 Instead of:
 
 ```cs
-if (e is Foo foo1) 
-    return foo1;
-if (e is Wrapper { P: Foo foo2 }) 
-    return foo2;
+if (e is C c1) 
+    return c1;
+if (e is Wrapper { Prop: C c2 }) 
+    return c2;
   
 Expr Simplify(Expr e)
 {
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -2,7 +2,7 @@
 
 ## Summary  
 
-Allow variable declarations under `or` patterns and across `case` labels in a `switch` section.
+Allow variable declarations under `or` and `not` patterns and across `case` labels in a `switch` section.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


alrz:
@jcouv Does this address your comment https://github.com/dotnet/csharplang/pull/4592#discussion_r721676260?

alrz:
>`Could we relax the redeclaration requirement in a switch section?`

Note: I think that could be done if we enforce the redeclaration in terms of definite-assignment instead of relying on a declarator. This way the switch example works as presented. That would also result in no error for something like `var x or 1`, but `x` will be left unassigned. I'll reflect this in a separate PR later.

333fred (approved):

333fred:
> I think that could be done if we enforce the redeclaration in terms of definite-assignment instead of relying on a declarator. 

I had a similar thought earlier, I think it's a good change that would be well received by LDM.

alrz:
I went ahead and made the change in this PR. @333fred @jcouv please take a look, thanks.

333fred:
This all seems much more complicated than I was envisioning for rules, particularly the in-pattern ones. What I was envisioning was simply: if a variable is not currently definitely assigned, it may be redeclared in a pattern. This would permit `a is T t || b is T t` because `t` is not definitely assigned in the false case, but would not permit `_ = a is T t; _ = b is T t;`

alrz:
>`a is T t || b is T t`

If I recall correctly, LDM didn't support redeclaration across expression boundaries.  I can readd the open question on that.
>This all seems much more complicated than I was envisioning for rules

I only reverted the initial PR and made some changes (this currently mirrors the prototype). 
I'm not sure how this could be specced in simpler terms while maintaining the same behavior. Suggestions?

333fred:
> If I recall correctly, LDM didn't support redeclaration across expression boundaries. I can readd the open question on that.

Right, we were leary of the complex rules involved and concerned about double assignments. I think that my proposed simplification, where it's just "you can assign if it's not definitely assigned", is much simpler and alleviates the double assignment concerns.

> I'm not sure how this could be specced in simpler terms while maintaining the same behavior. Suggestions?

My main thought is to simply remove all the things about where variables are allowed to be redeclared in patterns, and simply state that variables are allowed to be redeclared by patterns if they are not definitely assigned and the types all line up. We might have to define what definite assignment means _inside_ a pattern, but I think that's much more straightforward than the existing wording of the rules around where in patterns variables are allowed to be redeclared.

alrz:
Just to make sure we're on the same page, "you can assign if it's not definitely assigned" means pattern variables can be effectively redeclared *anywhere* within their scope; permitting `if (a is C c) ; else if (b is C c) ;` for example.

For binary patterns, we need to somehow flow the def assignment state from LHS so that the variables are considered assigned in the RHS (similar to expressions) to catch both `not (0, var x) or not (var x, 0)` or `(0, var x) and (var x, 0)` as errors. 

That doesn't prevent `_ = a is T t; _ = b is T t;` though, `t` could be unassigned when we enter the second statement.
Same for `case not var t: case var _ when e is var t:` because we've relaxed the redeclaration rule in the entire scope.

333fred:
> Just to make sure we're on the same page, "you can assign if it's not definitely assigned" means pattern variables can be effectively redeclared _anywhere_ within their scope; permitting `if (a is C c) ; else if (b is C c) ;` for example.

Yes.

> For binary patterns, we need to somehow flow the def assignment state from LHS so that the variables are considered assigned in the RHS (similar to expressions) to catch both `not (0, var x) or not (var x, 0)` or `(0, var x) and (var x, 0)` as errors.

Yes

> That doesn't prevent `_ = a is T t; _ = b is T t;` though, `t` could be unassigned when we enter the second statement. Same for `case not var t: case var _ when e is var t:` because we've relaxed the redeclaration rule in the entire scope.

Indeed. I don't especially care about that scenario: you can't use `t` anyway.

proposals/pattern-variables.md
```diff
@@ -1,73 +1,96 @@
...
-    case (int x, 0) a when Use(x, a): // ok
-    case (0, int x) b when Use(x, b): // ok
-        Use(x); // ok
-        Use(a); // error; not definitely assigned
-        Use(b); // error; not definitely assigned
-        break;
-    ```
+### Variable redeclaration
+
+- Pattern variables are allowed to be redeclared within their scope if not already definitely assigned. These names can possibly reference either of variables based on the result of the pattern-matching at runtime.
```
alrz, line 58: I think this should not include nested scoped, e.g.
```cs
if (e is C c) return;
if (e is C c) return; // allowed
{ if (e is C c) return; } // not allowed
```
Not completely sure if "within their scope" implies that.

333fred: Perhaps `in the same scope`?

333fred: It also would probably be worth moving it to an open questions section.

333fred (commented):
I like this version a lot more :). Just a couple small comments.

proposals/pattern-variables.md
alrz: There's a special error for that (CS0136) which I believe shouldn't be relaxed - in interaction with loops it gets too confusing. I think we should draw the line *somewhere* as the current rules are already too loose. Would like to see how all this play out.

alrz: @333fred Done. Any chance this could be scheduled for ldm review? Thanks.

333fred (approved):
@jcouv please take a look

jcouv:
Good to merge.
@alrz We were planning to review this morning, but ended up not having time. Likely will get discussed next week.

PR #5815, opened 2022-02-22
Replace spec content with links

BillWagner:
Replace the content of the spec folder with a list of anchored links to the ECMA draft spec.

Fixes dotnet/docs#27789 (final task).

Reviewing may be easier by [looking directly at the files in the source branch](https://github.com/BillWagner/csharplang/tree/remove-old-spec/spec). Or I'd suggest using the rich diff, and concentrating on the added (new) content. 

Question for reviewers: Should we create a tag before this PR gets merged to make it easier to go back to a point in time where the spec source was present in this repository?


BillWagner:
I did one final pass for any links to the existing spec content, and I didn't find any remaining.

jcouv (approved):

spec/attributes.md
```diff
@@ -1,846 +1,29 @@
-﻿# Attributes
+# Attributes
+
+This content has moved to the [`dotnet/csharpstandard`](https://github.com/dotnet/csharpstandard) repository.
+The list below provides links to each heading in this section. The links specify the C# 6 branch, which is version when the specifications merged.
+
+- <a id="attributes"></a>[Attributes](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/attributes.md#21-attributes)
```
jcouv, line 6: I'm a bit curious why we put the new docs into "draft-v6" folder. Why not just "v6" or no subfolder? Having "draft" means we'll have to fix a bunch of links once the docs are no longer draft.

BillWagner: @jcouv "draft-v6" is the branch name. When we submit to ECMA, we'll change the name to "standard-v6", and GitHub will automatically update the links.

333fred (approved):

BillWagner:
I added the tag `ms-spec-text` and added a link to that tag in each of the files where the text is removed. That will help us find the MS spec text if needed later.

PR #5817, opened 2022-02-23
Add main attributes proposal

chsienki:


proposals/global-attributes-for-main.md
```diff
@@ -0,0 +1,116 @@
...
+
+It is an error to use the `main` target location for an attribute in an assembly
+with no entry point (e.g. a class library).
+
+### Attribute locations
+
+There is no new
+[`AttributeTargets`](https://docs.microsoft.com/en-us/dotnet/api/system.attributetargets?view=net-6.0)
+enum value defined as part of this proposal, and it is an error to prefix an
+attribute with `[main:` that has any target other than `All` or `Method`.
```
333fred, line 50: The wording here implies that `AttributeTargets.Property | AttributeTargets.Method` cannot be put in `main:`. Consider rewording to state that the attribute must have a `Method` or `All` target, rather than talking about the things the attribute _doesn't_ have.

Side note: what does a missing `AttributeTargets` translate to? `All`?

chsienki: Good point. Yes, missing AttributeTargets implies `All`.

proposals/global-attributes-for-main.md
```diff
@@ -33,21 +33,23 @@ For example:
 Can be placed in any syntax tree of the compilation, and the `[STAThread]`
 attribute will be attached the entrypoint of the program.
 
-Both top level statements and regular entrypoints are supported by the attribute
-target. In the case of multiple entrypoints the attribute will only apply to the
-first chosen location.
+Only simple program entrypoints are supported. In the case of multiple
+entrypoints the attribute will only apply to the first chosen location.
```
333fred, line 37: There are not multiple entry points for programs with top level statements.

333fred: > Although its an error case, the compiler explicitly allows you to have multiple files with top level statements

It's an error case. The compiler explicitly does not allow it. Therefore, our handling of such an error case is an implementation detail, and doesn't belong in the specification.

jaredpar: I think we could simplify this to essentially 

> Only simple program entrypoints are supported. It is not supported when there is an explicit `Main` method which qualifies as the entry point. 

I agree we should avoid listing out how it works in cases like multiple files with top level statements. That is already an error scenario, basically an invalid program. Generally we don't spec out behaviors for scenarios like that.

PR #5828, opened 2022-02-25
Remove incorrect `override` keywords

333fred:
Fixes https://github.com/dotnet/docs/issues/26115. @AlekseyTs for review.

### DIFF
--- a/proposals/csharp-8.0/default-interface-methods.md
+++ b/proposals/csharp-8.0/default-interface-methods.md
@@ -109,11 +109,11 @@ interface IA
 }
 interface IB : IA
 {
-    override void IA.M() { WriteLine("IB.M"); } // explicitly named
+    void IA.M() { WriteLine("IB.M"); } // explicitly named
 }
 interface IC : IA
 {
-    override void M() { WriteLine("IC.M"); } // implicitly named
+    void M() { WriteLine("IC.M"); } // implicitly named
 }
 ```
 
--- a/proposals/csharp-8.0/default-interface-methods.md
+++ b/proposals/csharp-8.0/default-interface-methods.md
@@ -29,7 +29,7 @@ The syntax for an interface is extended to permit
 - member declarations using the explicit interface implementation syntax; and
 - Explicit access modifiers (the default access is `public`).
 
-Members with bodies permit the interface to provide a "default" implementation for the method in classes and structs that do not provide an overriding implementation.
+Members with bodies permit the interface to provide a "default" implementation for the method in classes and structs that do not provide their own implementation.
 
 Interfaces may not contain instance state. While static fields are now permitted, instance fields are not permitted in interfaces. Instance auto-properties are not supported in interfaces, as they would implicitly declare a hidden field.
 
@@ -119,9 +119,7 @@ interface IC : IA
 
 Explicit implementations in interfaces may not be declared `sealed`.
 
-Public `virtual` function members in an interface may be implemented in a derived interface explicitly (by qualifying the name in the override declaration with the interface type that originally declared the method, and omitting an access modifier).
-
-`virtual` function members in an interface may only be implemented explicitly (not implicitly) in derived interfaces, and members that are not `public` may only be implemented in a class or struct explicitly (not implicitly). In either case, the overridden or implemented member must be *accessible* where it is overridden.
+Public `virtual` function members in an interface may only be implemented in a derived interface explicitly (by qualifying the name in the declaration with the interface type that originally declared the method, and omitting an access modifier). The member must be *accessible* where it is implemented.
 
 ### Reabstraction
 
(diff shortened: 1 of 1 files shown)


AlekseyTs:
> ### Overrides in interfaces

Probably all references to "overrides" should be adjusted, not just in code examples.

---
Refers to: proposals/csharp-8.0/default-interface-methods.md:101 in a3cb067. [](commit_id = a3cb067b31a0fecc794a8a0a328b9c6ff537fae1, deletion_comment = False)

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -109,11 +109,11 @@ interface IA
 }
 interface IB : IA
 {
-    override void IA.M() { WriteLine("IB.M"); } // explicitly named
+    void IA.M() { WriteLine("IB.M"); } // explicitly named
 }
 interface IC : IA
 {
-    override void M() { WriteLine("IC.M"); } // implicitly named
+    void M() { WriteLine("IC.M"); } // implicitly named
```
AlekseyTs, line 116: >void M() { WriteLine("IC.M"); } // implicitly named [](http://example.com/codeflow?start=4&length=51)

This declares a new method in IC that has no relation to IA.M and doesn't implement it. I think the original example meant to express syntax for implicit implementation. There is no syntax that can be used for an implicit implementation in a derived interface. It looks like we can simply remove the IC interface from this example.

AlekseyTs:
> ### Overrides in interfaces

Interface methods cannot be overridden, they can only be implemented.

---
In reply to: [1068534052](https://github.com/dotnet/csharplang/pull/5828#issuecomment-1068534052) [](http://example.com/codeflow?ancestors=1068534052)

---
Refers to: proposals/csharp-8.0/default-interface-methods.md:101 in a3cb067. [](commit_id = a3cb067b31a0fecc794a8a0a328b9c6ff537fae1, deletion_comment = False)

AlekseyTs:

333fred:
@AlekseyTs please take another look.

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -109,23 +109,23 @@ interface IA
...
 
-Public `virtual` function members in an interface may be overridden in a derived interface explicitly (by qualifying the name in the override declaration with the interface type that originally declared the method, and omitting an access modifier).
+Public `virtual` function members in an interface may be implemented in a derived interface explicitly (by qualifying the name in the override declaration with the interface type that originally declared the method, and omitting an access modifier).
 
-`virtual` function members in an interface may only be overridden explicitly (not implicitly) in derived interfaces, and members that are not `public` may only be implemented in a class or struct explicitly (not implicitly). In either case, the overridden or implemented member must be *accessible* where it is overridden.
+`virtual` function members in an interface may only be implemented explicitly (not implicitly) in derived interfaces, and members that are not `public` may only be implemented in a class or struct explicitly (not implicitly). In either case, the overridden or implemented member must be *accessible* where it is overridden.
```
AlekseyTs, line 124: It looks like there is still a reference to "overridden" term in this paragraph.

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -109,23 +109,23 @@ interface IA
...
-    override void M() { WriteLine("IC.M"); } // implicitly named
+    void M() { WriteLine("IB.M"); } // Creates a new M, unrelated to `IA.M`. Warning
 }
 ```
 
-Override declarations in interfaces may not be declared `sealed`.
+Explicit implementations in interfaces may not be declared `sealed`.
 
-Public `virtual` function members in an interface may be overridden in a derived interface explicitly (by qualifying the name in the override declaration with the interface type that originally declared the method, and omitting an access modifier).
+Public `virtual` function members in an interface may be implemented in a derived interface explicitly (by qualifying the name in the override declaration with the interface type that originally declared the method, and omitting an access modifier).
```
AlekseyTs, line 122: >override declaration [](http://example.com/codeflow?start=134&length=20)

Consider rephrasing as well

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -189,9 +189,9 @@ abstract class E : IA, IB, IC // ok
 }
 ```
 
-> ***Open issue***: should we support explicit interface abstract overrides in classes?
+> ***Open issue***: should we support explicit interface abstract implementations in classes?
 
-In addition, it is an error if in a class declaration the most specific override of some interface method is an abstract override that was declared in an interface. This is an existing rule restated using the new terminology.
+In addition, it is an error if in a class declaration the most specific override of some interface method is an abstract implementation that was declared in an interface. This is an existing rule restated using the new terminology.
```
AlekseyTs, line 194: >the most specific override [](http://example.com/codeflow?start=54&length=26)

Consider rephrasing

AlekseyTs (approved):
LGTM (commit 2)

333fred:
@AlekseyTs could you take another look?

Chrobob-777:
can-ram-all-devices.pilots.jets

Chrobob-777 (commented):
can_view/all_devices.pilots.jets[System](system.pilots.jets)

PR #5829, opened 2022-02-25
Leave type parameters untouched by 'nameof(parameter)' feature

jcouv:
FYI @333fred 
I don't see benefit of allowing `[Attribute(nameof(TypeParameter))] void M<TypeParameter>() { }` so removing from the spec, since that's not the point of the feature. Will confirm with LDM when we review this speclet.

Relates to https://github.com/dotnet/roslyn/issues/40524

### DIFF
--- a/proposals/extended-nameof-scope.md
+++ b/proposals/extended-nameof-scope.md
@@ -14,14 +14,10 @@ Attributes like `NotNullWhen` or `CallerExpression` need to refer to parameters,
 
 ## Detailed design
 
-[Methods](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#methods)
-
-The method's *type_parameters* are in scope throughout the *method_declaration*, and can be used to form types throughout that scope in *return_type*, *method_body*, and *type_parameter_constraints_clauses* but not in *attributes*, **except within a `nameof` expression in *attributes*.**
-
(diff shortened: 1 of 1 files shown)


RikkiGibson:
I could see certain implementations of 'ref struct constraint' needing to refer to type parameters in attributes, for example, if they needed to do something like:

```cs
public ref struct Span<T> where T : ref struct
{
    [TypeParameterIsNotRefLike(nameof(T)]
    public Span(T[] arr) { ... }
}
```

But I don't consider it likely that we would arrive at that specific solution.

333fred:
> I don't see benefit of allowing [Attribute(nameof(TypeParameter))] void M<TypeParameter>() { } so removing from the spec, since that's not the point of the feature. 

I don't see the benefit in removing it :slightly_smiling_face:

jcouv:
It's more implementation and testing work. The motivating scenarios so far have been NRT attributes (none of which need type parameters).
It'd be great if we could identify a motivating scenario. 
I'll leave this as open question for LDM.

hez2010:
This is potentially useful for code analyzers.

jcouv:
@hez2010 Please elaborate

jcouv:
From email thread with LDM, let's keep type parameters in scope for this proposal.

PR #5838, opened 2022-02-27
Ref fields updates

jaredpar:
This is a few updates to the `ref` fields considerations discussions intended to facilitate LDM discussions this week.

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -780,13 +780,6 @@ The backing storage for the buffer will be generated using the `[InlineArray]` a
 
 ## Considerations
 
-### Keywords vs. attributes
-This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance: `scoped` and `escapes` could have been used instead of `DoesNotEscape` and `RefThisEscapes`.
-
-Keywords, even the contextual ones, have a much heavier weight in the language than attributes. The use cases these features solve, while very valuable, impact a small number of developers. Consider that only a fraction of high end developers are defining `ref struct` instances and then consider that only a fraction of those developers will be using these new lifetime features.  That doesn't seem to justify adding a new contextual keyword to the language.
-
-This does mean that program correctness will be defined in terms of attributes though. That is a bit of a gray area for the language side of things but an established pattern for the runtime. 
-
 ### Why do we need [RefFieldEscapes]?
 The biggest challenge posed by the [compat considerations](#compat-considerations) is that methods cannot capture and return `ref` parameters as `ref` fields. This is a hard assumption in the rules and there are many API patterns today that take advantage of this. In order to have methods that capture `ref` parameters as `ref` fields there must be some form of explicit opt-in that is visible to calling methods.
 
@@ -823,6 +816,14 @@ ref struct S<T> {
 ```
 
 ## Open Issues
+
+### Keywords vs. attributes
+This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance: `scoped` and `escapes` could have been used instead of `DoesNotEscape` and `RefThisEscapes`.
+
+Keywords, even the contextual ones, have a much heavier weight in the language than attributes. The use cases these features solve, while very valuable, impact a small number of developers. Consider that only a fraction of high end developers are defining `ref struct` instances and then consider that only a fraction of those developers will be using these new lifetime features.  That doesn't seem to justify adding a new contextual keyword to the language.
+
+This does mean that program correctness will be defined in terms of attributes though. That is a bit of a gray area for the language side of things but an established pattern for the runtime. 
+
 
 ### Take the breaking change
 Consider for a minute a design where the compat problem was approached from the other direction. Effectively make it such that every method was implicitly `[RefFieldsEscape]` and then have an attribute that restores the span safety rules in place today. Say `[RefFieldDoesNotEscape]`. 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -818,26 +818,71 @@ ref struct S<T> {
 ## Open Issues
 
 ### Keywords vs. attributes
-This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance: `scoped` and `escapes` could have been used instead of `DoesNotEscape` and `RefThisEscapes`.
+This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance `[DoesNotEscape]` could map to `scoped`
 
-Keywords, even the contextual ones, have a much heavier weight in the language than attributes. The use cases these features solve, while very valuable, impact a small number of developers. Consider that only a fraction of high end developers are defining `ref struct` instances and then consider that only a fraction of those developers will be using these new lifetime features.  That doesn't seem to justify adding a new contextual keyword to the language.
+Keywords, even the contextual ones, generally must meet a very high bar for inclusion. They take up valuable langauge real estate and are more prominent parts of the langauge.
 
-This does mean that program correctness will be defined in terms of attributes though. That is a bit of a gray area for the language side of things but an established pattern for the runtime. 
+This feature, while valuable, is going to serve a minority of C# developers. On the surface that would seem to favor not using keywords but there are two important points to consider: 
 
+1. The annotations will effect program semantics. This is a line C# is reluctant to cross and it's unlikley this is the feature that should justify it. 
+1. The developers most likely to use this feature intesect strongly with the set of developers that use function pointers. That feature, while also used by a minority of developers, did warrant a new syntax and that decision is still seen as sound. 
+
+As such syntax should be considered here. 
+
+A rough sketch of the syntax would be: 
+
+- `[DoesNotEscape]` maps to `scoped` 
+- `[RefThisEscapes]` maps to `escapes`
+- `[RefFieldsEscapes]` maps to `escapes` but attached to the `ref` of the method.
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -824,19 +817,69 @@ ref struct S<T> {
 
 ## Open Issues
 
+### Keywords vs. attributes
+This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance `[DoesNotEscape]` could map to `scoped`. However keywords, even the contextual ones, generally must meet a very high bar for inclusion. They take up valuable langauge real estate and are more prominent parts of the langauge. This feature, while valuable, is going to serve a minority of C# developers.
```
cston, line 821:
```suggestion
This design calls for using attributes to annotate the new lifetime rules. This also could've been done just as easily with contextual keywords. For instance `[DoesNotEscape]` could map to `scoped`. However keywords, even the contextual ones, generally must meet a very high bar for inclusion. They take up valuable language real estate and are more prominent parts of the language. This feature, while valuable, is going to serve a minority of C# developers.
```

proposals/low-level-struct-improvements.md
```diff
@@ -824,19 +817,69 @@ ref struct S<T> {
...
+
+On the surface that would seem to favor not using keywords but there are two important points to consider: 
+
+1. The annotations will effect program semantics. Having attributes impact program semantics is a line C# is reluctant to cross and it's unclear if this is the feature that should justify the language taking that step.
+1. The developers most likely to use this feature intesect strongly with the set of developers that use function pointers. That feature, while also used by a minority of developers, did warrant a new syntax and that decision is still seen as sound.
```
cston, line 826:
```suggestion
1. The developers most likely to use this feature intersect strongly with the set of developers that use function pointers. That feature, while also used by a minority of developers, did warrant a new syntax and that decision is still seen as sound. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -824,19 +817,69 @@ ref struct S<T> {
...
+1. The annotations will effect program semantics. Having attributes impact program semantics is a line C# is reluctant to cross and it's unclear if this is the feature that should justify the language taking that step.
+1. The developers most likely to use this feature intesect strongly with the set of developers that use function pointers. That feature, while also used by a minority of developers, did warrant a new syntax and that decision is still seen as sound. 
+
+Taken together this means syntax should be considered.
+
+A rough sketch of the syntax would be: 
+
+- `[DoesNotEscape]` maps to `scoped` 
+- `[RefThisEscapes]` maps to `escapes`
+- `[RefFieldsEscapes]` maps to `escapes` but attached to the `ref` of the method.
```
cston, line 834: We should consider attaching the attribute or keyword to the parameter rather than the method, to allow distinguishing between multiple `ref` parameters.

jaredpar: Agree. I was hesistant before because I thought it was overkill. Subsequent discussions though lead me to think otherwise. It's one of the changes I'm going to make after LDM today. I didn't want to rewrite the doc though until I had a sense of the syntax we will use (don't want to rewrite twice essentially). This section is rewritten with the assumption it will be on individual parameters

jaredpar:
Merging cause I want to have the doc for review in LDM today. Please feel free to keep adding comments though. I expect to make more edits after LDM and I can incorporate comments into that PR.

proposals/low-level-struct-improvements.md
```diff
@@ -824,19 +817,69 @@ ref struct S<T> {
...
+This change would impact methods that have the following properties:
+
+- Have a `Span<T>` or `ref struct`
+    - Where the `ref struct` is a return type, `ref` or `out` parameter
+    - Has an additional `in` or `ref` parameter
+
+To understand the impact it's helpful to break APIs into categories:
+
+1. Want consumers to account for `ref` being captured as a `ref` field. Prime example is the `Span(ref T value)` constructors 
+2. Do not want consumers to account for `ref` being captured as a `ref` field. These though break into two catogries
```
cston, line 868: >catogries [](http://example.com/codeflow?start=107&length=9)

`categories:`

proposals/low-level-struct-improvements.md
```diff
@@ -824,19 +817,69 @@ ref struct S<T> {
...
+    1. Unsafe APIs. These are APIS inside the `Unsafe` and `MemoryMarshal` types, of which `MemoryMarshal.CreateSpan` is the most prominent. These APIs do capture the `ref` unsafely but they are also known to be unsafe APIs. 
+    2. Safe APIs. These are APIs which take `ref` parameters for efficiency but it is not actually captured anywhere. Examples are small but one is `AsnDecoder.ReadEnumeratedBytes`
+
+This change primarily benefits (1) above. These are expected to make up the majority of APIs that take a `ref` and return a `ref struct` going forward. The changes negatively impact (2.1) and (2.2) as it breaks the existing calling semantics because the lifetime rules change. 
+
+The APIs in category (2.1) though are largely authored by Microsoft or by developers who stand the most to benefit from `ref` fields (the Tanner's of the world). It is reasonable to assume this class of developers would be ammenable to a compatability tax on upgrade to C# 11 in the form of a few annotations to retain the existing semantics if `ref` fields were provided in return.
```
cston, line 874: >(the Tanner's of the world) [](http://example.com/codeflow?start=133&length=27)

Consider omitting.

proposals/low-level-struct-improvements.md
```diff
@@ -824,19 +817,69 @@ ref struct S<T> {
...
 
-If this is very low then it's possible that a breaking change could be acceptable here. 
+Ideally the language could reduce the impact of silent breaking changes by issuing a warning when an API silently falls into the troublesome behavior. That would be a method that both takes a `ref`, returns `ref struct` but does not actually capture the `ref` in the `ref struct`. The compiler could issue a diagnostic in that case informing developers such `ref` should be annotated as `[RefFieldsDoNotEscape]` / `scoped`. 
 
-The challenge though is the breaking change and the fix for the breaking change are possible in different libraries. Consider the following:
+The challenge to keep in mind with respect to compat is thi breaking change and the impact of the breaking change can appear in different libraries. Consider the following:
```
cston, line 882: >thi [](http://example.com/codeflow?start=56&length=3)

"the"

proposals/low-level-struct-improvements.md
```diff
@@ -876,11 +931,13 @@ This holds together but does require us to extend the syntax for locals a bit.
...
 
 ### To use modreqs or not
-A decision needs to be made if methods marked with new life time attributes should or should not contain a `modreq`. The target of the `modreq` would simply be the attribute in question. Or there would be a single attribute, `System.Runtime.CompilerServices.LifetimeAnnotations`, for which a `modreq` is added if any of the lifetime attributes are used. 
+A decision needs to be made if methods marked with new lifetime attributes should or should not translate to `modreq` in emit. There would be effectively a 1:1 mapping between annotations and `modreq` if this approach was taken.
+
+The rationale for adding a `modreq` is the attributes change the semantics of span safety. Only languages which understand these semantics should be calling the methods in question. Further when applied to OHI scenarios, the lifetimes become a contract that all derived methods must implement. Having the annotations exist without `modreq` can lead to situations where `virtual` method chains with conflicting lifetime annotations are loaded (can happen if only one part of `virtual` chain is compiled and other is not).
```
cston, line 936: >OHI scenarios [](http://example.com/codeflow?start=206&length=13)

Perhaps "virtual or interface methods" instead, to avoid an acronym.

proposals/low-level-struct-improvements.md
```diff
@@ -876,11 +931,13 @@ This holds together but does require us to extend the syntax for locals a bit.
...
 
-The worry is whether or not this is overkill. No `modreq` were used in the initial span safety work and instead the framework relied on languages understanding and implementing the new rules. In cases like `[RefThisEscapes]` on a `struct` declaration would result in every single member having a `modreq`. It's questionable if that is worth the trade off
+The concern is whether or not this is overkill. It does have the negative impact that making signatures more flexible, by say adding `[DoesNotEscape]` to a paramater, will result in a binary compat change. That trade off means that over time frameworks like BCL likely won't be able to relax such signatures. It could be mitigated to a degree by taking some approach the language does with `in` parameters and only apply `modreq` in virtual positions.
```
cston, line 940: >paramater [](http://example.com/codeflow?start=156&length=9)

"parameter"

proposals/low-level-struct-improvements.md
tannergooding: This made me giggle. 10/10

Notably, in my own case I am amenable to the cost here and it should be possible to configure tooling (such as ClangSharp) to do this "automagically". In the case of my own usages of things like `MemoryMarshal`, they were mainly a workaround for the language missing the  `RefThisEscapes` and related functionality. The intent is to switch to those features once available and make this "safe" instead.

I expect the BCL couldn't/wouldn't update their own APIs and they would remain unsafe. There would just be less reason for users to need them.

One particular area of interest is in how it impacts users multi-targeting the last LTS (`.net 6`) and the latest (`.net 7`) and the experience users would see on downstream compilers.

proposals/low-level-struct-improvements.md
tannergooding, line 940: I think doing it only on virtuals would be good, if possible.

Having a binary break is never great and while I'd still take it, I'd much rather avoid having to do that to the several hundred property helpers I have in my interop library. Source breaks, on the other hand, are fine and I'm completely happy with new warnings showing up for existing users. -- This becomes particularly relevant when there are languages (like F#) that ignore modopt/modreq and so with that consideration, the existing considerations around ref safety, and the "usage numbers" for different languages, it seems like causing a break for something languages already need to be doing the "right" thing on to support an increasingly niche edge case.

PR #5852, opened 2022-03-01
Add docs task to the feature template

BillWagner:
@jcouv @jaredpar @MadsTorgersen @KathleenDollard

### DIFF
--- a/.github/ISSUE_TEMPLATE/proposal_template.md
+++ b/.github/ISSUE_TEMPLATE/proposal_template.md
@@ -14,6 +14,7 @@ New language feature proposals should fully fill out this template. This should 
 * [ ] Prototype: Not Started
 * [ ] Implementation: Not Started
 * [ ] Specification: Not Started
+* [ ] Feature docs: Not Started
 
 ## Summary
 [summary]: #summary


333fred:
To be entirely honest, we don't use the existing checklist items basically ever. I've been meaning to propose we find a different way to track progress.

333fred:
Well, this is part of docs item to better surface C# proposals. If we want to do that, I think we should do a bigger revamp of that progress section.

jcouv (approved):

jcouv:
Sure. We can make further changes. Would be happy to remove the Prototype and Implementation status which we didn't use much (just keep the Spec and Docs link).

BillWagner:
> Well, this is part of docs item to better surface C# proposals. If we want to do that, I think we should do a bigger revamp of that progress section.

The motivation for this was to make it easy for anyone that finds these issues (from SO, or whatever) to have a roadmap to the full story (as completed): the proposal, LDM notes on updates, the PRs for the change, and the docs for the final feature.

333fred:
> The motivation for this was to make it easy for anyone that finds these issues (from SO, or whatever) to have a roadmap to the full story (as completed): the proposal, LDM notes on updates, the PRs for the change, and the docs for the final feature.

Right. My concern is that these status sections are never updated and are not a good way to communicate info currently.

PR #5898, opened 2022-03-08
Clarify binding rules for `await foreach`

jcouv:
Relates to https://github.com/dotnet/roslyn/pull/60022

### DIFF
--- a/proposals/csharp-8.0/async-streams.md
+++ b/proposals/csharp-8.0/async-streams.md
@@ -230,7 +230,7 @@ is then expanded to:
 
 The body of the `finally` block is constructed according to the following steps:
 - If the type `E` has an appropriate `DisposeAsync` method:
-  - Perform member lookup on the type `E` with identifier `DisposeAsync` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for the disposal  interface as described below.
+  - Perform member lookup on the type `E` with identifier `DisposeAsync` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for the disposal interface as described below.
   - Perform overload resolution using the resulting method group and an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, check for the disposal interface as described below.
   - If the return type of the `DisposeAsync` method is not awaitable, an error is produced and no further steps are taken.
   - The `finally` clause is expanded to the semantic equivalent of:
--- a/proposals/csharp-8.0/async-streams.md
+++ b/proposals/csharp-8.0/async-streams.md
@@ -189,7 +189,7 @@ No syntax would be provided that would support using either the async or the syn
 
 ### Semantics
 
-The compile-time processing of an async `foreach` statement first determines the ***collection type***, ***enumerator type*** and ***iteration type*** of the expression (very similar to https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/statements.md#1295-the-foreach-statement). This determination proceeds as follows:
+The compile-time processing of an `await foreach` statement first determines the ***collection type***, ***enumerator type*** and ***iteration type*** of the expression (very similar to https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/statements.md#1295-the-foreach-statement). This determination proceeds as follows:
 
 - If the type `X` of *expression* is `dynamic` or an array type, then an error is produced an no further steps are taken.
 - Otherwise, determine whether the type `X` has an appropriate `GetAsyncEnumerator` method:
@@ -254,7 +254,7 @@ The body of the `finally` block is constructed according to the following steps:
     }
     ```
     except that if `E` is a value type, or a type parameter instantiated to a value type, then the conversion of `e` to `System.IAsyncDisposable` shall not cause boxing to occur.
-- Otherwise, if `E` is a sealed type, the `finally` clause is expanded to an empty block:
+- Otherwise, the `finally` clause is expanded to an empty block:
   ```csharp
   finally {
   }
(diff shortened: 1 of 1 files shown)


333fred (approved):

proposals/csharp-8.0/async-streams.md
```diff
@@ -187,52 +187,78 @@ await foreach (var i in enumerable)
...
 
-The compiler will bind to the pattern-based APIs if they exist, preferring those over using the interface (the pattern may be satisfied with instance methods or extension methods).  The requirements for the pattern are:
-- The enumerable must expose a `GetAsyncEnumerator` method that may be called with no arguments and that returns an enumerator that meets the relevant pattern.
-- The enumerator must expose a `MoveNextAsync` method that may be called with no arguments and that returns something which may be `await`ed and whose `GetResult()` returns a `bool`.
-- The enumerator must also expose `Current` property whose getter returns a `T` representing the kind of data being enumerated.
-- The enumerator may optionally expose a `DisposeAsync` method that may be invoked with no arguments and that returns something that can be `await`ed and whose `GetResult()` returns `void`.
+- If the type `X` of *expression* is `dynamic` or an array type, then an error is produced an no further steps are taken.
```
huanlin, line 194: ...is produced "**and**" no further steps....

proposals/csharp-8.0/async-streams.md
```diff
@@ -187,52 +187,78 @@ await foreach (var i in enumerable)
...
-- The enumerator must also expose `Current` property whose getter returns a `T` representing the kind of data being enumerated.
-- The enumerator may optionally expose a `DisposeAsync` method that may be invoked with no arguments and that returns something that can be `await`ed and whose `GetResult()` returns `void`.
+- If the type `X` of *expression* is `dynamic` or an array type, then an error is produced and no further steps are taken.
+- Otherwise, determine whether the type `X` has an appropriate `GetAsyncEnumerator` method:
+  - Perform member lookup on the type `X` with identifier `GetAsyncEnumerator` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for an enumerable interface as described below.
+  - Perform overload resolution using the resulting method group and an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, check for an enumerable interface as described below.
```
cston, line 197: Should a warning be reported if the match is not a method group or is ambiguous, similar to the recommendation for `foreach`?

proposals/csharp-8.0/async-streams.md
```diff
@@ -187,52 +187,78 @@ await foreach (var i in enumerable)
...
-finally
-{
-    await enumerator.DisposeAsync(); // omitted, along with the try/finally, if the enumerator doesn't expose DisposeAsync
 }
 ```
 
-If the iterated type doesn't expose the right pattern, the interfaces will be used.
+The body of the `finally` block is constructed according to the following steps:
+- If the type `E` has an appropriate `DisposeAsync` method:
+  - Perform member lookup on the type `E` with identifier `DisposeAsync` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for the disposal  interface as described below.
```
cston, line 233: >disposal  interface [](http://example.com/codeflow?start=234&length=19)

Extra space between words. And should this be "disposable interface" (here and below) instead?

jcouv: I'm trying to go for parallel with "enumeration interfaces".
"disposal" and "enumeration" are nouns, but "disposable" and "enumerable" are adjectives?

proposals/csharp-8.0/async-streams.md
```diff
@@ -187,52 +187,78 @@ await foreach (var i in enumerable)
...
+  - Perform overload resolution using the resulting method group and an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, check for the disposal interface as described below.
```
cston, line 234: >results in an ambiguity, or results in a single best method but that method is either static or not public [](http://example.com/codeflow?start=150&length=106)

Report a warning?

proposals/csharp-8.0/async-streams.md
```diff
@@ -187,52 +187,78 @@ await foreach (var i in enumerable)
...
+    ```
+  - Otherwise the `finally` clause is expanded to the semantic equivalent of:
+    ```csharp
+    finally {
+        System.IAsyncDisposable d = e as System.IAsyncDisposable;
+        if (d != null) await d.DisposeAsync();
+    }
+    ```
+    except that if `E` is a value type, or a type parameter instantiated to a value type, then the conversion of `e` to `System.IAsyncDisposable` shall not cause boxing to occur.
+- Otherwise, the `finally` clause is expanded to an empty block:
```
cston, line 257: If `E` is not a sealed type, should we generate the following, matching `foreach`?
```csharp
        System.IAsyncDisposable d = e as System.IAsyncDisposable;
        if (d != null) await d.DisposeAsync();
```

proposals/csharp-8.0/async-streams.md
jcouv: LDM said that we don't usually spec warnings, so that's why I didn't include this.

proposals/csharp-8.0/async-streams.md
jcouv: I don't think so. I don't remember why we'd decided that, but the code has:
`// For async foreach, we don't do the runtime check`

```
             if ((!enumeratorType.IsSealed && !isAsync) ||
                    this.Conversions.ClassifyImplicitConversionFromType(enumeratorType,
                        isAsync ? this.Compilation.GetWellKnownType(WellKnownType.System_IAsyncDisposable) : this.Compilation.GetSpecialType(SpecialType.System_IDisposable),
                        ref useSiteInfo).IsImplicit)
                {
                    builder.NeedsDisposal = true;
                }
```

cston (approved):

PR #5899, opened 2022-03-08
Update ref field proposal

jaredpar:
This update has two main parts:

- Move `[RefFieldsEscape]` from annotating methods to annotating the actual `ref` parameter that escapes. Initially this was done at the method level to remove some complexity from the rules. In reality though it's the individual parameters that matter and the correct decision is to annotate them. That gives the maximum flexibility to API designers.
- Significantly expand on the **Breaking Compatibility** section. In several one off discussions and [C# LDM meetings](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-02-28.md#conclusion-2) there was significant support for limited compat breaks to simplify this feature. Correspondingly this section was updated to have a more detailed design of what a compat break would look like and how it would impact the ref safety rules.

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -136,18 +136,23 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
 
 <a name="ref-fields-escape"></a>
 
-To accomplish this the span safety rules for method invocation will be changed. The method invocation rules presently state:
+To accomplish this two new annotations will be introduced to help control how arguments influence lifetime of method calls: `[RefFieldsEscape]` and `[DoesNotEscape]`. The annotation `[RefFieldsEscape]` when applied to a `ref` parameter signifies that it can be captured as a `ref` field in a returned `ref struct`.
+
+The `[DoesNotEscape]` annotation is not necessary for `ref` fields but more for reducing friction when dealing with specific classes of methods. It will be discussed in detail [later on](#does-not-escape) but needs to be introduced now as it impacts method invocation rules.
+
+To recognize these annotations the span safety rules for method invocation need to be updated. The first change is recognizing the impact the annotations have on arguments. For a given argument `a` that is passed to `ref` parameter `p`: 
+
+> 1. If `p` is `[RefFieldsEscape] ref` or `[RefFieldsEscape] in` and `a` is 
+>     1. A `[RefFieldEscapes]` parameter it contrbutes *safe-to-escape* to *calling method*
+>     2. A known heap location it contrbutes *safe-to-escape* to *calling method*
+>     3. It contributes *safe-to-escape* to *current method*
+> 2. If `p` is `[DoesNotEscape]` then `a` does not contribute *safe-to-escape* when considering arguments.
+
+This change allows us to keep our existng rules for rvalue method invocation:
 
 > An rvalue resulting from a method invocation e1.M(e2, ...) is *safe-to-escape* from the smallest of the following scopes:
-> 1. The entire enclosing method
-> 2. The *safe-to-escape* of all argument expressions (including the receiver)
-
-The change will be in how the *safe-to-escape* scope is calculated for arguments of method invocation when the return type is a ref-like type. The updated rule will read as follows:
-
-> When the rvalue is a ref-like type then *safe-to-escape* scope of arguments will be changed when they line up with a parameter marked as `[RefFieldEscapes]`
-> 1. When the argument is itself a parameter marked `[RefFieldEscapes]` it will be *calling method*
-> 2. When the argument is a known heap location it will be *calling method*
-> 3. Else it will be *current method*
+> * The entire enclosing method
+> * The *safe-to-escape* of all argument expressions (including the receiver)
 
 The rules for `ref` field assignment will be discussed in detail [later on](#ref-field-reassignment) the doc. For now readers only need to know constructors can initialize `ref` fields with parameters marked as `[RefFieldEscape]` or known heap locations.
 
@@ -184,28 +189,23 @@ ref struct RS
     static RS CreateRS([RefFieldEscapes] ref int p1, ref int p2)
     {
         // Okay: The RS argument `p1' lines up with a [RefFieldsEscape] argument
-        // but it is also a [RefFieldsEscape] parameter hence the return of the constructor 
-        // is safe-to-escape to the calling method (rule 1 above)
+        // but it is also a [RefFieldsEscape] parameter hence it contributes a safe-to-escape of 
+        // calling method (rule 1.1 above)
         return new RS(ref p1);
 
         // Okay: This is a known heap value hence it is safe-to-escape to the calling
-        // method (rule 2 above)
+        // method (rule 1.2 above)
         return new RS(new int[1]);
 
         // Error: This is the same analysis as above but in this case `p2` is not 
-        // [RefFieldsEscape] hence the return of the constructor is safe-to-escape to
-        // the current method (rule 3 above)
+        // [RefFieldsEscape] hence it contributes safe-to-escape of current method
+        // (rule 1.3 above)
         return new RS(ref p2);
 
-        // Error: This is the same analysis as above and once again by rule 3 is 
+        // Error: This is the same analysis as above and once again by rule 1.3 is 
         // only safe-to-escape to the current method
         int local = 42;
         return new RS(ref local);
-
-        // Okay: The RS(int[]) constructor has no arguments marked as `[RefFieldsEscape]` 
-        // hence arguments have their original safe-to-escape scope and this is 
-        // safe-to-escape to the calling method
-        return new RS(new int[1]);
     }
 }
 ```
@@ -528,6 +528,7 @@ This attribute can be applied to methods, constructors and operators. Applying t
 The semantics of this attribute, and how it impacts span safety rules, are described [above](#ref-fields-escape)
 
 #### DoesNotEscape
+<a name="does-not-escape">
 One source of repeated friction in low level code is the default escape for parameters is permissive. They are *safe-to-escape* to the *calling method*. This is a sensible default because it lines up with the coding patterns of .NET as a whole. In low level code though there is a larger use of  `ref struct` and this default can cause friction with other parts of the span safety rules.
 
 The main friction point occurs because of the following constraint around method invocations:
@@ -843,8 +844,8 @@ This small compat change reduces the overall compat impact of this change. The a
 
 The span safety rules for method invocation will be updated in several ways. The first is by recognizing the impact that `scoped` has on arguments. For a given argument `a` that is passed to parameter `p`:
 
-> If `p` is `scoped ref` or `scoped in` then `a` does not contribute *ref-safe-to-escape* when considering arguments
-> If `p` is `scoped` then `a` does not contribute *safe-to-escape* when considering arguments. Note that `ref scoped` is included here as the value is `scoped`
+> 1. If `p` is `scoped ref` or `scoped in` then `a` does not contribute *ref-safe-to-escape* when considering arguments
+> 2. If `p` is `scoped` then `a` does not contribute *safe-to-escape* when considering arguments. Note that `ref scoped` is included here as the value is `scoped`
 
 The language "does not contribute" means the arguments are simply not considered when calculating the *ref-safe-to-escape* or *safe-to-escape* value of the method return respectively. That is because the values can't contribute to that lifetime (the `scoped` prevents it).
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -136,21 +136,20 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
 
 <a name="ref-fields-escape"></a>
 
-To accomplish this the span safety rules will be changed as follows. First for the constructor of a `ref struct` that directly contains a `ref field`: 
-- If the constructor is annotated with `[RefFieldEscapes]` then `ref` fields can be initialized with any value that has *ref-safe-to-escape* to the *calling method*
-- Else `ref` fields can only be initialized with known heap values.
+To accomplish this the span safety rules for method invocation will be changed. The method invocation rules presently state:
 
-These rules ensure that the caller can understand when constructors will or will not capture parameters by `ref`. 
+> An rvalue resulting from a method invocation e1.M(e2, ...) is *safe-to-escape* from the smallest of the following scopes:
+> 1. The entire enclosing method
+> 2. The *safe-to-escape* of all argument expressions (including the receiver)
 
-Next the rules for method invocation will change as follows when the target method is annotated with `[RefFieldEscapes]`:
-- If the current method is not annotated with `[RefFieldEscapes]` then
-    - If all arguments have a *ref-safe-to-escape* scope of *heap* then the *safe-to-escape* scope of the return is to the *calling method*
-    - Else the *safe-to-escape* scope is to the *current method*
-- Else the *safe-to-escape* scope of the return is the minimum of 
-    - The existing *safe-to-escape* calculation for method invocation
-    - All of the *ref-safe-to-escape* values of `ref`, `in` and `out` arguments
+The change will be in how the *safe-to-escape* scope is calculated for arguments of method invocation when the return type is a ref-like type. The updated rule will read as follows:
 
-The design of `[RefFieldEscape]` will be discussed in detail [later in the proposal](#reffieldescapes).
+> When the rvalue is a ref-like type then *safe-to-escape* scope of arguments will be changed when they line up with a parameter marked as `[RefFieldEscapes]`
+> 1. When the argument is itself a parameter marked `[RefFieldEscapes]` it will be *calling method*
+> 2. When the argument is a known heap location it will be *calling method*
+> 3. Else it will be *current method*
+
+The rules for `ref` field assignment will be discussed in detail [later on](#ref-field-reassignment) the doc. For now readers only need to know constructors can initialize `ref` fields with parameters marked as `[RefFieldEscape]` or known heap locations.
 
 Let's examine these rules in the context of samples to better understand their impact and how they maintain the required compat considerations.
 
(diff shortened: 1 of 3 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -161,117 +160,59 @@ ref struct RS
...
-        // Okay: Both CreateRS2 and RS(ref int) are annotated as [RefFieldEscapes] hence the 
-        // safe-to-escape scope of the return is the min of the ref-safe-to-escape scope of 
-        // the ref / in arguments. In that case this is simply `parameter` which has a
-        // ref-safe-to-escape scope of the calling method
-        RS rs = new RS(ref parameter);
-        return rs;
+        // Error: This is the same analysis as above but in this case `p2` is not 
+        // [RefFieldsEscape] hence the return of the constructor is safe-to-escape to
+        // the current method (rule 3 above)
+        return new RS(ref p1);
```
333fred, line 198: Is this supposed to be p2?

proposals/low-level-struct-improvements.md
```diff
@@ -849,24 +787,129 @@ escapes struct S
...
+| `ref Span<int> s` | *calling method* | *calling method* | 
+| `scoped ref Span<int> s` | *current method* | *calling method* | 
+| `ref scoped Span<int> s` | *current method* | *current method* | 
+
+The declaration `scoped ref scoped Span<int>` is allowed but is redundant with `ref scoped Span<int>`. The *ref-safe-to-escape* of a value can never exceed the *safe-to-escape* hence once the *safe-to-escape* is restricted via `ref scoped` the *ref-safe-to-escape* is implicitly restricted.
+
+This also means that the `this` parameter of a `struct` can now be completely expressed as `scoped ref T`. Previously it had to be special cased as a `ref T` that had different *ref-safe-to-escape* rules than other `ref`. Now it can be expressed as a general concept vs. being a special cased item.
+
+These annotations can be applied to locals as well as parameters and have the same impact to lifetimes. They cannot be applied to any other location including fields, array elements, etc ... Further while `scoped` can be applied to any `ref` or `in` it can only be applied to values which are `ref struct`. Having `scoped int` adds no value to the rules and will be prevented to avoid confusion.
```
333fred, line 828:
```suggestion
These annotations can be applied to locals as well as parameters and have the same impact to lifetimes. They cannot be applied to any other location including fields, array elements, etc ... Further while `scoped` can be applied to any `ref` or `in` it can only be applied to by-values which are `ref struct`. Having `scoped int` adds no value to the rules and will be prevented to avoid confusion.
```

I found the "values" a bit confusing.

proposals/low-level-struct-improvements.md
```diff
@@ -136,21 +136,20 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
...
-- If the constructor is annotated with `[RefFieldEscapes]` then `ref` fields can be initialized with any value that has *ref-safe-to-escape* to the *calling method*
-- Else `ref` fields can only be initialized with known heap values.
+To accomplish this the span safety rules for method invocation will be changed. The method invocation rules presently state:
 
-These rules ensure that the caller can understand when constructors will or will not capture parameters by `ref`. 
+> An rvalue resulting from a method invocation e1.M(e2, ...) is *safe-to-escape* from the smallest of the following scopes:
+> 1. The entire enclosing method
+> 2. The *safe-to-escape* of all argument expressions (including the receiver)
+
+The change will be in how the *safe-to-escape* scope is calculated for arguments of method invocation when the return type is a ref-like type. The updated rule will read as follows:
```
cston, line 145: Why does the *safe-to-escape* scope change for the arguments to the method invocation, rather than the *safe-to-escape* scope of the arguments affecting the *safe-to-escape* scope of the ref-like return value?

proposals/low-level-struct-improvements.md
```diff
@@ -136,21 +136,25 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
...
-    - Else the *safe-to-escape* scope is to the *current method*
-- Else the *safe-to-escape* scope of the return is the minimum of 
-    - The existing *safe-to-escape* calculation for method invocation
-    - All of the *ref-safe-to-escape* values of `ref`, `in` and `out` arguments
+To recognize these annotations the span safety rules for method invocation need to be updated. The first change is recognizing the impact the annotations have on arguments. For a given argument `a` that is passed to `ref` parameter `p`: 
 
-The design of `[RefFieldEscape]` will be discussed in detail [later in the proposal](#reffieldescapes).
+> 1. If `p` is `[RefFieldsEscape] ref` or `[RefFieldsEscape] in` and `a` is 
+>     1. A `[RefFieldEscapes]` parameter it contrbutes *safe-to-escape* to *calling method*
+>     2. A known heap location it contrbutes *safe-to-escape* to *calling method*
```
cston, line 147:
```suggestion
>     2. A known heap location it contributes *safe-to-escape* to *calling method*
```

proposals/low-level-struct-improvements.md
```diff
@@ -849,24 +788,129 @@ escapes struct S
...
+    // Error: RValue Rule 3 specifies that the safe-to-escape be limited to the ref-safe-to-escape
+    // of the ref argument. That is the current method for value hence this is not allowed.
+    return new Span<int>(ref value);
+}
+
+Span<int> CreateAndCapture(ref int value)
+{
+    // Okay: RValue Rule 3 specifies that the safe-to-escape be limited to the ref-safe-to-escape
+    // of the ref argument. That is the calling method for value hence this is not allowed.
+    return new Span<int>(value)
```
cston, line 881:
```suggestion
    return new Span<int>(ref value)
```

proposals/low-level-struct-improvements.md
```diff
@@ -912,6 +956,8 @@ ref T StrangeIdentity<T>(out T value)
 
 This is an unlikely combination and the language could benefit strongly by considering `out` parameters as not returnable by `ref`. 
 
+Consideration was given in this section to also taking a breaking change for *ref-safe-to-escape* defaults for `this` on a `struct`. Specifically considering changing it to be `ref T` vs. `scoped ref T`. That has the advantage that there is no need for the `escapes` keyword anymore as every has maximum return as the default hence only `scoped` is needed to restrict the behavior. The impact of this change is far more significant. It impacts all `struct` instances that have a `ref` returning method which is likely too big of a surface area to take a casual break to.
```
cston, line 959: "... as every [...] has maximum ..."?

jaredpar, line 959:
```suggestion
Consideration was given in this section to also taking a breaking change for *ref-safe-to-escape* defaults for `this` on a `struct`. Specifically considering changing it to be `ref T` vs. `scoped ref T`. That has the advantage that there is no need for the `escapes` keyword anymore as every value already has maximum return as the default hence only `scoped` is needed to restrict the behavior. The impact of this change is far more significant. It impacts all `struct` instances that have a `ref` returning method which is likely too big of a surface area to take a casual break to.
```

proposals/low-level-struct-improvements.md
```diff
@@ -849,24 +788,129 @@ escapes struct S
...
+
+| Parameter | ref-safe-to-escape | safe-to-escape |
+|---|---|---|
+| `Span<int> s` | *current method* | *calling method* | 
+| `scoped Span<int> s` | *current method* | *current method* | 
+| `ref Span<int> s` | *calling method* | *calling method* | 
+| `scoped ref Span<int> s` | *current method* | *calling method* | 
+| `ref scoped Span<int> s` | *current method* | *current method* | 
+
+The declaration `scoped ref scoped Span<int>` is allowed but is redundant with `ref scoped Span<int>`. The *ref-safe-to-escape* of a value can never exceed the *safe-to-escape* hence once the *safe-to-escape* is restricted via `ref scoped` the *ref-safe-to-escape* is implicitly restricted.
```
333fred, line 825: I'm wanting explicit lifetimes more and more 🙂.

proposals/low-level-struct-improvements.md
```diff
@@ -849,24 +788,129 @@ escapes struct S
...
+    // Error: ref-safe-to-escape of out is now the current method
+    return ref i;
+}
+```
+
+This change to `out` reduces the overall compat impact of this change. The ability to return `out` by reference is effectively compiler trivia. However it negatively impacts analysis because the rules must consider the case that it is returned by `ref`. Hence `out` arguments, even though 99% of the time are not returned by ref must be considered as such and that conflates lifetime issues.
+
+The span safety rules for method invocation will be updated in several ways. The first is by recognizing the impact that `scoped` has on arguments. For a given argument `a` that is passed to parameter `p`:
+
+> 1. If `p` is `scoped ref` or `scoped in` then `a` does not contribute *ref-safe-to-escape* when considering arguments
```
333fred, line 847: And `ref scoped`?

proposals/low-level-struct-improvements.md
```diff
@@ -849,24 +788,129 @@ escapes struct S
...
+    // of the ref argument. That is the calling method for value hence this is not allowed.
+    return new Span<int>(ref value)
+}
+
+Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
+{
+    // Okay: the safe-to-escape of `span` is *calling method* hence this is legal.
+    return span;
+
+    // Okay: the parameter here is `scoped ref` hence `refLocal` does not contribute /
```
333fred, line 889: This example doesn't use the parameter?

proposals/low-level-struct-improvements.md
```diff
@@ -136,21 +136,25 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
...
-    - If all arguments have a *ref-safe-to-escape* scope of *heap* then the *safe-to-escape* scope of the return is to the *calling method*
-    - Else the *safe-to-escape* scope is to the *current method*
-- Else the *safe-to-escape* scope of the return is the minimum of 
-    - The existing *safe-to-escape* calculation for method invocation
-    - All of the *ref-safe-to-escape* values of `ref`, `in` and `out` arguments
+To recognize these annotations the span safety rules for method invocation need to be updated. The first change is recognizing the impact the annotations have on arguments. For a given argument `a` that is passed to `ref` parameter `p`: 
 
-The design of `[RefFieldEscape]` will be discussed in detail [later in the proposal](#reffieldescapes).
+> 1. If `p` is `[RefFieldsEscape] ref` or `[RefFieldsEscape] in` and `a` is 
+>     1. A `[RefFieldEscapes]` parameter it contrbutes *safe-to-escape* to *calling method*
```
cston, line 146:
```suggestion
>     1. A `[RefFieldEscapes]` parameter it contributes *safe-to-escape* to *calling method*
```

proposals/low-level-struct-improvements.md
jaredpar: The hidden advantage of this proposal is it opens the door for them in the future. Effectively the syntax can be extended to `scoped<'a>`  where `'a` represents a lifetime. From there it expands out to `where 'a > 'b` and what have you. At the moment though there is no explicit need for this hence not a part of the feature.

proposals/low-level-struct-improvements.md
jaredpar: A `ref scoped` is implicitly `scoped ref` hence it's covered in this rule.

proposals/low-level-struct-improvements.md
```diff
@@ -136,21 +136,25 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
 
 <a name="ref-fields-escape"></a>
 
-To accomplish this the span safety rules will be changed as follows. First for the constructor of a `ref struct` that directly contains a `ref field`: 
-- If the constructor is annotated with `[RefFieldEscapes]` then `ref` fields can be initialized with any value that has *ref-safe-to-escape* to the *calling method*
-- Else `ref` fields can only be initialized with known heap values.
+To accomplish this two new annotations will be introduced to help control how arguments influence lifetime of method calls: `[RefFieldsEscape]` and `[DoesNotEscape]`. The annotation `[RefFieldsEscape]` when applied to a `ref` parameter signifies that it can be captured as a `ref` field in a returned `ref struct`.
```
cston, line 139: >RefFieldsEscape [](http://example.com/codeflow?start=126&length=15)

Consider s/[RefFieldsEscape]/[RefFieldEscapes]/ since there are currently a mix of the two.

proposals/low-level-struct-improvements.md
```diff
@@ -136,21 +136,25 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
...
+>     3. It contributes *safe-to-escape* to *current method*
+> 2. If `p` is `[DoesNotEscape]` then `a` does not contribute *safe-to-escape* when considering arguments.
+
+This change allows us to keep our existng rules for rvalue method invocation:
+
+> An rvalue resulting from a method invocation e1.M(e2, ...) is *safe-to-escape* from the smallest of the following scopes:
+> * The entire enclosing method
+> * The *safe-to-escape* of all argument expressions (including the receiver)
+
+The rules for `ref` field assignment will be discussed in detail [later on](#ref-field-reassignment) the doc. For now readers only need to know constructors can initialize `ref` fields with parameters marked as `[RefFieldEscape]` or known heap locations.
```
cston, line 157:
```suggestion
The rules for `ref` field assignment will be discussed in detail [later on](#ref-field-reassignment) the doc. For now readers only need to know constructors can initialize `ref` fields with parameters marked as `[RefFieldEscapes]` or known heap locations.
```

proposals/low-level-struct-improvements.md
```diff
@@ -136,21 +136,25 @@ The rules we define for `ref` fields must ensure the `Span<T>` constructor prope
...
+To recognize these annotations the span safety rules for method invocation need to be updated. The first change is recognizing the impact the annotations have on arguments. For a given argument `a` that is passed to `ref` parameter `p`: 
 
-The design of `[RefFieldEscape]` will be discussed in detail [later in the proposal](#reffieldescapes).
+> 1. If `p` is `[RefFieldsEscape] ref` or `[RefFieldsEscape] in` and `a` is 
+>     1. A `[RefFieldEscapes]` parameter it contrbutes *safe-to-escape* to *calling method*
+>     2. A known heap location it contributes *safe-to-escape* to *calling method*
+>     3. It contributes *safe-to-escape* to *current method*
+> 2. If `p` is `[DoesNotEscape]` then `a` does not contribute *safe-to-escape* when considering arguments.
+
+This change allows us to keep our existng rules for rvalue method invocation:
```
cston, line 151:
```suggestion
This change allows us to keep our existing rules for rvalue method invocation:
```

proposals/low-level-struct-improvements.md
jaredpar: Correct.

333fred: Then I don't understand the explanation. If `refLocal` isn't involved at all in this example, why is its scope or lack thereof relevant?

jaredpar: `refLocal` is passed as an argument here so it's definitely involved. The point of this particular sample is to drive home that there are two lifetimes to consider here: the ref and the value. The parameter is `scoped ref` that means the ref lifetime does not contribute to the returned `ref struct` (the *ref-safe-to-escape* is not considered).

Compare that with the next sample where it concretely shows that the value does contribute here and prevent a return

proposals/low-level-struct-improvements.md
```diff
@@ -849,24 +788,129 @@ escapes struct S
...
+> 2. If `p` is `scoped` then `a` does not contribute *safe-to-escape* when considering arguments. Note that `ref scoped` is included here as the value is `scoped`
+
+The language "does not contribute" means the arguments are simply not considered when calculating the *ref-safe-to-escape* or *safe-to-escape* value of the method return respectively. That is because the values can't contribute to that lifetime (the `scoped` prevents it).
+
+The method invocation for lvalue returns can now be simplified. The receiver no longer needs to be special cased, in the case of `struct` it is now simply a `scoped ref T`, nor do `out` parameters need to be considered in the argument list. 
+
+> An lvalue resulting from a ref-returning method invocation `e1.M(e2, ...)` is *ref-safe-to-escape* the smallest of the following scopes:
+> 1. The *calling method*
+> 2. The *ref-safe-to-escape* of all `ref` arguments
+> 3. The *ref-safe-to-escape* of all `in` parameters when the argument is an lvalue otherwise *current method*
```
cston, line 857: To be precise, should this be "arguments" rather than "parameters"?
```suggestion
> 3. The *ref-safe-to-escape* of all `in` arguments when the argument is an lvalue otherwise *current method*
```

proposals/low-level-struct-improvements.md
333fred: I think I'm confused because you say "parameter", which is ambiguous: did you mean `span` that is currently in scope, or did you mean the parameter of the called method (which is also this method, making it confusing to talk about)? Consider clarifying what parameter you mean.

cston (approved):

proposals/low-level-struct-improvements.md
jaredpar: Is this explanation clearer? 

```
    // Okay: the local `refLocal` has a ref-safe-to-escape of *current method* and a 
    // safe-to-escape of *calling method*. In the call below it is passed to a 
    // parameter that is `scoped ref` which means it does not contribute 
    // ref-safe-to-escape. It only contributes its safe-to-escape hence the returned
    // rvalue ends up as safe-to-escape of *calling method*
```

333fred: Much clearer, thanks!

jaredpar: Awesome. Will push that change shortly

333fred (approved):

eyalalonn:
@jaredpar I'm not sure whether the syntax is still an open question and I know it's a minor but you proposed the keywords `scoped` and `escapes` might worth considering `scoped` and `unscoped` as opposed to `escapes`.

jaredpar:
@eyalalonn the particular syntax is very much up for consideration. Thanks for the suggestion here I agree that is a reasonable way to approach it.

PR #5940, opened 2022-03-22
Specify behavior of default this initializer with field initializers

RikkiGibson:
Closes dotnet/roslyn#58790

### DIFF
--- a/proposals/csharp-10.0/parameterless-struct-constructors.md
+++ b/proposals/csharp-10.0/parameterless-struct-constructors.md
@@ -67,7 +67,7 @@ We also add the following language to [§15.4.9](https://github.com/dotnet/cshar
 
 > If the struct includes field initializers, then any constructors within the struct which have a `this()` constructor initializer referring to the default parameterless constructor are treated as though they do not have any constructor initializer.
 
-This means we get an error in a scenario like the following:
+This means an error is reported for the following:
 
 ```cs
 public struct S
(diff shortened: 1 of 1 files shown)


proposals/csharp-10.0/parameterless-struct-constructors.md
```diff
@@ -63,6 +63,24 @@ Constructors cannot be `partial`.
...
 > When an instance constructor has no constructor initializer, **or when the constructor initializer `this()` represents the default parameterless constructor**, ... that constructor implicitly performs the initializations specified by the _variable_initializers_ of the instance fields ... . This corresponds to a sequence of assignments that are executed immediately upon entry to the constructor ... . The variable initializers are executed in the textual order in which they appear in the ... declaration.
 
+We also add the following language to [§15.4.9](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/structs.md#1549-constructors):
+
+> If the struct includes field initializers, then any constructors within the struct which have a `this()` constructor initializer referring to the default parameterless constructor are treated as though they do not have any constructor initializer.
+
+This means we get an error in a scenario like the following:
```
cston, line 70: Minor: Perhaps "This means an error is reported for the following:"

cston (approved):

proposals/csharp-10.0/parameterless-struct-constructors.md
RikkiGibson:
```suggestion
This means an error is reported for the following:
```

RikkiGibson:
It looks like we want to change this behavior in 17.2 so that referencing the default parameterless constructor with `: this()` means that we first emit the 'initobj', then the field initializers.

```cs
struct S
{
    int x = 1;
    int y;
    public S(bool unused) : this() { }
}

// compiles to (roughly)
struct S
{
    int x;
    int y;
    public S(bool unused)
    {
        this = default;
        x = 1;
    }
}
```

jcouv:
Should we close this?

RikkiGibson:
Sure. We can open a new PR which specifies the behavior we really want.

PR #5943, opened 2022-03-22
Fix broken links

gewarren:
Contributes to dotnet/docs#28282

### DIFF
--- a/proposals/csharp-9.0/function-pointers.md
+++ b/proposals/csharp-9.0/function-pointers.md
@@ -1,4 +1,4 @@
-﻿# Function Pointers
+# Function Pointers
 
 ## Summary
 
@@ -274,7 +274,7 @@ In unsafe code, the following changes are made to the type inference algorithms:
 
 #### Input types
 
-[§11.6.3.4](https://github.com/dotnet/csharpstandard/blob/draft-v6/expressions.md#11634-input-types)
+[§11.6.3.4](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#11634-input-types)
 
 The following is added:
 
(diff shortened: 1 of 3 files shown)


proposals/csharp-9.0/function-pointers.md
```diff
@@ -1,4 +1,4 @@
-﻿# Function Pointers
+# Function Pointers
```
333fred, line 1: What changed here? Did the file encoding get set to something different?

gewarren: It was a red dot that showed up in the edit view. You can see it here: https://github.com/dotnet/csharplang/edit/main/proposals/csharp-9.0/function-pointers.md

333fred: Right, I believe that is the UTF-8 BOM. What I'm trying to determine is whether it should be removed 🙂

BillWagner (approved):
This LGTM @gewarren 

Let's :shipit:

proposals/csharp-9.0/function-pointers.md
333fred: @gewarren did you determine what this change was?

gewarren: @333fred It was the BOM, as you suspected. (You can [see it in the blue status bar](https://github.dev/dotnet/csharplang/blob/26e70504f3505398fe8b21b85ff5f74ab3f07e7e/proposals/csharp-9.0/function-pointers.md) for the previous version of the file.) I looked at a couple other .md files in this repo and they didn't have the BOM.

333fred: Thanks.

PR #5959, opened 2022-03-28
Ref fields major revision

jaredpar:
This revision of `ref` fields leans into the decisions made in recent LDM meetings. Specifically:

- Use `syntax` over attributes
- Accept the minor breaking changes to existing API signatures

It's possible this will be easier to read as a new document vs. diffing against the old one. The changes are pretty significant. 

Overall though I'm really happy with the changes. I think the design is much easier to explain now and fits in nicely with the existing span safety rules.


proposals/todo.txt
```diff
@@ -0,0 +1,292 @@
+move to unscoped instead of escapes
```
jaredpar, line 1: Probably should have deleted my todo notes before PRing 😄

proposals/low-level-struct-improvements.md
```diff
@@ -158,7 +158,7 @@ First the rules establishing *ref-safe-to-escape* values for fields needs to be
 > 3. Else if `e` is of a reference type, it has *ref-safe-to-escape* of *calling method*
 > 4. Else its *ref-safe-to-escape* is taken from the *ref-safe-to-escape* of `e`.
 
-This is does not represent rule change though as the rules have always accounted for `ref` state to exist inside a `ref struct`. This is in fact how the `ref` state in `Span<T>` has always worked and the consumption rules correctly account for this. The change here is just accounting for developers to be able to now declare their own `ref` fields and ensure they do so by the existing rules implicitly applied to `Span<T>`. 
+This is does not represent rule change though as the rules have always accounted for `ref` state to exist inside a `ref struct`. This is in fact how the `ref` state in `Span<T>` has always worked and the consumption rules correctly account for this. The change here is just accounting for developers to be able to access `ref` fields directly and ensure they do so by the existing rules implicitly applied to `Span<T>`.
```
huanlin, line 161: This is does not...
This does not...

YairHalberstadt:
Should `ref readonly` fields be supported? Would an `in` be convertible to a `ref readonly` field?

It seems this should be legal:

```csharp
public struct ReadOnlySpan<T>
{
    readonly ref readonly T _t;
    public ReadOnlySpan(in T t)
    {
        _t = t;
    }
}
```

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,125 @@ readonly ref struct Span<T>
...
-
-Consider the following method signature:
+The challenge in this proposal is the compatibility implications this design has to our existing [span safety rules](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md). While those rules fully support the concept of a `ref struct` having having `ref` fields they do not allow for APIs, other than `stackalloc`, to capture `ref` state that refers to the stack. The span safety rules have a [hard assumption](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md#span-constructor) that a constructor of the form `Span(ref T value)` does not exist. That means the safety rules do not account for a `ref` parameter being able to escape as a `ref` field hence it allows for code like the following.
```
cston, line 45: >having having [](http://example.com/codeflow?start=266&length=13)

Typo

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,125 @@ readonly ref struct Span<T>
...
+The existing rules only account for (1) and (2). They do not account for (3) hence gaps like returning locals as `ref` fields are not accounted for. This design must change the rules to account for (3). This will have a small impact to compatibility for existing APIs. Specifically it will impact APIs that have the following properties.
 
-    // Okay
-    int[] array = new int[42];
-    return CreatSpan(ref array[0]);
-}
-```
-
-All of the above `return` statements are legal by our existing rules because the return of `CreateSpan` is always *safe-to-escape* to the *calling method*. The challenge in this proposal, which comes up many times, is they **must** remain legal in the version of the language which implements these features and / or when `Span<T>` moves to using `ref` fields. The above patterns can legally exist today and cannot become errors when moving to a new language version. 
+- Have a `Span<T>` or `ref struct` in the signature
```
cston, line 67: >`Span<T>` or  [](http://example.com/codeflow?start=9&length=13)

Consider removing "`Span<T>` or" since that's covered by "`ref struct`".

jaredpar:
> Should ref readonly fields be supported? Would an in be convertible to a ref readonly field?

Yes that should be supported. It's implied by the spec but could've called it out a bit better.

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,125 @@ readonly ref struct Span<T>
...
+    return CreateSpan(ref local);
+
+    // Legal in C# 10 but would be illegal with ref fields
+    Span<int> span = stackalloc int[42];
+    return CreateSpan(ref span[0]);
 }
 ```
 
-It will be possible for such methods to exist. Specifically methods which take `ref` parameters, capture them in `ref` fields and return them. But they require a declarative opt-in to let the compiler, and developer, know that it is happening and to adjust the span safety rules accordingly.
+The impact of this compatibility break is expected to be very small. The impacted API shape made little sense in the abscene of `ref` fields hence it is unlikely customers created many of these. Experiments running tools to spot this API shape over existing repositories back up that assertion. The only repository with any significant counts of this shape is [dotnet/runtime](https://github.com/dotnet/runtime) and that is because that repo can create `ref` fields via the `ByReference<T>` intrinsic type.
```
cston, line 99: >abscene [](http://example.com/codeflow?start=117&length=7)

Typo

proposals/low-level-struct-improvements.md
jaredpar: clearly i'm having having some issues here

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,125 @@ readonly ref struct Span<T>
...
 
-It is important to understand these compat considerations before diving too far into the proposal here as they are central to parts of the design.
+Even so the design must account for such APIs existing because it express a valid pattern, just not a common one. Hence the design must give developers the tools to restore the existing lifetime rules when upgrading to C# 10. Specifically it must provide mechanisms that allow developers to annotate `ref` parameters as unable to escape by `ref` or `ref` field. That allows customers to define APIs in C# 11 that have the same C# 10 callsite rules.
```
cston, line 101: >express [](http://example.com/codeflow?start=66&length=7)

"expresses"?

jaredpar:
@YairHalberstadt @cston added a section on how `readonly` works with `ref` fields and example for `ReadOnlySpan<T>`

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,125 @@ readonly ref struct Span<T>
...
-> * The *safe-to-escape* of all argument expressions (including the receiver)
-
-The rules for `ref` field assignment will be discussed in detail [later on](#ref-field-reassignment) the doc. For now readers only need to know constructors can initialize `ref` fields with parameters marked as `[RefFieldEscapes]` or known heap locations.
-
-Let's examine these rules in the context of samples to better understand their impact and how they maintain the required compat considerations.
-
-```c#
-ref struct RS
+``` C#
+struct S
```
cston, line 110: >struct [](http://example.com/codeflow?start=0&length=6)

`ref struct` here, and in other examples below.

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,131 @@ readonly ref struct Span<T>
...
-Consider as an example the indexer on `Span<T>` which is returning `ref` fields by `ref` today. The fourth item here is what provides the safety here:
+A `ref` field can be combined with `readonly` modifiers in the following ways:
 
-```c#
-ref int Examples()
-{
-    Span<int> s1 = stackalloc int[5];
+- `readonly ref`: this is a field that cannot be ref re-assigned outside a constructor or `init` methods. It can be value assigned though outside those contexts
+- `ref readonly`: this is a field that can be ref re-assigned but cannot be value assigned at any point. This how an `in` parameter could be ref re-assigned to a `ref` field.
+- `readonly ref readonly`: a combination of `ref readonly` and `readonly ref.
```
cston, line 153: >ref [](http://example.com/codeflow?start=73&length=3)

Missing backquote.

proposals/low-level-struct-improvements.md
```diff
@@ -45,235 +42,131 @@ readonly ref struct Span<T>
...
 
 > An lvalue designating a reference to a field, e.F, is *ref-safe-to-escape* (by reference) as follows:
 > 1. If `F` is a `ref` field and `e` is `this`, it is *ref-safe-to-escape* from the enclosing method.
 > 2. Else if `F` is a `ref` field its *ref-safe-to-escape* scope is the *safe-to-escape* scope of `e`.
-> 3. Else if `e` is of a reference type, it is *ref-safe-to-escape* from the enclosing method.
+> 3. Else if `e` is of a reference type, it has *ref-safe-to-escape* of *calling method*
 > 4. Else its *ref-safe-to-escape* is taken from the *ref-safe-to-escape* of `e`.
 
-This explicitly allows for `ref` fields being returned as `ref` from a `ref struct` but not normal fields (that will be covered later). 
+This does not represent rule change though as the rules have always accounted for `ref` state to exist inside a `ref struct`. This is in fact how the `ref` state in `Span<T>` has always worked and the consumption rules correctly account for this. The change here is just accounting for developers to be able to access `ref` fields directly and ensure they do so by the existing rules implicitly applied to `Span<T>`.
```
cston, line 167: >rule change [](http://example.com/codeflow?start=24&length=11)

"a rule change"?

proposals/low-level-struct-improvements.md
```diff
@@ -284,403 +177,340 @@ ref struct RS
...
+        // safe-to-escape of `this` and ref-safe-to-escape of `value` is *calling method* hence 
+        // this is legal.
+        _field = ref value;
+        _length = 1;
     }
 }
 ```
 
-<a name="ref-field-reassignment"></a>
+The change to ref re-assignment rules means `ref` parameters can now escape from a method as a `ref` field in a `ref struct` value. As discussed in the [compat considerations section](#new-span-challenges) this can change the rules for existing APIs that never intended for `ref` parameters to escape as a `ref` field. The lifetime rules for parameters are based soley on their declaration not on their usage. All `ref` and `in` parameters are *ref-safe-to-escape* to the *calling method* and hence can now be returned by `ref` or a `ref` field. In order to support APIs having `ref` parameters that can be escaping or non-escaping, and thus restore C# 10 call site semantics, the langauge will introduce limited lifetime annotations.
```
cston, line 215: >langauge [](http://example.com/codeflow?start=681&length=8)

Typo

proposals/low-level-struct-improvements.md
```diff
@@ -743,170 +563,231 @@ Given those rules here is a valid reference assembly transformation for a `ref s
...
+        span[2] = 'g';
 
-The method invocation for rvalue returns needs to change as follows to account for `ref` field returns.
+        // Error: The safe-to-escape of `span` is the current method scope 
+        // while `reader` is outside the current method scope hence this fails
+        // by the above rule.
+        if (reader.TextEquals(span)
+        {
+            ...
+        })
```
PathogenDavid, line 734: > ```csharp
> })
> ```

Typo, `)` should be up with `if`.

proposals/low-level-struct-improvements.md
```diff
@@ -743,170 +563,231 @@ Given those rules here is a valid reference assembly transformation for a `ref s
...
-    // Error: similar analysis as above but the safe-to-escape scope of `stackLocal` is 
-    // *current method* hence this is illegal
-    Span<int> stackLocal = stackalloc int[42];
-    return ComplexScopedRefExample(ref stackLocal);
+        // Okay: the compiler never considers `span` as capturable here hence it doesn't
+        // contribute to the method arguments must match rule
+        if (reader.TextEquals(span)
+        {
+            ...
+        })
```
PathogenDavid, line 770: > ```csharp
> })
> ```

Typo, `)` should be up with `if`.

proposals/low-level-struct-improvements.md
```diff
@@ -284,403 +177,340 @@ ref struct RS
...
-        // scope by the revised constructor rules. This falls into bullet two 
-        // above and hence based on that allowed scope.
-        int local3 = 42;
-        var local4 = new RS(ref local3);
-        return ref local4.Prop1;
+```c#
+readonly ref struct Span<T>
+{
+    readonly ref T _field;
+    int _length;
```
PathogenDavid, line 202: `_length` should probably be `readonly`.

proposals/low-level-struct-improvements.md
huanlin: "having having" is still there. how come this is marked as resolved?

jaredpar: It was removed in this commit which was pushed yesterday

https://github.com/dotnet/csharplang/pull/5959/commits/584743fd17aec94c2d46084bdf64e61df473008e

Are you not seeing that in your view?

proposals/low-level-struct-improvements.md
jaredpar: yep, good catch

proposals/low-level-struct-improvements.md
huanlin: > It was removed in this commit which was pushed yesterday
> 
> [584743f](https://github.com/dotnet/csharplang/commit/584743fd17aec94c2d46084bdf64e61df473008e)
> 
> Are you not seeing that in your view?

I can see it in 584743f, and I thought it would show "here".  It turns out I was wrong. Sorry! my bad.

proposals/low-level-struct-improvements.md
```diff
@@ -743,170 +563,231 @@ Given those rules here is a valid reference assembly transformation for a `ref s
...
+        Span<char> span = stackalloc char[4];
+        span[0] = 'd';
+        span[1] = 'o';
+        span[2] = 'g';
 
-The method invocation for rvalue returns needs to change as follows to account for `ref` field returns.
+        // Error: The safe-to-escape of `span` is the current method scope 
+        // while `reader` is outside the current method scope hence this fails
+        // by the above rule.
+        if (reader.TextEquals(span)
```
PathogenDavid, line 731: I should've made it clearer in my previous comments, but there's a `)` missing here.

proposals/low-level-struct-improvements.md
```diff
@@ -743,170 +563,231 @@ Given those rules here is a valid reference assembly transformation for a `ref s
...
+        span[1] = 'o';
+        span[2] = 'g';
 
-    // Error: similar analysis as above but the safe-to-escape scope of `stackLocal` is 
-    // *current method* hence this is illegal
-    Span<int> stackLocal = stackalloc int[42];
-    return ComplexScopedRefExample(ref stackLocal);
+        // Okay: the compiler never considers `span` as capturable here hence it doesn't
+        // contribute to the method arguments must match rule
+        if (reader.TextEquals(span)
```
PathogenDavid, line 767: Another missing `)`

Thealexbarney:
> When indexing into a `fixed` buffer of type `T` the `readonly` state of the container must be taken into account.  If the container is `readonly` then the indexer returns `ref readonly T` else it returns `ref T`.

Will this concept be something that could be applied elsewhere as well, or would it be specially handled in the case of fixed buffers? Being able to overload based on the `readonly` state of a struct would be pretty useful, allowing you to return a `ref readonly` or `ref` based on that.

jaredpar:
> Will this concept be something that could be applied elsewhere as well, or would it be specially handled in the case of fixed buffers?

It will be special cased to this specific feature. 

> Being able to overload based on the readonly state of a struct would be pretty useful, allowing you to return a ref readonly or ref based on that.

Personally don't have any interest in replicating the `const` pattern from C++ here where you have diff methods for `const` and non-const. Particularly cause 99% of time it's the same method. The other 1% is usually something evil 😄 

Instead of an overload I think of this more as a permission parameter. Essentially the indexer looks like the following: 

```c#
ref P T GetValue<P>(ref P TheStruct this) where P : permission 
  => ref ... 
```

The idea being that the permission of the receiver dictates the permission of the return. In this case the set of permissions is pretty limited: `readonly` or implicit mutable. But under the hood it's one method where the receiver is dictating the return of the method.

jaredpar:
Going to go ahead and merge this as several efforts are underway that depend on the updated design. Want to have a checked in doc to point at (much more readable than the PR). Happy to continue addressing feedback and can start a new PR to address it.

proposals/low-level-struct-improvements.md
```diff
@@ -284,403 +177,340 @@ ref struct RS
...
-    {
-        int[] array = new int[] { 42, 13 };
-        SmallSpan s = default;
+The `scoped` annotation also means that the `this` parameter of a `struct` can now be defined as `scoped ref T`. Previously it had to be special cased in the rules as `ref` parameter that had different *ref-safe-to-escape* rules than other `ref` parameters (see all the references to including or excluding the receiver in the span safety rules). Now it can be expressed as a general concept throughout the rules which further simplifies them.
 
-        // Okay: the value being assigned here is known to refer to the heap 
-        // hence it is allowed by our rules above because it requires no changes
-        // to existing method invocation rules (hence preserves compat)
-        s._field = ref array[i];
+In addition to parameters the `scoped` annotation can be applied to locals or `struct` instance methods. These annotations have the same impact to lifetimes when applied to locals. For locals these annotations concretely define the lifetimes and override the lifetime that would be inferred via the initializer.
```
RikkiGibson, line 256: "locals *of* `struct` instance methods"?

RikkiGibson: If this actually means that 'scoped' can be applied to the signature of a struct instance method, consider including an example.

RikkiGibson: If this actually means that 'scoped' can be applied to the signature of a struct instance method, consider including an example.

proposals/low-level-struct-improvements.md
```diff
@@ -284,403 +177,340 @@ ref struct RS
...
-    Span<int> _field;
-    void Set(Span<int> p)
-    {
-        _field = p;
-    }
-
-    static void DangerousCode(ref RS p)
-    {
-        Span<int> span = stackalloc int[] { 42 };
+     int _field;
```
RikkiGibson, line 412: nit: looks like there is an extra space here

RikkiGibson, line 412: nit: looks like there is an extra space here

Thealexbarney:
> Personally don't have any interest in replicating the `const` pattern from C++ here where you have diff methods for `const` and non-const. Particularly cause 99% of time it's the same method. The other 1% is usually something evil 😄
> 
> Instead of an overload I think of this more as a permission parameter. Essentially the indexer looks like the following:
> 
> ```cs
> ref P T GetValue<P>(ref P TheStruct this) where P : permission 
>   => ref ... 
> ```
> 
> The idea being that the permission of the receiver dictates the permission of the return. In this case the set of permissions is pretty limited: `readonly` or implicit mutable. But under the hood it's one method where the receiver is dictating the return of the method.

You're right. I don't think I've ever seen a `const` overload that wasn't the same thing as its non-`const` version. And unless you just like being evil, no situations come to mind where you'd want them to be different.

The permission parameter concept seems like it could be a good solution. One (admittedly niche) advantage I see is if C# ever gains support for representing controlled-mutability managed pointers then that could maybe be added to the set of permissions.

In any case, not needing different members like `Prop` and `PropRo` if you want both mutable and `readonly` references would be a welcome change.

jaredpar:
> The permission parameter concept seems like it could be a good solution. One (admittedly niche) advantage I see is if C# ever gains support for representing controlled-mutability managed pointers then that could maybe be added to the set of permissions.

Permission parameters is a feature we got a lot of experience with in System C#. In that variant of C# we had a large permission set and could flow it through generics. It was quite extensive. Having it for `readonly` only is pretty straight forward compared to the madness that existed in System C#. It would likely end up being compatible with this pattern though. 

Then again, I'd rate the chance we do permission parameters as very low. It's really neat but requires a collection of other features to make it useful. The cost benefit of doing all that work is questionable.

PR #6002, opened 2022-04-10
Update pattern-variables.md

alrz:
https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-03-28.md

> #### Conclusion
> We would like to scope down the allowed location for redeclaration/assignment to:
> 
> 1. Within case labels for the same switch section.
> 2. Within a single expression.
> 3. Within related statements: an if/else if, for example, but not in two unconnected if statements.

Tentatively excluded the third bullet point and moved to open questions.

### DIFF
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -53,7 +53,7 @@
 
 ### Variable redeclaration
 
-- Pattern variables are allowed to be redeclared in the same expression if not already definitely assigned. These names can possibly reference either of variables based on the result of the pattern-matching at runtime.
+- Pattern variables are allowed to be redeclared in the same expression or switch section if not already definitely assigned. These names can possibly reference either of variables based on the result of the pattern-matching at runtime.
 - Pattern variables with multiple declarations must be of the same type, excluding tuple names and nullability for reference types.
 
 ### Definite assignment
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -43,10 +43,9 @@
 	}
 	```
 
-- Also relax single-declaration rules across the entire declaring scope as long as each variable is assigned once.
+- Also relax single-declaration rules within expression boundaries as long as each variable is assigned once.
 	```cs
 	if (e is C c || e is Wrapper { Prop: C c }) ;
-	if (e is C c) ; else if (e is Wrapper { Prop: C c }) ;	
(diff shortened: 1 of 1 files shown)


333fred:
> Tentatively excluded the third bullet point and moved to open questions to reiterate (https://github.com/dotnet/csharplang/discussions/5965#discussioncomment-2459888)

Please don't. LDM made a decision here, and we'd like to see a proposal on it.

alrz:
From the discussion I thought you wanted to take another look at it.

I'll see if I can come up with a draft proposal on that, let me know if you have any immediate ideas to start with. Thanks!

alrz:
I couldn't think of a way to generalize the definition of "related statements" to anything beyond `if`/`else` so the merging of variables needs to be done in four distinct locations:

- Within case labels for the same switch section (includes `when` clauses)
- Within a single expression (includes `is` expressions)
- Within a single pattern (includes `switch` expression arms)
- Within a single `if` statement (includes nested `if` statements)

Thoughts?

333fred:
> I couldn't think of a way to generalize the definition of "related statements" 

Yeah, I agree with that assessement, there's no wording I can think of either. Your proposed rules look good to me, with the addendum about nested `if` statements; I agree that seems undesirable.

proposals/pattern-variables.md
```diff
@@ -43,18 +43,23 @@
...
 ## Detailed design
 
 ### Variable redeclaration
 
-- Pattern variables are allowed to be redeclared in the same scope if not already definitely assigned. These names can possibly reference either of variables based on the result of the pattern-matching at runtime.
+- Pattern variables are allowed to be redeclared in the following locations if not already definitely assigned:
+	- Within case labels for the same switch section (includes `when` clauses)
+	- Within a single expression (includes `is` expressions)
+	- Within a single pattern (includes `switch` expression arms)
+	- Across top-level condition expressions within a single `if` statement (includes `else if`)
```
alrz, line 60: >Within a single expression (includes `is` expressions)

:memo: This should exclude `(e is bool x || e switch { bool x => x })` because there's a nested scope within switch.

>Across top-level condition expressions within a single `if` statement 

:memo: This should exclude `if (e is not int x) if (e is int x)` even though embedded statements won't introduce a new scope.

333fred (approved):

333fred:
@jcouv, did you want to look at this?

jcouv (approved):

PR #6042, opened 2022-04-20
Ref fields update

jaredpar:
Updates based on discussions and LDM meeting

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -426,9 +426,10 @@ Detailed Notes:
 - A `ref` field cannot be declared `static`, `volatile` or `const`
 - The reference assembly generation process must preserve the presence of a `ref` field inside a `ref struct` 
 - A `readonly ref struct` must declare its `ref` fields as `readonly ref`
-- The span safety rules
-    - Will be updated as outlined in this document.
-    - Will take effect whenever it's determined the core library supports `ref` fields
+- The span safety rules document will be updated as outlined in this document.
+- The new span safety rules will be in effect when either 
+    - The core library contains the feature flag indicating support for `ref` fields
+    - The `langversion` value is 11 or higher
 
 ### Provide unscoped
 One of the most notable friction points is the inability to return fields by `ref` in instance members of a `struct`. This means developers can't create `ref` returning methods / properties and have to resort to exposing fields directly. This reduces the usefulness of `ref` returns in `struct` where it is often the most desired. 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -173,6 +173,8 @@ ref struct ReadOnlyExample
 
 A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today.
 
+A `readonly ref` will be emitted to metadata using the `initonly` flag, same as any other field. A `ref readonly` field will be attributed with `System.Runtime.CompilerServices.IsReadOnlyAttribute`. A `readonly ref readonly` will be emitted with both items.
+
 This feature requires runtime support and changes to the ECMA spec. As such these will only be enabled when the corresponding feature flag is set in corelib. The issue tracking the exact API is tracked here https://github.com/dotnet/runtime/issues/64165
 
 The set of changes to our span safety rules necessary to allow `ref` fields is small and targeted. The rules already account for `ref` fields existing and being consumed from APIs. The changes need to focus on only two aspects: how they are created and how they are ref re-assigned. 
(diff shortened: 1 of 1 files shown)


jaredpar:
@cston, @RikkiGibson PTAL

proposals/low-level-struct-improvements.md
```diff
@@ -916,6 +918,17 @@ This will work but is likely going to result in unnecessary code generation. One
 1. `Unsafe.AsRef<T>(in T value)` could expand its existing purpose by changing to `scoped in T value`. This would allow it to both remove `in` and `scoped` from parameters. It then becomes the universal "remove ref safety" method
 2. Introduce a new method whose entire purpose is to remove `scoped`: `ref T Unsafe.AsUnscoped<T>(scoped in T value)`. This removes `in` as well because if it did not then callers still need a combination of method calls to "remove ref safety" at which point the existing solution is likely sufficient.
 
+### What will make C# 11.0
+The fetaures outlined in this document don't need to be implemented in a single pass. Instead they can be implemented in phases across several language releases in the following buckets:
```
jkotas, line 922:
```suggestion
The features outlined in this document don't need to be implemented in a single pass. Instead they can be implemented in phases across several language releases in the following buckets:
```

cston (approved):

proposals/low-level-struct-improvements.md
```diff
@@ -173,6 +173,8 @@ ref struct ReadOnlyExample
 
 A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today.
 
+A `readonly ref` will be emitted to metadata using the `initonly` flag, same as any other field. A `ref readonly` field will be attributed with `System.Runtime.CompilerServices.IsReadOnlyAttribute`. A `readonly ref readonly` will be emitted with both items.
```
RikkiGibson, line 176: It looks like this is consistent with how 'in' parameters are emitted and how 'ref readonly' parameters are planned to be emitted. Wanted to say this out loud so someone can correct me if I'm wrong. :)

proposals/low-level-struct-improvements.md
```diff
@@ -421,10 +423,12 @@ The section on `ref` field and `scoped` is long so wanted to close with a brief
...
 Detailed Notes:
 - A `ref` field can only be declared inside of a `ref struct` 
-- A `ref` field cannot be declared `static`
+- A `ref` field cannot be declared `static`, `volatile` or `const`
 - The reference assembly generation process must preserve the presence of a `ref` field inside a `ref struct` 
 - A `readonly ref struct` must declare its `ref` fields as `readonly ref`
-- The span safety rules for method invocation, fields and assignment must be updated as outlined in this document.
+- The span safety rules
+    - Will be updated as outlined in this document.
+    - Will take effect whenever it's determined the core library supports `ref` fields
```
RikkiGibson, line 431: Consider being more specific here. Are we talking about compile-time feature detection similar to what we do for DIMs?

proposals/low-level-struct-improvements.md
```diff
@@ -421,10 +423,12 @@ The section on `ref` field and `scoped` is long so wanted to close with a brief
 
 Detailed Notes:
 - A `ref` field can only be declared inside of a `ref struct` 
-- A `ref` field cannot be declared `static`
+- A `ref` field cannot be declared `static`, `volatile` or `const`
```
RikkiGibson, line 426: It might be good to include some discussion on why `volatile` is not supported here.

RikkiGibson (approved):

proposals/low-level-struct-improvements.md
jaredpar: It's different from `in` in that `ref` fields will never have a `modreq`.

proposals/low-level-struct-improvements.md
jaredpar: Updated this a bit

proposals/low-level-struct-improvements.md
jaredpar: The reason is that `volatile` only applies when the address can be accessed from multiple threads:

> volatile. specifies that the address is a volatile address (that is, it can be referenced externally to the current thread of execution)

A `ref` field cannot be accessed in such a way though because it's a stack bound value hence it's a bit non-sensical. 

Not sure where this would be put in the doc. I'm somewhat warry of continuing to grow this section.

tannergooding: > A ref field cannot be accessed in such a way though because it's a stack bound value hence it's a bit non-sensical.

I think more importantly is that it can't be accessed that way because `T*` for a struct that contains a `ref field` isn't allowed (it isn't "unmanaged"). It being a `ref struct` likewise prevents passing it around as a `void*` via `Unsafe.AsPointer`.

Some dev with enough want could use `fnptr` reinterpretation tricks to make it work and then get a `ref T` local on multiple threads, but that's probably stretching the need for language support 😄

PR #6059, opened 2022-04-25
Capture implemented native int conversions

jcouv:
We've reviewed this PR with @cston and @tannergooding. Our conclusion is that the roslyn implementation is correct, except for:
- the `nuint -> double` and `nuint -> float` cases (https://github.com/dotnet/roslyn/issues/60714)
- `nuint->nint` (https://github.com/dotnet/roslyn/issues/60944)

### DIFF
--- a/proposals/csharp-9.0/native-integers.md
+++ b/proposals/csharp-9.0/native-integers.md
@@ -46,31 +46,32 @@ The tables below cover the conversions between special types.
 | Operand | Target | Conversion | IL |
 |:---:|:---:|:---:|:---:|
 | `object` | `nint` | Unboxing | `unbox` |
-| `void*` | `nint` | PointerToVoid | `conv.i` |
+| `void*` | `nint` | PointerToVoid | `conv.i`/`conv.ovf.i.un` |
 | `sbyte` | `nint` | ImplicitNumeric | `conv.i` |
 | `byte` | `nint` | ImplicitNumeric | `conv.u` |
 | `short` | `nint` | ImplicitNumeric | `conv.i` |
 | `ushort` | `nint` | ImplicitNumeric | `conv.u` |
 | `int` | `nint` | ImplicitNumeric | `conv.i` |
-| `uint` | `nint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `uint` | `nint` | ExplicitNumeric | `conv.u` / `conv.ovf.i.un` |
 | `long` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
-| `ulong` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
-| `char` | `nint` | ImplicitNumeric | `conv.i` |
+| `ulong` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i.un` |
+| `char` | `nint` | ImplicitNumeric | `conv.u` |
 | `float` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
 | `double` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
 | `decimal` | `nint` | ExplicitNumeric | `long decimal.op_Explicit(decimal) conv.i` / `... conv.ovf.i` |
 | `IntPtr` | `nint` | Identity | |
 | `UIntPtr` | `nint` | None | |
+| | | | |
 | `object` | `nuint` | Unboxing | `unbox` |
 | `void*` | `nuint` | PointerToVoid | `conv.u` |
-| `sbyte` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `sbyte` | `nuint` | ExplicitNumeric | `conv.i` / `conv.ovf.u` |
 | `byte` | `nuint` | ImplicitNumeric | `conv.u` |
-| `short` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `short` | `nuint` | ExplicitNumeric | `conv.i` / `conv.ovf.u` |
 | `ushort` | `nuint` | ImplicitNumeric | `conv.u` |
-| `int` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `int` | `nuint` | ExplicitNumeric | `conv.i` / `conv.ovf.u` |
 | `uint` | `nuint` | ImplicitNumeric | `conv.u` |
 | `long` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
-| `ulong` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `ulong` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u.un` |
 | `char` | `nuint` | ImplicitNumeric | `conv.u` |
 | `float` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
 | `double` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
@@ -83,7 +84,7 @@ The tables below cover the conversions between special types.
 | Operand | Target | Conversion | IL |
 |:---:|:---:|:---:|:---:|
 | `nint` | `object` | Boxing | `box` |
-| `nint` | `void*` | PointerToVoid | `conv.i` |
+| `nint` | `void*` | PointerToVoid | `conv.i`/`conv.ovf.u` |
 | `nint` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
 | `nint` | `sbyte` | ExplicitNumeric | `conv.i1` / `conv.ovf.i1` |
 | `nint` | `byte` | ExplicitNumeric | `conv.u1` / `conv.ovf.u1` |
@@ -92,29 +93,30 @@ The tables below cover the conversions between special types.
 | `nint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4` |
 | `nint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4` |
 | `nint` | `long` | ImplicitNumeric | `conv.i8` / `conv.ovf.i8` |
-| `nint` | `ulong` | ExplicitNumeric | `conv.i8` / `conv.ovf.i8` |
+| `nint` | `ulong` | ExplicitNumeric | `conv.i8` / `conv.ovf.u8` |
 | `nint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2` |
 | `nint` | `float` | ImplicitNumeric | `conv.r4` |
 | `nint` | `double` | ImplicitNumeric | `conv.r8` |
-| `nint` | `decimal` | ImplicitNumeric | `conv.i8 decimal decimal.op_Implicit(long)` |
+| `nint` | `decimal` | ImplicitNumeric | `decimal decimal.op_Implicit(long)` |
 | `nint` | `IntPtr` | Identity | |
 | `nint` | `UIntPtr` | None | |
 | `nint` |Enumeration|ExplicitEnumeration||
+| | | | |
 | `nuint` | `object` | Boxing | `box` |
 | `nuint` | `void*` | PointerToVoid | `conv.u` |
-| `nuint` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
-| `nuint` | `sbyte` | ExplicitNumeric | `conv.i1` / `conv.ovf.i1` |
-| `nuint` | `byte` | ExplicitNumeric | `conv.u1` / `conv.ovf.u1` |
-| `nuint` | `short` | ExplicitNumeric | `conv.i2` / `conv.ovf.i2` |
-| `nuint` | `ushort` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2` |
-| `nuint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4` |
-| `nuint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4` |
-| `nuint` | `long` | ExplicitNumeric | `conv.i8` / `conv.ovf.i8` |
-| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.ovf.u8` |
+| `nuint` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i.un` |
+| `nuint` | `sbyte` | ExplicitNumeric | `conv.i1` / `conv.ovf.i1.un` |
+| `nuint` | `byte` | ExplicitNumeric | `conv.u1` / `conv.ovf.u1.un` |
+| `nuint` | `short` | ExplicitNumeric | `conv.i2` / `conv.ovf.i2.un` |
+| `nuint` | `ushort` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
+| `nuint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4.un` |
+| `nuint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4.un` |
+| `nuint` | `long` | ExplicitNumeric | `conv.u8` / `conv.ovf.i8.un` |
+| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.u8` |
 | `nuint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
-| `nuint` | `float` | ImplicitNumeric | `conv.r.un conv.r4` |
-| `nuint` | `double` | ImplicitNumeric | `conv.r.un conv.r8` |
-| `nuint` | `decimal` | ImplicitNumeric | `conv.u8 decimal decimal.op_Implicit(ulong)` |
+| `nuint` | `float` | ImplicitNumeric | `conv.r4` |
+| `nuint` | `double` | ImplicitNumeric | `conv.r8` |
+| `nuint` | `decimal` | ImplicitNumeric | `decimal decimal.op_Implicit(ulong)` |
 | `nuint` | `IntPtr` | None | |
 | `nuint` | `UIntPtr` | Identity | |
 | `nuint` |Enumeration|ExplicitEnumeration||
--- a/proposals/csharp-9.0/native-integers.md
+++ b/proposals/csharp-9.0/native-integers.md
@@ -56,7 +56,7 @@ General notes:
   - `conv.u*` for `any to unsigned *` (where * is the target width)
 - Taking a few examples:
   - `sbyte->nint` and `sbyte->nuint` use `conv.i` while `byte->nint` and `byte->nuint` use `conv.u` because they are all **widening**.
-  - `nint->byte` and `nuint->byte` use `conv.u1` while `nint->sbyte` and `nuint->sbyte` use `conv.i1` as a _pre-existing semantic_. For `byte`, `sbyte`, `short`, and `ushort` the "stack type" is `int32`. So `conv.i1` is effectively "downcast to a signed byte and then sign-extend up to int32" while `conv.u1` is effectively "downcast to an unsigned byte and then zero-extend up to int32"
+  - `nint->byte` and `nuint->byte` use `conv.u1` while `nint->sbyte` and `nuint->sbyte` use `conv.i1`. For `byte`, `sbyte`, `short`, and `ushort` the "stack type" is `int32`. So `conv.i1` is effectively "downcast to a signed byte and then sign-extend up to int32" while `conv.u1` is effectively "downcast to an unsigned byte and then zero-extend up to int32"
   - `checked(void*->nint)` uses `conv.ovf.i.un` the same way that `checked(void*->long)` uses `conv.ovf.i8.un`.
 
 | Operand | Target | Conversion | IL |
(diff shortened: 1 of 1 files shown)


proposals/csharp-9.0/native-integers.md
```diff
@@ -92,29 +93,30 @@ The tables below cover the conversions between special types.
 | `nint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4` |
 | `nint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4` |
 | `nint` | `long` | ImplicitNumeric | `conv.i8` / `conv.ovf.i8` |
-| `nint` | `ulong` | ExplicitNumeric | `conv.i8` / `conv.ovf.i8` |
+| `nint` | `ulong` | ExplicitNumeric | `conv.i8` / `conv.ovf.u8` |
 | `nint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2` |
 | `nint` | `float` | ImplicitNumeric | `conv.r4` |
 | `nint` | `double` | ImplicitNumeric | `conv.r8` |
-| `nint` | `decimal` | ImplicitNumeric | `conv.i8 decimal decimal.op_Implicit(long)` |
+| `nint` | `decimal` | ImplicitNumeric | `decimal decimal.op_Implicit(long)` |
```
cston, line 100: Why drop the `conv.i8` from `nint -> decimal`?

proposals/csharp-9.0/native-integers.md
```diff
@@ -92,29 +93,30 @@ The tables below cover the conversions between special types.
...
+| `nuint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4.un` |
+| `nuint` | `long` | ExplicitNumeric | `conv.u8` / `conv.ovf.i8.un` |
+| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.u8` |
 | `nuint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
-| `nuint` | `float` | ImplicitNumeric | `conv.r.un conv.r4` |
-| `nuint` | `double` | ImplicitNumeric | `conv.r.un conv.r8` |
-| `nuint` | `decimal` | ImplicitNumeric | `conv.u8 decimal decimal.op_Implicit(ulong)` |
+| `nuint` | `float` | ImplicitNumeric | `conv.r4` |
+| `nuint` | `double` | ImplicitNumeric | `conv.r8` |
+| `nuint` | `decimal` | ImplicitNumeric | `decimal decimal.op_Implicit(ulong)` |
```
cston, line 119: Why drop `conv.u8` from `nuint -> decimal`?

proposals/csharp-9.0/native-integers.md
jcouv: Oops. It's indeed there in implementation. Restored both. Thanks

cston (approved):

jcouv:
@cston I added some general notes. Please take another look.

We've reviewed this PR with @cston and @tannergooding. Our conclusion is that the roslyn implementation is correct, except for a tracked issue in the `nuint -> double` and `nuint -> float` cases (https://github.com/dotnet/roslyn/issues/60714).

proposals/csharp-9.0/native-integers.md
```diff
@@ -92,28 +93,29 @@ The tables below cover the conversions between special types.
...
-| `nuint` | `long` | ExplicitNumeric | `conv.i8` / `conv.ovf.i8` |
-| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.ovf.u8` |
+| `nuint` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i.un` |
+| `nuint` | `sbyte` | ExplicitNumeric | `conv.i1` / `conv.ovf.i1.un` |
+| `nuint` | `byte` | ExplicitNumeric | `conv.u1` / `conv.ovf.u1.un` |
+| `nuint` | `short` | ExplicitNumeric | `conv.i2` / `conv.ovf.i2.un` |
+| `nuint` | `ushort` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
+| `nuint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4.un` |
+| `nuint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4.un` |
+| `nuint` | `long` | ExplicitNumeric | `conv.u8` / `conv.ovf.i8.un` |
```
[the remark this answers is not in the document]
cston: I suspect `conv.i8` would sign-extend the `nuint` value on 32-bit platforms, which would be incorrect.

proposals/csharp-9.0/native-integers.md
```diff
@@ -92,28 +97,29 @@ The tables below cover the conversions between special types.
...
+| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.u8` |
 | `nuint` | `char` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
-| `nuint` | `float` | ImplicitNumeric | `conv.r.un conv.r4` |
-| `nuint` | `double` | ImplicitNumeric | `conv.r.un conv.r8` |
+| `nuint` | `float` | ImplicitNumeric | `conv.r4` |
+| `nuint` | `double` | ImplicitNumeric | `conv.r8` |
```
cston, line 122: Let's keep the `conv.r.un` for these cases, since that is needed (see https://github.com/dotnet/roslyn/issues/60714). (I didn't realize the table included `conv.r.un` when we discussed this earlier.)

jcouv: Restored

proposals/csharp-9.0/native-integers.md
```diff
@@ -43,34 +43,39 @@ There is an identity conversion between compound types that differ by native int
...
 
+General notes:
+- Most docs explain `conv.u` as conversion to __native unsigned integer__ and `conv.i` as conversion to __native integer__. But a more appropriate description is that `conv.u` is zero-extending and `conv.i` is sign-extending (if the most significant bit is `1`, then the conversion fills with `1`s).
+- Some non-obvious choices are guided by pre-existing semantics. For example, `sbyte->nint` uses `conv.i` and `byte->nint` uses `conv.u`, the same way that `sbyte->long` uses `conv.i8` and `byte->long` uses `conv.u8`. Also, `checked(void*->nint)` uses `conv.ovf.i.un` the same way that `checked(void*->long)` uses `conv.ovf.i8.un`.
+
 | Operand | Target | Conversion | IL |
 |:---:|:---:|:---:|:---:|
 | `object` | `nint` | Unboxing | `unbox` |
-| `void*` | `nint` | PointerToVoid | `conv.i` |
+| `void*` | `nint` | PointerToVoid | `conv.i`/`conv.ovf.i.un` |
```
cston, line 53: Minor: Considering adding spaces around `/` for consistency with the other cases.

proposals/csharp-9.0/native-integers.md
```diff
@@ -83,7 +88,7 @@ The tables below cover the conversions between special types.
 | Operand | Target | Conversion | IL |
 |:---:|:---:|:---:|:---:|
 | `nint` | `object` | Boxing | `box` |
-| `nint` | `void*` | PointerToVoid | `conv.i` |
+| `nint` | `void*` | PointerToVoid | `conv.i`/`conv.ovf.u` |
```
cston, line 91: Consider adding spaces around `/` for consistency.

cston (approved):

proposals/csharp-9.0/native-integers.md
```diff
@@ -43,34 +43,39 @@ There is an identity conversion between compound types that differ by native int
 The tables below cover the conversions between special types.
 (The IL for each conversion includes the variants for `unchecked` and `checked` contexts if different.)
 
+General notes:
+- Most docs explain `conv.u` as conversion to __native unsigned integer__ and `conv.i` as conversion to __native integer__. But a more appropriate description is that `conv.u` is zero-extending and `conv.i` is sign-extending (if the most significant bit is `1`, then the conversion fills with `1`s).
+- Some non-obvious choices are guided by pre-existing semantics. For example, `sbyte->nint` uses `conv.i` and `byte->nint` uses `conv.u`, the same way that `sbyte->long` uses `conv.i8` and `byte->long` uses `conv.u8`. Also, `checked(void*->nint)` uses `conv.ovf.i.un` the same way that `checked(void*->long)` uses `conv.ovf.i8.un`.
```
tannergooding, line 48: I think you've got this inversed.

`sbyte->nint` and `sbyte->nuint` use `conv.i` while `byte->nint` and `byte->nuint` use `conv.u` because they are all **widening** and therefore the `i` effectively means `sign-extend` while the `u` effectively means `zero-extend`.

`nint->byte` and `nuint->byte` use `conv.u1` while `nint->sbyte` and `nuint->sbyte` use `conv.i1` as a `pre-existing semantic`. These are both **narrowing** and so there isn't a specific need to pick `signed` (`i1`) or `unsigned` (`u1`), its just a semantic of what's already done.

The overflow cases do have semantics that are important and that's `conv.ovf.*` where the input is treated as `signed` and the output is `*` and `conv.ovf.*.un` where the input is treated as `unsigned` and the output is `*`.

tannergooding: So basically, `checked` contexts for both **widening** and **narrowing** are:
* `conv.ovf.*` for `signed to *`
* `conv.ovf.*.un` for `unsigned to *`

While, `unchecked` contexts for **widening** are:
* `conv.i*` for `signed to *` (where * is the target width)
* `conv.u*` for `unsigned to *` (where * is the target width)

And, `unchecked` contexts for **narrowing** are:
* `conv.i*` for `any to signed *` (where * is the target width)
* `conv.u*` for `any to unsigned *` (where * is the target width)

proposals/csharp-9.0/native-integers.md
tannergooding: Why do we emit `conv.i` for `void*` to `nint`?

Assuming `void*` is treated as unsigned; then this should be similar to `uint to int` or `ulong to long` and therefore be a `nop`, correct?

You can see for https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgZgAIS6BhOgbyrs4fuwDsM6AWQAUAVz4CEASjoBeAHx1hEqQgDcHLrTq4IvAOZCxug3WlzFwk/tUbKXbnV4SjvcfzMyFS5/1ubOAMcXADkxF3NvMAALGDAAaxgAE2FlP2k7B21rOjDRHMjFGLjElKs9Gwyg7V8BMLcIryLYhOTU2tUpOwBfIA===:
* `unchecked uint to int` is a `nop`
* `unchecked ulong to long` is a `nop`
* `unchecked nuint to nint` is a `conv.i`
*
* `unchecked uint to int` is a `conv.ovf.i4.un`
* `unchecked ulong to long` is a `conv.ovf.i8.un`
* `unchecked nuint to nint` is a `conv.ovf.i.un`

So `nuint to nint` is the "inconsistent" thing.

proposals/csharp-9.0/native-integers.md
```diff
@@ -43,34 +43,39 @@ There is an identity conversion between compound types that differ by native int
...
-| `uint` | `nint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `uint` | `nint` | ExplicitNumeric | `conv.u` / `conv.ovf.i.un` |
 | `long` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
-| `ulong` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
-| `char` | `nint` | ImplicitNumeric | `conv.i` |
+| `ulong` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i.un` |
+| `char` | `nint` | ImplicitNumeric | `conv.u` |
 | `float` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
 | `double` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i` |
 | `decimal` | `nint` | ExplicitNumeric | `long decimal.op_Explicit(decimal) conv.i` / `... conv.ovf.i` |
```
tannergooding, line 65: Is this one where we should expose `nint decimal.op_Explicit(decimal)` and `nint decimal.op_CheckedExplicit(decimal)` explicitly for the language to take advantage of?

proposals/csharp-9.0/native-integers.md
```diff
@@ -43,34 +43,39 @@ There is an identity conversion between compound types that differ by native int
...
 | `IntPtr` | `nint` | Identity | |
 | `UIntPtr` | `nint` | None | |
```
tannergooding, line 67: Is this table missing the following?
| `nuint` | `nint` | ExplicitNumeric | None / `conv.ovf.i.un`  |

proposals/csharp-9.0/native-integers.md
```diff
@@ -43,34 +43,39 @@ There is an identity conversion between compound types that differ by native int
...
+| | | | |
 | `object` | `nuint` | Unboxing | `unbox` |
 | `void*` | `nuint` | PointerToVoid | `conv.u` |
```
tannergooding, line 70: Similar question to https://github.com/dotnet/csharplang/pull/6059/files#r858012318

Why do we emit `conv.u` for `void* to nuint`?

Based on the other conversions, it looks like `void*` is treated as `unsigned` and therefore this should effectively be a "nop" (unless there is some weird stack evaluation requirement I'm missing).

proposals/csharp-9.0/native-integers.md
```diff
@@ -43,34 +43,39 @@ There is an identity conversion between compound types that differ by native int
...
-| `sbyte` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `sbyte` | `nuint` | ExplicitNumeric | `conv.i` / `conv.ovf.u` |
 | `byte` | `nuint` | ImplicitNumeric | `conv.u` |
-| `short` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `short` | `nuint` | ExplicitNumeric | `conv.i` / `conv.ovf.u` |
 | `ushort` | `nuint` | ImplicitNumeric | `conv.u` |
-| `int` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
+| `int` | `nuint` | ExplicitNumeric | `conv.i` / `conv.ovf.u` |
 | `uint` | `nuint` | ImplicitNumeric | `conv.u` |
 | `long` | `nuint` | ExplicitNumeric | `conv.u` / `conv.ovf.u` |
```
tannergooding, line 77: Just commenting that this one makes sense because on a 32-bit platform, its a narrowing and so `conv.u` makes sense.

While on a 64-bit platform, it'd be the same as `nint -> nuint` which would be a `nop`. Since we can't use a `nop` due to 32-bit, keeping `conv.u` then makes the most sense.

proposals/csharp-9.0/native-integers.md
tannergooding: Same as for the other two, in an `unchecked` context I think this could just be a `nop`.

tannergooding: And same for `nint to nuint` right below on L92

proposals/csharp-9.0/native-integers.md
```diff
@@ -92,25 +97,26 @@ The tables below cover the conversions between special types.
 | `nint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4` |
 | `nint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4` |
 | `nint` | `long` | ImplicitNumeric | `conv.i8` / `conv.ovf.i8` |
```
tannergooding, line 99: `nint to long` cannot overflow. It is either a sign-extension (on 32-bit) or the same size (on 64-bit).

Unless the language is trying to future proof for some .NET where pointers could be larger than 64-bit (in which case the entire BCL and more would need to update to handle this stuff...), then I think it can just be `conv.i8`

proposals/csharp-9.0/native-integers.md
```diff
@@ -92,25 +97,26 @@ The tables below cover the conversions between special types.
...
-| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.ovf.u8` |
+| `nuint` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i.un` |
+| `nuint` | `sbyte` | ExplicitNumeric | `conv.i1` / `conv.ovf.i1.un` |
+| `nuint` | `byte` | ExplicitNumeric | `conv.u1` / `conv.ovf.u1.un` |
+| `nuint` | `short` | ExplicitNumeric | `conv.i2` / `conv.ovf.i2.un` |
+| `nuint` | `ushort` | ExplicitNumeric | `conv.u2` / `conv.ovf.u2.un` |
+| `nuint` | `int` | ExplicitNumeric | `conv.i4` / `conv.ovf.i4.un` |
+| `nuint` | `uint` | ExplicitNumeric | `conv.u4` / `conv.ovf.u4.un` |
+| `nuint` | `long` | ExplicitNumeric | `conv.u8` / `conv.ovf.i8.un` |
+| `nuint` | `ulong` | ImplicitNumeric | `conv.u8` / `conv.u8` |
```
tannergooding, line 119: This should just be `conv.u8` for consistency with the other entries when checked/unchecked is the same.

proposals/csharp-9.0/native-integers.md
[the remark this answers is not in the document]
tannergooding: Right. On a 32-bit platform this needs to be the same as `uint to long` which would be a `zero-extension`. While on 64-bit platforms, this is a `ulong to long` which is a "nop".

This means we need `conv.u8` on both to be correct. This is also why `nint to long` and `nint to ulong` are `conv.i8`

proposals/csharp-9.0/native-integers.md
tannergooding: > so there isn't a specific need to pick signed (i1) or unsigned (u1), its just a semantic of what's already done.

After re-reading the ECMA-335 bit, there is actually an important semantic here for **narrowing**.

For `byte`, `sbyte`, `short`, and `ushort` the "stack type" is `int32`. So `conv.i1` is effectively "downcast to a signed byte and then sign-extend up to int32" while `conv.u1` is effectively "downcast to an unsigned byte and then zero-extend up to int32"

proposals/csharp-9.0/native-integers.md
cston: It looks like the table is incorrect. The implementation uses `conv.i8` even in a `checked` context.

proposals/csharp-9.0/native-integers.md
jcouv: We indeed emit nop for `void*->nint`. I'll update the table to reflect.

jcouv: Filed [https://github.com/dotnet/roslyn/issues/60944](https://github.com/dotnet/roslyn/issues/60944) for `nuint->nint` (should be `nop` instead of `conv.i`)

proposals/csharp-9.0/native-integers.md
jcouv: `nuint->nint` is covered on line 111:

> | `nuint` | `nint` | ExplicitNumeric | `conv.i` / `conv.ovf.i.un` |

proposals/csharp-9.0/native-integers.md
jcouv: Will fix the table.
The implementation indeed uses `conv.i8` for both unchecked and checked:
`            conversions(sourceType: "nint", destType: "long", ImplicitNumeric, expectedImplicitIL: conv("conv.i8"), expectedExplicitIL: conv("conv.i8"));`

proposals/csharp-9.0/native-integers.md
jcouv: That seems sensible, but my inclination is to keep the `decimal->nint` conversion as-is. 
I don't think the complexity of conditionally converting with either `long decimal.op_Explicit(decimal) conv.i` or  `nint decimal.op_Explicit(decimal)` is worthwhile.

jcouv:
@tannergooding Thanks for the thorough review. I addressed your feedback and tried to incorporate your summary. Please take another look.

proposals/csharp-9.0/native-integers.md
```diff
@@ -45,12 +45,25 @@ The tables below cover the conversions between special types.
...
+  - `conv.ovf.*.un` for `unsigned to *`
+- `unchecked` contexts for **widening** are:
+  - `conv.i*` for `signed to *` (where * is the target width)
+  - `conv.u*` for `unsigned to *` (where * is the target width)
+- `unchecked` contexts for **narrowing** are:
+  - `conv.i*` for `any to signed *` (where * is the target width)
+  - `conv.u*` for `any to unsigned *` (where * is the target width)
+- Taking a few examples:
+  - `sbyte->nint` and `sbyte->nuint` use `conv.i` while `byte->nint` and `byte->nuint` use `conv.u` because they are all **widening**.
+  - `nint->byte` and `nuint->byte` use `conv.u1` while `nint->sbyte` and `nuint->sbyte` use `conv.i1` as a _pre-existing semantic_. These are both **narrowing** and so there isn't a specific need to pick `signed` (`i1`) or `unsigned` (`u1`), its just a semantic of what's already done.
```
tannergooding: This bit should be reworded based on what I put as a follow up: https://github.com/dotnet/csharplang/pull/6059#discussion_r858027728

The semantic is that the stack type is `int32` not `byte` or `sbyte` and so we have to convert to the target sign so when its widened to `int32` on the stack, the value is still correct

tannergooding (approved):

cston (approved):

PR #6062, opened 2022-04-26
Update ref fields for restricted types

jaredpar:
This PR has two changes:

1. Moving compat considerations into considerations section. 
2. Detail how restricted types will be removed from the compiler

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -93,7 +93,7 @@ A `ref` field can be combined with `readonly` modifiers in the following ways:
 - `ref readonly`: this is a field that can be ref re-assigned but cannot be value assigned at any point. This how an `in` parameter could be ref re-assigned to a `ref` field.
 - `readonly ref readonly`: a combination of `ref readonly` and `readonly ref`. 
 
-```
+```c#
 ref struct ReadOnlyExample
 {
     ref readonly int Field1;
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -431,6 +372,22 @@ Detailed Notes:
...
+Once `ref` fields are available these types can be correctly defined in C# using a combination of `ref struct` and `ref` fields. Therefore when the compiler detects that a runtime supports `ref` fields it will no longer have a notion of restricted types. It will instead use the types as they are defined in the code. 
+
+To support this our span safety rules will be updated as follows:
+
+- `__makeref` will be treated as a method with the signature `TypedReference __makeref<T>(ref T value)`
+- `__refvalue` will be treated as a method with the signature `ref T __refvalue<T>(TypedReference tr)`. The expression `__refvalue(tr, int)` will effectively use the second argument as the type parameter.
+- `__arglist` as a parameter will have a *ref-safe-to-escape* and *safe-to-escape* of *current method*. 
+- `__arglist(...)` as an expression will have a *ref-safe-to-escape* and *safe-to-escape* of *current method*. 
+
+Conforming runtimes will ensure that `TypedReference`, `RuntimeArgumentHandle` and `ArgIterator` are defined as `ref struct`. That combined with the above rules will ensure references to the stack do escape beyond their lifetime.
```
AaronRobinsonMSFT, line 387:
```suggestion
Conforming runtimes will ensure that `TypedReference`, `RuntimeArgumentHandle` and `ArgIterator` are defined as `ref struct`. That combined with the above rules will ensure references to the stack do not escape beyond their lifetime.
```

proposals/low-level-struct-improvements.md
```diff
@@ -579,6 +536,65 @@ The backing storage for the buffer will be generated using the `[InlineArray]` a
 ## Considerations
 There are considerations other parts of the development stack should consider when evaluating this feature.
 
+### Compat Considerations
+The challenge in this proposal is the compatibility implications this design has to our existing [span safety rules](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md). While those rules fully support the concept of a `ref struct` having `ref` fields they do not allow for APIs, other than `stackalloc`, to capture `ref` state that refers to the stack. The span safety rules have a [hard assumption](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md#span-constructor) that a constructor of the form `Span(ref T value)` does not exist. That means the safety rules do not account for a `ref` parameter being able to escape as a `ref` field hence it allows for code like the following.
```
AaronRobinsonMSFT, line 540:
```suggestion
The challenge in this proposal is the compatibility implications this design has to our existing [span safety rules](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md). While those rules fully support the concept of a `ref struct` having `ref` fields they do not allow for APIs, other than `stackalloc`, to capture `ref` state that refers to the stack. The span safety rules have a [hard assumption](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md#span-constructor) that a constructor of the form `Span(ref T value)` does not exist. That means the safety rules do not account for a `ref` parameter being able to escape as a `ref` field hence it allows for code like the following.
```

proposals/low-level-struct-improvements.md
```diff
@@ -431,6 +372,22 @@ Detailed Notes:
     - The core library contains the feature flag indicating support for `ref` fields
     - The `langversion` value is 11 or higher
 
+### Sunset restricted types
+The compiler has a concept of a set of "restricted types" which is largely undocumented. These types were given a special status because in C# 1.0 there was no general purpose way to express their behavior. Most notably the fact that the types can contain references to the execution stack. Instead the compiler had special knowledge of them and restricted their usage in ways that would always be safe: disallowed returns, cannot use as array elements, cannot use in generics, etc ...
```
cston, line 376: >restricted their usage in ways that would always be safe [](http://example.com/codeflow?start=346&length=56)

Perhaps "restricted their **use to** ways that would always be safe" or "prevented their use in ways that would be unsafe".

proposals/low-level-struct-improvements.md
```diff
@@ -431,6 +372,22 @@ Detailed Notes:
...
+
+Once `ref` fields are available these types can be correctly defined in C# using a combination of `ref struct` and `ref` fields. Therefore when the compiler detects that a runtime supports `ref` fields it will no longer have a notion of restricted types. It will instead use the types as they are defined in the code. 
+
+To support this our span safety rules will be updated as follows:
+
+- `__makeref` will be treated as a method with the signature `TypedReference __makeref<T>(ref T value)`
```
cston, line 382: >TypedReference  [](http://example.com/codeflow?start=62&length=15)

`static TypedReference ...`?

proposals/low-level-struct-improvements.md
```diff
@@ -431,6 +372,22 @@ Detailed Notes:
...
+- `__refvalue` will be treated as a method with the signature `ref T __refvalue<T>(TypedReference tr)`. The expression `__refvalue(tr, int)` will effectively use the second argument as the type parameter.
```
cston, line 383: >`ref T  [](http://example.com/codeflow?start=62&length=7)

`static ref T ...`?

cston (approved):

RikkiGibson:
I think it would be valuable if this proposal could mention the way `TypedReference` and possibly the other formerly-restricted structs will be declared. e.g.

```cs
readonly ref struct TypedReference {
  readonly ref byte _pointer;
  readonly IntPtr _type;
} 
```

It seems very key that the 'ref byte' is able to keep whatever is referenced alive, even if it's "not really a byte" that's being referenced--the GC just "does the right thing" based on the address that is being referenced, without needing to be given a correct static type for the reference.

PR #6116, opened 2022-05-12
Describe generic lifetimes and a proposed mapping for new syntax

agocke:
The existing ref field proposals have to formalize new safety rules. This is more of a spec proposal than a language proposal. If the language introduces an explicit notion of generic lifetimes, it may be significantly easier to verify the safety and correctly implement the feature than attempting to define smaller but more complicated rules.

### DIFF
--- a/proposals/ref-lifetimes.md
+++ b/proposals/ref-lifetimes.md
@@ -239,17 +239,17 @@ Span<int> BadUseExamples(int parameter)
 // Introduce a new lifetime for the parameter, but don't constrain it to the
 // local scope.
 Span<'a, int> CreateSpan<'a, 'b>(ref<'b> int parameter)
-Span<'a, int> BadUseExamples<'a>(int parameter)
+Span<'global, int> BadUseExamples(int parameter)
 {
-    return CreateSpan<'global, '_>(ref parameter);
+    return CreateSpan(ref parameter);
 
     // Legal in C# 10 and legal in C# 11 due to scoped ref
     int local = 42;
-    return CreateSpan<'global, '_>(ref local);
+    return CreateSpan(ref local);
 
     // Legal in C# 10 and legal in C# 11 due to scoped ref
     Span<int> span = stackalloc int[42];
-    return CreateSpan<'global, '_>(ref span[0]);
+    return CreateSpan(ref span[0]);
 }
 ```
 
--- a/proposals/ref-lifetimes.md
+++ b/proposals/ref-lifetimes.md
@@ -238,7 +238,7 @@ Span<int> BadUseExamples(int parameter)
 
 // Introduce a new lifetime for the parameter, but don't constrain it to the
 // local scope.
-Span<'a, int> CreateSpan<'a, 'b>(ref<'b> int parameter)
+Span<'global, int> CreateSpan<'a>(ref<'a> int parameter)
 Span<'global, int> BadUseExamples(int parameter)
 {
     return CreateSpan(ref parameter);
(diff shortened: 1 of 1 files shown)


agocke:
FYI @jaredpar @MadsTorgersen, this is my attempt at defining lifetimes as a type system feature, and then reducing the proposed rules to it. Not many problems for the most part, although the description of "scoped" as meaning "current lifetime" looks pretty strange if you take it literally, since it doesn't really make sense for a value coming from outside the method to have a lifetime local to the current method.

hez2010:
Cheers! 
The original `scoped` with its coming rules is really hard to understand. The original proposed simplified version of lifetime annotation `scoped` intends to simplify lifetime usage but it instead making it more complicated and confusing.
Looking forward to explicit lifetime annotations as proposed in C# 11 :)

proposals/ref-lifetimes.md
```diff
@@ -0,0 +1,294 @@
...
+longer than the shortest lifetime, there would be no valid lifetime across all call sites.
+
+Instead, we can introduce generic lifetimes, which allow each call site to substitute a given
+lifetime for each lifetime parameter. To represent such a scheme, we must introduce a syntax for
+specifying the lifetime of a ref variable. To differentiate from conventional generics, we will add
+the generic parameters as belonging to the `ref` keyword for ref variables, instead of the type
+syntax. We will also use the `'` token to prefix the names of each parameter. For instance,
+
+```C#
+ref<'a> T M<'a,'b, T>(ref<'a> T p1, ref<'b> int p2) { ... }
```
jaredpar, line 32: I think the generic lifetime is better suited for being on the `scoped` modifier, not `ref`. The rationale being that when you have a `ref` on a `ref struct` type then there are actually two lifetimes involved: the lifetime of the `ref` and of the value it points to. Even in the C# 11 syntax these can be differentiated by the placement of the `scoped` modifier. 

https://raw.githubusercontent.com/dotnet/csharplang/main/proposals/low-level-struct-improvements.md#rules-scoped

Putting the lifetime on the `ref` modifier means you don't have the maximum flexibility supported by the underlying data.

agocke:
One thing I want to clarify: I don't really take any position on whether we want this syntax in the language itself. Mainly I think having the syntactical representation (and the semantic representation in the type system) of lifetimes as generics is a bug-finding tool for the *spec*. I find the type representation much simpler to prove correct than individual uses of `scoped`, so I was interested in creating this to represent whatever actual syntax we choose in the type form, so that we can prove it sound via reduction.

proposals/ref-lifetimes.md
agocke: Seems fine to me, but I don't think you need this. My proposal has two additions of where you can put lifetime annotations for values:

1. On the `ref` keyword.
2. At the beginning of a struct's (ref struct or regular) generic parameter list.

This lets you differentiate between the lifetime of the `ref`, and the lifetime of the struct/stuff inside the struct. For example,

```C#
ref struct Span<'a, T> { ... }

// lifetime on Span
Span<'a, int> M<'a>(Span<'a, int> s) { ... }

// lifetime on ref
ref<'a> Span<'global, int> M<'a>(ref<'a> Span<'global, int> s) { ... }

File: proposals/ref-lifetimes.md [hunk 2]
```diff
@@ -0,0 +1,294 @@
...
+
+```C#
+// Case 1
+ref int M1() { ... }
+// Case 2
+void M2(ref int x) { ...}
+// Case 3
+Span<int> M3() { ... }
+// Case 4
+void M4<'a>(Span<'a, int> s) { ... }
```
Clonkex, line 113: Is this a typo? Case 4 maps to case 4? (Not sure if I'm using this Github feature properly, let's see.)

agocke:
I think this is mostly obsoleted by my write-up in https://gist.github.com/agocke/dea8adbe105101b6a397925c6984428b

I do think we should translate that into a spec when we're ready to spec ref fields.

RikkiGibson:
I am hoping to make 'ref field of ref struct type' happen this year (through unsafe). Hopefully I can do it then.

PR #6154, opened 2022-05-23
Update utf8-string-literals.md based on recent LDM decisions

AlekseyTs:


proposals/utf8-string-literals.md
```diff
@@ -28,87 +28,52 @@ To fix this we will allow for UTF8 literals in the language and encode them into
...
-The predominant usage of this feature is expected to be with literals but it will work with any `string` constant value.
-A conversion from a `string` constant with `null` value will be supprted as well. The result of the conversion will be `default`
-value of the target type.
+### `u8` suffix on string literals
 
-```c#
-const string data = "dog"
-ReadOnlySpan<byte> span = data;     // new byte[] { 0x64, 0x6f, 0x67 }
-```
+Language will provide the `u8` suffix on string literals to force the type to be UTF8.
```
333fred, line 33:
```suggestion
The language will provide the `u8` suffix on string literals to force the type to be UTF8.
```

proposals/utf8-string-literals.md
```diff
@@ -28,87 +28,52 @@ To fix this we will allow for UTF8 literals in the language and encode them into
...
 
-While the inputs to these conversions are constants and the data is fully encoded at compile time, the conversion is **not** considered constant by the language. That is because arrays are not constant today. If the definition of `const` is expanded in the future to consider arrays then these conversions should also be considered. Practically though this means a result of these conversions cannot be used as the default value of an optional parameter. 
+A `u8` literal doesn't have a constant value. That is because ```ReadOnlySpan<byte>``` cannot be type of a constant today. If the definition of `const` is expanded
+in the future to consider ```ReadOnlySpan<byte>```, then this value should also be considered a constant. Practically though this means a `u8`
+literal cannot be used as the default value of an optional parameter.
 
 ```c#
 // Error: The argument is not constant
-void Write(ReadOnlySpan<byte> message = "missing") { ... } 
+void Write(```ReadOnlySpan<byte>``` message = "missing"u8) { ... }
```
333fred, line 57:
```suggestion
void Write(ReadOnlySpan<byte> message = "missing"u8) { ... } 
```

proposals/utf8-string-literals.md
```diff
@@ -154,6 +119,54 @@ It seems unlikely that we would regret the target type conversion between string
 
 It seems more likely that we'd regret the `u8` suffix pointing to `ReadOnlySpan<byte>` instead of `Utf8String`. It would be similar to how we regret that `stackalloc int[]` has a natural type of `int*` instead of `Span<int>`. This is not a deal breaker though, just an inconvenience.
 
+### Conversions between `string` constants and `byte` sequences
```
333fred, line 122: I'm a little confused by this section: is this part of the feature still happening? Or was it removed?

jcouv: nit: I'd probably just leave a one sentence note in alternatives ("We experimented with a conversion ... but decided against it"), but wouldn't keep all this detail. I got a bit confused too ;-)

jcouv (approved):
LGTM Thanks (iteration 4)

proposals/utf8-string-literals.md
```diff
@@ -28,87 +28,52 @@ To fix this we will allow for UTF8 literals in the language and encode them into
...
-```c#
-const string data = "dog"
-ReadOnlySpan<byte> span = data;     // new byte[] { 0x64, 0x6f, 0x67 }
-```
+The language will provide the `u8` suffix on string literals to force the type to be UTF8.
+The suffix is case-insensitive, `U8` suffix will be supported and will have the same meaning as `u8` suffix.
 
-In the case of any constant operation on strings, such as `+`, the encoding to UTF8 will occur on the final `string` vs. happening for the individual parts and then concatenating the results. This ordering is important to consider because it can impact whether or not the conversion succeeds. 
+When the `u8` suffix is used, the value of the literal is a ```ReadOnlySpan<byte>``` containing a UTF-8 byte representation of the string.
+A null terminator is placed beyond the last byte in memory (and outside the length of the ```ReadOnlySpan<byte>```) in order to handle some
```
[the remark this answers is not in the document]
AlekseyTs: > Is the null terminator part of the current implementation?

No, it is not.

PR #6157, opened 2022-05-24
Address issues in records feature spec

BillWagner:
Fixes dotnet/docs#28648

**Addresses dotnet/docs#28642 Comment 1.**

Use the grammar from the rejected Records proposal to describe positional records.

### DIFF
--- a/proposals/csharp-9.0/records.md
+++ b/proposals/csharp-9.0/records.md
@@ -9,13 +9,30 @@ The syntax for a record is as follows:
 ```antlr
 record_declaration
     : attributes? class_modifier* 'partial'? 'record' identifier type_parameter_list?
-      parameter_list? record_base? type_parameter_constraints_clause* record_body
+      record_parameters? record_base? type_parameter_constraints_clause* record_body
     ;
 
 record_base
     : ':' class_type argument_list?
     | ':' interface_type_list
     | ':' class_type argument_list? ',' interface_type_list
+    ;
+
+record_parameters
+    : '(' record_parameter_list? ')'
+    ;
+
+record_parameter_list
+    : record_parameter
+    | record_parameter ',' record_parameter_list
+    ;
+
+record_parameter
+    : attributes? type identifier record_property_name? default_argument?
+    ;
+
+record_property_name
+    : ':' identifier
     ;
 
 record_body


BillWagner:
I didn't address Comment 2 in the issue (as withdrawn).

I also didn't address Comment 3, per https://github.com/dotnet/docs/issues/28648#issuecomment-1136361489

proposals/csharp-9.0/records.md
```diff
@@ -9,13 +9,34 @@ The syntax for a record is as follows:
...
+    : record_parameter
+    | record_parameter ',' record_parameter_list
+    ;
+
+record_parameter
+    : attributes? type identifier record_property_name? default_argument?
+    ;
+
+record_property_name
+    : ':' identifier
```
CyrusNajmabadi, line 39: wait... this is a thing?!  how did i not know about this?

proposals/csharp-9.0/records.md
```diff
@@ -9,13 +9,34 @@ The syntax for a record is as follows:
...
-      parameter_list? record_base? type_parameter_constraints_clause* record_body
+      record_parameters? record_base? type_parameter_constraints_clause* record_body
     ;
 
 record_base
-    : ':' class_type argument_list?
+    : ':' class_type record_base_arguments?
     | ':' interface_type_list
-    | ':' class_type argument_list? ',' interface_type_list
+    | ':' class_type record_base_arguments? ',' interface_type_list
```
jcouv, line 18: Can `record_base` be simplified to the following? 
`| ':' class_type record_base_arguments? (',' interface_type_list)?`

proposals/csharp-9.0/records.md
```diff
@@ -9,13 +9,34 @@ The syntax for a record is as follows:
...
+    : '(' record_parameter_list? ')'
+    ;
+
+record_parameter_list
+    : record_parameter
+    | record_parameter ',' record_parameter_list
+    ;
+
+record_parameter
+    : attributes? type identifier record_property_name? default_argument?
```
jcouv, line 35: If we can't re-use an existing parameter list, I would at least avoid the word "property" in `record_property_name` here.

jcouv (commented):
Done with review pass (iteration 2)

proposals/csharp-9.0/records.md
BillWagner: Only as part of the `record_parameter` production. So, the `record_property_name` production indicates that the parameter name of a positional record primary constructor lowers to the property name of the record.

jcouv: Thanks @CyrusNajmabadi for calling attention to this. There is no such thing as controlling the parameter name and property name separately. Records just use a normal parameter list after the record's name.

proposals/csharp-9.0/records.md
BillWagner: Yes! making that change.

proposals/csharp-9.0/records.md
```diff
@@ -9,13 +9,34 @@ The syntax for a record is as follows:
 ```antlr
 record_declaration
     : attributes? class_modifier* 'partial'? 'record' identifier type_parameter_list?
-      parameter_list? record_base? type_parameter_constraints_clause* record_body
+      record_parameters? record_base? type_parameter_constraints_clause* record_body
```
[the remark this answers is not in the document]
BillWagner: Aha! Neither Rex nor I thought this was legal:

```csharp
public record R() // note empty parameter list.
{
    public int P { get; init; }
}
```
Yes, that makes this makes using the `parameter_list` easier.

proposals/csharp-9.0/records.md
```diff
@@ -13,9 +13,11 @@ record_declaration
...
 
 record_base
-    : ':' class_type argument_list?
-    | ':' interface_type_list
-    | ':' class_type argument_list? ',' interface_type_list
+    : ':' class_type record_base_arguments? (',' interface_type_list)?
+    ;
+
+record_base_arguments
+    : '(' argument_list? ')'
```
BillWagner, line 20: This can be further simplified because `argument_list` does include the `(` and `)`:

```suggestion
    : ':' class_type argument_list? (',' interface_type_list)?
    ;
```

BillWagner:
@jcouv This has now gotten a lot simpler, and is optional. I think the production for `record_base` is simpler in this form, but I think the new and old versions are equivalent grammars.

(And today I learned that positional records can have empty parameter lists).

Even if we don't merge this, I'll close dotnet/docs#28648. I don't see further action for that issue.

proposals/csharp-9.0/records.md
```diff
@@ -13,9 +13,7 @@ record_declaration
     ;
 
 record_base
-    : ':' class_type argument_list?
-    | ':' interface_type_list
-    | ':' class_type argument_list? ',' interface_type_list
+    : ':' (class_type argument_list?)? (',' interface_type_list)?
```
333fred, line 16: This grammar would allow `: , IInterface` (and indeed, require it).

BillWagner:
Thanks for the reviews.

I'll close this and update https://github.com/dotnet/docs/issues/28648 with comments and close it.

PR #6188, opened 2022-06-01
fix: wrong class-name in Open Questions code example

Irubataru:
The class name seems to be wrong which was a bit confusing at the first read.


333fred (approved):

333fred:
Thanks @Irubataru!

PR #6209, opened 2022-06-09
ref field updates

jaredpar:
Fixing a few issues found during the development of `ref` fields

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -301,15 +301,15 @@ The method invocation for lvalue returns can now be simplified. The receiver no 
 
 > An lvalue resulting from a ref-returning method invocation `e1.M(e2, ...)` is *ref-safe-to-escape* the smallest of the following scopes:
 > 1. The *calling method*
-> 2. The *ref-safe-to-escape* of all `ref` arguments
-> 3. The *safe-to-escape* of all argument expressions
+> 2. The *ref-safe-to-escape* contributed byall `ref` arguments
+> 3. The *safe-to-escape* contributed by all argument expressions
 
 The method invocation for rvalue returns needs to change as follows to account for `ref` field returns.
 
 > An rvalue resulting from a method invocation `e1.M(e2, ...)` is *safe-to-escape* from the smallest of the following scopes:
 > 1. The *calling method*
-> 2. The *safe-to-escape* of all argument expressions
-> 3. When the return is a `ref struct` then *ref-safe-to-escape* of all `ref` arguments
+> 2. The *safe-to-escape* conrtibuted by all argument expressions
+> 3. When the return is a `ref struct` then *ref-safe-to-escape* contributed by all `ref` arguments
 
 This rule now lets us define the two variants of desired methods:
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -112,7 +112,23 @@ ref struct ReadOnlyExample
 }
 ```
 
-A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today.
+A `readonly` on a `ref struct` will instance will force `ref` fields to be `readonly ref readonly`. This also means a `readonly ref struct` must mark `ref` fields as `readonly ref readonly`. This maintains the invariant that a `ref struct` which is `readonly` cannot store ref-like state.
+
+```c#
+ref struct S
+{
+    ref int Field;
+
+    void Uses(in S s1, ref S s2)
+    {
+        int local = default;
+        s1.Field = 42;          // Error: can't assign ref readonly value
+        s1.Field = ref loal;    // Error: can't repoint readonly ref
+        s2.Field = 42;          // Okay
+        s2.Field = ref loal;    // Okay
+    }
+}
+```
 
 A `readonly ref` will be emitted to metadata using the `initonly` flag, same as any other field. A `ref readonly` field will be attributed with `System.Runtime.CompilerServices.IsReadOnlyAttribute`. A `readonly ref readonly` will be emitted with both items.
 
@@ -932,6 +948,37 @@ struct Dimensions
 ```
 
 **Decision** Do not allow for now
+
+### How readonly impacts ref fields
+When a `ref struct` instance that has a `ref` field is in a `readonly` location there are two different ways the field could be represented: `readonly ref` and `readonly ref readonly`. Both of these interpretations are valid but they have significantly different impacts to existing span safety rules. 
+
+In the original span safety rules a `ref struct` in a `readonly` location cannot be used to store ref-like state. This is a deliberate intent of the span safety rules and one that customers take specific advantage of. For example:
+
+```c#
+readonly ref struct Reader
+{
+    readonly ref readonly Span<char> field;
+    public void Read(Span<char> span);
+}
+```
+
+Customers understand, and rely on, the fact that this API does not qualify for the method arguments must match rule. That is because a `readonly ref struct` cannot store ref-like state. If this decision was reversed though this invariant would no longer hold because it would allow for the following: 
+
+```c#
+readonly ref struct Reader
+{
+    // Consider the case the language allowed this approach
+    readonly ref Span<char> field;
+    public void Read(Span<char> span)
+    {
+        field = span;
+    }
+}
+```
+
+To support this in a safe way it would mean that `in` parameters must qualify for method arguments must match rule. That would be a potentially significant breaking change to existing code.
+
+**Decision** The language will treat `ref` fields as `readonly ref readonly` when the instance is in a `readonly` location.
 
 ### Violating scoped
 The runtime repository has several non-public APIs that capture `ref` parameters as `ref` fields. These are unsafe because the lifetime of the resulting value is not tracked. For example the `Span<T>(ref T value, int length)` constructor.
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -112,7 +112,7 @@ ref struct ReadOnlyExample
 }
 ```
 
-A `readonly` on a `ref struct` will instance will force `ref` fields to be `readonly ref readonly`. This also means a `readonly ref struct` must mark `ref` fields as `readonly ref readonly`. This maintains the invariant that a `ref struct` which is `readonly` cannot store ref-like state.
+A `readonly` on a `ref struct` instance will cause all `ref` fields to be considered `readonly ref readonly`. By extension this means that `readonly` on a `ref struct` declaration will force all `ref` fields to be declared as `ref readonly ref`. These rules maintains the invariant that a `ref struct` which is `readonly` cannot be used to store additional ref-like values.
 
 ```c#
 ref struct S
@@ -122,10 +122,13 @@ ref struct S
     void Uses(in S s1, ref S s2)
     {
         int local = default;
+        // s1 is readonly by virtue of `in` 
         s1.Field = 42;          // Error: can't assign ref readonly value
-        s1.Field = ref loal;    // Error: can't repoint readonly ref
+        s1.Field = ref local;   // Error: can't repoint readonly ref
+
+        // s2 is same type in a position that allows mutations
         s2.Field = 42;          // Okay
-        s2.Field = ref loal;    // Okay
+        s2.Field = ref local;   // Okay
     }
 }
 ```
@@ -381,10 +384,14 @@ The updated rule is:
 
 Impact of this change is discussed more deeply [below](#examples-method-arguments-must-match). Overall this will allow developers to make call sites more flexible by annotating non-escaping ref-like values with `scoped`.
 
-The `scoped` modifier on parameters also impacts our OHI and `delegate` conversion rules. The signature for an override, interface implementation or `delegate` conversion can: 
+The `scoped` modifier on parameters also impacts our object overriding, interface implementation and `delegate` conversion rules. The signature for an override, interface implementation or `delegate` conversion can: 
 - Add `scoped` to a `ref` or `in` parameter
 - Add `scoped` to a `ref struct` parameter 
-Any other difference with respect to `scoped` will be considered an error.
+Any other difference with respect to `scoped` will be considered an error. 
+
+The `scoped` modifier also has the following effects on method signatures:
+- The `scoped` modifier though does not affect hiding.
+- Overloads cannot differ only on `scoped`
 
 The section on `ref` field and `scoped` is long so wanted to close with a brief summary of the proposed breaking changes:
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -377,8 +377,8 @@ The presence of `scoped` allows us to refine the rule to reduce the friction the
 The updated rule is:
 
 > For a method invocation if there is an argument for a `ref`, `scoped ref` or `out` parameter of a `ref struct` type where the argument has *safe-to-escape* E1 then 
-> 1. No `ref struct` argument may contribute a narrower *safe-to-escape* than E1.
-> 2. No `ref` or `in` argument may contribute a narrower *ref-safe-to-escape* than E1.
+> 1. No other `ref struct` argument may contribute a narrower *safe-to-escape* than E1.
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -125,10 +125,9 @@ First the rules establishing *ref-safe-to-escape* values for fields need to be u
 <a name="rules-field-lifetimes"></a>
 
 > An lvalue designating a reference to a field, e.F, is *ref-safe-to-escape* (by reference) as follows:
-> 1. If `F` is a `ref` field and `e` is `this`, it is *ref-safe-to-escape* from the enclosing method.
-> 2. Else if `F` is a `ref` field its *ref-safe-to-escape* scope is the *safe-to-escape* scope of `e`.
-> 3. Else if `e` is of a reference type, it has *ref-safe-to-escape* of *calling method*
-> 4. Else its *ref-safe-to-escape* is taken from the *ref-safe-to-escape* of `e`.
+> 1. If `F` is a `ref` field its *ref-safe-to-escape* scope is the *safe-to-escape* scope of `e`.
+> 2. Else if `e` is of a reference type, it has *ref-safe-to-escape* of *calling method*
+> 3. Else its *ref-safe-to-escape* is taken from the *ref-safe-to-escape* of `e`.
```
jaredpar, line 130: Rules 1 and 2 were redundant so collapsed them

proposals/low-level-struct-improvements.md
```diff
@@ -157,7 +156,7 @@ Next the rules for ref re-assignment need to be adjusted for the presence of `re
 
 > For a ref reassignment in the form ...
 > 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* to *calling method* then `e2` must be *ref-safe-to-escape* to the *calling method*
-> 2. `e1 = ref e2`: where `e1` is a local or parameter, the *ref-safe-to-escape* of `e2` must be at least as wide a scope as the *ref-safe-to-escape* of `e1`.
+> 2. `e1 = ref e2`: where `e1` is a `ref` local or parameter then `e2` must have a *ref-safe-to-escape* at least as big as the *ref-safe-to-escape* of `e1`.
```
jaredpar, line 159: No change here. Just tried to get the language consistent between the rules.

proposals/low-level-struct-improvements.md
```diff
@@ -285,7 +284,9 @@ Span<int> Use()
 }
 ```
 
-The span safety rules will be written in terms of `scoped ref` and `ref`. For span safety purposes an `in` parameter is equivalent to `ref` and `out` is equivalent to `scoped ref`. Both `in` and `out` will only be specifically called out when it is important to the semantic of the rule. Otherwise they are just considered `ref` and `scoped ref` respectively.
+The span safety rules will be written in terms of `scoped ref` and `ref`. For span safety purposes an `in` parameter is equivalent to `ref` and `out` is equivalent to `scoped ref`. Both `in` and `out` will only be specifically called out when it is important to the semantic of the rule. Otherwise they are just considered `ref` and `scoped ref` respectively. 
+
+When discussing the *ref-safe-to-escape* of arguments that correspond to `in` parameters they will be generalized as `ref` arguments in the spec. In the case the argument is an lvalue then the *ref-safe-to-escape* is that of the lvalue, otherwise it is *current method*. Again `in` will only be called out here when it is important to the semantic of the current rule.
```
jaredpar, line 289: I continued to be frustrated at having to call `in` out differently than `ref` in a couple of places. Hoping to fix that by clarifying that the *ref-safe-to-escape* of `in` changes based on whether it's an lvalue or compiler generated temp. Allowed me to remove `in` language from a few places.

proposals/low-level-struct-improvements.md
RikkiGibson: My issue with using "big" here is that in the implementation, smaller numbers are used to represent "wider", also known as "bigger" scopes. :)

jaredpar: It is unfortunate that the original implementation occurred before the spec and the spec didn't re-use the implementation language. When we look at cleaning up the ref saftey implementation in MQ we should also think a lot about getting the language in sync with the spec.

proposals/low-level-struct-improvements.md
RikkiGibson: I didn't understand why we needed a distinction for a compiler-generated temp. It's just another local variable, right?

proposals/low-level-struct-improvements.md
cston, line 159: >`ref` local or parameter [](http://example.com/codeflow?start=36&length=24)

Perhaps "`ref` local or `ref` parameter" if that is what is intended here, just to be extra clear.

cston, line 159: >big [](http://example.com/codeflow?start=116&length=3)

Consider reverting to "wide" for consistency with other statements in this spec and span-safety.md which use "narrow" or "wide".

proposals/low-level-struct-improvements.md
```diff
@@ -300,16 +301,15 @@ The method invocation for lvalue returns can now be simplified. The receiver no
 
 > An lvalue resulting from a ref-returning method invocation `e1.M(e2, ...)` is *ref-safe-to-escape* the smallest of the following scopes:
 > 1. The *calling method*
-> 2. The *ref-safe-to-escape* of all `ref` arguments
-> 3. The *ref-safe-to-escape* of all `in` parameters when the argument is an lvalue otherwise *current method*
-> 4. The *safe-to-escape* of all argument expressions
+> 2. The *ref-safe-to-escape* contributed byall `ref` arguments
```
cston, line 304: >byall [](http://example.com/codeflow?start=42&length=5)

Typo: "by all"

cston, line 304:
```suggestion
> 2. The *ref-safe-to-escape* contributed by all `ref` arguments
```

proposals/low-level-struct-improvements.md
```diff
@@ -351,12 +351,25 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
 
 <a name="rules-method-arguments-must-match"></a>
 
-The presence of `scoped` allows us to also refine the method arguments must match rule. It can now be reduced to 
+The presence of `ref` fields means the rules around method arguments must match need to be updated as a `ref` parameter can now be stored as fields in a `ref struct` argument to the method. Previously the rule only had to consider other `ref struct` being stored as fields. The impact of this is discussed in [the compat considerations](#compat-considerations).
```
cston, line 354:
```suggestion
The presence of `ref` fields means the rules around method arguments must match need to be updated as a `ref` parameter can now be stored as a field in a `ref struct` argument to the method. Previously the rule only had to consider another `ref struct` being stored as a field. The impact of this is discussed in [the compat considerations](#compat-considerations). 
```

proposals/low-level-struct-improvements.md
jaredpar: The spec often refers to the *ref-safe-to-escape* of the `ref` argument. In the case we're referring to `in` though it could be the expression or the compiler generated temp. I think it's understood it's 99% talking about "what we actually pass" vs. "what is actually typed in the editor" but it's a spec so need to be specific.

proposals/low-level-struct-improvements.md
```diff
@@ -351,12 +351,25 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
...
-> For a method invocation if there is an argument for a `ref` or `scoped ref` parameter of a `ref struct` type where the argument has *safe-to-escape* E1 then no argument may contribute a narrower *safe-to-escape* than E1.
+> For a method invocation if there is an argument for a `ref`, `scoped ref` or `out` parameter of a `ref struct` type where the argument has *safe-to-escape* E1 then 
+> 1. No `ref struct` argument may contribute a narrower *safe-to-escape* than E1.
+> 2. No `ref` or `in` argument may contribute a narrower *ref-safe-to-escape* than E1.
+>
+> This rule does not apply to `ref struct` arguments for `in` parameters
 
 Impact of this change is discussed more deeply [below](#examples-method-arguments-must-match). Overall this will allow developers to make call sites more flexible by annotating non-escaping ref-like values with `scoped`.
 
+The `scoped` modifier on parameters also impacts our OHI and `delegate` conversion rules. The signature for an override, interface implementation or `delegate` conversion can:
```
cston, line 368: >OHI [](http://example.com/codeflow?start=53&length=3)

Consider using "overriding and interface implementation" instead. It would avoid introducing a new "OHI" term, and also `scoped` does not affect hiding.

proposals/low-level-struct-improvements.md
```diff
@@ -301,15 +301,15 @@ The method invocation for lvalue returns can now be simplified. The receiver no
...
+> 2. The *ref-safe-to-escape* contributed by all `ref` arguments
+> 3. The *safe-to-escape* contributed by all argument expressions
 
 The method invocation for rvalue returns needs to change as follows to account for `ref` field returns.
 
 > An rvalue resulting from a method invocation `e1.M(e2, ...)` is *safe-to-escape* from the smallest of the following scopes:
 > 1. The *calling method*
-> 2. The *safe-to-escape* of all argument expressions
-> 3. When the return is a `ref struct` then *ref-safe-to-escape* of all `ref` arguments
+> 2. The *safe-to-escape* conrtibuted by all argument expressions
```
RikkiGibson, line 311:
```suggestion
> 2. The *safe-to-escape* contributed by all argument expressions
```

proposals/low-level-struct-improvements.md
```diff
@@ -112,7 +112,23 @@ ref struct ReadOnlyExample
 }
 ```
 
-A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today.
+A `readonly` on a `ref struct` will instance will force `ref` fields to be `readonly ref readonly`. This also means a `readonly ref struct` must mark `ref` fields as `readonly ref readonly`. This maintains the invariant that a `ref struct` which is `readonly` cannot store ref-like state.
```
cston, line 115: >will instance [](http://example.com/codeflow?start=31&length=13)

Typo?

proposals/low-level-struct-improvements.md
```diff
@@ -112,7 +112,23 @@ ref struct ReadOnlyExample
...
+```c#
+ref struct S
+{
+    ref int Field;
+
+    void Uses(in S s1, ref S s2)
+    {
+        int local = default;
+        s1.Field = 42;          // Error: can't assign ref readonly value
+        s1.Field = ref loal;    // Error: can't repoint readonly ref
```
cston, line 126: >loal [](http://example.com/codeflow?start=23&length=4)

`local` here and below?

proposals/low-level-struct-improvements.md
```diff
@@ -112,7 +112,23 @@ ref struct ReadOnlyExample
...
+
+```c#
+ref struct S
+{
+    ref int Field;
+
+    void Uses(in S s1, ref S s2)
+    {
+        int local = default;
+        s1.Field = 42;          // Error: can't assign ref readonly value
```
cston, line 125: >s1.Field = 42; [](http://example.com/codeflow?start=8&length=14)

Does the `ref readonly`-ness apply through properties as well? For instance:
```csharp
ref int Property => ref Field;
...
s1.Property = 42;
```

jcouv:
Spec-related notes from yesterday. It's fine if they are addressed in a separate PR:
- grammar
- general rule: `scoped` can only be applied to a `ref` type or type that is a `ref struct`
- OHI
- ref in object initializers
- "unscoped" and "safe fixed size buffers" are not in C# 11 (probably should remove sections from this doc)

proposals/low-level-struct-improvements.md
jcouv, line 115: >A `readonly` on a `ref struct` will instance will force `ref` fields to be `readonly ref readonly`. This also means a `readonly ref struct` must mark `ref` fields as `readonly ref readonly`. [](http://example.com/codeflow?start=0&length=190)

The first two sentences in this paragraph seem near-duplicates

proposals/low-level-struct-improvements.md
```diff
@@ -112,7 +112,23 @@ ref struct ReadOnlyExample
 }
 ```
 
-A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today.
+A `readonly` on a `ref struct` will instance will force `ref` fields to be `readonly ref readonly`. This also means a `readonly ref struct` must mark `ref` fields as `readonly ref readonly`. This maintains the invariant that a `ref struct` which is `readonly` cannot store ref-like state.
+
+```c#
+ref struct S
```
jcouv, line 118: This code snippet doesn't seem related to the previous paragraph. Consider adding a blurb to explain the general rule for `ref` vs. `ref readonly`/`in` parameters.

proposals/low-level-struct-improvements.md
```diff
@@ -351,12 +367,25 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
 
 <a name="rules-method-arguments-must-match"></a>
 
-The presence of `scoped` allows us to also refine the method arguments must match rule. It can now be reduced to 
+The presence of `ref` fields means the rules around method arguments must match need to be updated as a `ref` parameter can now be stored as a field in a `ref struct` argument to the method. Previously the rule only had to consider another `ref struct` being stored as a field. The impact of this is discussed in [the compat considerations](#compat-considerations).
```
jcouv, line 370: >The presence of `ref` fields means the rules around method arguments must match need to be updated [](http://example.com/codeflow?start=0&length=98)

"The presence of `ref` fields means the rules around method arguments must match need to be updated ..."
This part of the sentence seems incomplete (I couldn't parse it).

proposals/low-level-struct-improvements.md
```diff
@@ -351,12 +367,25 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
...
+> 2. No `ref` or `in` argument may contribute a narrower *ref-safe-to-escape* than E1.
+>
+> This rule does not apply to `ref struct` arguments for `in` parameters
 
 Impact of this change is discussed more deeply [below](#examples-method-arguments-must-match). Overall this will allow developers to make call sites more flexible by annotating non-escaping ref-like values with `scoped`.
 
+The `scoped` modifier on parameters also impacts our OHI and `delegate` conversion rules. The signature for an override, interface implementation or `delegate` conversion can: 
+- Add `scoped` to a `ref` or `in` parameter
+- Add `scoped` to a `ref struct` parameter 
+Any other difference with respect to `scoped` will be considered an error.
```
jcouv, line 387: It would be good to add that overloads differing only by `scoped` are errors.

jcouv (commented):
Done with review pass (iteration 5)

proposals/low-level-struct-improvements.md
jaredpar: The first is describing how `readonly` on a value will impact `ref` fields, the other is describing how `readonly` on a declaration will do so. Let me try and clean that up a bit.

proposals/low-level-struct-improvements.md
jaredpar: For properties or methods it depends on the type of `this`. In a `readonly` property or method `this` is `ref readonly` hence all `ref` fields will be `ref readonly`.

proposals/low-level-struct-improvements.md
jaredpar: Embarrassing ...

proposals/low-level-struct-improvements.md
jaredpar: The docs uses the non `-` form for this.

proposals/low-level-struct-improvements.md
jaredpar: Tried to clear this up.

proposals/low-level-struct-improvements.md
jcouv: The other places with "method arguments must match" were not as hard to parse.
Consider: "... means the method arguments must match rule ..."

cston (approved):

jcouv (approved):
LGTM Thanks (iteration 7). Thanks for the clarifications

RikkiGibson:
Does this spec capture the rule changes we were going for more recently, i.e. disallowing a ref field of a ref struct type?

jaredpar:
> Does this spec capture the rule changes we were going for more recently, i.e. disallowing a ref field of a ref struct type?

Updating that part later this week

PR #6233, opened 2022-06-24
Add speclet for file-local types

RikkiGibson:
Related to dotnet/roslyn#60819

### DIFF
--- a/proposals/file-types.md
+++ b/proposals/file-types.md
@@ -33,7 +33,6 @@ Our primary motivation is from source generators. Source generators work by addi
 1. Those files should be able to contain implementation details which are hidden from the rest of the compilation, yet are usable throughout the file they are declared in.
 2. We want to reduce the need for generators to "search" for type names which won't collide with declarations in user code or code from other generators.
 
-
 ## Detailed design
 [design]: #detailed-design
 
--- a/proposals/file-types.md
+++ b/proposals/file-types.md
@@ -151,6 +151,8 @@ class C
 ### Usage in signatures
 There is a general need to prevent `file` types from appearing in member parameters, returns, and type parameter constraints where the `file` type might not be in scope at the point of usage of the member.
 
+Note that non-file types are permitted to implement file interfaces, similar to how types can implement less-accessible interfaces. Depending on the types present in the interface members, it could result in a violation of the rules in the following section.
+
 #### Only allow signature usage in members of `file` types
 Perhaps the simplest way to ensure this is to enforce that `file` types can only appear in signatures or as base types of other `file` types:
 
--- a/proposals/file-types.md
+++ b/proposals/file-types.md
@@ -137,14 +137,22 @@ namespace NS1
 Therefore, we don't specify the feature in terms of which scope the type is contained in, but rather as additional "filtering rules" in member lookup.
 
 ### Attributes
-In the initial iteration of the feature, a file type is not allowed to be used as an attribute. Such attributes are of limited utility for users who are unaware of the internal name generation strategy used in the implementation.
+File classes are permitted to be attribute types, and can be used as attributes within both file types and non-file types, just as if the attribute type were a non-file type. The metadata name of the file-local attribute type still goes through the same name generation strategy as other file-local types. This means detecting the presence of a file-type by a hard-coded string name is likely to be impractical, because it requires depending on the internal name generation strategy of the compiler, which may change over time. However, detecting via `typeof(MyFileLocalAttribute)` works.
 
 ```cs
-file class FC : System.Attribute { }
+using System;
+using System.Linq;
 
-[FC] // error
-class C
+file class MyFileLocalAttribute : Attribute { }
+
+[MyFileLocalAttribute]
+public class C
 {
+    public static void Main()
+    {
+        var attribute = typeof(C).CustomAttributes.Where(attr => attr.AttributeType == typeof(MyFileLocalAttribute)).First();
+        Console.Write(attribute); // outputs the generated name of the file-local attribute type
+    }
 }
 ```
 
--- a/proposals/file-types.md
+++ b/proposals/file-types.md
@@ -130,20 +130,15 @@ namespace NS1
 Therefore, we don't specify the feature in terms of which scope the type is contained in, but rather as additional "filtering rules" in member lookup.
 
 ### Attributes
-A type which is both an attribute type and a file type is said to be a *file attribute*. Much like an ordinary *file types*, a *file attribute* can only be used in the file where it is declared.
+In the initial iteration of the feature, a file type is not allowed to be used as an attribute. Such attributes are of limited utility for users who are unaware of the internal name generation strategy used in the implementation.
 
 ```cs
-// File1.cs
-file class Attr : System.Attribute { }
+file class FC : System.Attribute { }
 
-[Attr] // ok
-class Program { }
-```
-
-```cs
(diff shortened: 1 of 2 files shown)


proposals/file-types.md
```diff
@@ -0,0 +1,151 @@
...
+var widget = new Widget(); // error: The type or namespace name 'Widget' could not be found.
+```
+
+## Motivation
+[motivation]: #motivation
+
+Our primary motivation is from source generators. Source generators work by adding files to the user's compilation.
+1. Those files should be able to contain implementation details which are hidden from the rest of the compilation, yet are usable throughout the file they are declared in.
+2. We want to reduce the need for generators to "search" for type names which won't collide with declarations in user code or code from other generators.
+
```
RikkiGibson, line 35:
```suggestion
```

proposals/file-types.md
```diff
@@ -0,0 +1,150 @@
...
+```cs
+public file class C1 { } // error
+internal file class C2 { } // error
+file class C3 { } // ok
+```
+
+### Naming
+The implementation guarantees that `file` types in different files with the same name will be distinct to the runtime. The type's accessibility and name in metadata is implementation-defined. The intention is to permit the compiler to adopt any future access-limitation features in the runtime which are suited to the feature. It's expected that in the initial implementation, an `internal` accessibility would be used and an unspeakable generated name will be used which depends on the file the type is declared in.
+
+### Lookup
```
jcouv, line 54: May need to adjust the "scopes" section as well: [https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/basic-concepts.md#77-scopes](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/basic-concepts.md#77-scopes)

proposals/file-types.md
```diff
@@ -0,0 +1,150 @@
...
+class Program
+{
+    static void Main()
+    {
+        C.M(); // refers to the 'C' in File2.cs
+    }
+}
+```
+
+### Attributes
```
jcouv, line 89: This makes sense, but a concern is that one main scenario for attributes is to be checked via reflection. For file types, that doesn't seem practical. 
I'm wondering whether we should disallow this (using file-types as attributes). Do we have a motivating scenario?

proposals/file-types.md
```diff
@@ -0,0 +1,188 @@
...
+class Program
+{
+    static void Main()
+    {
+        C.M(); // refers to the 'C' in File2.cs
+    }
+}
+```
+
+Note that we don't update the [scopes](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/basic-concepts.md#77-scopes) section of the spec. This is because, as the spec states:
```
jcouv, line 89: I think "scopes" should be adjusted. It currently says:
"The scope of a namespace member declared by a namespace_member_declaration (§13.6) with no enclosing namespace_declaration is the entire program text."

jcouv: Upon re-reading, I think this is fine. We're saying that file-types are in scope in the entire program, but the lookup rules may filter them out. I think that works

jcouv (approved):
LGTM Thanks (iteration 4)

proposals/file-types.md
```diff
@@ -0,0 +1,188 @@
...
+[motivation]: #motivation
+
+Our primary motivation is from source generators. Source generators work by adding files to the user's compilation.
+1. Those files should be able to contain implementation details which are hidden from the rest of the compilation, yet are usable throughout the file they are declared in.
+2. We want to reduce the need for generators to "search" for type names which won't collide with declarations in user code or code from other generators.
+
+## Detailed design
+[design]: #detailed-design
+
+- We add the `file` modifier to the [class](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/classes.md#1422-class-modifiers), [struct](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/structs.md#1522-struct-modifiers), [interface](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/interfaces.md#1722-interface-modifiers), and [enum](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/enums.md#183-enum-modifiers) modifier sets in the grammar.
```
svick, line 39: What about [delegates](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/delegates.md#192-delegate-declarations)? For some reason, they don't have a separate modifiers section in the spec, like the other type kinds do, but they do have their own grammar rule `delegate_modifier`.

cston, line 39: And `record`?

proposals/file-types.md
```diff
@@ -0,0 +1,188 @@
...
+
+Note that this does restrict usage in explicit implementations, even though such usages are safe. We do this in order to simplify the rules for the initial iteration of the feature.
+
+```cs
+file interface I
+{
+    void M(I i);
+}
+
+class C : I
```
cston, line 174: >: I [](http://example.com/codeflow?start=8&length=3)

Is it an error to implement the interface, even if there are no interface members?
```csharp
file interface I { }
public class C : I { } // error?
```

RikkiGibson: It is not an error.

proposals/file-types.md
RikkiGibson: Yeah, I think record, record struct and delegate should be included here.

proposals/file-types.md
```diff
@@ -0,0 +1,188 @@
...
+file class FC : System.Attribute { }
+
+[FC] // error
+class C
+{
+}
+```
+
+### Usage in signatures
+There is a general need to prevent `file` types from appearing in member signatures where the `file` type might not be in scope at the point of usage of the member.
```
jcouv, line 145: Let's mention that "signature" includes type constraints

proposals/file-types.md
RikkiGibson: Fixed, thanks.

RikkiGibson:
It was suggested in email thread with LDT that we do nothing re: attributes. i.e. you can declare 'file class C : System.Attribute' and use it in '[C]', and we won't give a diagnostic or do anything else unusual. It wasn't clear if we had consensus about that.

proposals/file-types.md
```diff
@@ -137,14 +137,22 @@ namespace NS1
 Therefore, we don't specify the feature in terms of which scope the type is contained in, but rather as additional "filtering rules" in member lookup.
 
 ### Attributes
-In the initial iteration of the feature, a file type is not allowed to be used as an attribute. Such attributes are of limited utility for users who are unaware of the internal name generation strategy used in the implementation.
+File classes are permitted to be attribute types, and can be used as attributes within both file types and non-file types, just as if the attribute type were a non-file type. The metadata name of the file-local attribute type still goes through the same name generation strategy as other file-local types. This means detecting the presence of a file-type by a hard-coded string name is likely to be impractical, because it requires depending on the internal name generation strategy of the compiler, which may change over time. However, detecting via `typeof(MyFileLocalAttribute)` works.
```
cston, line 140: There are several different names used: "file type", "file-local ... type", and "file-type". Consider standardizing on one or perhaps two names.

cston (approved):

proposals/file-types.md
RikkiGibson: Yeah, I'm going to make a pass to change everything to "file-local" types, here and in the implementation.

cston (approved):

PR #6249, opened 2022-06-28
ref fields updates

jaredpar:
This updates the proposal based on shipping decisions for C# 11 and 12. Several changes are still pending LDM review but have been tentatively okayed by the various stakeholders.

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -379,6 +379,7 @@ The section on `ref` field and `scoped` is long so wanted to close with a brief 
 Detailed Notes:
 - A `ref` field can only be declared inside of a `ref struct` 
 - A `ref` field cannot be declared `static`, `volatile` or `const`
+- A `ref` field cannot have a type that is `ref struct`
 - The reference assembly generation process must preserve the presence of a `ref` field inside a `ref struct` 
 - A `readonly ref struct` must declare its `ref` fields as `readonly ref`
 - The span safety rules document will be updated as outlined in this document.
@@ -389,7 +390,7 @@ Detailed Notes:
 ### Sunset restricted types
 The compiler has a concept of a set of "restricted types" which is largely undocumented. These types were given a special status because in C# 1.0 there was no general purpose way to express their behavior. Most notably the fact that the types can contain references to the execution stack. Instead the compiler had special knowledge of them and restricted their use to ways that would always be safe: disallowed returns, cannot use as array elements, cannot use in generics, etc ...
 
-Once `ref` fields are available these types can be correctly defined in C# using a combination of `ref struct` and `ref` fields. Therefore when the compiler detects that a runtime supports `ref` fields it will no longer have a notion of restricted types. It will instead use the types as they are defined in the code. 
+Once `ref` fields are available and extended to support `ref struct` these types can be correctly defined in C# using a combination of `ref struct` and `ref` fields. Therefore when the compiler detects that a runtime supports `ref` fields it will no longer have a notion of restricted types. It will instead use the types as they are defined in the code. 
 
 To support this our span safety rules will be updated as follows:
 
@@ -398,7 +399,7 @@ To support this our span safety rules will be updated as follows:
 - `__arglist` as a parameter will have a *ref-safe-to-escape* and *safe-to-escape* of *current method*. 
 - `__arglist(...)` as an expression will have a *ref-safe-to-escape* and *safe-to-escape* of *current method*. 
 
-Conforming runtimes will ensure that `TypedReference`, `RuntimeArgumentHandle` and `ArgIterator` are defined as `ref struct`. That combined with the above rules will ensure references to the stack do not escape beyond their lifetime.
+Conforming runtimes will ensure that `TypedReference`, `RuntimeArgumentHandle` and `ArgIterator` are defined as `ref struct`. Further `TypedReference` must be viewed as having a `ref` field to a `ref struct` for any possible type (it can store any value). That combined with the above rules will ensure references to the stack do not escape beyond their lifetime.
 
 Note: strictly speaking this is a compiler implementation detail vs. part of the language. But given the relationship with `ref` fields it is being included in the language proposal for simplicity.
 
@@ -1008,17 +1009,78 @@ Some thought was given to the idea of having `this` have different defaults base
 
 **Decision** Keep `this` as `scoped ref`
 
+### ref fields to ref struct
+This feature opens up a new set of ref safety rules because it allows for a `ref` field to refer to a `ref struct`. This generic nature of `ByReference<T>` meant that up until now the runtime could not have such a construct. As a result all of our rules are written under the assumption this is not possible. The `ref` field feature is largely not about making new rules but codifying the existing rules in our system. Allowing `ref` fields to `ref struct` requires us to codify new rules because there are several new scenarios to consider.
+
+The first is that a `readonly ref` is now capable of storing `ref` state. For example:
+
+```c#
+readonly ref Container
+{
+    readonly ref Span<int> Span;
+
+    void Store(Span<int> span)
+    {
+        Span = span;
+    }
+}
+```
+
+This means when thinking about method arguments must match rules we must consider `readonly ref T` is potential method output sources when `T` potentially has a `ref` field to a `ref struct`.
+
+The second issue is language must consider a new type of escape scope: *ref-field-safe-to-escape*. All `ref struct` which transitively contain a `ref` field have another escape scope representing the value(s) in the `ref` field(s). In the case of multiple `ref` fields they can be collectively tracked as a single value. The default value for this for parameters is *calling method*. 
+
+```c#
+ref struct Nested
+{
+    ref Span<int> Span;
+}
+
+Span<int> M(ref Nested nested) => nested.Span;
+```
+
+This value is not related to the escape scope of the container; that is as it the container scope gets smaller it has no impact on the *ref-field-safe-to-escape* of the `ref` field values. 
+
+```c#
+ref struct Nested
+{
+    ref Span<int> Span;
+}
+
+void M(ref Nested nested)
+{
+    scoped ref Nested refLocal = nested;
+
+    // the ref-field-safe-to-escape of local is still *calling method* which means the following
+    // is illegal
+    refLocal.Span = stackalloc int[42];
+
+    scoped Nested valLocal = nested;
+
+    // the ref-field-safe-to-escape of local is still *calling method* which means the following
+    // is still illegal
+    valLocal.Span = stackalloc int[42];
+}
+```
+
+This *ref-field-safe-to-escape-scope* has essentially always existed. Up until now `ref` fields could only point to normal `struct` hence it was trivially collapsed to *calling method*.  To support `ref` fields to `ref struct` our existing rules need to be updated to take into account this new escape scope.
+
+These problems are very solvable. The compiler team has sketched out a few versions of these rules and they largely fall out from our existing analysis. The problem is there is no consuming code for such rules that helps prove out there correctness and usability. This makes us very hesitant to add support because of the fear we'll pick wrong defaults and back the runtime into usability corner when it does take advantage of this. This concern is particularly strong because .NET 8 likely pushes is in this direction with `allow T: ref struct` and `Span<Span<T>>`. The rules would be better written if it's done in conjunction with consumption code.
+
+**Decision** Delay allowing `ref` field to `ref struct` until .NET 8 where we have scenarios that will help drive the rules around these scenarios.
+
 ### What will make C# 11.0?
 The features outlined in this document don't need to be implemented in a single pass. Instead they can be implemented in phases across several language releases in the following buckets:
 
 1. `ref` fields and `scoped`
-2. Sunset restricted types
-3. `unscoped` 
-4. fixed sized buffers
+2. `ref` fields to `ref struct`
+3. Sunset restricted types
+4. `unscoped` 
+5. fixed sized buffers
 
 What gets implemented in which release is merely a scoping exercise. 
 
-**Decision** Only `ref` fields, `scoped` and sunsetting restricted types will make C# 11.0. LDM is happy to revisit `unscoped` if a more natural keyword can be settled on or data suggests it's possible to make `this` `unscoped` by default in all cases.
+**Decision** Only (1) will make C# 11.0. The expectation is (2) and (3) are enabled very early in C# 12.0 to enable dogfooding by runtime throughout the .NET 8 cycle
 
 ## Future Considerations
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -112,7 +112,7 @@ ref struct ReadOnlyExample
 }
 ```
 
-A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today.
+A `readonly ref struct` will require that `ref` fields are marked as `readonly ref`. There is no requirement that they are marked as `readonly ref readonly`. This does allow a `readonly struct` to have indirect mutations via such a field but that is no different than a `readonly` field that pointed to a reference type today ([more details](#reason-readonly-shallow))
 
 A `readonly ref` will be emitted to metadata using the `initonly` flag, same as any other field. A `ref readonly` field will be attributed with `System.Runtime.CompilerServices.IsReadOnlyAttribute`. A `readonly ref readonly` will be emitted with both items.
 
@@ -1132,3 +1132,44 @@ The proposal prevents this though because it violates the span safety rules. Con
 At that point the line `r = ref i` is illegal by [ref re-assignment rules](#rules-ref-re-assignment). 
 
 These rules were not intended to prevent this behavior but do so as a side effect. It's important to keep this in mind for any future rule update to evaluate the impact to scenarios like this.
+
+#### readonly cannot be deep through ref fields
+<a name="reason-readonly-shallow"></a>
+
+Consider the below code sample:
+
+```c#
+ref struct S
+{
+    ref int Field;
+
+    readonly void Method()
+    {
+        // Legal or illegal?
+        Field = 42;
+    }
+}
+```
+
+When designing the rules for `ref` fields on `readonly` instances in a vacuum the rules can be validly designed such that the above is legal or illegal. Essentially `readonly` can validly be deep through a `ref` field or it can apply only to the `ref`. Applying only to the `ref` prevents ref re-assignment but allowing normal assignment which changes the referred to value.
+
+This design does not exist in a vacuum though, it is designing rules for types that already effectively have `ref` fields. The most prominent of which, `Span<T>`, already has a strong dependency on `readonly` not being deep here. It's primary scenario is the ability to assign to the `ref` field through a `readonly` instance. 
+
+```c#
+readonly ref struct SpanOfOne
+{
+    readonly ref int Field;
+
+    public ref int this[int index]
+    {
+        get
+        {
+            if (index != 1)
+                throw new Exception();
+            return ref Field;
+        }
+    }
+}
+```
+
+This means we must choose the shallow interpretation of `readonly`.
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -1244,13 +1244,13 @@ readonly ref struct SpanOfOne
 This means we must choose the shallow interpretation of `readonly`.
 
 #### Method arguments must match
-The method arguments must match rule is a common source of confusion for developers. It's a rule which has a number of special cases that are hard to understand unless you are familiar with the reasoning behind the rule.
+The method arguments must match rule is a common source of confusion for developers. It's a rule which has a number of special cases that are hard to understand unless you are familiar with the reasoning behind the rule. For the sake of better understanding the reasons for the rule we will simplify ref-safe-to-escape* and *safe-to-escape* to simply *escape-scope*. 
 
 Methods can pretty liberally return state passed to them as parameters. Essentially any reachable state which is `unscoped` can be returned (including returning by `ref`). This can be returned directly through a `return` statement or indirectly by assigning into a `ref` value. 
 
-Direct returns don't pose much problems for ref safety. The compiler simply needs to look at all the returnable inputs to a method and then it effectively restricts the return value to be the minimum *safe-to-escape* of the input. That return value then goes through normal processing.
+Direct returns don't pose much problems for ref safety. The compiler simply needs to look at all the returnable inputs to a method and then it effectively restricts the return value to be the minimum *escape-scope* of the input. That return value then goes through normal processing.
 
-Indirect returns pose a significant problem because all `ref` are both an input and output to the method. The compiler has to look at every single `ref` which is assignable in the called method, evaluate it's *safe-to-escape*, and then verify no returnable input to the method has a smaller *safe-to-escape* than that `ref`. If any such case exists then the method call must be illegal because it could violate `ref` safety. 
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -263,6 +262,16 @@ ref int Sneaky(out int i)
 }
 ```
 
+Languages that support `ref struct` must ensure the original value passed into an `out` parameter for a `ref struct` is not directly returned or indirectly through another `ref` just as they would for a `ref scoped` parameter. Doing so would allow for type safety holes to exist. C# achieves this via it's definite assignment rules. That both achieves our ref safety as well as allowing for existing code which assigns and then returns `out` parameters values.
+
+```c#
+ref Span<int> StrangeButLegal(out Span<int> span)
+{
+    span = default;
+    return span;
```
cston, line 271: >return span; [](http://example.com/codeflow?start=4&length=12)

The method signature returns a `ref` rather than a value. Also, the `out Span<int> span` cannot be returned by `ref` nor by value since it is `out scoped`.

proposals/low-level-struct-improvements.md
```diff
@@ -285,8 +294,14 @@ Span<int> Use()
...
+
+This means going forward `out` will match developers intuition when it comes to ref safety: `out` means `out`.
+
 The span safety rules will be written in terms of `scoped ref` and `ref`. For span safety purposes an `in` parameter is equivalent to `ref` and `out` is equivalent to `scoped ref`. Both `in` and `out` will only be specifically called out when it is important to the semantic of the rule. Otherwise they are just considered `ref` and `scoped ref` respectively.
 
+When discussing the *ref-safe-to-escape* of arguments that correspond to `in` parameters they will be generalized as `ref` arguments in the spec. In the case the argument is an lvalue then the *ref-safe-to-escape* is that of the lvalue, otherwise it is *current method*. Again `in` will only be called out here when it is important to the semantic of the current rule.
```
cston, line 303: >is that of the lvalue [](http://example.com/codeflow?start=214&length=21)

"is the *calling method**"?

proposals/low-level-struct-improvements.md
```diff
@@ -247,11 +246,11 @@ Other uses for `scoped` on locals are discussed [below](#examples-scoped-locals)
 
 When `scoped` is applied to an instance method the `this` parameter will have the type `scoped ref T`. By default this is redundant as `scoped ref T` is the default type of `this`. It is useful in the case the `struct` is declared as `unscoped` (detailed [below](#return-fields-by-ref)).
 
-The `scoped` annotation cannot be applied to any other location including returns, fields, array elements, etc ... Further while `scoped` can be applied to any `ref`, `in` or `out` it can only be applied to values which are `ref struct`. Having declarations like `scoped int` adds no value to the rules and will be prevented to avoid developer confusion.
+The `scoped` annotation cannot be applied to any other location including returns, fields, array elements, etc ... Further while `scoped` has impact when applied to any `ref`, `in` or `out` it only has impact when applied to values which are `ref struct`. Having declarations like `scoped int` has no impact because a non `ref struct` is always safe to return. The compiler will create a diagnostic on such cases to avoid developer confusion.
```
jcouv, line 249: >The compiler will create a diagnostic on such cases to avoid developer confusion. [](http://example.com/codeflow?start=361&length=81)

We could be more specific: It is an error to use `scoped` for a value that isn't a `ref` or a `ref struct`.

proposals/low-level-struct-improvements.md
```diff
@@ -351,20 +365,34 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
...
 
 Impact of this change is discussed more deeply [below](#examples-method-arguments-must-match). Overall this will allow developers to make call sites more flexible by annotating non-escaping ref-like values with `scoped`.
 
+The `scoped` modifier on parameters also impacts our object overriding, interface implementation and `delegate` conversion rules. The signature for an override, interface implementation or `delegate` conversion can: 
+- Add `scoped` to a `ref` or `in` parameter
+- Add `scoped` to a `ref struct` parameter 
+Any other difference with respect to `scoped` will be considered an error. 
+
+The `scoped` modifier also has the following effects on method signatures:
+- The `scoped` modifier though does not affect hiding.
```
jcouv, line 384: >though [](http://example.com/codeflow?start=24&length=6)

nit: The "though" seems strange here

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
-### What will make C# 11.0
+### Unscoped this by default?
+The design only has two locations which are `scoped` by default: 
+
+- `this` is `scoped ref` 
+- `out` is `ref scoped`
+
+The decision on `out` is to significantly reduce the compat burden of `ref` fields and at the same time is a more natural default. It lets developers actually think of `out` as data flowing outward only where as if it's `scoped ref` or even `ref` then the rules must consider data flowing in both directions. This leads to significant developer confusion where as the `ref scoped` interpretation means that `out` means `out`. 
+
+The decision on `this` this though is undesirable because it means a `struct` cannot return a field by `ref`. This is an important scenario to high perf developers and the `unscoped` keyword was added effectively for this one scenario.
```
jcouv, line 980: >this though [](http://example.com/codeflow?start=23&length=11)

?

jcouv, line 979: >was added effectively [](http://example.com/codeflow?start=191&length=21)

would help

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +968,135 @@ This will work but is likely going to result in unnecessary code generation. One
...
-### What will make C# 11.0
+### Unscoped this by default?
+The design only has two locations which are `scoped` by default: 
+
+- `this` is `scoped ref` 
+- `out` is `scoped ref`
+
+The decision on `out` is to significantly reduce the compat burden of `ref` fields and at the same time is a more natural default. It lets developers actually think of `out` as data flowing outward only where as if it's `ref` then the rules must consider data flowing in both directions. This leads to significant developer confusion.
+
+The decision on `this` this though is undesirable because it means a `struct` cannot return a field by `ref`. This is an important scenario to high perf developers and the `unscoped` keyword was added effectively for this one scenario.
```
Sergio0694, line 979: While it is true that unscoped was primarily meant for this scenario, should this also mention that there are other valid cases where it would also be useful? I made the eample the other day of some type à la ValueStringBuilder, returning a span with the content written so far. Even if that buffer was wrapping an array in the heap and therefore would be safe to escape, it would be nice to be able to mark that property as unscoped so that callers would not be able to escape that span outside of the builder scope, as the builder would be disposed when out of scope (say, with a using declaration), and that disposal logic would cause the buffer to be returned to the array pool. Without that keyword, there would be no way to prevent consumers from misusing that and incorrectly escaping that span.

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +968,135 @@ This will work but is likely going to result in unnecessary code generation. One
...
-4. fixed sized buffers
+2. `ref` fields to `ref struct`
+3. Sunset restricted types
+4. `unscoped` 
+5. fixed sized buffers
 
 What gets implemented in which release is merely a scoping exercise. 
 
-**Decision** Only `ref` fields, `scoped` and sunsetting restricted types will make C# 11.0. LDM is happy to revisit `unscoped` if a more natural keyword can be settled on or data suggests it's possible to make `this` `unscoped` by default in all cases.
+**Decision** Only (1) will make C# 11.0. The expectation is (2) and (3) are enabled very early in C# 12.0 to enable dogfooding by runtime throughout the .NET 8 cycle
```
tannergooding, line 1099: 4 and 5 are then possible future?

Is there any workaround for 4? That is, could a dev manually apply `LifetimeAnnotation` with the relevant parameters and still achieve the ability to return a `ref` field on a struct or is this non-achievable until support is later added?

(Which would mean devs in need would need to continue with unsafe alternatives like `MemoryMarshal` and `Unsafe` APIs?)

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
+
+**Decision** Keep `this` as `scoped ref`
+
+### ref fields to ref struct
+This feature opens up a new set of ref safety rules because it allows for a `ref` field to refer to a `ref struct`. This generic nature of `ByReference<T>` meant that up until now the runtime could not have such a construct. As a result all of our rules are written under the assumption this is not possible. The `ref` field feature is largely not about making new rules but codifying the existing rules in our system. Allowing `ref` fields to `ref struct` requires us to codify new rules because there are several new scenarios to consider.
+
+The first is that a `readonly ref` is now capable of storing `ref` state. For example:
+
+```c#
+readonly ref Container
```
cston, line 1032: >readonly ref [](http://example.com/codeflow?start=0&length=12)

`readonly ref struct`?

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
+ - `this` as `scoped ref`: `ref struct`
+
+ This minimizes compat breaks and maximizes flexibility but at the cost of complicating the story for customers. It also doesn't fully solve the problem because future features, like safe `fixed` buffers, require that a mutable `ref struct` have `ref` returns for fields which don't work by this design alone as it would fall into the `scoped ref` category. 
+
+**Decision** Keep `this` as `scoped ref`
+
+### ref fields to ref struct
+This feature opens up a new set of ref safety rules because it allows for a `ref` field to refer to a `ref struct`. This generic nature of `ByReference<T>` meant that up until now the runtime could not have such a construct. As a result all of our rules are written under the assumption this is not possible. The `ref` field feature is largely not about making new rules but codifying the existing rules in our system. Allowing `ref` fields to `ref struct` requires us to codify new rules because there are several new scenarios to consider.
+
+The first is that a `readonly ref` is now capable of storing `ref` state. For example:
```
cston, line 1029: >`readonly ref` [](http://example.com/codeflow?start=20&length=14)

`readonly ref struct`?

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
+    readonly ref Span<int> Span;
+
+    void Store(Span<int> span)
+    {
+        Span = span;
+    }
+}
+```
+
+This means when thinking about method arguments must match rules we must consider `readonly ref T` is potential method output sources when `T` potentially has a `ref` field to a `ref struct`.
```
cston, line 1043: >potential method output sources [](http://example.com/codeflow?start=102&length=31)

"output" and "sources" together seems confusing. Perhaps just "a potential method output".

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
+```c#
+ref struct Nested
+{
+    ref Span<int> Span;
+}
+
+Span<int> M(ref Nested nested) => nested.Span;
+```
+
+This value is not related to the escape scope of the container; that is as it the container scope gets smaller it has no impact on the *ref-field-safe-to-escape* of the `ref` field values.
```
cston, line 1056: >This value is not related to the escape scope of the container; that is as it the container scope gets smaller it has no impact on the *ref-field-safe-to-escape* of the `ref` field values. [](http://example.com/codeflow?start=0&length=188)

Is this correct? Can't we say the *ref-field-safe-to-escape* scope is no smaller than the container scope?

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
+
+```c#
+ref struct Nested
+{
+    ref Span<int> Span;
+}
+
+void M(ref Nested nested)
+{
+    scoped ref Nested refLocal = nested;
```
cston, line 1066: >nested [](http://example.com/codeflow?start=33&length=6)

`ref nested`

proposals/low-level-struct-improvements.md
```diff
@@ -939,17 +969,133 @@ This will work but is likely going to result in unnecessary code generation. One
...
+
+    // the ref-field-safe-to-escape of local is still *calling method* which means the following
+    // is still illegal
+    valLocal.Span = stackalloc int[42];
+}
+```
+
+This *ref-field-safe-to-escape-scope* has essentially always existed. Up until now `ref` fields could only point to normal `struct` hence it was trivially collapsed to *calling method*.  To support `ref` fields to `ref struct` our existing rules need to be updated to take into account this new escape scope.
+
+These problems are very solvable. The compiler team has sketched out a few versions of these rules and they largely fall out from our existing analysis. The problem is there is no consuming code for such rules that helps prove out there correctness and usability. This makes us very hesitant to add support because of the fear we'll pick wrong defaults and back the runtime into usability corner when it does take advantage of this. This concern is particularly strong because .NET 8 likely pushes is in this direction with `allow T: ref struct` and `Span<Span<T>>`. The rules would be better written if it's done in conjunction with consumption code.
```
cston, line 1084: >pushes is in [](http://example.com/codeflow?start=491&length=12)

"pushes us in this ..." or perhaps just "pushes in this ..."

proposals/low-level-struct-improvements.md
```diff
@@ -1132,3 +1278,156 @@ The proposal prevents this though because it violates the span safety rules. Con
...
+
+    readonly void Method()
+    {
+        // Legal or illegal?
+        Field = 42;
+    }
+}
+```
+
+When designing the rules for `ref` fields on `readonly` instances in a vacuum the rules can be validly designed such that the above is legal or illegal. Essentially `readonly` can validly be deep through a `ref` field or it can apply only to the `ref`. Applying only to the `ref` prevents ref re-assignment but allowing normal assignment which changes the referred to value.
```
cston, line 1301: >allowing [](http://example.com/codeflow?start=311&length=8)

"allows"?

proposals/low-level-struct-improvements.md
```diff
@@ -1132,3 +1278,156 @@ The proposal prevents this though because it violates the span safety rules. Con
...
+
+This design does not exist in a vacuum though, it is designing rules for types that already effectively have `ref` fields. The most prominent of which, `Span<T>`, already has a strong dependency on `readonly` not being deep here. It's primary scenario is the ability to assign to the `ref` field through a `readonly` instance.
```
cston, line 1303: >It's [](http://example.com/codeflow?start=230&length=4)

"Its"?

proposals/low-level-struct-improvements.md
```diff
@@ -263,6 +261,16 @@ ref int Sneaky(out int i)
 }
 ```
 
+Languages that support `ref struct` must ensure the original value passed into an `out` parameter for a `ref struct` is not read. That in turn means it cannot be directly or indirectly returned.Doing so would allow for type safety holes to exist. C# achieves this via it's definite assignment rules. That both achieves our ref safety as well as allowing for existing code which assigns and then returns `out` parameters values.
```
cston, line 264: >. [](http://example.com/codeflow?start=193&length=1)

Missing space.

proposals/low-level-struct-improvements.md
```diff
@@ -351,31 +364,46 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
...
+* A `out` parameter would be considered `safe-to-escape` within the *current method*.
 
 Detailed Notes:
 - A `ref` field can only be declared inside of a `ref struct` 
 - A `ref` field cannot be declared `static`, `volatile` or `const`
+- A `ref` field cannot have a type that is `ref struct`
 - The reference assembly generation process must preserve the presence of a `ref` field inside a `ref struct` 
 - A `readonly ref struct` must declare its `ref` fields as `readonly ref`
-- The span safety rules document will be updated as outlined in this document.
+- A `scoped` modifier must appear before `in`, `out`, or `ref`
```
cston, line 397: >must appear before [](http://example.com/codeflow?start=22&length=18)

"may appear before"?

proposals/low-level-struct-improvements.md
jaredpar: Made it more specific but chose warning over error. The reasoning is that making it an error places burden on generator tools. They have to do semantic analysis before knowing if a type is or is not `ref struct` before they can emit `scoped` on the parameter. Historically in LDM we've preferred warning in these scenarios to benefit generators. Basically it's not wrong, it's just not doing what you want.

proposals/low-level-struct-improvements.md
```diff
@@ -285,8 +294,14 @@ Span<int> Use()
 }
 ```
 
+Further treating the input to an `out` parameter as returnable is extremely confusing to developers. It essentially subverts the intent of `out` to force developers to think about the original value that the language intends to be assigned by the caller. That is why going forward languages implementing `ref` safety will be required to ensure that `out` parameters are never returned.
```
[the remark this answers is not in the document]
jaredpar: Tried cleaning this up a bit, let me know if it's clearer now.

proposals/low-level-struct-improvements.md
jaredpar: In the case arguments to `in` parameters are not lvalue the compiler generates a temporary and passes that by `ref`. This is passing a local by `ref` hence *ref-safe-to-escape* should be *current method*. Does that make sense or am I missing something?

proposals/low-level-struct-improvements.md
jaredpar: I prefer **must** here because it leaves open the door to do `ref scoped` in the future.

proposals/low-level-struct-improvements.md
jaredpar: yeah that made no sense, removed

jaredpar: Tried rewording this a bit. Let me know if it reads better now.

proposals/low-level-struct-improvements.md
jaredpar: > Is this correct?

The scenario I'm trying to get at is the following: 

```c#
// The RFSTE and STE of `nested` is *calling method*
void M(Nested nested)
{
  // The STE is *current method* 
  // The RFST is still *calling method* 
  scoped Nested local = nested; 

}
```

That is correct, otherwise it allows for us to do actions like `local.Span = stackalloc`. If I can make this clearer happy to do so. 

> Can't we say the ref-field-safe-to-escape scope is no smaller than the container scope?

Can add that too.

proposals/low-level-struct-improvements.md
Sergio0694: Sure. For instance, in the Store, we built a custom ETW provider, which allows users to pass in an interpolated string expression and then have it be null-terminated and passed directly to the native API without having to allocate an intermediary string (we're using this all over the place for all of our ETW logging). The logging APIs exposed to callers look look something like this (simplified):

```csharp
void ILogger.LogInformation(ref ILogger.InterpolatedStringHandler handler)
{
    using var context = handler.DangerousGetContext();

    TheNativePInvokeEtwAPI(context.Text, ...);
}
```

Now, the `ILogger.InterpolatedStringHandler` is essentially very similar to `DefaultInterpolatedStringHandler`, except it exposes that `DangerousGetContext()` API which looks like this, conceptually (this is currently implemented with unsafe hacks):

```csharp
ref struct InterpolatedStringHandler
{
    public Context DangerousGetContext()
    {
        return new(ref this);
    }

    public readonly ref struct Context
    {
        private readonly ref InterpolatedStringHandler _handler;

        internal Context(ref InterpolatedStringHandler handler)
        {
            _handler = handler;
            Text = handler.AppendNullTerminatorAndGetBuffer();
        }

        public ReadOnlySpan<char> Text { get; } // This should be unscoped

        public void Dispose()
        {
            _handler.Clear();
        }
    }
}
```

Essentially, `Context` is the only way to access the raw span, and it ensures the null-terminator is added before you get it. When that's disposed, the parent handler is also disposed so its buffer is returned. Now, there's two caveats:
- The `Context` value returned by `DangerousGetContext` escapes `this`, so it should be `unscoped`. This would be the classic "we're escaping `ref this`" where `unscoped` was meant for.
- The `Text` span exposed by `Context` **also** cannot escape `Context`, as its buffer is owned by `Context` (and transitively, by its parent handler). As such, you'd want to also annotate that `Text` property with `unscoped`, to ensure that its lifetime cannot escape that of its parent `Context`, even though it's not technically escaping a reference to it.

The latter is what I meant for `unscoping` also being very useful for expressing some sort of "ownership" semantics, even in cases where the issue is not strictly speaking an actual ref to an instance escaping out of scope.

Does that make sense?

proposals/low-level-struct-improvements.md
cston: I was referring to the previous clause: 
> In the case the argument is an lvalue then the *ref-safe-to-escape* is that of the lvalue

Should that be ... ?
> In the case the argument is an lvalue then the *ref-safe-to-escape* is **the calling method**

proposals/low-level-struct-improvements.md
jaredpar: > That is, could a dev manually apply LifetimeAnnotation with the relevant parameters and still achieve the ability to return a ref field on a struct or is this non-achievable until support is later added?

Possibly. I will think on that a bit. 

The problem we're facing at the moment is: 

1. `unscoped` is really only useful for a single case. There are others but they're pretty esoteric (at least the ones we've considered). It's not really at the bar for a keyword.
2. `unscoped ref` is a good natural default for `struct` but pretty awful for `ref struct` hence can't categorically invert the default here. 

Using the attribute directly on the type / member is possibly a good middle ground. The concern I would have is that will force us to essentially find all the other uses, which are incorrect, and mark them as an error. Basically I don't want to get into a situation where devs start adding them to fields, type parameters, etc ... places we may want to use lifetime annotations in the future, and suddenly we cause friction on upgrade.

proposals/low-level-struct-improvements.md
jaredpar: I will allows this change :smile:

proposals/low-level-struct-improvements.md
```diff
@@ -486,13 +514,11 @@ The type will have the following definition:
 [AttributeUsage(AttributeTargets.Struct | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Parameter)]
 public sealed class LifetimeAnnotationAttribute : Attribute
 {
```
jaredpar, line 516: @cston wanted to make sure you saw the change here. Think for a while we'll need to accetp both ctors since the attribute is out there but we ourselves should move to the single value case (assuming you agree with my design change here).

proposals/low-level-struct-improvements.md
cston: "Must" could imply that all `in`, `out`, and `ref` modifiers require a preceding `scoped` modifier. Would it make sense to qualify the statement? Perhaps: "For by-ref values, the `scoped` modifier must appear before `in`, `out`, or `ref`"

proposals/low-level-struct-improvements.md
cston: We should consider a more general `int` value rather than one or two `bool` values, to allow for more expanded use in the future.

proposals/low-level-struct-improvements.md
jcouv: I'm pointing out that the current implementation. So either the spec can say error, or the implementation should be discussed/adjusted.

proposals/low-level-struct-improvements.md
Sergio0694: Clarified on Discord, `unscoped` would not support this specific scenario anyway.
Thanks @jaredpar for bearing with me and explaining this in detail 😄
Resolving this.

proposals/low-level-struct-improvements.md
RikkiGibson: I haven't had the chance to look closely here but I do think we should have some discussion for how the representation might evolve and how "older" and "newer" versions of the representation would coexist.

jaredpar: This is why we changed the design from `bool` to `int`. That should allow us future expansion points into new lifetime areas that won't completely bust older compilers that encounter them.

Issue #6051, opened 2022-04-22
[Proposal]: Support default parameter values in lambdas


captainsafia:
# Support default parameter values in lambdas

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

## Summary

[summary]: #summary

To build on top of the lambda improvements introduced in C#11, we propose adding support for default parameter values in lambdas. This would enable users to implement the following lambdas.

```csharp
var addWithDefault = (int addTo = 2) => addTo + 1;
addWithDefault(); // 3
addWithDefault(5); // 6
```

## Motivation

[motivation]: #motivation

App frameworks in the .NET ecosystem leverage lambdas heavily to allow users to quickly write business logic associated with an endpoint.

```csharp
var app = WebApplication.Create(args);

app.MapPost("/todos/{id}", (int id, string task, TodoService todoService) => {
  var todo = todoService.Create(id, task);
  return Results.Created(todo);
});
```

Lambdas don't currently support setting default values on parameters, so if a developer wanted to build an application that was resilient to scenarios where users didn't provide data, they're left to either use local functions or set the default values within the lambda body, as opposed to the more succinct proposed syntax.

```csharp
var app = WebApplication.Create(args);

app.MapPost("/todos/{id}", (int id, string task = "foo", TodoService todoService) => {
  var todo = todoService.Create(id, task);
  return Results.Created(todo);
});
```

The proposed syntax also has the benefit of reducing confusing differences between lambdas and local functions, making it easier to reason about constructs and "grow up" lambdas to functions without compromising features.

## Detailed design

[design]: #detailed-design

Currently, when a user implements a lambda with a default value, the compiler raises a `CS1065` error.

```csharp
var addWithDefault = (int addTo = 2) => addTo + 1;
```

Default values can be applied to lambda parameters with the following behavior.

```csharp
var addWithDefault = (int addTo = 2) => addTo + 1;
addWithDefault(); // 3
addWithDefault(5); // 6
```

The default value will be emitted to metadata. Users can introspect the `DefaultValue` in the `ParameterInfo` associated with the lambda's `MethodInfo`.

```csharp
var addWithDefault = (int addTo = 2) => addTo + 1;
addWithDefault.Method.GetParameters()[0].DefaultValue; // 2
```

Similar to the behavior for delegates with `out` parameters or , a new natural type is generated for each lambda defined with any default parameter values.

```csharp
[...]

PR #6274, opened 2022-07-13
Add lambda/method group defaults proposal

adamperlin:
Championed Issue: #6051. This is a draft for a proposal to add default parameters to lambda functions and method groups, based heavily on the description in the championed issue. This topic has been discussed several times, but the proposal has yet to be merged.

### DIFF
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -322,3 +322,15 @@ var a = (int i = 13) => 1;
 var b = ([DefaultParameterValue(13)] int i) => 1;
 b = a; // Allowed
 ```
+
+**Open question:** First, note that this is outside the scope of the current proposal but it might be worth discussing in the future.
+Do we want to support defaults with implicitly typed lambda parameters? I.e., 
+
+```csharp
+delegate void M1(int i = 3);
+M1 m = (x = 3) => x + x; // Ok
+
+delegate void M2(long i = 2);
+M2 m = (x = 3.0) => ...; //Error: cannot convert implicitly from long to double
+```
+This inference leads to some tricky conversion issues which would require more discussion.
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-To build on top of the lambda improvements introduced in C#10, we propose adding support for default parameter values in lambdas. This would enable users to implement the following lambdas.
+To build on top of the lambda improvements introduced in C#10 (see [relevant background](#relevant-background)), we propose adding support for default parameter values in lambdas. This would enable users to implement the following lambdas:
 
 ```csharp
 var addWithDefault = (int addTo = 2) => addTo + 1;
@@ -23,6 +23,11 @@ int addWithDefaultMethod(int addTo = 2) {
 }
 ```
 
+## Relevant Background
+[Lambda Improvements in C# 10](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md)
+
+[Method group conversion specification §10.8](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/conversions.md#108-method-group-conversions)
+
 ## Motivation
 
 [motivation]: #motivation
@@ -32,7 +37,7 @@ App frameworks in the .NET ecosystem leverage lambdas heavily to allow users to 
 ```csharp
 var app = WebApplication.Create(args);
 
-app.MapPost("/todos/{id}", (int id, string task, TodoService todoService) => {
+app.MapPost("/todos/{id}", (TodoService todoService, int id, string task) => {
   var todo = todoService.Create(id, task);
   return Results.Created(todo);
 });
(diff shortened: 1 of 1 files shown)


jcouv:
Let's add some pointers to relevant background:
- 10.8 method group conversions: https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/conversions.md#108-method-group-conversions
- Lambda improvements from C# 10: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md

proposals/lambda-method-group-defaults.md
```diff
@@ -0,0 +1,280 @@
...
+## Motivation
+
+[motivation]: #motivation
+
+App frameworks in the .NET ecosystem leverage lambdas heavily to allow users to quickly write business logic associated with an endpoint.
+
+```csharp
+var app = WebApplication.Create(args);
+
+app.MapPost("/todos/{id}", (int id, string task, TodoService todoService) => {
```
333fred, line 35:
```suggestion
app.MapPost("/todos/{id}", (TodoService todoService, int id, string task) => {
```

proposals/lambda-method-group-defaults.md
```diff
@@ -0,0 +1,280 @@
...
+  return Results.Created(todo);
+});
+```
+
+Lambdas don't currently support setting default values on parameters, so if a developer wanted to build an application that was resilient to scenarios where users didn't provide data, they're left to either use local functions or set the default values within the lambda body, as opposed to the more succinct proposed syntax.
+
+```csharp
+var app = WebApplication.Create(args);
+
+app.MapPost("/todos/{id}", (int id, string task = "foo", TodoService todoService) => {
```
333fred, line 46:
```suggestion
app.MapPost("/todos/{id}", (TodoService todoService, int id, string task = "foo") => {
```

svick:
This is a breaking change, right? Currently, [the following code compiles](https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoTCAsACAWQApYAXPKPAXjxQEo8BvAX00wDcIAnPMawgNzYUSADxkAfHgj8wAoA===):

```c#
void M(int i = 1) {}

var m = M;
Action<int> a = m;
```

But with this proposal, this will be an error, since `m` will no longer be `Action<int>`.

adamperlin:
I have now revised the proposal based on LDM, PR, and other feedback so I think it is ready for a second look

davidfowl:
Excellent write up @adamperlin!

333fred (approved):
@cston, please take a look as well.

proposals/lambda-method-group-defaults.md
```diff
@@ -0,0 +1,324 @@
...
+// internal delegate int b'(int arg0 = 13);
+var d = (int c = 13) => 1;
+// internal delegate int b'(int arg0 = 13);
+
+var e = E;
+// internal delegate int b'(int arg0 = 13);
+var f = F;
+// internal delegate int c'(int arg0 = 0);
+var g = G;
+// internal delegate int c'(int arg0 = 0);
```
cston, line 201: >c'(int arg0 = 0) [](http://example.com/codeflow?start=25&length=16)

Should this be `b'(int arg0 = 13)`?

proposals/lambda-method-group-defaults.md
```diff
@@ -0,0 +1,324 @@
...
+
+a = b; // Not allowed
+a = c; // Allowed
+a = d; // Allowed
+c = e // Allowed
```
cston, line 206: >e [](http://example.com/codeflow?start=4&length=1)

Missing `;` after these next few assignment statements.

proposals/lambda-method-group-defaults.md
```diff
@@ -0,0 +1,324 @@
...
+  return a;
+}
+
+delegate int Del(int a = 1);
+
+Del del = (int x = 100) => x; // Error: Default parameter does not match, so no conversion can be performed
+                              
+Del del1 = (int x = 1) => x; // Allowed, because default parameter value in lambda matches default parameter value in delegate
+
+Del del2 = D; // This behavior does not change and compiles as before as per the method group conversion rules
```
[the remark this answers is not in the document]
cston: It looks like it's covered below:

> `d = M; // Allowed with warning.`

proposals/lambda-method-group-defaults.md
```diff
@@ -0,0 +1,336 @@
...
+Do we want to support defaults with implicitly typed lambda parameters? I.e., 
+
+```csharp
+delegate void M1(int i = 3);
+M1 m = (x = 3) => x + x; // Ok
+
+delegate void M2(long i = 2);
+M2 m = (x = 3.0) => ...; //Error: cannot convert implicitly from long to double
+```
+This inference leads to some tricky conversion issues which would require more discussion.
```
jaredpar, line 336: It's not just inference that is a problem, it would increase the look ahead for our parser. Consider the state of the parser when it's looking at the expression that starts with a `(`. It has to decide if that is a lambda expression or not. Today the parser can bail out the moment it gets to `(x =` as that disqualifies it from being a lambda. If we allowed defaults on implicit then it has to parse all the way through the `=>`.

adamperlin: Ah thanks for pointing this out. I will add a note about that as well

proposals/lambda-method-group-defaults.md
adamperlin: Yes it should be. Thanks for catching that

proposals/lambda-method-group-defaults.md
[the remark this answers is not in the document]
adamperlin: Yep the case is covered there

cston (approved):

PR #6277, opened 2022-07-14
Spec `ref` in object initializers

jcouv:
Relates to https://github.com/dotnet/roslyn/issues/62120

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -562,6 +562,50 @@ This also has the added benefit that it will make `fixed` buffers easier to cons
 
 The backing storage for the buffer will be generated using the `[InlineArray]` attribute. This is a mechanism discussed in [issue 12320](https://github.com/dotnet/runtime/issues/12320) which allows specifically for the case of efficiently declaring sequence of fields of the same type. This particular issue is still under active discussion and the expectation is that the implementation of this feature will follow however that discussion goes.
 
+### Initializers with `ref` values in `new` and `with` expressions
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -562,6 +562,50 @@ This also has the added benefit that it will make `fixed` buffers easier to cons
...
+    : 'ref' expression // added
+    : expression
+    | object_or_collection_initializer
+    ;
+```
+
+In the section for [`with` expression](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/records.md#with-expression), we update the grammar to:
+```antlr
+member_initializer
+    : identifier '=' 'ref' expression // added
```
333fred, line 580: We don't require `ref` before the identifier as well? Don't we normally require `ref` on the lvalue for `ref` assignment?

jcouv: We don't require `ref` on the LHS:
```
public ref struct S
{
    ref int refField;
    public void M(ref int x)
    {
        refField = ref x; // ok
        ref refField = ref x; // error CS1001: Identifier expected
    }
}
```

Here's spec for local ref re-assignment as a special-case of ref assignment: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.3/ref-local-reassignment.md

333fred: Huh. Totally misremembering it then.

cston (approved):

PR #6278, opened 2022-07-14
Remove GC hole from `low-level-struct-improvements.md`

DaZombieKiller:
While reading through the "violating scoped" section, I noticed that it (presumably accidentally) suggests a workaround that contains a GC hole, as it attempts to roundtrip a `ref` through an unmanaged pointer in order to get around span safety rules.

It appears that ultimately the decision was made to go through with one of the other considerations instead (changing `Unsafe.AsRef` to use `scoped`), so this PR simply removes the section containing the GC hole in the hopes that it will prevent future readers from copying it.

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -756,17 +756,7 @@ The runtime repository has several non-public APIs that capture `ref` parameters
 
 The majority of these APIs will likely choose to have proper lifetime tracking on the return which will be achieved simply by updating to C# 11. A few though will want to keep their current semantics of not tracking the return value because their entire intent is to be unsafe. The most notable examples are `MemoryMarshal.CreateSpan` and `MemoryMarshal.CreateReadOnlySpan`. This will be achieved by marking the parameters as `scoped`.
 
-That means the runtime needs an established pattern for unsafely removing `scoped` from a parameter. This can be done today via a combination of existing methods:
-
-```c#
-Span<T> CreateSpan<T>(scoped ref T value, int length)
-{
-    ref T local = Unsafe.AsRef<T>(Unsafe.AsPointer(ref value));
-    return new Span<T>(local, length);
-}
-```
-
-This will work but is likely going to result in unnecessary code generation. One other consideration is that either:
+That means the runtime needs an established pattern for unsafely removing `scoped` from a parameter:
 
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -756,17 +756,7 @@ The runtime repository has several non-public APIs that capture `ref` parameters
...
-
-```c#
-Span<T> CreateSpan<T>(scoped ref T value, int length)
-{
-    ref T local = Unsafe.AsRef<T>(Unsafe.AsPointer(ref value));
-    return new Span<T>(local, length);
-}
-```
-
-This will work but is likely going to result in unnecessary code generation. One other consideration is that either:
```
saucecontrol, line 769: I would argue that the fact one can bounce through a pointer to remove `scoped` is, in fact, an important consideration.  In the absence of proper support from the language for removing `scoped`, developers will end up with exactly this code.

This note is incorrect, however, as it will definitely not work in some cases because it's a potential GC hole.  The fact that this is difficult for most developers to reason about makes it a serious pit of failure with the current `scoped ref` feature.

RikkiGibson:
The issue here is that we may run a GC between the ref being converted to pointer and the pointer being converted back to ref, right?

saucecontrol:
Yep, exactly.

333fred:
> The issue here is that we may run a GC between the ref being converted to pointer and the pointer being converted back to ref, right?

Maybe I'm missing something, but why would that be a GC hole? The method has a managed ref to `T` as a parameter the whole time. Does it need a `GC.KeepAlive` on that `T` after the conversion?

saucecontrol:
It's not just the liveness. The managed ref parameter would be updated by GC but the pointer local wouldn't, so then it's converted to a managed ref pointing at the wrong location.

DaZombieKiller:
If the object moves between the `AsPointer`/`AsRef` calls, then the pointer passed to `AsRef` may be incorrect, as it's not tracked by the GC and won't be updated.

333fred:
Ah, I see, thanks. It's not the object being collected, it's being moved. Presumably it would need a `fixed` in between to prevent it?

DaZombieKiller:
A `fixed` block would be one potential solution, additional restrictions that imposes aside (`unmanaged`, etc.)

saucecontrol:
In other words, it's really only safe if you pin and then use `MemoryMarshal.CreateSpan` instead of `new Span`.

cston (approved):

jaredpar (approved):

RikkiGibson:
Thanks @DaZombieKiller!

PR #6361, opened 2022-08-10
Correct sample code for explicit implementation

peter-dolkens:
Confirmed intention of sample code with Bill Wagner on Twitter


BillWagner (approved):
Thanks @peter-dolkens 

This looks good to me. Thanks for spotting it and making the change..

PR #6365, opened 2022-08-11
Update list-patterns-enumerables.md

alrz:
[Rendered](https://github.com/alrz/csharplang/blob/patch-15/proposals/list-patterns-enumerables.md)

### DIFF
--- a/proposals/list-patterns-enumerables.md
+++ b/proposals/list-patterns-enumerables.md
@@ -5,6 +5,7 @@
 Lets you to match an enumerable with a sequence of patterns e.g. `enumerable is [ 1, 2, 3 ]` will match a sequence of the length three with 1, 2, 3 as its elements.
 
 ## Detailed design
+
 The pattern syntax is unchanged:
 
 ```antlr
--- a/proposals/list-patterns-enumerables.md
+++ b/proposals/list-patterns-enumerables.md
@@ -2,16 +2,14 @@
 
 ## Summary
 
-Lets you to match an enumerable with a sequence of patterns e.g. `enumerable is { 1, 2, 3 }` will match a sequence of the length three with 1, 2, 3 as its elements.
+Lets you to match an enumerable with a sequence of patterns e.g. `enumerable is [ 1, 2, 3 ]` will match a sequence of the length three with 1, 2, 3 as its elements.
 
 ## Detailed design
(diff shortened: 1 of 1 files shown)


alrz:
cc @jcouv @333fred

333fred:
@alrz what's the reason for the large amount of removals?

alrz:
Because we no longer need to deal with trailing patterns for enumerable types.

I confirmed with Julien if we want to do that, at least initially.

alrz:
I tried to keep things that are still relevant. Let me know if there's other things you'd like to keep.

333fred:
> Because we no longer need to deal with trailing patterns for enumerable types.

I'm not sure what you mean by this?

proposals/list-patterns-enumerables.md
```diff
@@ -11,214 +11,53 @@ The pattern syntax is unchanged:
...
 ```
 
 ### Pattern compatibility
 
-A *length_pattern* is now also compatible with any type that is not *countable* but is *enumerable* — it can be used in `foreach`.
+When the type is not *indexable* or *countable*, pattern compatibility is defined as follows:
 
-A *list_pattern* is now also compatible with any type that is not *indexable* but is *enumerable*.
+- A *list_pattern* is compatible with any type that is *enumerable* — it can be used in `foreach`.
+- A *slice_pattern* is compatible with any type that is compatible with a *list_pattern* but no subpatterns are allowed and it can only appear as the last item in the list.
```
jcouv, line 24: >and it can only appear as the last item in the list. [](http://example.com/codeflow?start=120&length=52)

I don't think we should have this limitation. 
Let's add a note to open questions below. I'll queue this for LDM since didn't get a strong answer from Mads in email.

jcouv (approved):
LGTM Thanks (iteration 3)

alrz:
To not depend on the impl details, I think the compiler could only generate a typeclass for struct-enumerators by implementing:
```cs
interface Iterator<TSelf, T> {
  static abstract bool MoveNext(TSelf);
  static abstract T GetCurrent(TSelf);
}
```
and pass it on to the helper fucntion as a type argument.

A different overload could be used for `IEnumerator<T>` so we don't need to generate additional code for the base scenario.

PR #6367, opened 2022-08-11
Update required members spec

333fred:
Update with changes from https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-08-10.md#required-property-default-nullability.

### DIFF
--- a/proposals/required-members.md
+++ b/proposals/required-members.md
@@ -210,17 +210,21 @@ this proposal should attempt to address it.
 
 ### Effect on nullable analysis
 
-Members that are marked `required` are not required to be initialized to a non-nullable state at the end of a constructor. All `required` members from this type and any base types are considered
-by nullable analysis to be maybe-null at the beginning of any constructor in that type, unless chaining to a `this` or `base` constructor that is attributed with `SetsRequiredMembersAttribute`.
+Members that are marked `required` are not required to be initialized to a valid nullable state at the end of a constructor. All `required` members from this type and any base types are considered
+by nullable analysis to be default at the beginning of any constructor in that type, unless chaining to a `this` or `base` constructor that is attributed with `SetsRequiredMembersAttribute`.
 
-Constructors attributed with `SetsRequiredMembersAttribute` will warn about all required fields and properties (including from base types) that do not have a null state matching their declared
-null state.
+Nullable analysis will warn about all `required` members from the current and base types that do not have a valid nullable state at the end of a constructor attributed with `SetsRequiredMembersAttribute`.
 
 ```cs
 #nullable enable
 public class Base
 {
     public required string Prop1 { get; set; }
+
+    public Base() {}
+
+    [SetsRequiredMembers]
+    public Base(int unused) { Prop1 = ""; }
 }
 public class Derived : Base
 {
@@ -243,6 +247,13 @@ public class Derived : Base
     {
         Prop1.ToString(); // Ok
         Prop2.ToString(); // Ok
+    }
+
+    [SetsRequiredMembers]
+    public Derived(int unused1, int unused2, int unused3) : base(unused1)
+    {
+        Prop1.ToString(); // Ok
+        Prop2.ToString(); // Warning: possibly null dereference
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


333fred:
@RikkiGibson for a review.

proposals/required-members.md
```diff
@@ -210,10 +210,42 @@ this proposal should attempt to address it.
 
 ### Effect on nullable analysis
 
-Members that are marked `required` are not required to be initialized to a non-nullable state at the end of a constructor. These members are additionally considered by nullable analysis to
-be maybe-null at the beginning of any constructor in a type, unless chaining to a `this` or `base` constructor that is attributed with `SetsRequiredMembersAttribute`.
+Members that are marked `required` are not required to be initialized to a non-nullable state at the end of a constructor. All `required` members from this type and any base types are considered
+by nullable analysis to be maybe-null at the beginning of any constructor in that type, unless chaining to a `this` or `base` constructor that is attributed with `SetsRequiredMembersAttribute`.
```
RikkiGibson, line 214: I think *non-nullable* and *maybe-null* could be replaced with more precise terms, but we don't need to fix that in this PR. For example, if the member's type is an unconstrained type parameter we could have a *maybe-default* state at the start of the constructor and require either a *maybe-null* or *not-null* state at the end of the constructor. (I'll convert this comment to issue and we can resolve separately at some point when we have time.)

proposals/required-members.md
```diff
@@ -210,10 +210,42 @@ this proposal should attempt to address it.
...
+
+    [SetsRequiredMembers]
+    public Derived(int unused) : base()
+    {
+        Prop1.ToString(); // Warning: possibly null dereference
+        Prop2.ToString(); // Warning: possibly null dereference
+    }
+
+    [SetsRequiredMembers]
+    public Derived(int unused, int unused2) : this()
```
RikkiGibson, line 242: it would be useful to have a sample which uses a `[SetsRequiredMembers] public Base() { }` constructor, then chains to it from Derived, and indicates the expected initial state for `Prop1`. (would expect not-null if we chain to a Base constructor with SetsRequiredMembers.)

proposals/required-members.md
```diff
@@ -210,10 +210,42 @@ this proposal should attempt to address it.
...
 ### Effect on nullable analysis
 
-Members that are marked `required` are not required to be initialized to a non-nullable state at the end of a constructor. These members are additionally considered by nullable analysis to
-be maybe-null at the beginning of any constructor in a type, unless chaining to a `this` or `base` constructor that is attributed with `SetsRequiredMembersAttribute`.
+Members that are marked `required` are not required to be initialized to a non-nullable state at the end of a constructor. All `required` members from this type and any base types are considered
+by nullable analysis to be maybe-null at the beginning of any constructor in that type, unless chaining to a `this` or `base` constructor that is attributed with `SetsRequiredMembersAttribute`.
 
-NB: `SetsRequiredMembersAttribute` does not bring back nullable warnings at the end of a constructor for members, as it is an escape hatch intended to inform the compiler to stop enforcing things.
+Constructors attributed with `SetsRequiredMembersAttribute` will warn about all required fields and properties (including from base types) that do not have a null state matching their declared
+null state.
```
RikkiGibson, line 217:
```suggestion
Constructors attributed with `SetsRequiredMembersAttribute` will warn about all required fields and properties (including from base types) that do not have a null state matching their declared
null state on exit.
```

RikkiGibson (approved):

proposals/required-members.md
333fred: Do you have suggestions for what those terms should be? I can fix them now.

RikkiGibson: I wish I had introduced compact terms for these in https://github.com/dotnet/csharplang/blob/main/proposals/nullable-constructor-analysis.md#an-alternative-approach. We can start now by saying:

- instead of saying members are *maybe-null* at the beginning, say that they have the *default state* at the beginning.
- instead of saying members must be *non-null* at the end, say that they must have a *valid state* at the end (using *a* to indicate that more than one flow state could be *valid*).

I might go ahead and send a PR to define and use these terms over in that doc as well.

PR #6442, opened 2022-09-08
Update syntax for `out scoped` variable declaration

cston:
Use `out scoped` rather than `scoped out` since `scoped` applies to the value in the calling method rather than the reference in the method being called.


AlekseyTs (approved):

333fred (commented):
LGTM (commit 1)

PR #6443, opened 2022-09-09
Fix typo

dayo05:
Change name of variable about multiplication of unchecked to r6
Maybe he/she wants to say r6


333fred (approved):

333fred:
Thanks @dayo05!

PR #6483, opened 2022-09-23
Spec out as return only 

jaredpar:
Considering how the spec changes would look if we went forward with `out` parameters having *safe-to-return* of *return only*

https://github.com/dotnet/roslyn/issues/64155

### DIFF
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -387,18 +387,17 @@ The presence of `ref` fields means the rules around method arguments must match 
 
 > For any method invocation `e.M(a1, a2, ... aN)`
 > 1. Calculate the *safe-to-escape* of the method return.
->     - Ignore the *ref-safe-to-escape* of arguments to `in / ref / out` parameters of `ref struct` types. The corresponding parameters *ref-safe-to-escape* are at most *return only* and hence cannot be returned via `ref` or `out` parameters.
+>     - Ignore the *ref-safe-to-escape* of arguments whose corresponding parameters have a *ref-safe-to-escape* of *return-only* or smaller.
 >     - Assume the method has a `ref struct` return type.
-> 2. All `ref` or `out` arguments of `ref struct` types must be assignable by a value with that *safe-to-escape*. This applies even when the `ref` argument matches a `scoped ref` parameter.
+> 2. All `ref` arguments of `ref struct` types must be assignable by a value with that *safe-to-escape*. This applies even when the `ref` argument matches a `scoped ref` parameter.
 
-> Further for any method invocation `e.M(a1, a2, ... aN)` where the target method has at least one `[UnscopedRef] out` parameter
+> Further for any method invocation `e.M(a1, a2, ... aN)`
 > 1. Calculate the *safe-to-escape* of the method return.
->     - Include the *ref-safe-to-escape* of arguments to `[UnscopedRef] out` parameters
->     - Ignore the *ref-safe-to-escape* of arguments to all other `in / ref / out` parameters
+>     - Ignore the *ref-safe-to-escape* of arguments whose corresponding parameters are `scoped`
 >     - Assume the method has a `ref struct` return type.
 > 2. All `out` arguments of `ref struct` types must be assignable by a value with that *safe-to-escape*.
 
-The presence of `scoped` allows developers to reduce the friction this rule creates by marking parameters which are not returned as `scoped`. This removes their arguments from (1) above and provides greater flexibility to callers.
+The presence of `scoped` allows developers to reduce the friction this rule creates by marking parameters which are not returned as `scoped`. This removes their arguments from (1) in both cases above and provides greater flexibility to callers.
 
 Impact of this change is discussed more deeply [below](#examples-method-arguments-must-match). Overall this will allow developers to make call sites more flexible by annotating non-escaping ref-like values with `scoped`.
 
@@ -1504,9 +1503,9 @@ S Usage()
 }
 ```
 
-To make these APIs usable the compiler has to create separation between how freely `ref` and their values can be returned by default. This is the rational for having `ref` to `ref struct` and `out` be [implicitly scoped](#implicitly-scoped).
+To make these APIs usable the compiler ensures that `ref` have a smaller lifetime than their associated parameters. This is the rational This is the rational for having *ref-safe-to-escape* for `ref` to `ref struct` be *return only* and `out` be *containing method*. That prevents cyclic assignment because of the difference in lifetimes.
 
-It is also why `[UnscopedRef]` only promotes the *ref-safe-to-escape* of such values to *return only* and not *calling method*. Consider that using *calling method* forces a viral use of `[UnscopedRef]` for a `ref struct`:
+It is also why `[UnscopedRef]` only promotes the *ref-safe-to-escape* of any `ref` to `ref struct` values to *return only* and not *calling method*. Consider that using *calling method* allows for cyclic assignment and would force a viral use of `[UnscopedRef]` for a `ref struct`:
 
 ```c#
 ref struct S
@@ -1525,6 +1524,30 @@ void M(ref S s)
 ```
 
 This is correctly illegal in that case because the compiler has to consider the pathological case that `S.Data` could cyclic assign via `this`. That forces methods all methods that call `S.Data` to further mark their `ref` parameters as `[UnscopedRef]`. This is viral until the method which creates the value as a local. This is why *return only* exists as an escape scope. It does complicate the spec / implementation but it serves to make the feature significantly more usable.
+
+Note: this cyclic assignment problem does continue to exist for `[UnscopedRef] out` to `ref struct` because that causes the *safe-to-escape* and *ref-safe-to-escape* to be equivalent. 
+
+```c#
+ref struct RS
+{
+    int field;
+    ref int refField;
+}
+
+void M1(out RS p)
+{
+    // Error: method arguments must match rules do the following
+    // Step 1 calculate would calculate the smallest escape as *containing method*
+    // Step 2 would fail the assignment check because p safe-to-escape is *return only*
+    M1(out p);
+}
+
+void M2([UnscopedRef] out RS p)
+{
+    // The lifetimes of LHS and RHS are equivalent here and hence this is legal
+    p.refField = ref p.Field;
+}
+```
 
 In terms of advanced annotations the `[UnscopedRef]` design creates the following:
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -333,17 +333,17 @@ The language "does not contribute" means the arguments are simply not considered
 
 The method invocation rules can now be simplified. The receiver no longer needs to be special cased, in the case of `struct` it is now simply a `scoped ref T`. The value rules need to change to account for `ref` field returns:
 
-> A value resulting from a method invocation `e1.M(e2, ...)` is *safe-to-escape* from the smallest of the following scopes:
+> A value resulting from a method invocation `e1.M(e2, ...)` is *safe-to-escape* from the narrowest of the following scopes:
 > 1. The *calling method*
-> 2. When the return is a `ref struct`:
->    1. The *safe-to-escape* contributed by all argument expressions
->    2. The *ref-safe-to-escape* contributed by all `ref` arguments
+> 2. When the return is a `ref struct` the *safe-to-escape* contributed by all argument expressions
+> 3. When the return is a `ref struct` the *ref-safe-to-escape* contributed by all `ref` arguments
 
 The `ref` calling rules can be simplified to:
 
-> A value resulting from a method invocation `ref e1.M(e2, ...)` is *ref-safe-to-escape* the smallest of the following scopes:
-> 1. The *safe-to-escape* of the rvalue of `e1.M(e2, ...)`
-> 2. The *ref-safe-to-escape* contributed by all `ref` arguments
+> A value resulting from a method invocation `ref e1.M(e2, ...)` is *ref-safe-to-escape* the narrowest of the following scopes:
+> 1. The *calling method*
+> 2. The *safe-to-escape* contributed by all argument expressions
+> 3. The *ref-safe-to-escape* contributed by all `ref` arguments
 
 This rule now lets us define the two variants of desired methods:
 
--- a/proposals/low-level-struct-improvements.md
+++ b/proposals/low-level-struct-improvements.md
@@ -157,7 +157,7 @@ Next the rules for ref reassignment need to be adjusted for the presence of `ref
 The left operand of the `= ref` operator must be an expression that binds to a ref local variable, a ref parameter (other than `this`), an out parameter, **or a ref field**.
 
 > For a ref reassignment in the form ...
-> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* at least *return only* then `e2` must have at least the same *ref-safe-to-escape* as `x`.
+> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* at least *return only* then `e2` must have at least *ref-safe-to-escape* at least as large as `x`
 > 2. `e1 = ref e2`: where `e1` is a `ref` local or `ref` parameter then `e2` must have a *safe-to-escape* equal to *safe-to-escape* for `e1` and `e2` must have *ref-safe-to-escape* at least as large as *ref-safe-to-escape* of the *ref-safe-to-escape* of `e1`
 
 That means the desired `Span<T>` constructor works without any extra annotation:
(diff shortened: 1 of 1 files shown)


proposals/low-level-struct-improvements.md
```diff
@@ -1513,6 +1525,30 @@ void M(ref S s)
...
+
+void M1(out RS p)
+{
+    // Error: method arguments must match rules do the following
+    // Step 1 calculate would calculate the smallest escape as *containing method*
+    // Step 2 would fail the assignment check because p safe-to-escape is *return only*
+    M1(out p);
+}
+
+void M2([UnscopedRef] out RS p)
```
jaredpar, line 1545: I gave a lot of consideration to `[UnscopedRef] out` being a compilation error. Basically just cut off the problem. But it is realistic, although very very corner, that someone would have assigned and returned an `out` to a `ref struct` by `ref` so decided to leave it.

proposals/low-level-struct-improvements.md
```diff
@@ -157,7 +157,7 @@ Next the rules for ref reassignment need to be adjusted for the presence of `ref
 The left operand of the `= ref` operator must be an expression that binds to a ref local variable, a ref parameter (other than `this`), an out parameter, **or a ref field**.
 
 > For a ref reassignment in the form ...
-> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* to *calling method* then `e2` must be *ref-safe-to-escape* to the *calling method*
+> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* at least *return only* then `e2` must have at least the same *ref-safe-to-escape* as `x`.
```
cston, line 160: >have at least the same *ref-safe-to-escape* as `x`. [](http://example.com/codeflow?start=90&length=51)

Perhaps "must have *ref-safe-to-escape* at least as large as `x`", to match the second bullet, and so it's clear that larger is allowed rather than smaller.

proposals/low-level-struct-improvements.md
```diff
@@ -383,11 +387,17 @@ The presence of `ref` fields means the rules around method arguments must match
 
 > For any method invocation `e.M(a1, a2, ... aN)`
 > 1. Calculate the *safe-to-escape* of the method return.
->     - Ignore the *ref-safe-to-escape* of arguments to `in / ref / out` parameters of `ref struct` types. The corresponding parameters *ref-safe-to-escape* are at most *return only* and hence cannot be returned via `ref` or `out` parameters.
+>     - Ignore the *ref-safe-to-escape* of arguments whose corresponding parameters have a *ref-safe-to-escape* of *return-only* or smaller.
+>     - Assume the method has a `ref struct` return type.
+> 2. All `ref` arguments of `ref struct` types must be assignable by a value with that *safe-to-escape*. This applies even when the `ref` argument matches a `scoped ref` parameter.
+
+> Further for any method invocation `e.M(a1, a2, ... aN)`
+> 1. Calculate the *safe-to-escape* of the method return.
```
cston, line 395: >*safe-to-escape* of the method return [](http://example.com/codeflow?start=19&length=37)

It's a little confusing that we're calling this the "*safe-to-escape* of the method return", and also in the previous set of bullets, but we ignore different arguments in each case.

I'm not sure how to make it clearer though. Perhaps just add a "where:" to the end of bullet 1 in each case.
> 1. Calculate the *safe-to-escape* of the method return *where*:

proposals/low-level-struct-improvements.md
RikkiGibson: One thing I considered briefly was having the STE of `[UnscopedRef] out` be CallingMethod, to ensure that `outParam = M(ref outParam)` is an error. But it just started to feel too complicated. I'm happy with *play stupid games, win stupid prizes* as a resolution here.

proposals/low-level-struct-improvements.md
```diff
@@ -1491,9 +1503,9 @@ S Usage()
 }
 ```
 
-To make these APIs usable the compiler has to create separation between how freely `ref` and their values can be returned by default. This is the rational for having `ref` to `ref struct` and `out` be [implicitly scoped](#implicitly-scoped).
+To make these APIs usable the compiler ensures that `ref` have a smaller lifetime than their associated parameters. This is the rational This is the rational for having *ref-safe-to-escape* for `ref` to `ref struct` be *return only* and `out` be *containing method*. That prevents cyclic assignment because of the difference in lifetimes.
```
RikkiGibson, line 1506: Looks like *this is the rational* is repeated and was probably meant to be *this is the rationale*.

proposals/low-level-struct-improvements.md
```diff
@@ -157,7 +157,7 @@ Next the rules for ref reassignment need to be adjusted for the presence of `ref
 The left operand of the `= ref` operator must be an expression that binds to a ref local variable, a ref parameter (other than `this`), an out parameter, **or a ref field**.
 
 > For a ref reassignment in the form ...
-> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* to *calling method* then `e2` must be *ref-safe-to-escape* to the *calling method*
+> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* at least *return only* then `e2` must have at least *ref-safe-to-escape* at least as large as `x`
```
RikkiGibson, line 160:
```suggestion
> 1. `x.e1 = ref e2`: where `x` is *safe-to-escape* at least *return only* then `e2` must have *ref-safe-to-escape* at least as large as `x`
```

proposals/low-level-struct-improvements.md
RikkiGibson:
```suggestion
To make these APIs usable the compiler ensures that `ref` have a smaller lifetime than their associated parameters. This is the rationale for having *ref-safe-to-escape* for `ref` to `ref struct` be *return only* and `out` be *containing method*. That prevents cyclic assignment because of the difference in lifetimes.
```

cston, line 1506: >ensures that `ref` have a smaller lifetime than their associated parameters. [](http://example.com/codeflow?start=39&length=76)

Perhaps:
> ensures that the `ref` lifetime for a `ref` parameter is smaller than lifetime of any references in the associated parameter value.

proposals/low-level-struct-improvements.md
```diff
@@ -1513,6 +1525,30 @@ void M(ref S s)
...
+    int field;
+    ref int refField;
+}
+
+void M1(out RS p)
+{
+    // Error: method arguments must match rules do the following
+    // Step 1 calculate would calculate the smallest escape as *containing method*
+    // Step 2 would fail the assignment check because p safe-to-escape is *return only*
+    M1(out p);
```
cston, line 1542: >M1 [](http://example.com/codeflow?start=4&length=2)

`M2`?

proposals/low-level-struct-improvements.md
```diff
@@ -1513,6 +1525,30 @@ void M(ref S s)
...
+ref struct RS
+{
+    int field;
+    ref int refField;
+}
+
+void M1(out RS p)
+{
+    // Error: method arguments must match rules do the following
+    // Step 1 calculate would calculate the smallest escape as *containing method*
```
cston, line 1540: >calculate  [](http://example.com/codeflow?start=14&length=10)

Remove leading "calculate".

proposals/low-level-struct-improvements.md
```diff
@@ -1513,6 +1525,30 @@ void M(ref S s)
...
+```c#
+ref struct RS
+{
+    int field;
+    ref int refField;
+}
+
+void M1(out RS p)
+{
+    // Error: method arguments must match rules do the following
```
cston, line 1539: >do the following [](http://example.com/codeflow?start=48&length=16)

Perhaps:
> // Error: from method arguments must match:

proposals/low-level-struct-improvements.md
```diff
@@ -1569,6 +1605,42 @@ readonly ref struct SpanOfOne
 
 This means we must choose the shallow interpretation of `readonly`.
 
+#### Modeling constructors 
+One subtle tricky design point is how are constructors bodies modeled for ref safety? Essentially how is the following constructor analyzed?
```
cston, line 1609: >? [](http://example.com/codeflow?start=84&length=1)

"." rather than "?"

cston, line 1609: >One subtle tricky design point is how are [](http://example.com/codeflow?start=0&length=41)

Perhaps:
> One subtle design question is: How are ... ?

proposals/low-level-struct-improvements.md
```diff
@@ -1569,6 +1605,42 @@ readonly ref struct SpanOfOne
...
+    public S(ref int f)
+    {
+        field = ref f;
+    }
+}
+```
+
+There are roughly two approaches:
+
+1. Model as a `static` method where `this` is a local where it's *safe-to-escape* is *calling method*
```
cston, line 1625: >a local where it's [](http://example.com/codeflow?start=46&length=18)

Should this be "a return value where *safe-to-escape* is ..."?

proposals/low-level-struct-improvements.md
```diff
@@ -1569,6 +1605,42 @@ readonly ref struct SpanOfOne
...
+2. Model as a `static` method where `this` is a `ref` or `out` parameter.
```
cston, line 1626: >where `this` is a `ref` or `out` parameter [](http://example.com/codeflow?start=30&length=42)

Could we drop the `ref` case and use "... where `this` is an `out` parameter" instead?

proposals/low-level-struct-improvements.md
jaredpar: I was really trying to lean on method invocation rules here and essentially re-use them. Cause really that is what this is doing, using the method invocation rules to calculate the narrowest possible return and validate it's okay to assign to every `ref` / `out` parameter. The method invocation rules are simple enough now, maybe just restate them here and note that this is effectively applyiyng method escape rules.

jaredpar: @cston here is the version where I don't rely on re-using method invocation rules. Do you think this is clearer? 

> For any method invocation `e.M(a1, a2, ... aN)`
> 1. Calculate the minimum of the following:
>     - *calling method*
>     - The *safe-to-escape* of all arguments
>     - The *ref-safe-to-escape* of all ref arguments whose corresponding parameters have a *ref-safe-to-escape* of *calling method*
> 2. All `ref` arguments of `ref struct` types must be assignable by a value with that *safe-to-escape*.

proposals/low-level-struct-improvements.md
jaredpar: Are you accusing me of mis-spelling a word? ... yeah I'm guilty

proposals/low-level-struct-improvements.md
jaredpar: At least you all are doing a good job reveiwing my changes 

:face-palm:

proposals/low-level-struct-improvements.md
cston: That sounds good. One concern though is step 2 refers to "a value with that *safe-to-escape*" but step 1 doesn't call the result a *safe-to-escape* value. Perhaps step 1 could be:
> Calculate a *safe-to-escape* from the minimum of:

proposals/low-level-struct-improvements.md
jaredpar: I think it has to be a local. Consider that a constructor body has assignments, method calls, etc ... All of those need to happen on a value. If it's return only when binding the method body how do we evaluate the calls? A local works because it can be STS return only and RSTS containing method. 

Yes that means there is a copy at the end. That would matter if we actually implemented ctors that way. Instead this is just how do we model ref safety analysis. A local mostly works.

proposals/low-level-struct-improvements.md
jaredpar: @RikkiGibson are you okay with this change of the MAMM language?

RikkiGibson: I think of calling method as the "minimum" scope, so I found the language confusing.

RikkiGibson: In general I'd like if we could avoid saying bigger, smaller, minimum, etc. for scope, and say narrower, wider, narrowest, widest. Otherwise mentally switching between the spec and implementation becomes very difficult.

I imagine that when we invest in this area of the compiler later on, we could consider introducing a scope type where "narrower" scopes are also considered "smaller" per the comparison operators. I think that's one of several aspects of the implementation which makes understanding it more difficult.

jaredpar: Okay, updated this section.

cston (approved):

proposals/low-level-struct-improvements.md
```diff
@@ -313,9 +313,10 @@ The design also requires that the introduction of a new escape scope: *return on
...
 
-There are two locations which default to *return only*:
-- A `ref` or `in` parameter for a `ref struct` will have a *ref-safe-to-escape* of *return only*.  This is necessary to prevent [silly cyclic assignment](#cyclic-assignment) issues.
+There are three locations which default to *return only*:
+- A `ref` or `in` parameter. This is done in part for `ref struct` to prevent [silly cyclic assignment](#cyclic-assignment) issues. It is done uniformly though to simplify the model as well as minimize compat changes.
 - A `out` parameter for a `ref struct` will have *safe-to-escape* of *return only*. This allows for return and `out` to be equally expressive. This does not have the silly cyclic assignment problem because `out` is implicitly `scoped` so the *ref-safe-to-escape* is still smaller than the *safe-to-escape*.
+- A `this` parameter for a `struct` constructor. This falls out due them being modeled as `out` parameters.
```
cston, line 319: "due to being modeled..."?

proposals/low-level-struct-improvements.md
```diff
@@ -548,7 +550,9 @@ The [rationale](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-
 
 <a name="rules-unscoped"></a>
 
-To fix this the  language will provide the opposite of the `scoped` lifetime annotation by supporting an `UnscopedRefAttribute`. This can be applied to any `ref` which is [implicitly `scoped`](#implicitly-scoped). The *ref-safe-to-escape* of a `ref` annotated with `[UnscopedRef]` is that of a normal `ref` of the same type. For example if the target type the `ref` modifies is a `ref struct` then the *ref-safe-to-escape* is *return only*, otherwise it is *calling method*.
+To fix this the  language will provide the opposite of the `scoped` lifetime annotation by supporting an `UnscopedRefAttribute`. This can be applied to any `ref` and it will change the *ref-safe-to-escape* to be one level wider than its default. For example:
+- if applied to a `struct` instance method it will become *return only* where previously it was *containing method*.
+- if applied to a `ref` parameter it will become *containing method* where previously it was *return only*
```
cston, line 555: "it will become *calling method* where ..."?

RikkiGibson (approved):

RikkiGibson:
@cston think it would be fine to just commit the changes you suggested to the PR and merge.

proposals/low-level-struct-improvements.md
```diff
@@ -313,10 +313,15 @@ The design also requires that the introduction of a new escape scope: *return on
...
 
-A `ref` or `in` parameter for a `ref struct` will have a *ref-safe-to-escape* of *return only*.  This is necessary to prevent [silly cyclic assignment](#cyclic-assignment) issues.
+There are three locations which default to *return only*:
+- A `ref` or `in` parameter. This is done in part for `ref struct` to prevent [silly cyclic assignment](#cyclic-assignment) issues. It is done uniformly though to simplify the model as well as minimize compat changes.
+- A `out` parameter for a `ref struct` will have *safe-to-escape* of *return only*. This allows for return and `out` to be equally expressive. This does not have the silly cyclic assignment problem because `out` is implicitly `scoped` so the *ref-safe-to-escape* is still smaller than the *safe-to-escape*.
+- A `this` parameter for a `struct` constructor. This falls out due them being modeled as `out` parameters.
```
cston, line 319:
```suggestion
- A `this` parameter for a `struct` constructor. This falls out due to being modeled as `out` parameters. 
```

proposals/low-level-struct-improvements.md
```diff
@@ -536,7 +550,9 @@ The [rationale](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-
 
 <a name="rules-unscoped"></a>
 
-To fix this the  language will provide the opposite of the `scoped` lifetime annotation by supporting an `UnscopedRefAttribute`. This can be applied to any `ref` which is [implicitly `scoped`](#implicitly-scoped). The *ref-safe-to-escape* of a `ref` annotated with `[UnscopedRef]` is that of a normal `ref` of the same type. For example if the target type the `ref` modifies is a `ref struct` then the *ref-safe-to-escape* is *return only*, otherwise it is *calling method*.
+To fix this the  language will provide the opposite of the `scoped` lifetime annotation by supporting an `UnscopedRefAttribute`. This can be applied to any `ref` and it will change the *ref-safe-to-escape* to be one level wider than its default. For example:
+- if applied to a `struct` instance method it will become *return only* where previously it was *containing method*.
+- if applied to a `ref` parameter it will become *containing method* where previously it was *return only*
```
cston, line 555:
```suggestion
- if applied to a `ref` parameter it will become *calling method* where previously it was *return only*
```

PR #6542, opened 2022-10-13
Add lambda `params` proposal

jjonescz:
Relates to test plan https://github.com/dotnet/roslyn/issues/62485

### DIFF
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -376,7 +376,7 @@ The default parameter values will be emitted to metadata. The IL for this featur
 inherits from `System.Delegate` or similar will be generated, and the `Invoke` method will include `.param` directives to set default parameters or `System.ParamArrayAttribute` &ndash; 
 just as would be the case for a standard named delegate with default or `params` parameters.
 
-These delegate types are able to be inspected at runtime, as normal.
+These delegate types can be inspected at runtime, as normal.
 In code, users can introspect the `DefaultValue` in the `ParameterInfo` associated with the lambda or method group by using the associated `MethodInfo`.
 
 ```csharp
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -81,7 +81,7 @@ app.MapPost("/todos/{id}", todoHandler);
 ```
 
 ## Current Behavior
-Currently, when a user implements a lambda with a default or `params` parameter, the compiler raises an error stating that it's not allowed. 
+Currently, when a user implements a lambda with a default or `params` parameter, the compiler raises an error.
 
 ```csharp
 var addWithDefault = (int addTo = 2) => addTo + 1; // error CS1065: Default values are not valid in this context.
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -212,6 +212,19 @@ Same rules as for method parameters ([§14.6.2](https://github.com/dotnet/csharp
 
 No changes to the grammar are necessary for method groups since this proposal would only change their semantics.
 
+### Updates of prior proposals
+
+The following addition (in bold) is required to the [function types](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md#natural-function-type) specification in a prior proposal:
+
+> The natural type of an anonymous function expression or method group is a *function_type*. A *function_type* represents a method signature: the parameter types, **default values, ref kinds, `params` modifiers**, and return type and ref kind. Anonymous function expressions or method groups with the same signature have the same *function_type*.
+
+The following addition (in bold) is required to the [delegate types](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md#delegate-types) specification in a prior proposal:
+
+> The delegate type for the anonymous function or method group with parameter types `P1, ..., Pn` and return type `R` is:
+>
+> - if any parameter or return value is not by value, **or any parameter is optional or `params`**, or there are more than 16 parameters, or any of the parameter types or return are not valid type arguments (say, `(int* p) => { }`), then the delegate is a synthesized `internal` anonymous delegate type with signature that matches the anonymous function or method group, and with parameter names `arg1, ..., argn` or `arg` if a single parameter;
+> [...]
+
 ### Binder changes
 
 #### Synthesizing new delegate types
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -287,7 +287,7 @@ d = (int c = 10) => 2; // Error: default parameter value is different between ne
                        // and synthesized delegate b'. We won't do implicit conversion
 ```
 
-Anonymous delegates with an array as the last parameter will be unified when the last parameter has the same `params` annotation, regardless of its name.
+Anonymous delegates with an array as the last parameter will be unified when the last parameter has the same `params` modifier, regardless of its name.
 
 ```csharp
 int C(int[] xs) {
@@ -344,7 +344,7 @@ int E(params int[] xs) {
 
 delegate int DelParams(params int[] xs);
 
-DelParams del4 = (int[] xs) => xs.Length; // Error: `params` annotation does not match, so no conversion can be performed
+DelParams del4 = (int[] xs) => xs.Length; // Error: `params` modifier does not match, so no conversion can be performed
 
 DelParams del5 = (params int[] xs) => xs.Length; // Allowed
 
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -354,6 +354,17 @@ var e = E;
 // e is inferred as internal delegate int b'(params int[] arg);
 
 DelParams del7 = e; // Not allowed. Cannot convert internal delegate type to DelParams.
+
+delegate int DelArray(int[] xs);
+
+DelArray del8 = (params int[] xs) => xs.Length; // Error: conversion from `params` to non-`params`
+
+del8 = E; // Allowed: both have `params`
+
+var del9 = (params int[] xs) => xs.Length;
+// del9 is inferred as internal delegate int b'(params int[] arg);
+
+del8 = del9; // Not allowed. Cannot convert internal delegate type to DelArray.
 ```
 
 Since lambdas and method groups with optional and `params` parameters are typed as anonymous delegates, the
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -11,9 +11,9 @@ var addWithDefault = (int addTo = 2) => addTo + 1;
 addWithDefault(); // 3
 addWithDefault(5); // 6
 
-var count = (params int[] xs) => xs.Length;
-count(); // 0
-count(1, 2, 3); // 3
+var counter = (params int[] xs) => xs.Length;
+counter(); // 0
+counter(1, 2, 3); // 3
 ```
 
 Similarly, we will allow the same kind of behavior for method groups:
@@ -108,9 +108,9 @@ var addWithDefault = (int addTo = 2) => addTo + 1;
 addWithDefault(); // 3
 addWithDefault(5); // 6
 
-var count = (params int[] xs) => xs.Length;
-count(); // 0
-count(1, 2, 3); // 3
+var counter = (params int[] xs) => xs.Length;
+counter(); // 0
+counter(1, 2, 3); // 3
 ```
 
 Default values and `params` can be applied to method group parameters by specifically defining such method group:
@@ -123,12 +123,12 @@ int addWithDefault(int addTo = 2) {
 var add1 = addWithDefault; 
 add1(); // ok, default parameter will be used
 
-int count(params int[] xs) {
+int counter(params int[] xs) {
   return xs.Length;
 }
 
-var count1 = count;
-count1(1, 2, 3); // ok, `params` will be used
+var counter1 = counter;
+counter1(1, 2, 3); // ok, `params` will be used
 ```
 
 ## Breaking Change
@@ -219,7 +219,7 @@ var addWithDefault = (int addTo = 2) => addTo + 1;
 // internal delegate int a'(int arg = 2);
 var printString = (string toPrint = "defaultString") => Console.WriteLine(toPrint);
 // internal delegate void b'(string arg = "defaultString");
-var count = (params int[] xs) => xs.Length;
+var counter = (params int[] xs) => xs.Length;
 // internal delegate int c'(params int[] arg);
 string pathJoin(string s1, string s2, string sep = "/") { return $"{s1}{sep}{s2}"; }
 var joinFunc = pathJoin;
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -4,7 +4,7 @@
 
 [summary]: #summary
 
-To build on top of the lambda improvements introduced in C# 10 (see [relevant background](#relevant-background)), we propose adding support for default parameter values and [`params` arrays](https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/params) in lambdas. This would enable users to implement the following lambdas:
+To build on top of the lambda improvements introduced in C# 10 (see [relevant background](#relevant-background)), we propose adding support for default parameter values and `params` arrays in lambdas. This would enable users to implement the following lambdas:
 
 ```csharp
 var addWithDefault = (int addTo = 2) => addTo + 1;
--- a/proposals/lambda-method-group-defaults.md
+++ b/proposals/lambda-method-group-defaults.md
@@ -269,6 +269,36 @@ d = (int c = 10) => 2; // Error: default parameter is different between new lamb
                        // and synthesized delegate b'. We won't do implicit conversion
 ```
 
+Anonymous delegates with an array as the last parameter will be unified when the last parameter has the same `params` annotation, regardless of its name.
+
+```csharp
+int C(int[] xs) {
+  return xs.Length;
+}
+
+int D(params int[] xs) {
+  return xs.Length;
+}
+
+var a = (int[] xs) => xs.Length;
+// internal delegate int a'(int[] xs);
+var b = (params int[] xs) => xs.Length;
+// internal delegate int b'(params int[] xs);
+
+var c = C;
+// internal delegate int a'(int[] xs);
+var d = D;
+// internal delegate int b'(params int[] xs);
+
+a = b; // Not allowed
+a = c; // Allowed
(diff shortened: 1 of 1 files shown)


proposals/lambda-method-group-defaults.md
```diff
@@ -69,41 +81,54 @@ app.MapPost("/todos/{id}", todoHandler);
 ```
 
 ## Current Behavior
-Currently, when a user implements a lambda with a default value, the compiler raises an error stating that default values are not allowed. 
+Currently, when a user implements a lambda with a default or `params` parameter, the compiler raises an error stating that it's not allowed.
```
cston, line 84: >raises an error stating that it's not allowed.  [](http://example.com/codeflow?start=94&length=47)

Since there are examples below, perhaps just "reports an error."

proposals/lambda-method-group-defaults.md
```diff
@@ -18,9 +22,16 @@ var addWithDefault = addWithDefaultMethod;
 addWithDefault(); // 3
 addWithDefault(5) // 6
 
+var counter = countMethod;
```
cston, line 25: >counter [](http://example.com/codeflow?start=4&length=7)

Minor: It looks like we switch between `var count = ...` and `var counter = ...` throughout the document. Unless there is a reason for having two distinct names, consider choosing one.

proposals/lambda-method-group-defaults.md
```diff
@@ -145,31 +190,44 @@ only been supported since C# 10, so only code which has been written since then
 ### Grammar and Parser Changes
 This enhancement requires the following changes to the grammar for lambda expressions.
 ```diff
-explicit_anonymous_function_parameter
+ explicit_anonymous_function_parameter_list
++    : explicit_anonymous_function_parameters
++    | explicit_anonymous_function_parameters ',' parameter_array
```
cston, line 195: Consider combining:
```
    : explicit_anonymous_function_parameters (',' parameter_array)?
```

proposals/lambda-method-group-defaults.md
```diff
@@ -145,31 +190,44 @@ only been supported since C# 10, so only code which has been written since then
...
 
 #### Synthesizing New Delegate Types
-As with the behavior for delegates with `ref` or `out` parameters, a new natural type is generated for each lambda or method group defined with any default parameter values.
+As with the behavior for delegates with `ref` or `out` parameters, a new natural type is generated for each lambda or method group defined with any default or `params` parameters.
 Note that in the below examples, the notation `a'`, `b'`, etc. is used to represent these anonymous delegate types.
 
 ```csharp
 var addWithDefault = (int addTo = 2) => addTo + 1;
-// internal delegate int b'(int arg0 = 2);
+// internal delegate int a'(int arg0 = 2);
```
cston, line 219: >int arg0 [](http://example.com/codeflow?start=28&length=8)

`int arg`

proposals/lambda-method-group-defaults.md
```diff
@@ -145,31 +190,44 @@ only been supported since C# 10, so only code which has been written since then
...
+var count = (params int[] xs) => xs.Length;
+// internal delegate int c'(params int[] arg0);
 string pathJoin(string s1, string s2, string sep = "/") { return $"{s1}{sep}{s2}"; }
 var joinFunc = pathJoin;
 // internal delegate string d'(string arg0, string arg1, string arg3 = " ");
 ```
 
 #### Conversion and Unification Behavior 
-The delegates described previously will be unified when the same parameter (based on position) has the same default value, regardless of parameter name.
+Anonymous delegates with default parameters will be unified when the same parameter (based on position) has the same default value, regardless of parameter name.
```
cston, line 231: Do we need a similar statement for delegates for `params` cases?

proposals/lambda-method-group-defaults.md
```diff
@@ -292,10 +373,10 @@ D1 d = Fun; // Allowed WITHOUT warning, since Fun and its default parameter are
 ### IL/Runtime Behavior
 
 The default parameter values will be emitted to metadata. The IL for this feature will be very similar in nature to the IL emitted for lambdas with `ref` and `out` parameters. A class which
-inherits from `System.Delegate` or similar will be generated, and the `Invoke` method will include `.param` directives to set default parameters -- 
-just as would be the case for a standard named delegate with default parameters.
+inherits from `System.Delegate` or similar will be generated, and the `Invoke` method will include `.param` directives to set default parameters or `System.ParamArrayAttribute` &ndash; 
+just as would be the case for a standard named delegate with default or `params` parameters.
 
-These delegate types are able to inspected at runtime, as normal.
+These delegate types are able to be inspected at runtime, as normal.
```
cston, line 379: >are able to be [](http://example.com/codeflow?start=21&length=14)

Perhaps "can be".

proposals/lambda-method-group-defaults.md
```diff
@@ -1,15 +1,19 @@
-# Default Parameters for Lambdas and Method Groups
+# Default and Variable Parameters for Lambdas and Method Groups
```
333fred, line 1:
```suggestion
# Default and Parameter Array Parameters for Lambdas and Method Groups
```

proposals/lambda-method-group-defaults.md
```diff
@@ -1,15 +1,19 @@
...
 
 ## Summary
 
 [summary]: #summary
 
-To build on top of the lambda improvements introduced in C#10 (see [relevant background](#relevant-background)), we propose adding support for default parameter values in lambdas. This would enable users to implement the following lambdas:
+To build on top of the lambda improvements introduced in C# 10 (see [relevant background](#relevant-background)), we propose adding support for default parameter values and [`params` arrays](https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/params) in lambdas. This would enable users to implement the following lambdas:
```
333fred, line 7: I don't think we really need the docs link. If we want to link to anything, we should link to the csharpstandard repo, but I don't think a link is necessary at all here.

proposals/lambda-method-group-defaults.md
```diff
@@ -69,41 +81,54 @@ app.MapPost("/todos/{id}", todoHandler);
...
-void addWithDefault(int addTo = 2) {
+int addWithDefault(int addTo = 2) {
   return addTo + 1;
 }
 
 var add1 = addWithDefault; 
-add1(); // ok, default parameter will be used.
+add1(); // ok, default parameter will be used
+
+int count(params int[] xs) {
```
333fred, line 126: While we're here, can you do a pass to make all of these method names follow C# `PascalCase` convention?

proposals/lambda-method-group-defaults.md
jcouv, line 84: >default or `params` parameter [](http://example.com/codeflow?start=50&length=29)

nit and not directly related to this PR: the spec uses "default_argument" and "optional parameter", but I've not seen "default parameter" before.

proposals/lambda-method-group-defaults.md
```diff
@@ -145,31 +190,44 @@ only been supported since C# 10, so only code which has been written since then
...
-    ;
+     ;
 ```
 No changes to the grammar are necessary for method groups since this proposal would only change their semantics.
 
 ### Binder Changes
 
 #### Synthesizing New Delegate Types
-As with the behavior for delegates with `ref` or `out` parameters, a new natural type is generated for each lambda or method group defined with any default parameter values.
+As with the behavior for delegates with `ref` or `out` parameters, a new natural type is generated for each lambda or method group defined with any default or `params` parameters.
```
jcouv, line 214: We should spell out how we are updating the prior spec on [function types](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md#natural-function-type) and [delegate types](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md#delegate-types)?

For instance, the delegate types should now say something like:

> The delegate type for the anonymous function or method group with parameter types P1, ..., Pn and return type R is:
> - if any parameter or return value is not by value, **or any parameter is optional or `params`,** or there are more than 16 parameters, or any of the parameter types or return are not valid type arguments (say, (int* p) => { }), then the delegate is a synthesized internal anonymous delegate type with signature that matches the anonymous function or method group, and with parameter names arg1, ..., argn or arg if a single parameter;
> ...


The function type section would now say something like:
> A function_type represents a method signature: the parameter types, **default value, and ref kind and `params` modifier,** and return type and ref kind.

The function type section also says:
```
Function_types are used in a few specific contexts only:
- implicit and explicit conversions
- method type inference (§11.6.3) and best common type (§11.6.3.15)
- var initializers
````
So those sections should be reviewed (and updated as needed) for impact of default values and `params` as well.

File: proposals/lambda-method-group-defaults.md [hunk 11]
```diff
@@ -145,31 +190,44 @@ only been supported since C# 10, so only code which has been written since then
...
+ explicit_anonymous_function_parameter_list
++    : explicit_anonymous_function_parameters
++    | explicit_anonymous_function_parameters ',' parameter_array
++    | parameter_array
++    ;
++
++explicit_anonymous_function_parameters
+     : explicit_anonymous_function_parameter
+       (',' explicit_anonymous_function_parameter)*
+     ;
```
jcouv, line 202: Looking at the baseline spec for [method parameters](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1462-method-parameters), consider whether some parts should be repeated here.

For instance, "A parameter with a ref, out or this modifier cannot have a default_argument.", "A parameter_array may occur after an optional parameter, but cannot have a default value – the omission of arguments for a parameter_array would instead result in the creation of an empty array."

Or we could just refer to that section and say that the same rules apply.

jcouv (commented):

proposals/lambda-method-group-defaults.md
jjonescz: Good catch, thanks. 0661f1fb0b893c301f2d8255fcd1fd08d1a04939

proposals/lambda-method-group-defaults.md
jjonescz: Right, thanks. Fixed also in other places: bc44898d4a61c1efe13e00c486dc3bb03323c040

proposals/lambda-method-group-defaults.md
jjonescz: Added: 8f426e696075d4da1bc724b7a52ababee9acf8c6

proposals/lambda-method-group-defaults.md
jjonescz: Thanks. 0286e34e15bc1b1fb67c091c7364fa5db108f003

proposals/lambda-method-group-defaults.md
jjonescz: What about "Optional and Parameter Array Parameters for Lambdas and Method Groups" (since "default parameters" are not a thing https://github.com/dotnet/csharplang/pull/6542#discussion_r994975532)? 3351f2725b0b524cd9d7000cfbd8d96d5b0c82be. And wouldn't "Optional and `params` Array Parameters for Lambdas and Method Groups" be even better?

Also, should I change the filename (currently `lambda-method-group-defaults.md`)?

proposals/lambda-method-group-defaults.md
jjonescz: OK, thanks. Removed the link: b811c74f82222febb9630d215ef68698edea21d2

proposals/lambda-method-group-defaults.md
jjonescz: Good idea. 322230de0c13a49e980d67dc0ebe5e54ed26a04d

proposals/lambda-method-group-defaults.md
jjonescz: Good catch, I have also been wondering whether to change that. So, now I did: 9c3e329cb5c26b955774f73f2f89714c2a28dbbb

proposals/lambda-method-group-defaults.md
jjonescz: Alright, thanks, I added how the prior proposal should be updated: 9c3828f58a451d8a8e502933034e6f4032a6727e.

> The function type section also says:
> 
> ```
> Function_types are used in a few specific contexts only:
> - implicit and explicit conversions
> - method type inference (§11.6.3) and best common type (§11.6.3.15)
> - var initializers
> ```
> 
> So those sections should be reviewed (and updated as needed) for impact of default values and `params` as well.

I don't follow. Do you mean to update the section that defines the contexts where *function_type*s are used? I think those stay the same.

proposals/lambda-method-group-defaults.md
jjonescz: Makes sense, thanks. 3649b5bf0a6060442c998d8c46167a23504d65b8

proposals/lambda-method-group-defaults.md
```diff
@@ -134,42 +169,82 @@ DoAction(writeInt, 3); // Error, cannot convert from anonymous delegate type to
...
+>
+> - if any parameter or return value is not by value, **or any parameter is optional or `params`**, or there are more than 16 parameters, or any of the parameter types or return are not valid type arguments (say, `(int* p) => { }`), then the delegate is a synthesized `internal` anonymous delegate type with signature that matches the anonymous function or method group, and with parameter names `arg1, ..., argn` or `arg` if a single parameter;
+> [...]
 
-#### Synthesizing New Delegate Types
-As with the behavior for delegates with `ref` or `out` parameters, a new natural type is generated for each lambda or method group defined with any default parameter values.
+### Binder changes
+
+#### Synthesizing new delegate types
+As with the behavior for delegates with `ref` or `out` parameters, a new natural type is generated for each lambda or method group defined with any optional or `params` parameters.
```
cston, line 231: >a new natural type is generated for each lambda or method group [](http://example.com/codeflow?start=67&length=63)

This text wasn't added in this PR but it's not clear what "natural type" means. Assuming this refers to the synthesized delegate type, perhaps it should say "delegate types are synthesized for lambdas or method groups with optional or `params` parameters."

proposals/lambda-method-group-defaults.md
```diff
@@ -208,33 +283,80 @@ e = f; // Not Allowed
 b = f; // Allowed
 e = g; // Allowed
 
-d = (int c = 10) => 2; // Error: default parameter is different between new lambda
+d = (int c = 10) => 2; // Error: default parameter value is different between new lambda
                        // and synthesized delegate b'. We won't do implicit conversion
 ```
 
-Similarly, there is of course compatibility with named delegates that have default parameters as well: 
+Anonymous delegates with an array as the last parameter will be unified when the last parameter has the same `params` modifier, regardless of its name.
```
cston, line 290: >the same `params` modifier, regardless of its name [](http://example.com/codeflow?start=100&length=50)

Perhaps "the same `params` modifier and array type, regardless of parameter name".

proposals/lambda-method-group-defaults.md
```diff
@@ -208,33 +283,80 @@ e = f; // Not Allowed
...
+
+DelParams del5 = (params int[] xs) => xs.Length; // Allowed
+
+DelParams del6 = E; // Allowed as before per the method group conversion rules
+
+var e = E;
+// e is inferred as internal delegate int b'(params int[] arg);
+
+DelParams del7 = e; // Not allowed. Cannot convert internal delegate type to DelParams.
 ```
```
cston, line 358: Consider including a conversion to non-`params` delegate if not already covered:
```csharp
delegate int DelArray(int[] xs);
DelArray d8 = (params int[] xs) => xs.Length; // Allowed
d8 = E; // Allowed
var e = (params int[] xs) => xs.Length;
d8 = e; // Not allowed. Cannot convert internal delegate type to DelArray.
```

cston: Should an error be reported converting a `params` lambda expression to a non-`params` delegate?

proposals/lambda-method-group-defaults.md
jcouv: I meant, make sure to read the rest of the [C# 10 lambda spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md?rgh-link-date=2022-10-13T18%3A30%3A26Z#natural-function-type) and make similar edits to other sections that might be affected.

333fred (approved):

proposals/lambda-method-group-defaults.md
jjonescz: It's not clear to me either. Changed: 69d5c7ecbbbc173f9280083af57491fcfb3b7245 Thanks.

proposals/lambda-method-group-defaults.md
jjonescz: That sounds better, thanks. 755b66af3cfdd99d8ef804eea3f1299fc01fd965

proposals/lambda-method-group-defaults.md
jjonescz: > Consider including a conversion to non-`params` delegate

Good catch, thanks. Included: 728d5b5a8a5ab34e4a0f7dfd100c0faefe1fc107

> Should an error be reported converting a `params` lambda expression to a non-`params` delegate?

I believe yes for similar reason as converting a lambda expression with optional parameter to a delegate without: the default value / `params` modifier would effectively useless.

cston (approved):

333fred (approved):

proposals/lambda-method-group-defaults.md
```diff
@@ -134,42 +169,82 @@ DoAction(writeInt, 3); // Error, cannot convert from anonymous delegate type to
...
 
-### Grammar and Parser Changes
+### Grammar and parser changes
 This enhancement requires the following changes to the grammar for lambda expressions.
 ```diff
-explicit_anonymous_function_parameter
+ explicit_anonymous_function_parameter_list
++    : explicit_anonymous_function_parameters (',' parameter_array)?
++    | parameter_array
++    ;
```
jaredpar, line 196: The grammar changes imply that we support `params` and optional parameters in anonymous methods. Essentially `delegate (params int[] a)`. I don't see any examples of this in the spec though. Is it intended to be allowed?

jjonescz: Yes, I think that lambdas and anonymous methods are the same in this regard. I don't know if that was the intention. But if yes, I can mention `delegate () {}` anonymous methods in the proposal...

jjonescz: I changed the grammar to disallow these improvements in `delegate { }` syntax: 8585a197713535978d072b975737835fe149a03f

I based my syntax on [the one from `lambda-improvements.md`](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md#syntax). But I don't quite understand why it mentions `equals_value_clause`, it's not defined anywhere and seems unused in that proposal. So, in my diff of the syntax, I act as if it wasn't there...

jaredpar (approved):

jcouv (approved):
LGTM Thanks (iteration 19)

PR #6545, opened 2022-10-14
Fix typos

jnm2:


meetings/2022/LDM-2022-10-05.md
```diff
@@ -38,7 +38,7 @@ https://github.com/dotnet/roslyn/issues/64344.
 
 #### Return-only scope
 
-In response to breaking changes seen by customers with `ref` parameters, we added a new scope in-between the method scope and the calling function
+In response to breaking changes seen by customers with `ref` parameters, we added a new scope in between the method scope and the calling function
```
jnm2, line 41: Apparently the rule is that phrases like this are hyphenated when used as an adjective in front of a noun. In this case the pair of words is used as a preposition.

jcouv (approved):

jcouv:
Thanks :-)

PR #6547, opened 2022-10-14
Move generic attributes to C# 11 sub-folder

jcouv:
FYI @BillWagner, I had incorrectly put the "generic attributes" speclet into C# 10, but we actually held the feature in preview until C# 11.


333fred (approved):

BillWagner:
Addressed in dotnet/docs#31787

PR #6555, opened 2022-10-17
Require types to match between pattern variable redeclarations

jcouv:
FYI @alrz @333fred 
If we allow some type variation, we're going to have to merge types (find the best type, by eliminating tuple name differences and resolving nullability differences) and I don't see the benefit. I'll go ahead and merge for purpose of LDM discussion today.


alrz:
>we're going to have to merge types

True, but that's rather limiting because variable sources can be entirely unrelated.

It feels like a nice-to-have feature though, if it's not too much costly to support.

jcouv:
LDM agreed with merging nullabilities. Will update shortly.

PR #6556, opened 2022-10-17
Allow types to differ in terms of nullability

jcouv:
Also allow between conditions of an `if/else`.
FYI @alrz @333fred

### DIFF
--- a/proposals/pattern-variables.md
+++ b/proposals/pattern-variables.md
@@ -66,7 +66,7 @@
 	- Across top-level condition expressions within a single `if` statement (includes `else if`)
 
   These names can possibly reference either of variables based on the result of the pattern-matching at runtime.
-- Pattern variables with multiple declarations must be of the same type, excluding nullability for reference types.
+- Pattern variables with multiple declarations must be of the same type, excluding top-level nullability for reference types. Differences in nested nullability are subject to standard nullability conversion warnings.
 
         if (e is C c || e is Wrapper { Prop: var c }) { /* c may be null */ }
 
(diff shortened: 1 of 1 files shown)


proposals/pattern-variables.md
```diff
@@ -60,7 +66,9 @@
 	- Across top-level condition expressions within a single `if` statement (includes `else if`)
 
   These names can possibly reference either of variables based on the result of the pattern-matching at runtime.
-- Pattern variables with multiple declarations must be of the same type, including tuple names and nullability for reference types.
+- Pattern variables with multiple declarations must be of the same type, excluding nullability for reference types.
```
333fred, line 69:
```suggestion
- Pattern variables with multiple declarations must be of the same type, excluding top-level nullability for reference types. Differences in nested nullability are subject to standard nullability conversion warnings.
```

333fred (approved):

proposals/pattern-variables.md
```diff
@@ -101,4 +109,4 @@ These rules cover the existing top-level `is not` pattern variables. However, in
 
 ## Unresolved questions
 - Would it be possible to permit different types for each variable especially within `if`/`else` chains?
-  And if so, would it be a part of this proposal or a separate feature?
+  And if so, would it be a part of this proposal or a separate feature? (discussed in LDM 2022/10/17)
```
alrz, line 112: No notes on that discussion?

TBH I don't know what scenario is targeted here if it's limited to identical types.

333fred: Notes were published yesterday.

alrz: I'm talking about if/else chains (edited my comment to clarify).

PR #6561, opened 2022-10-18
Define safe-to-escape inference for out vars

RikkiGibson:
We should have done this at the same time as dotnet/roslyn#64414, but better late than never.

cc @cston @jaredpar @jcouv.

### DIFF
--- a/proposals/csharp-11.0/low-level-struct-improvements.md
+++ b/proposals/csharp-11.0/low-level-struct-improvements.md
@@ -303,43 +303,15 @@ Further treating the input to an `out` parameter as returnable is extremely conf
 This means going forward `out` will match developers intuition when it comes to ref safety: `out` means `out`.
 
 #### Infer *safe-to-escape* of declaration expressions
-We adjust the span-safety proposal to automatically infer a *safe-to-escape* of a declaration expression like the following.
-    
-```cs
-ref struct RS
-{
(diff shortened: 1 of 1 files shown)


proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -1700,6 +1700,45 @@ public static void ctor(out S @this, ref int f)
...
+
+```
+
+The *safe-to-escape* of a declaration variable from an `out` argument or deconstruction (`(var x, var y) = M()`) is the *narrowest* of the following:
+* calling method
+* if out variable is marked `scoped`, then the current local scope (i.e. current method or narrower).
+* if out variable's type is ref struct, consider all arguments to the containing invocation, including the receiver:
+  * STE of any argument where its corresponding parameter is not `out` and has STE of ReturnOnly or wider
+  * RSTE of any argument where its corresponding parameter has RSTE of ReturnOnly or wider
+
```
jaredpar, line 1739: This change is all in the remarks while this is really a spec change. Would move this up into the actual spec section. Think we should add a section for declarations right below where we talk about the changes to `out` parameters.

RikkiGibson: Done, thanks!

RikkiGibson:
Split the examples and spec as discussed offline.

jaredpar (approved):

PR #6599, opened 2022-10-26
Fix generic parameter rendering

Youssef1313:
This is how it previously rendered both in GitHub and in [docs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements)
 
image

Escaping the `<` so that `<T>` isn't treated as HTML tag.

Another way to fix it is to wrap `ReadOnlySpan<T>` with backticks

### DIFF
--- a/proposals/csharp-11.0/low-level-struct-improvements.md
+++ b/proposals/csharp-11.0/low-level-struct-improvements.md
@@ -1174,7 +1174,7 @@ This snippet wants to mutate a parameter by escaping elements of the data. The e
 
 ### Fun Samples
 
-#### ReadOnlySpan<T>
+#### ReadOnlySpan\<T>
 
 ```c#
 public readonly ref struct ReadOnlySpan<T>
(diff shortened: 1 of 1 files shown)


proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -1174,7 +1174,7 @@ This snippet wants to mutate a parameter by escaping elements of the data. The e
 
 ### Fun Samples
 
-#### ReadOnlySpan<T>
+#### ReadOnlySpan\<T>
 
 ```c#
 public readonly ref struct ReadOnlySpan<T>
```
cston, line 1180: Thanks @Youssef1313. Since you're making changes here, perhaps also remove the `<T>` from the constructor signature below since the syntax is incorrect.

cston (approved):

cston:
Thanks @Youssef1313.

PR #6608, opened 2022-10-27
add comment on doc updates

BillWagner:
I know our docs don't cover this use of the `!` operator. Adding that would help this situation.

After more thinking after we met, I'm thinking that might be the only necessary changes.


RikkiGibson (approved):

RikkiGibson:
Thanks!

PR #6633, opened 2022-10-31
Consider lambda optional and `params` parameters in anonymous function conversions spec

jjonescz:
(As suggested by @jcouv in https://github.com/dotnet/roslyn/pull/64961#discussion_r1006166198.)


jcouv:
I think we should update the spec regardless of the behavior we land on.

PR #6655, opened 2022-11-03
Create struct-generic-overloads

Skyppex:
Adding a proposal for a feature which helps reduce code duplication by using Compiler Generated code. This feature allows for structs to be generated with concrete types based on a struct of the same name with different generic arguments.


HaloFour:
Proposals for language features should be submitted as [Discussions](https://github.com/dotnet/csharplang/discussions).

PR #6657, opened 2022-11-04
Add notes from params improvements working group meeting 2022-11-03

cston:


meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
+# Params Improvements Working Group Meeting for November 3, 2022
+The second meeting of the _`params` improvements_ working group was focused on the code generated for arguments to `params ReadOnlySpan<T>`. 
+
+## Span allocation
+The compiler _may allocate a span on the stack_ when:
+- the span is implicitly allocated,
```
CyrusNajmabadi, line 6: might want to define what 'implicit' means here.

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+
+// WriteLine(string format, params ReadOnlySpan<object> args);
+WriteLine("{0}, {1}", _tmp2);
+```
+
+Ideally the BCL will provide types such as `FixedSizeArray1`, `FixedSizeArray2`, etc. for a limited number of array lengths.
+And if the compilation requires spans for other argument lengths, the compiler would generate and emit the additional types.
+
+The compiler will implicitly allocate `params` argument spans on the stack regardless of the number of arguments.
+To avoid implicit stack allocation at a call site, the calling code should allocate the span explicitly.
```
CyrusNajmabadi, line 38: oh, i did'nt realize we made that determination.  this seems risky, if we are going to just add all these new apis.  someone could upgrade, recompile, and potentially blow stack.  e.g. a call to some string.format that takes 1000 args, right?

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+For arguments to a `params ReadOnlySpan<T>` parameter (which is implicitly `scoped`), the conditions above are satisfied.
+
+The compiler will reuse a single span across all calls to `params ReadOnlySpan<T>` methods where the span element types are considered identical by the runtime.
+
+The length of the reused span will be the length of the longest `params` argument list from all uses.
+(The actual span argument passed to a `params` method will be a slice of that reused span, with the expected length for the particular call.)
+
+Reuse may be across distinct call sites _or_ repeated calls from the same call site.
+Reuse is per thread of execution and within the same method only.
+No reuse across lambda expressions, local functions, and the containing methods.
```
CyrusNajmabadi, line 58: the local function case seems unfortunate, esp if we can see that it is never captured as a delegate.  could we distinguish those cases and allow reuse across local function for when that doesn't happen?

CyrusNajmabadi (commented):
LGTM.  a couple of questions, but minor.

meetings/working-groups/params-improvements/PI-2022-11-03.md
RikkiGibson: I don't think this is the full picture. I expected that we would also request a JIT feature where large fixed-size arrays can be heap-allocated instead if heuristics owned by the runtime determine that's the right thing to do.

meetings/working-groups/params-improvements/PI-2022-11-03.md
davidwrighton: The cost of these allocations (assuming they end up on the stack) is likely to be exceedingly low. Possibly just the cost of zero'ing the memory region. Of course this needs measurement to be absolutely confident, but it shouldn't be all that expensive.

meetings/working-groups/params-improvements/PI-2022-11-03.md
davidwrighton: I think we need justification for worrying about that scenario. If we really want large fixed size locals to be heap allocated we need concrete evidence this is a real problem. Not speculation. I know I said we *could* do this. However, we won't build such a heuristic without evidence, as we won't be able to build a good heuristic without customer experience, as this sort of magic behavior is likely have negative effects as well. In applications that I've seen, deep stacks are very much in the minority, and applications could easily tolerate a 1000 arg stack allocation. (Without strong evidence, I believe we would declare the heuristic to be .... we always allocate on the stack.)

CyrusNajmabadi: @davidwrighton i'm ok with that approach.  But i do think we need to be careful here as we're talking about a silent source-change for existing apps.

RikkiGibson: It could be that the cases where people risk using up the stack are usually advanced scenarios already. Given that, it wouldn't necessarily be that difficult for people to identify the particularly large stack frames and replace `M(lots, of, args)` with `M(new[] { lots, of, args })` where needed.

meetings/working-groups/params-improvements/PI-2022-11-03.md
RikkiGibson: It sounds like this would just be a stack space optimization. I think the compiler should be permitted to perform such an optimization, but that we should punt out actually doing it until somebody needs it. I could imagine it getting difficult to do the optimization correctly in reentrant scenarios, i.e. where local functions have `params ROS` parameters and call into themselves or each other.

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+To avoid implicit stack allocation at a call site, the calling code should allocate the array explicitly with `new[] { ... }`.
+
+_What is the fallback for other target frameworks: heap allocation or compiler-generated types with contiguous fields: `struct FixedSizeArrayN { T T0; ... T TN; }`?_
+
+## Span reuse
+The compiler _may reuse a span_ when:
+- the span is implicitly allocated (as a `params` argument or as a _collection literal_),
+- the span length is known at compile time,
+- the compiler can determine _through escape analysis_ that no references to the span escape the containing method, and
+- the compiler can determine there are _no aliases_ to the span in user code.
```
jaredpar, line 47:
```suggestion
- the compiler can determine there are _no reachable aliases_ to the span in user code.
```

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+
+Reuse may be across distinct call sites _or_ repeated calls from the same call site.
+Reuse is per thread of execution and within the same method only.
+No reuse across lambda expressions, local functions, and the containing methods.
+
+The reused span will be cleared after each use.
+
+To opt out of reuse at a call site, the calling code should allocate the span explicitly.
+
+_Should we only reuse spans that are allocated on the stack? If we also allow reuse of heap allocated buffers, that will require completely different code gen for managing allocations._
```
jaredpar, line 64: I think the spec should definitely allow for re-use of the buffer no matter how it is allocated. Even if we don't do it now, it should be available to us in the future as an option.

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+
+    WriteLine("Count = {0}", dictionary.Count);
+}
+```
+
+The method could be lowered to:
+```csharp
+static void WriteDictionary<K, V>(Dictionary<K, V> dictionary)
+{
+    FixedSizeArray3 _tmp1 = new FixedSizeArray3();
```
jaredpar, line 107:
```suggestion
    FixedSizeArray2 _tmp1 = new FixedSizeArray2();
```

Only two arguments are used here so seems like this is the more appropriate type

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+The compiler will reuse a single span across all calls to `params ReadOnlySpan<T>` methods where the span element types are considered identical by the runtime.
+
+The length of the reused span will be the length of the longest `params` argument list from all uses.
+(The actual span argument passed to a `params` method will be a slice of the reused span, with the expected length for the call.)
+
+Reuse may be across distinct call sites _or_ repeated calls from the same call site.
+Reuse is per thread of execution and within the same method only.
+No reuse across lambda expressions, local functions, and the containing methods.
+
+The reused span will be cleared after each use.
```
jaredpar, line 60: This seems a bit strong. Generally our guarantees around code that generates implicit references that _root_ objects is that we ensure that we remove the implicit references when we logically exit the scope that created them. For example when exiting a scope in an `async` method we clear out fields that represented locals before the state machine rewrite.

The logic as written here forces us to clear when:

1. Exiting a method. No need to do this cause exiting the method implicitly clears. 
2. When it is a span of `unmanaged` types. Clearing here is just spinning CPU cycles.
3. When a span is immediately re-used. Consider consecutive calls to `WriteLine()`. No need to clear between the calls if the same slots are taken. 

I think we should consider mimicking our existing rules here. Essentially the guarantee is that when we exit a scope we ensure that no implicitly allocated span holds a references created in the scope. That gives us a lot more freedom for managing the span.

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+    public T[2] Array; // pseudo-code for inline array
+}
+```
+
+With that type, a call to `WriteLine("{0}, {1}", x, y)` could be emitted as:
+```csharp
+var _tmp1 = new FixedSizeArray2<object>();
+_tmp1.Array[0] = x;
+_tmp1.Array[1] = y;
+var _tmp2 = new ReadOnlySpan<object>(_tmp.Array);
```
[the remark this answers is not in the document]
cston: The public constructors are `ReadOnlySpan(T[])` and `ReadOnlySpan(T[], int, int)`.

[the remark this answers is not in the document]
jaredpar: You're right. Forgot that API was internal. Yeah we can use `MemoryMarshal` instead as it provides the behavior we need. 

Really though I think what APIs we use will fall out from the discussion on fixed sized arrays with the runtime team.

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+
+// WriteLine(string format, params ReadOnlySpan<object> args);
+WriteLine("{0}, {1}", _tmp2);
+```
+
+Ideally the BCL will provide types such as `FixedSizeArray1`, `FixedSizeArray2`, etc. for a limited number of array lengths.
+And if the compilation requires spans for other argument lengths, the compiler will generate and emit the additional types.
+
+The compiler will implicitly allocate `params` argument spans on the stack regardless of the number of arguments.
```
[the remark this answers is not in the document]
cston: Updated to state that the number of arguments to the `params` parameter is not considered.

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,126 @@
...
+To avoid implicit stack allocation at a call site, the calling code should allocate the array explicitly with `new[] { ... }`.
+
+_What is the fallback for other target frameworks: heap allocation or compiler-generated types with contiguous fields: `struct FixedSizeArrayN { T T0; ... T TN; }`?_
```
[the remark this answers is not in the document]
cston: Removed this question.

meetings/working-groups/params-improvements/PI-2022-11-03.md
cston: Thanks. I've updated the statement to: "When exiting a scope, the compiler will ensure that no implicitly allocated span holds references from the scope."

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,125 @@
+# Params Improvements Working Group Meeting for November 3, 2022
+The second meeting of the _`params` improvements_ working group was focused on the code generated for arguments to `params ReadOnlySpan<T>`. 
+
+## Span allocation
+The compiler _may allocate a span on the stack_ when:
+- the span is implicitly allocated (as a `params` argument or as a _collection literal_),
```
CyrusNajmabadi, line 6:
```suggestion
- the span is implicitly allocated (as a `params` argument or as a [_collection literal_](https://github.com/dotnet/csharplang/issues/5354)),
```

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,125 @@
...
+- the span length is known at compile time, and
+- the compiler can determine _through escape analysis_ that no references to the span escape the containing method.
+
+For arguments to a `params ReadOnlySpan<T>` parameter (which is implicitly `scoped`), the conditions above are satisfied.
+
+We have a request with the runtime team to support _fixed size arrays_ of managed types, at least for fields.
```
CyrusNajmabadi, line 12:
```suggestion
We have a [request](https://github.com/dotnet/runtime/issues/61135) with the runtime team to support _fixed size arrays_ of managed types, at least for fields.
```

meetings/working-groups/params-improvements/PI-2022-11-03.md
```diff
@@ -0,0 +1,125 @@
...
+
+// WriteLine(string format, params ReadOnlySpan<object> args);
+WriteLine("{0}, {1}", _tmp2);
+```
+
+Ideally the BCL will provide types such as `FixedSizeArray1`, `FixedSizeArray2`, etc. for a limited number of array lengths.
+And if the compilation requires spans for other argument lengths, the compiler will generate and emit the additional types.
+
+The number of arguments passed to a `params` parameter is _not considered_ when determining whether to implicitly allocate the span on the stack.
+To avoid implicit stack allocation at a call site, the calling code should allocate the array explicitly with `new[] { ... }`.
```
CyrusNajmabadi, line 39: I would def add a comment that we believe this is acceptable, but htat we will be observing the fallout of this to see if it causes problems for the ecosystem on source-update.

meetings/working-groups/params-improvements/PI-2022-11-03.md
CyrusNajmabadi: i agree with @RikkiGibson .  I think it's too strong to say "no reuse". I think it would be better to say that it is permitted.

PR #6674, opened 2022-11-09
Nullability improvements notes 2022-11-07

RikkiGibson:


svick:
Not sure this is the right place to discuss this, but I had a look at filtering operators that likely wouldn't work well with a `[Filter]`. The ones I found:

1. Partition ([MoreLinq](https://morelinq.github.io/3.3/ref/api/html/M_MoreLinq_MoreEnumerable_Partition__1.htm), [F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html#partition)\*): Splits a sequence into a 2-tuple of sequences: one for which a predicate is `true`, the other for which it's `false`.
2. Choose ([MoreLinq](https://morelinq.github.io/3.3/ref/api/html/M_MoreLinq_MoreEnumerable_Choose__2.htm)): Effectively a fused combination of `Select` and `Where`. E.g. `customers.Choose(c => (c.Coupon is not null && c.Email is not null, (c.Coupon, c.Email)))` could be used to extract specific non-null properties from an object.

Designing a more complicated version of `[Filter]` which could accommodate these operators is probably not worth it, but I thought it's worth considering anyway.

\* Obviously F# wouldn't directly benefit from this, I'm just using its standard library as indication of which operators might be common in real user code.

RikkiGibson:
Thanks for taking a look. We'll definitely consider the existence of these as we look at specific designs for a filtering attribute.

RikkiGibson:
There hasn't been further feedback or objection from the working group, so will merge these notes for clerical purposes.

PR #6688, opened 2022-11-10
Update ref-safe-to-escape of ref parameters to return only

cston:
Update a couple of a cases where the _ref-safe-to-escape_ of `ref` or `in` parameters was _calling method_ rather than _return only_.


jcouv (approved):
LGTM Thanks (iteration 1)

jaredpar (approved):

PR #6691, opened 2022-11-11
ref struct working group notes

jaredpar:


meetings/working-groups/ref-improvements/REF-2022-11-11.md
```diff
@@ -0,0 +1,134 @@
...
+    ...
+}
+
+'b Span<int> Create('b ref 'a RS i) {
+    where 'b <= 'a
+
+}
+```
+
+By establishing a relationship between `'a` and `'b` it allows for both the value and the `ref` to be return from the method. It also prevents cyclic assignment issues because the relationship does not allow for it from a lifetime (`'b` could be smaller than `'a`).
```
cston, line 45: >return [](http://example.com/codeflow?start=102&length=6)

"returned"

meetings/working-groups/ref-improvements/REF-2022-11-11.md
```diff
@@ -0,0 +1,134 @@
...
+> Every `ref` argument to a method be assignable *from* every other `ref` argument in terms of lifetimes. 
+
+Before there were just two layers of lifetimes, now there are basically infinite and they **all** have to match. The problem with `ref` fields to `ref struct` is now virtually everything is a MAMM nightmare. Even an instance method on a `struct` is dealing with MAMM insanity.
+
+The only way for MAMM to work in this scenario is for all the lifetimes involved to be equivalent. That really defeats the purpose of adding them. 
+
+# Paths Forward 
+
+## Implement RFRS
+The concensus is that to fully support RFRS would likely require us to have full lifetime notations in the langauge. For example putting the lifetimes on `scoped` modifier. That would allow for example `scoped<a>`, `scoped<b>`, etc ...
```
cston, line 115: >concensus [](http://example.com/codeflow?start=4&length=9)

"consensus"

meetings/working-groups/ref-improvements/REF-2022-11-11.md
cston, line 45:
```suggestion
By establishing a relationship between `'a` and `'b` it allows for both the value and the `ref` to be returned from the method. It also prevents cyclic assignment issues because the relationship does not allow for it from a lifetime (`'b` could be smaller than `'a`).
```

meetings/working-groups/ref-improvements/REF-2022-11-11.md
cston, line 115:
```suggestion
The consensus is that to fully support RFRS would likely require us to have full lifetime notations in the language. For example putting the lifetimes on `scoped` modifier. That would allow for example `scoped<a>`, `scoped<b>`, etc ... 
```

meetings/working-groups/ref-improvements/REF-2022-11-11.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Implement RFRS in a limited fashion
+Much of the complexity of RFRS comes from the fields being as flexible as a `ref` parameter. The ability to ref re-assign, assign new values, etc ... drive the complexity particularly around MAMM. Perhaps if we limited their capabilities we could satisfy key scenarios without jumping off the complexity cliff. For example if we forced them to be `readonly ref`, `ref scoped`, etc ... would that help? 
+
+The `ref scoped` is likely the most plausible approach here. The key is whether we can define its meaning in such a way that it remains useful while keeping complexity low. The rules would also need to apply to `ref scoped` parameters too which increase the scenarios a bit.
+
+## Punt it 
+The explicit decision to not support RFRS for now should open the door for us exploring `ref struct` as generic arguments and implementing interfaces. There are existing scenarios that will sufficiently help drive this feature. 
+
+Need to do some more dilegence to be certain that we don't close the door on RFRS in the future.
```
cston, line 127: Perhaps:
```suggestion
Need more due diligence to be certain that we don't close the door on RFRS in the future. 
```

meetings/working-groups/ref-improvements/REF-2022-11-11.md
```diff
@@ -0,0 +1,134 @@
...
+
+```csharp
+ref struct RS { }
+Span<int> Create(ref RS i) {
+    ...
+}
+
+'b Span<int> Create('b ref 'a RS i) {
+    where 'b <= 'a
+
```
cston, line 41: Minor: Consider switching `'a` and `'b` for consistency with the previous example, and move the `where` clause before the opening `{`.
```
'a Span<int> Create('a ref 'b RS i)
    where 'a <= 'b
{
}
```

meetings/working-groups/ref-improvements/REF-2022-11-11.md
```diff
@@ -0,0 +1,134 @@
...
+# Complications around ref fields of ref struct
+The immediate issue this causes is it creates layers of lifetimes in our types. Consider concretely: 
+
+```csharp
+ref struct RS { 
+  ref 'a Span<byte> Field; 
+}
+```
+
+The value of the field has a lifetime that is independent of the containing instance. Typically when we look at a value there are two lifetimes to consider: the safe to escape and ref safe to escape. In the case of `RS` though there is also the lifetime of `Field`.
```
cston, line 56: >is independent of the containing instance [](http://example.com/codeflow?start=43&length=41)

Is it worth stating that the lifetime of the field (the `ref` and any `ref` within the value) cannot be shorter than the lifetime of the containing instance though?

meetings/working-groups/ref-improvements/REF-2022-11-11.md
```diff
@@ -0,0 +1,134 @@
...
+The `ref scoped` is likely the most plausible approach here. The key is whether we can define its meaning in such a way that it remains useful while keeping complexity low. The rules would also need to apply to `ref scoped` parameters too which increase the scenarios a bit.
+
+## Punt it 
+The explicit decision to not support RFRS for now should open the door for us exploring `ref struct` as generic arguments and implementing interfaces. There are existing scenarios that will sufficiently help drive this feature. 
+
+Need to do some more dilegence to be certain that we don't close the door on RFRS in the future. 
+
+# Next Steps
+
+1. Andy and Aaron are going to provide simplified samples where RFRS would be beneficial.
```
cston, line 131: >Andy and Aaron are going to provide [](http://example.com/codeflow?start=3&length=35)

Perhaps just "Provide simplified samples ...".

meetings/working-groups/ref-improvements/REF-2022-11-11.md
jjonescz: Shouldn't it be `lifetime` keyword instead of `where` as defined a few paragraphs above?

PR #6738, opened 2022-11-22
Provide spec for V8 proposal constraints-in-overrides 

RexJaeschke:


proposals/csharp-8.0/constraints-in-overrides.md
```diff
@@ -1,3 +1,81 @@
 ﻿## Override with constraints
 
 In C# 8.0, we added a feature to permit the specification of certain type parameter constraints in an `override` method declaration. This is a placeholder for its specification.
+
+
+Rex Jaeschke's contribution to this spec, using section numbering from the draft-v8 Ecma standard
```
333fred, line 6: The spec files are shown on docs.microsoft.com, so we should leave these descriptions out.

PR #6747, opened 2022-11-26
Update params-span.md based on working group

cston:


proposals/params-span.md
```diff
@@ -21,125 +22,156 @@ public static class Console
...
-Otherwise the array will be allocated on the heap.
+### Array allocation and reuse
+The compiler will include the following optimizations for implicitly allocated `params` arrays. Additional optimizations may be added in future for cases where the compiler can determine there are _no reachable aliases_ to the array.
 
-```csharp
-Span<int> s = new[] { i, j, k };   // int[] on the stack
-WriteLine(fmt, new[] { x, y, z }); // object[] on the stack for WriteLine(string fmt, ReadOnlySpan<object> args);
-```
+The compiler _will allocate the array on the stack_ for a `params ReadOnlySpan<T>` argument when
+- the parameter is implicitly or explicitly `scoped`,
```
RikkiGibson, line 76: Could you remind me why we won't allocate on the stack when the parameter is marked unscoped? Is it because we think we can't reuse the array across calls, etc. in that case?

proposals/params-span.md
```diff
@@ -21,125 +22,156 @@ public static class Console
...
 
-Direct runtime support for stack allocation of arrays of managed types might be useful for lowering implicit allocation as well.
+### Support `params Span<T>`?
+Support `params Span<T>` to allow the `params` method to modify the span contents, even though the effects are only observable at call sites that explicitly allocate the argument array?
 
-The GC does not currently track the lifetime of a `stackalloc` array so if the contents of the array have a shorter lifetime than the method, the compiler will need to zero the contents of the array so the lifetime of elements matches expectations.
+### Support `params IEnumerable<T>`, etc.?
+If we're extending `params` to support `ReadOnlySpan<T>`, should we also support `params` parameters of other collection types, including interfaces and concrete types? The reason to support `params ReadOnlySpan<T>` is to improve the performance of existing callers. And other collection types are already well supported by having non-`params` overloads for the other types in addition to a `params T[]` overload. That said, this proposal doesn't prevent extending `params` to other types in the future.
```
RikkiGibson, line 174: We should mention that we think collection literals reduces the motivation here. We basically want params ROS in order to make existing call sites behave better in more cases. But in general and for new code we think it's good for people to just wrap the params elements in `[]`.

proposals/params-span.md
cston: Yes, the parameter must be `scoped` to allow the array to be reused. I've added that here, thanks.

proposals/params-span.md
```diff
@@ -21,125 +22,162 @@ public static class Console
...
 
 ### Extending `params`
-`params` parameters will be supported with types `Span<T>`, `ReadOnlySpan<T>`, and `IEnumerable<T>`.
-
-A call in _expanded form_ ([§11.6.4.2](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#11642-applicable-function-member)) to a method with a `params T[]` or `params IEnumerable<T>` parameter will result in an array `T[]` allocated on the heap.
+A `params` parameter type may be `System.ReadOnlySpan<T>` for a valid type argument `T`.
 
-A call in _expanded form_ ([§11.6.4.2](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#11642-applicable-function-member)) to a method with a `params ReadOnlySpan<T>` or `params Span<T>` parameter will result in an array `T[]` created on the stack _if the `params` array is within limits (if any) set by the compiler_.
-Otherwise the array will be allocated on the heap.
+A call in [_expanded form_](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#11642-applicable-function-member) to a method with a `params ReadOnlySpan<T>` parameter will result in a `ReadOnlySpan<T>` instance implicit created by the compiler.
```
jjonescz, line 38:
```suggestion
A call in [_expanded form_](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#11642-applicable-function-member) to a method with a `params ReadOnlySpan<T>` parameter will result in a `ReadOnlySpan<T>` instance implicitly created by the compiler.
```

proposals/params-span.md
```diff
@@ -21,125 +22,162 @@ public static class Console
...
-Are scenarios for `params IEnumerable<T>` sufficiently compelling to justify that?
+To _opt out_ of compiler optimizations at a call site, the calling code should allocate the span explicitly.
 
-### Array limits
-The compiler may use heuristics to determine when to fallback to heap allocation for the underlying data for spans.
-If heuristics are necessary, experimentation should establish the limits we agree on.
-
-### Lowering approach
-We need to determine the particular approach used to lower `params` and array creation expressions to avoid heap allocation.
+The code emitted for an implicitly allocated `params` span should be identical to the code emitted when using a [_collection literal_](https://github.com/dotnet/csharplang/issues/5354)) for the `params` argument.
```
jjonescz, line 97:
```suggestion
The code emitted for an implicitly allocated `params` span should be identical to the code emitted when using a [_collection literal_](https://github.com/dotnet/csharplang/issues/5354) for the `params` argument.
```

jjonescz (approved):

RikkiGibson (approved):

PR #6756, opened 2022-11-29
Nullability improvements notes 2022-11-22

RikkiGibson:


meetings/working-groups/nullability-improvements/NI-2022-11-22.md
```diff
@@ -0,0 +1,194 @@
...
+        [NullableFilter] Func<TSource, bool> predicate)
+    {
+        // ...
+    }
+}
+```
+
+When a `[NullableFilter]` parameter is used, it is associated with the parameter type of the delegate. In this case, that parameter is `predicate`, the delegate parameter's type is `TSource`, and the containing method is `Where`. All values of type `TSource` which *flow out* of `Where` are assumed to have been passed in to a call to `predicate`, and a `true` value was returned from each call.
+
+One of the first bits of feedback was on the name. We think this attribute is not itself about nullability. It's just that if lambdas passes as arguments for the predicate parameter do null tests, then null state information about the collection elements can propagate through. Other forms of static analysis could easily take advantage of the same assumption, that the predicate returned true for all elements returned from the containing method. Therefore, we think perhaps the attribute name `[Filter]` or `[Predicate]` would be more suitable.
```
jjonescz, line 38:
```suggestion
One of the first bits of feedback was on the name. We think this attribute is not itself about nullability. It's just that if lambdas passed as arguments for the predicate parameter do null tests, then null state information about the collection elements can propagate through. Other forms of static analysis could easily take advantage of the same assumption, that the predicate returned true for all elements returned from the containing method. Therefore, we think perhaps the attribute name `[Filter]` or `[Predicate]` would be more suitable.
```

jjonescz (approved):

PR #6772, opened 2022-12-02
Adjust grammar specification for Primary Constructors.

AlekseyTs:


333fred (approved):

proposals/primary-constructors.md
```diff
@@ -64,7 +64,7 @@ class_body
   
 struct_declaration
   : attributes? struct_modifier* 'partial'? 'record'? 'struct' identifier type_parameter_list?
-    parameter_list? struct_interfaces? type_parameter_constraints_clause* struct_body ';'?
+    parameter_list? struct_interfaces? type_parameter_constraints_clause* struct_body
```
CyrusNajmabadi, line 67: i thought we allowed trailing semicolon on thsee types in the language.  or is this only for error recovery?

CyrusNajmabadi (approved):

proposals/primary-constructors.md
MadsTorgersen: We do, but it's already built into the `class_body` and `struct_body` productions, so also having it here weirdly allows up to two trailing semicolons!

MadsTorgersen (approved):
This change seems right. The `class_body` and `struct_body` productions already allow trailing semicolon, so `class_declaration` and `struct_declaration` shouldn't also do so.

CyrusNajmabadi:
ah!  that makes sense.  Thanks! :)

PR #6785, opened 2022-12-08
limit disposable pattern

BillWagner:
The proposal spec'ed this out in general terms for any type, and included a provision to allow extension `Dispose` methods.

The implementation only allows `ref` struct types to use the pattern-based Dispose. Furthermore, the `Dispose` method can't be an extension method.

Fixes dotnet/docs#31815

I made edits to make the proposal match the implementation. As an alternative, I could make edits to indicate which portions of the proposal were implemented and which may be delivered in the future. Let me know which you'd prefer.

### DIFF
--- a/proposals/csharp-8.0/using.md
+++ b/proposals/csharp-8.0/using.md
@@ -102,12 +102,12 @@ Restrictions around `using` declaration:
 
 ### pattern-based using
 
-The language will add the notion of a disposable pattern: that is a type which has an accessible 
+The language will add the notion of a disposable pattern for `ref struct` types: that is a `ref struct` which has an accessible 
 `Dispose` instance method. Types which fit the disposable pattern can participate in a `using` 
 statement or declaration without being required to implement `IDisposable`. 
 
 ```csharp
-class Resource
+ref struct Resource
 { 
     public void Dispose() { ... }
 }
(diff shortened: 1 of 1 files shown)


proposals/csharp-8.0/using.md
```diff
@@ -118,16 +118,8 @@ using (var r = new Resource())
 }
 ```
 
-This will allow developers to leverage `using` in a number of new scenarios:
-
-- `ref struct`: These types can't implement interfaces today and hence can't participate in `using`
+This will allow developers to leverage `using` for `ref struct` types. These types can't implement interfaces today and hence can't participate in `using`
 statements.
-- Extension methods will allow developers to augment types in other assemblies to participate
```
jcouv, line 125: I confirmed from the code (`TryFindDisposePatternMethod`) that extension methods don't count. Tagging @chsienki to confirm this was intentional and to fix the speclet.

jcouv (approved):

proposals/csharp-8.0/using.md
chsienki: Yes, extension methods we're explicitly disallowed, even for ref structs. The decision being that it would have been legal to have an extension method called 'Dispose' that previously wouldn't have been called if you used the type in a foreach. With the extension method lookup, it would now get called, which was considered too big a breaking change. 

LDM notes are here: https://github.com/dotnet/csharplang/blob/3c8559f186d4c5df5d1299b0eaa4e139ae130ab6/meetings/2018/LDM-2018-12-12.md#when-we-start-recognizing-pattern-dispose-do-we-call-it-in-foreach

chsienki (approved):

PR #6855, opened 2023-01-04
Clarify some aspects of primary constructor parameters capturing

AlekseyTs:


AlekseyTs:
@cston, @jjonescz FYI

proposals/primary-constructors.md
```diff
@@ -128,7 +128,7 @@ Primary constructor parameters in class/struct declarations can be declared `ref
 
 All instance member initializers in the class body will become assignments in the generated constructor.
 
-If a primary constructor parameter is referenced from within an instance member, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. 
+If a primary constructor parameter is referenced from within an instance member, and the refernce is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name.
```
333fred, line 131:
```suggestion
If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. 
```

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+
+With respect to primary constructors, the rule has effect on whether an identifier within an instance member should be treated as a type reference, or as a primary constructor parameter reference, which, in turn, captures the parameter into the the state of the enclosing type. Even though "the member lookup of `E.I` is never ambiguous", when lookup yields a member group, in some cases it is impossible to make determination whether a member access refers to a static, or to an instance member without fully resolving (binding) the member access. At the same time, the fact of capturing a primary constructor parameter, changes properties of enclosing type in a way that affects semantic analysis. For example, the type might become unmanaged and fail certain constraints because of that.
```
333fred, line 196:
```suggestion
With respect to primary constructors, the rule affects whether an identifier within an instance member should be treated as a type reference, or as a primary constructor parameter reference, which, in turn, captures the parameter into the the state of the enclosing type. Even though "the member lookup of `E.I` is never ambiguous", when lookup yields a member group, in some cases it is impossible to make determination whether a member access refers to a static, or to an instance member without fully resolving (binding) the member access. At the same time, the fact of capturing a primary constructor parameter, changes properties of enclosing type in a way that affects semantic analysis. For example, the type might become unmanaged and fail certain constraints because of that. 
```

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+
+With respect to primary constructors, the rule has effect on whether an identifier within an instance member should be treated as a type reference, or as a primary constructor parameter reference, which, in turn, captures the parameter into the the state of the enclosing type. Even though "the member lookup of `E.I` is never ambiguous", when lookup yields a member group, in some cases it is impossible to make determination whether a member access refers to a static, or to an instance member without fully resolving (binding) the member access. At the same time, the fact of capturing a primary constructor parameter, changes properties of enclosing type in a way that affects semantic analysis. For example, the type might become unmanaged and fail certain constraints because of that. 
+There are even scenarios for which binding can succeed either way, depending on whether the parameter is considered captured, or not. For example:
```
333fred, line 197:
```suggestion
There are even scenarios for which binding can succeed either way, depending on whether the parameter is considered captured or not. For example:
```

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+    }
+    
+    public static void M1<T>(T x) where T : unmanaged
+    {
+        System.Console.WriteLine(""static"");
+    }
+}
+```
+
+If we treat receiver ```Color``` as a value, we capture the parameter and 'S1' becomes managed. Then static method becomes inapplicable due to constraint and we would call instance method. However, if we treat the receiver as a type, we don't capture the parameter and 'S1' remains unmanaged, then both methods are applicable, but the static method is "better" because it doesn't have an optional parameter. Neither choice leads to an error, but each would result in distinct behavior.
```
333fred, line 221:
```suggestion
If we treat receiver ```Color``` as a value, we capture the parameter and 'S1' becomes managed. Then the static method becomes inapplicable due to the constraint and we would call instance method. However, if we treat the receiver as a type, we don't capture the parameter and 'S1' remains unmanaged, then both methods are applicable, but the static method is "better" because it doesn't have an optional parameter. Neither choice leads to an error, but each would result in distinct behavior.
```

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+
+Given this, compiler is going to produce an ambiguity error for a member access `E.I` when all the following conditions are met:
```
333fred, line 223:
```suggestion
Given this, compiler will produce an ambiguity error for a member access `E.I` when all the following conditions are met:
```

333fred (commented):
Done review pass (commit 2)

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+    {
+        Color.M1(this);
+    }
+}
+
+class Color
+{
+    public void M1<T>(T x, int y = 0)
+    {
+        System.Console.WriteLine(""instance"");
```
jjonescz, line 211:
```suggestion
        System.Console.WriteLine("instance");
```

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+    }
+    
+    public static void M1<T>(T x) where T : unmanaged
+    {
+        System.Console.WriteLine(""static"");
```
jjonescz, line 216:
```suggestion
        System.Console.WriteLine("static");
```

proposals/primary-constructors.md
```diff
@@ -187,6 +187,43 @@ Records produce a warning if a primary constructor parameter isn't read within t
...
+    }
+}
+```
+
+If we treat receiver ```Color``` as a value, we capture the parameter and 'S1' becomes managed. Then static method becomes inapplicable due to constraint and we would call instance method. However, if we treat the receiver as a type, we don't capture the parameter and 'S1' remains unmanaged, then both methods are applicable, but the static method is "better" because it doesn't have an optional parameter. Neither choice leads to an error, but each would result in distinct behavior.
+
+Given this, compiler is going to produce an ambiguity error for a member access `E.I` when all the following conditions are met:
+- Member lookup of `E.I` yields a member group containing instance and static members at the same time. Extension methods applicable to the receiver type are treated as instance methods for the purpose of this check.
+- If `E` is treated as a simple name, rather than a type name, it would refer to a primary constructor parameter and would capture the parameter into the state of the enclosing type.
```
jjonescz, line 225: What if the primary constructor parameter `E` is already definitely captured elsewhere? Will that make the member access `E.I` unambiguous?

AlekseyTs: > What if the primary constructor parameter E is already definitely captured elsewhere? Will that make the member access E.I unambiguous?

No. Each occurrence is examined in isolation.  

> Or how else can one resolve the ambiguity error?

The type can be qualified, the parameter can be taken into parenthesis, or renamed.

cston (approved):

PR #6877, opened 2023-01-12
Move and add more details for "Double storage warning" section

AlekseyTs:


cston (approved):

proposals/primary-constructors.md
```diff
@@ -337,10 +346,6 @@ public class C(bool b, int i, string s) : B(b)
 }
 ```
 
-### Double storage warning
-
-If a primary constructor parameter is passed to the base and *also* captured, there's a high risk that it is inadvertently stored twice in the object. It might make sense to issue a warning about this, if we can settle on good conditions for the warning, as well as recommended ways to shut it up if the code was intentional.
```
jjonescz, line 342: > recommended ways to shut it up if the code was intentional

So are there any recommended ways to suppress the warning except `#pragma`?

jjonescz (approved):

proposals/primary-constructors.md
AlekseyTs: > So are there any recommended ways to suppress the warning except `#pragma`?

Don't do this, i.e. change the code so that conditions are not met.

PR #6880, opened 2023-01-13
Add roles spec covering static extensions

jcouv:
In short, for phase A:
1.	We'd modify simple name and member access rules, so that when a member lookup is performed we fall back to either doing an “extension member lookup” (for types that are not extensions) or a member lookup on underlying type (for types that are extensions)
2.	We'd add an “extension member lookup”, that finds an extension member `X.I` given an underlying type `U` and an identifier `I`. This gets tricky with generics and definitely needs refining.

For additional context, here are the [extension method invocation rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11783-extension-method-invocations) and [member lookup rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#115-member-lookup).


proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+C. roles and extensions that implement interfaces
+
+The syntax for a roles is as follows:
+
+```antr
+type_declaration
+	: role_declaration // add
+	| extension_declaration // add
+	| ...
+	;
```
CyrusNajmabadi, line 15: dedent

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+
+role_declaration
+    : modifier* 'role' identifier type_parameter_list? role_underlying_type type_parameter_constraints_clause* role_body
+    ;
+
+extension_declaration
+    : modifier* 'extension' identifier type_parameter_list? role_underlying_type type_parameter_constraints_clause* role_body
+    ;
```
CyrusNajmabadi, line 23: @MadsTorgersen Continued feedback that i think that role/extension are not good terms.

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+
+role_underlying_type
+    : ':' type
+    ;
```
CyrusNajmabadi, line 27: note for spec.  We likely will want to have similar thoughts on if you need to use `unsafe` to write `: int*`.  And whether or not you can do `: string?` (probably not?).

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+
+role_member_declaration
+    : constant_declaration
+    | field_declaration
+    | method_declaration
+    | property_declaration
+    | event_declaration
+    | indexer_declaration
+    | operator_declaration
+    | type_declaration
```
CyrusNajmabadi, line 41: allowing type-decls in roles is def interesting (but seems like it could be useful).  What is teh fully qualified name of that type-decl though?  

are roles also allowed to be defined in roles?

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+In this first subset of the feature, the syntax is restricted to non-nested **extension_declaration**
+and containing only **constant_declaration**, **field_declaration**, **method_declaration** and **property_declaration** members.
+TODO: events?
+
+### General
+
+The permitted modifiers are `partial`, `static`, `unsafe` and the accessibility modifiers `public` and `internal`.  
+An extension declaration must have a `static` modifier. If the declaration is partial, then at least one part must have the `static` modifier. 
+The standard rules for modifiers apply (valid combination of access modifiers, no duplicates).
+The **role_underlying_type** type may not be `dynamic`, a pointer or an extension type.
```
CyrusNajmabadi, line 56: interesting.  curious why not pointer (unless that's just an initial limitation that would be relaxed later).

proposals/roles.md
jcouv: Current proposal disallows pointer type as underlying type. This is following a restriction that we have on extension methods.
Good question on nullability.

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+The standard rules for modifiers apply (valid combination of access modifiers, no duplicates).
+The **role_underlying_type** type may not be `dynamic`, a pointer or an extension type.  
+This declares a static class type. It inherits from type `object`. It has `internal` declared accessibility by default.
+TODO: attributes?
+TODO: emitted name?
+
+### Members
+
+The extension type members may not use the `new` modifier.
+Accessibility modifiers including `protected` are disallowed.
```
CyrusNajmabadi, line 64: oh, interesting.  seems a pity as i could imagine having internal members that are useful for when looking through the role in its own assembly, versus public members when seen externally.

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+The extension type does not inherit members from its underlying type.
```
CyrusNajmabadi, line 65: def surprising.  but i guess would be good to see how that works out in practice.

jcouv: We want to allow value types and sealed types as the "underlying type", so we can't rely on inheritance relationship. Instead, we rely on lookup rules to make things work.

CyrusNajmabadi: ah, so maybe it's a question fo terminology?
the members are not 'inherited' per-se, but users would still be able to use those members at the use site?  Could def maybe be clarified if that's so.

proposals/roles.md
```diff
@@ -0,0 +1,155 @@
...
+  - Otherwise, `E.I` is an invalid member reference, and a compile-time error occurs.
+- **Otherwise, if `E` is classified as a type, if `E` is not a type parameter or an extension type, and if an ***extension member lookup*** of `I` in `E` with `K` type parameters produces a match, then `E.I` is evaluated and classified as follows:**  
+  ...
+- **Otherwise, if `E` is classified as an extension type, and if a member lookup of `I` in underlying type `U` with `K` type parameters produces a match, then `E.I` is evaluated and classified as follows:** 
+  ...
+- If `E` is a property access, indexer access, variable, or value, the type of which is `T`, and a member lookup of `I` in `T` with `K` type arguments produces a match, then `E.I` is evaluated and classified as follows:
+  ...
+- Otherwise, an attempt is made to process `E.I` as an extension method invocation. If this fails, `E.I` is an invalid member reference, and a binding-time error occurs.
+
+### Extension member lookup
```
jcouv, line 137: :memo: Maybe the "extension member lookup" should be factored into the normal "member lookup".
So other scenarios in the spec that rely on "member lookup", such as resolution of `Add` in collection initializers ([here](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#117154-collection-initializers)) would benefit.

A "member lookup" would become a "simple member lookup" (current rules for member lookup) followed by either an "extension member lookup" or a simple member lookup in the underlying type.

jcouv: :memo: The process of finding a "compatible substituted extension type" is currently proposed in a similar way that we deal with interface unification (detecting potential duplicate/ambiguous interface implementations). We could simplify this problem by placing restrictions on extension types. 
For example, we could:
- require extension types to be non-generic. 
- require the extension's type parameters to match the type parameters of the underlying type.

An example of issue:
`static extension Evil<T, U> : Dictionary<IDisposable, T> where U : IDisposable`
There is no way to go from an instance of `Dictionary'2` to a substituted `Evil'2`.

Another example of complexity:
```
extension E<T, U> : Dictionary<List<T>, U> // T used in nested position
Dictionary<List<int>, string>
```

proposals/roles.md
jcouv: > What is teh fully qualified name of that [nested] type-decl though?

Just like any other nested type-decl. You can refer to it with `Extension.NestedType`.

> are roles also allowed to be defined in roles?

See below, "In this first subset of the feature, the syntax is restricted to non-nested extension_declaration"

CyrusNajmabadi: > You can refer to it with Extension.NestedType.

Nifty.  Is that in metadata as well?

jcouv: TBD. There's an open question of what name we use to emit the extension type.

proposals/roles.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Phase A: Adding static constants, fields, methods and properties
+
+In this first subset of the feature, the syntax is restricted to non-nested **extension_declaration**
+and containing only **constant_declaration**, **field_declaration**, **method_declaration** and **property_declaration** members.
+TODO: events?
+
+### General
+
+The permitted modifiers are `partial`, `unsafe` and the accessibility modifiers `public` and `internal`.
```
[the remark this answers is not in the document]
jcouv: This section is about the extension type, not the members. Will clarify the section headers.

proposals/roles.md
```diff
@@ -0,0 +1,165 @@
...
+The extension type does not inherit members from its underlying type.  
+
+#### Constants
+
+Existing [rules for constants](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#144-constants) apply (so duplicates or the `static` modifier are disallowed).
+
+#### Fields
+
+A **field_declaration** in an **extension_declaration** shall explicitly include a `static` modifier.  
+Otherwise, existing [rules for fields](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#145-fields) apply.
```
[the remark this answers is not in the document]
jcouv: Correct. That's the line above: "A **field_declaration** in an **extension_declaration** shall explicitly include a `static` modifier."

proposals/roles.md
```diff
@@ -0,0 +1,165 @@
...
+Existing [rules for constants](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#144-constants) apply (so duplicates or the `static` modifier are disallowed).
+
+#### Fields
+
+A **field_declaration** in an **extension_declaration** shall explicitly include a `static` modifier.  
+Otherwise, existing [rules for fields](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#145-fields) apply.  
+
+#### Methods
+
+A **method_declaration** in an **extension_declaration** shall explicitly include a `static` modifier.
```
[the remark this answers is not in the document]
jcouv: That is correct and that's allowed by the proposal so far. The whole point of Phase A is to cover this scenario (extension method to add a static method to an existing type, including a static type like `System.Math`).

proposals/roles.md
```diff
@@ -0,0 +1,165 @@
...
+#### Methods
+
+A **method_declaration** in an **extension_declaration** shall explicitly include a `static` modifier.  
+Parameters with the `this` modifier are disallowed.
+Otherwise, existing [rules for methods](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#146-methods) apply.
+In particular, a static method does not operate on a specific instance, and it is a compile-time error to refer to this in a static method.
+
+#### Properties
+
+A **property_declaration** in an **extension_declaration** shall explicitly include a `static` modifier.
```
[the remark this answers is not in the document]
jcouv: I may be missing something. Let me know if my reply above didn't resolve the question.

proposals/roles.md
```diff
@@ -0,0 +1,165 @@
...
+and containing only **constant_declaration**, **field_declaration**, **method_declaration** and **property_declaration** members.
+TODO: events?
+
+### General
+
+The permitted modifiers are `partial`, `unsafe` and the accessibility modifiers `public` and `internal`.  
+Note that `static` is disallowed.  
+The standard rules for modifiers apply (valid combination of access modifiers, no duplicates).  
+The **role_underlying_type** type may not be `dynamic`, a pointer or an extension type.  
+This declares a static class type. It inherits from type `object`. It has `internal` declared accessibility by default.
```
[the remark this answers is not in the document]
jcouv: I agree. We'll likely change this to `ref struct`. Need to confirm with Mads (he proposed class for the static extension scenario).

proposals/roles.md
```diff
@@ -0,0 +1,165 @@
...
+The standard rules for modifiers apply (valid combination of access modifiers, no duplicates).  
+The **role_underlying_type** type may not be `dynamic`, a pointer or an extension type.  
+This declares a static class type. It inherits from type `object`. It has `internal` declared accessibility by default.  
+TODO: attributes?
+TODO: emitted name?
+
+### Members
+
+The extension type members may not use the `new` modifier.
+Accessibility modifiers including `protected` are disallowed.
```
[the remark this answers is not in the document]
jcouv: In phase A as proposed, there's only extensions, no roles. But this will need more discussion.

proposals/roles.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Phase A: Adding static constants, fields, methods and properties
+
+In this first subset of the feature, the syntax is restricted to non-nested **extension_declaration**
+and containing only **constant_declaration**, **field_declaration**, **method_declaration** and **property_declaration** members.
+TODO: events?
+
+### Extension type
+
+The permitted modifiers on an extension type are `partial`, `unsafe` and the accessibility modifiers `public` and `internal`.
```
KathleenDollard, line 54: Does it have to match or be no more broad than the accessibility of the underlying role type?

KathleenDollard (commented):
I look forward to seeing some examples code about both declaring and using these. I think I see it, but it will be great to confirm when that is ready.

proposals/roles.md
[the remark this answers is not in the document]
333fred: I think my comments are being resolved to the wrong places. These should have been on the static keyword requirements for methods and properties.

proposals/roles.md
jcouv: Added. Thanks

proposals/roles.md
```diff
@@ -143,21 +143,20 @@ We modify the [member access rules](https://github.com/dotnet/csharpstandard/blo
...
   - If `I` identifies an enumeration member, then the result is a value, namely the value of that enumeration member.
   - Otherwise, `E.I` is an invalid member reference, and a compile-time error occurs.
-- **If `E` is classified as a type, if `E` is not a type parameter or an extension type, if the member is not invoked, and if an ***extension member lookup*** of `I` in `E` with `K` type parameters produces a match, then `E.I` is evaluated and classified as follows:**  
+- **If `E` is classified as a type, if `E` is not a type parameter or an extension type, and if an ***extension member lookup*** of `I` in `E` with `K` type parameters produces a match, then `E.I` is evaluated and classified as follows:**  
   ...
-- **If `E` is classified as a role (only relevant in phase B) or extension type, and if a member lookup of `I` in underlying type `U` with `K` type parameters produces a match, then `E.I` is evaluated and classified as follows:** 
+- **If `E` is classified as a role (only relevant in phase B) or extension type, and if a member lookup of `I` in underlying type `U` with `K` type parameters produces a match, then `E.I` is evaluated and classified as follows:**
```
333fred, line 148: Some trailing newlines

proposals/roles.md
```diff
@@ -0,0 +1,244 @@
...
+### Extension type
+
+An extension type is declared by a non-nested *extension_declaration*.  
+The permitted modifiers on an extension type are `partial`, `unsafe` and the accessibility modifiers `public` and `internal`.  
+Note that `static` is disallowed.  
+The standard rules for modifiers apply (valid combination of access modifiers, no duplicates).  
+The *role_underlying_type* type may not be `dynamic`, a pointer, a nullable reference, a ref struct type, or an extension type.  
+The *role_underlying_type* type must include all the type parameters from the extension type.  
+The *role_underlying_type* may not include an *interface_type_list* (this is part of Phase C).  
+An extension declaration must include an underlying type, unless it is partial. It is a compile-time error if no part of a partial extension type includes an underlying type, or the underlying type from differs amongst all the parts.
```
AlekseyTs, line 61: >type from differs [](http://example.com/codeflow?start=193&length=17)

"type from differs". Typo?

proposals/roles.md
```diff
@@ -52,14 +58,15 @@ TODO: events?
 
 ### Extension type
 
-An extension type is declared by a non-nested *extension_declaration*.  
-The permitted modifiers on an extension type are `partial`, `unsafe` and 
-the accessibility modifiers `public` and `internal`.  
-Note that `static` is disallowed.  
+An extension type is a role type (new kind of type) declared by an *extension_declaration*.  
+The permitted modifiers on an extension type are `partial`, `unsafe`, `file` and 
+the accessibility modifiers `public`, `internal` and `private`.
```
FaustVX, line 63: I thought `file` is an accessibility modifier, like `public`, `internal`, `protected` or `private`

jcouv: `file` affects scope/visibility, but not accessibility. It's a subtle distinction.
The [spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/file-local-types.md) says it's a modifier, not an accessibility modifier.

FaustVX: OK, but what's the difference ?
If something is not visible, it's not accessible and if it's not accessible, it's not visible neither.
As `file` is mutually exclusive to accessibility modifier, it seems to be an accessibility modifier as well.

333fred: > As file is mutually exclusive to accessibility modifier, it seems to be an accessibility modifier as well.

It is only mutually exclusive because we wanted to leave the space open for things like `file private` nested classes, or `file public` implementations of interface members that can be referenced by name in the current file and implement interface members, but not be visible outside the file.

FaustVX: Ok, thanks

proposals/roles.md
```diff
@@ -0,0 +1,449 @@
...
+
+An extension type is a role type (new kind of type) declared by an *extension_declaration*.  
+The permitted modifiers on an extension type are `partial`, `unsafe`, `file` and 
+the accessibility modifiers.  
+Note that `static` is disallowed. The extension type is static if its underlying type
+is static.  
+The standard rules for modifiers apply (valid combination of access modifiers, no duplicates).  
+
+The *role_underlying_type* type may not be `dynamic`, a pointer, a nullable reference (no top-level nullability), 
+a ref struct type.
```
jaredpar, line 106:
```suggestion
or a ref struct type.  
```

proposals/roles.md
```diff
@@ -0,0 +1,630 @@
+﻿# Roles
+
+This proposal is divided into three parts, all relating to extending existing types:  
+A. static extensions  
+B. roles and extensions with members  
+C. roles and extensions that implement interfaces  
+
+The syntax for a roles is as follows:
```
jjonescz, line 8:
```suggestion
The syntax for roles is as follows:
```

proposals/roles.md
```diff
@@ -0,0 +1,665 @@
...
+extension_declaration
+    : modifier* 'extension' identifier type_parameter_list? role_underlying_type? type_parameter_constraints_clause* role_body
+    ;
+
+role_underlying_type
+    : ':' type
+    : ':' type ',' role_or_interface_type_list
+    ;
+
+role_or_interface_type_list
```
KathleenDollard, line 30: I am having some trouble with this syntax. Will interfaces and roles be mixed or will roles follow interfaces?

I am uncomfortable with them being mixed.

proposals/roles.md
```diff
@@ -0,0 +1,665 @@
...
+```
+Role r = default(UnderlyingType); // emitted as a local of type `UnderlyingType`
+r.RoleMember(); // emitted as `new Role(ref r).RoleMember();`
+```
+
+Roles appearing in signatures are emitted as the role's underlying type
+marked with a modopt of the role type.
+
+```
+void M(Role r) // emitted as `void M(modopt(Role) UnderlyingType r)`
```
RikkiGibson, line 595: More detail is needed for scenarios where roles are used as nested types--array types, pointed-at types, type arguments, etc.

proposals/roles.md
```diff
@@ -0,0 +1,665 @@
...
+##### Conversions
+
+TODO
+```
+role R : U { } 
+R r = default;
+object o = r; // what conversion is that? if R doesn't have `object` as base type. What about interfaces?
+```
+
+Should allow conversion operators. Extension conversion is useful.
```
RikkiGibson, line 646: neo going through the looping train station

RikkiGibson: I have been reviewing this for a few hours so the sense of deja vu is getting strong.. 😉

proposals/roles.md
```diff
@@ -0,0 +1,665 @@
...
+void M(Role r) // emitted as `void M(modopt(Role) UnderlyingType r)`
+```
+
+TODO how do we emit an extension type versus a handcrafted ref struct?  
+TODO how do we emit the relationship to underlying type? base type or special constructor?  
+TODO our emit strategy should allow using pointer types and ref structs as underlying types
+in the future.  
+TODO issues in async code with ref structs
+
+## Phase A: Adding static constants, fields, methods and properties
```
RikkiGibson, line 604: It feels like this header should say *Adding extensions with static constants, ...*

proposals/roles.md
```diff
@@ -0,0 +1,665 @@
...
+
+In this first subset of the feature, the syntax is restricted to *extension_declaration*
```
RikkiGibson, line 606: One thing that I really feel is missing from here is discussion on *which way to go* in terms of extending extensions 🙂. We have an extension methods feature already. The two ways to go with a general extensions feature are to:

1. build out more kinds of extension declarations using the existing scheme, of a static containing type and use of a `this` modifier
2. build a completely new type kind with an expanded set of capabilities (which is what we are doing).

I'm not deeply against (2), but I think by default there should be an explanation of *what we get out of it* that we could never get out of (1), and agreement established for why it is fine to eventually supplant the existing extension methods feature in the language.

When `extension` declarations finally come along with the ability to declare instance methods, the existing extension methods feature will be obsolete in a similar sense that `delegate() { }` syntax was obsoleted by `() => {}`. This is my understanding of the proposal as written--please correct me if I'm wrong.

proposals/roles.md
```diff
@@ -0,0 +1,704 @@
...
+    : extension_type (',' extension_or_interface_type_list)
+    ;
+
+extension_body
+    : '{' extension_member_declaration* '}'
+    ;
+
+extension_member_declaration
+    : constant_declaration
+    | field_declaration
```
CyrusNajmabadi, line 57: how do fields work.  how can you add a field through an extension?  i get constant fields, but i don't get fields.

proposals/roles.md
```diff
@@ -0,0 +1,704 @@
...
+    | method_declaration
+    | property_declaration
+    | event_declaration
```
CyrusNajmabadi, line 60: events also seem problematic.  or is this just add/remove events, and not field-style events?

proposals/roles.md
jcouv: Below in the document, we call out that only static fields are allowed.
This wasn't yet spelled out for events, but the same thing will be applicable. Anything that involves an instance field (such as auto- or semi-auto-property) is disallowed.

CyrusNajmabadi: ah... static fields.  Got it.  Thanks!

PR #6883, opened 2023-01-15
Update using-alias-types.md

CyrusNajmabadi:


proposals/using-alias-types.md
```diff
@@ -38,20 +38,17 @@ The sections that do need updating are:
...
+# Supporting aliases to types containing pointers.
 
+A new [unsafe context](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/unsafe-code.md#222-unsafe-contexts) is added 
 
---
-
-Similarly what should be done about `using NullablePerson = Person?; // Person is a reference type`?  My intuition is that this is fine (though should only be legal if the *using* is in a `#nullable enable` section).  The meaning of `NullablePerson` in all reference locations is `Person?` (even if that location is `#nullable disable`).  However, depending on the nullability region where it is referenced you may or may not get nullable warnings around it.
+```diff
+unsafe_using_alias_directive
+using_alias_directive
```
CyrusNajmabadi, line 51:
```suggestion
```

CyrusNajmabadi:
@333fred @jcouv ptal.  Thanks!

proposals/using-alias-types.md
```diff
@@ -38,20 +38,22 @@ The sections that do need updating are:
 + The namespace_name referenced by a using_namespace_directive is resolved in the same way as the namespace_or_type_name referenced by a using_alias_directive. Thus, using_namespace_directives in the same compilation unit or namespace body do not affect each other and can be written in any order.
 ```
 
+```diff
++ It is illegal for a using alias type to be a nullable reference type.  e.g. `using X = string?;` is not legal.
```
333fred, line 42: Does this need to specify top-level in some fashion? Ie, `List<string?>` will still be legal to alias.

CyrusNajmabadi: Happy to add that.

333fred (approved):

jcouv:
> +    : 'using' identifier '=' (namespace_name | type) ';'

Add `unsafe?` (line 19)

---
Refers to: proposals/using-alias-types.md:19 in 0d4acd9. [](commit_id = 0d4acd97ab469b371b130e1b0212acee03387450, deletion_comment = False)

proposals/using-alias-types.md
```diff
@@ -38,20 +38,25 @@ The sections that do need updating are:
...
 
-The spec is currently unclear if this would be ok or not.  Technically, the `using_alias_directive` here is not in an `unsafe` context, so the `My*` could be considered an error.  However, the spirit of this specification is that should be allowed, and only the *usages* of `MyPointer` would themselves have to either be in another `using_alias_directive` or in an `unsafe` context.  Another way this could be formalized is that the `(namespace_name | type)` portion of a `using_alias_directive` would always be an `unsafe` context, but that wouldn't negate the fact that any place that alias was referenced would also need to be an `unsafe` context.
+```diff
+using_alias_directive
++    : 'using' 'unsafe'? identifier '=' (namespace_name | type) ';'
+    ;
 
++ The 'unsafe' modifier specified in the alias declaration causes the entire textual extent of the 'type' portion to become an unsafe context. 
++ 'unsafe' cannot be used with a 'using static' declaration. 
++ 'unsafe' can only be used with an using_alias_directive, not a using_directive.
```
jcouv, line 59: Need to disallow `unsafe` with a namespace

proposals/using-alias-types.md
```diff
@@ -38,20 +38,25 @@ The sections that do need updating are:
...
 ```
 
+```diff
++ It is illegal for a using alias type to be a nullable reference type.
 
-## Design meeting open questions.
-
-This section needs to be resolved in a design meeting.
+    1. `using X = string?;` is not legal.
+    2. `using X = List<string?>;` is legal.  The alias is to `List<...>` which is itself not a nullable reference type itself, even though it contains one as a type argument.
```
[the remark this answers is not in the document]
CyrusNajmabadi: will do.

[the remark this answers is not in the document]
CyrusNajmabadi: Done!

proposals/using-alias-types.md
```diff
@@ -38,20 +38,25 @@ The sections that do need updating are:
...
-```
+A new [unsafe context](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/unsafe-code.md#222-unsafe-contexts) is added through an optional 'unsafe' modifier in the using_alias_directive production:
 
-The spec is currently unclear if this would be ok or not.  Technically, the `using_alias_directive` here is not in an `unsafe` context, so the `My*` could be considered an error.  However, the spirit of this specification is that should be allowed, and only the *usages* of `MyPointer` would themselves have to either be in another `using_alias_directive` or in an `unsafe` context.  Another way this could be formalized is that the `(namespace_name | type)` portion of a `using_alias_directive` would always be an `unsafe` context, but that wouldn't negate the fact that any place that alias was referenced would also need to be an `unsafe` context.
+```diff
+using_alias_directive
++    : 'using' 'unsafe'? identifier '=' (namespace_name | type) ';'
+    ;
 
++ The 'unsafe' modifier specified in the alias declaration causes the entire textual extent of the 'type' portion to become an unsafe context.
```
jcouv, line 58: >modifier [](http://example.com/codeflow?start=15&length=8)

nit: I might avoid the word "modifier" here, since we're not dealing with a modifier list. For comparison, section 22.2 only uses "`unsafe` modifier" for positions that are actually modifiers.

jcouv (commented):
Done with review pass (iteration 8)

proposals/using-alias-types.md
CyrusNajmabadi: That's what the last line does?  Or do you mean that we should say that even for an alias that it explicitly doesn't cover the namespace_name portion?

CyrusNajmabadi: Updated the text to say it *only* covers the 'type' portion, not the 'namespace_name' portion.

proposals/using-alias-types.md
CyrusNajmabadi: done.

proposals/using-alias-types.md
```diff
@@ -38,20 +38,26 @@ The sections that do need updating are:
...
+    3. `using X = int?;` is legal.  This is a nullable *value* type, not a nullable *reference* type.
+```
 
-The intent of this specification is to allow one to write something like:
+# Supporting aliases to types containing pointers.
 
-```
-using MyPointer = My*;
-```
+A new [unsafe context](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/unsafe-code.md#222-unsafe-contexts) is added through an optional 'unsafe' modifier in the using_alias_directive production:
```
jcouv, line 51:
```suggestion
A new [unsafe context](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/unsafe-code.md#222-unsafe-contexts) is added through an optional 'unsafe' in the using_alias_directive production:
```

jcouv (approved):
LGTM Thanks (iteration 9). One nit left to consider (removing one last mention of "modifier")

proposals/using-alias-types.md
```diff
@@ -38,20 +38,26 @@ The sections that do need updating are:
...
 
-The spec is currently unclear if this would be ok or not.  Technically, the `using_alias_directive` here is not in an `unsafe` context, so the `My*` could be considered an error.  However, the spirit of this specification is that should be allowed, and only the *usages* of `MyPointer` would themselves have to either be in another `using_alias_directive` or in an `unsafe` context.  Another way this could be formalized is that the `(namespace_name | type)` portion of a `using_alias_directive` would always be an `unsafe` context, but that wouldn't negate the fact that any place that alias was referenced would also need to be an `unsafe` context.
+```diff
+using_alias_directive
++    : 'using' 'unsafe'? identifier '=' (namespace_name | type) ';'
```
jcouv, line 55: Did LDM discuss `unsafe using` versus `using unsafe`? On one hand we have `global using` (keyword before `using`) and on the other we have `using static` (keyword after).

CyrusNajmabadi: We didn't explicitly discuss the two options. We can bring to see how people think about this.

proposals/using-alias-types.md
CyrusNajmabadi, line 51:
```suggestion
A new [unsafe context](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/unsafe-code.md#222-unsafe-contexts) is added through an optional 'unsafe' keyword in the using_alias_directive production:
```

proposals/using-alias-types.md
CyrusNajmabadi: I've added it to the next meeting.

PR #6919, opened 2023-01-25
Publish notes for roles WG 2023-01-25

jcouv:
FYI @333fred @AlekseyTs @MadsTorgersen @captainsafia

### DIFF
--- a/meetings/working-groups/roles/roles-2023-01-25.md
+++ b/meetings/working-groups/roles/roles-2023-01-25.md
@@ -1,0 +1,108 @@
+# Roles WG 2023-01-25
+
+## Language for relationships
+
+We'll use "augments" for relationship to underlying type 
+(comparable to "inherits" for relationship to base type).  
+We'll use "inherits" for relationship to inherited roles 
+(comparable to "implements" for relationship to implemented interfaces).  
+
+```csharp
+struct U { }
+role X : U { }
+role Y : U, X, X1 { }
+```
+"Y has underlying type U"  
+"Y augments U"  
+"Y inherits X and X1"  
+"Derived role Y inherits members from inherited roles X and X1"  
+
+## Syntax for role underlying type list
+
+We should assume multiple inheritance for now. Will update syntax to allow multiple inherited roles.  
+In that syntax, the first type will be the underlying type.  
+`role R : I { } // First type is underlying type (not an implemented interface)`
+
+## Type parameters in extensions
+
+The purpose of requiring all extension type parameters to 
+appear in the underlying type is to help find a unique substitution 
+when looking for compatible extensions.
+
+For example, if you have `extension X<T> : U<T>` and then need to decide
+whether it is compatible with the `U<int>` type, you need to find a unique
+compatible substituted extension type `X<int>`.  
+
+This rule does not apply to roles (assuming that we don't let consumers turn 
+a role into an extension after the fact).
+
+## Constraints 
+
+Will need to spell out rules for all kinds of constraints, such as `struct`, `class`, etc.
+
+`where T : Extension`, `where T : Role`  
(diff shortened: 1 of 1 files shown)


RikkiGibson:
I haven't read these notes closely, so apologies if I missed any related discussion.

It seems like if roles are "backed" by ref structs, it could complicate their use in async code. Is there a need to more formally/rigorously "loosen" the restrictions around ref structs in async methods, in order for a ref-struct-backed roles feature to be generally usable?

meetings/working-groups/roles/roles-2023-01-25.md
```diff
@@ -0,0 +1,108 @@
...
+"Y has underlying type U"  
+"Y augments U"  
+"Y inherits X and X1"  
+"Derived role Y inherits members from inherited roles X and X1"  
+
+## Syntax for role underlying type list
+
+We should assume multiple inheritance for now. Will update syntax to allow multiple inherited roles.  
+In that syntax, the first type will be the underlying type.  
+`role R : I { } // First type is underlying type (not an implemented interface)`
```
333fred, line 24: I do feel like this syntax is a bit unwieldy. It mostly works ok for class inheritance, because there are only two categories (inherit and implement), and in all contexts the length of one of those categories is statically known (0 or 1 inherits for reference types, 0 implements for interfaces). For roles, this is not true: 1 underlying type, M inherited roles, and N implemented interfaces. I think that's likely to introduce complexity in understanding the code, for both roslyn and the human reader. Might be worth taking a note of this issue as an open question for the eventual syntax debate.

meetings/working-groups/roles/roles-2023-01-25.md
```diff
@@ -0,0 +1,108 @@
...
+
+This will need more discussion, and we prefer to separate it from language-level questions.  
+`System.Role<T>` is harder to use for T being pointer, ref struct. Also implies runtime support.  
+We don't want a breaking change when we reach phase C and you add an interface to an existing role.  
+We'd rather limit amount of runtime work in phase B, if possible. 
+
+A typeref is preferrable to some string.  
+We brainstormed using a modopt for encoding the underlying type in a way that would support pointers and
+ref structs. Maybe something like `System.Role<modopt(pointer) void>` or `ref struct S : modopt(pointer) System.Role`.  
+But there is a deep restriction in compiler codebase. Internal/public API expose custom modifiers as named types.
```
333fred, line 66: Might also be worth calling out that this restriction stems from the published version of 335, so if we take a dependency on being able to put `Type`s in a `modopt` and not just `TypeRef`s, the runtime will want to doc the spec change in their 335 augments doc.

meetings/working-groups/roles/roles-2023-01-25.md
FaustVX: I agree with you, it's not the clearer syntax.
I've already proposed using `for` https://github.com/dotnet/csharplang/issues/5497#issuecomment-984900268
But with the addition of multiples `roles`, maybe `role Role for UnderlyingType : Interface1, Interface2 with Role1, Role2 {}`
Everything is well separated and reads well for the user and the compiler.

jcouv: Yup. That's definitely a concern. I have an open issue on syntax in speclet.
C# went with a flat list (rather than separators and keywords) in the past, but it's definitely getting harder for this scenario.
A possible mitigation is to lean on naming convention (`I` prefix for interfaces helps, maybe a similar naming convention for roles?).

FaustVX: Personally, I'm OK with a flat list for `class`es because it can only have 1 other `class`, the rest are `interface`s.
But for Roles, even with a potential prefix (`R...`) or suffix (`...Role`), I still prefer to have some sort of separator.

PR #6938, opened 2023-02-02
Clarify restrictions for accessing captured primary constructor parameters in a readonly context.

AlekseyTs:


proposals/primary-constructors.md
```diff
@@ -128,7 +128,7 @@ Primary constructor parameters in class/struct declarations can be declared `ref
 
 All instance member initializers in the class body will become assignments in the generated constructor.
 
-If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. 
+If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. In a readonly struct the capture fields will be readonly. Therefore, access to captured parameters of a readony struct will have similar restriction as access to readonly fields. Access to captured parameters within a readonly member will have similar restrictions as access to instance fields in the same context.
```
cston, line 131:
```suggestion
If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. In a readonly struct the capture fields will be readonly. Therefore, access to captured parameters of a readonly struct will have similar restriction as access to readonly fields. Access to captured parameters within a readonly member will have similar restrictions as access to instance fields in the same context.
```

cston (approved):

proposals/primary-constructors.md
jjonescz:
```suggestion
If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. In a readonly struct the capture fields will be readonly. Therefore, access to captured parameters of a readonly struct will have similar restrictions as access to readonly fields. Access to captured parameters within a readonly member will have similar restrictions as access to instance fields in the same context.
```

proposals/primary-constructors.md
```diff
@@ -128,7 +128,7 @@ Primary constructor parameters in class/struct declarations can be declared `ref
 
 All instance member initializers in the class body will become assignments in the generated constructor.
 
-If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. 
+If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. In a readonly struct the capture fields will be readonly. Therefore, access to captured parameters of a readonly struct will have similar restriction as access to readonly fields. Access to captured parameters within a readonly member will have similar restrictions as access to instance fields in the same context.
```
AlekseyTs, line 131:
```suggestion
If a primary constructor parameter is referenced from within an instance member, and the reference is not inside of a `nameof` argument, it is captured into the state of the enclosing type, so that it remains accessible after the termination of the constructor. A likely implementation strategy is via a private field using a mangled name. In a readonly struct the capture fields will be readonly. Therefore, access to captured parameters of a readonly struct will have similar restrictions as access to readonly fields. Access to captured parameters within a readonly member will have similar restrictions as access to instance fields in the same context.
```

PR #6993, opened 2023-02-22
Delete non-record-primary-constructors.md

jcouv:
FYI @MadsTorgersen @AlekseyTs


RikkiGibson:
Is there a specific reason we want to delete rather than move to the rejected folder?

jcouv:
The proposal was not rejected, it was absorbed into a more spelled out proposal: https://github.com/dotnet/csharplang/blob/main/proposals/primary-constructors.md

333fred (approved):

PR #7038, opened 2023-03-08
Add more details to extensions spec

jcouv:


333fred (approved):

proposals/extensions.md
```diff
@@ -92,48 +101,56 @@ TODO should we have a naming convention like `Extension` suffixes? (`DataObjectE
...
+An extension type satisfies the constraints satisfied by its underlying type (see section on constraints). 
+In phase C, some additional constraints can be satisfied (additional implemented interfaces).  
 
-It is a compile-time error if the `implicit` or `explicit` modifiers differ amongst
-all the parts of an extension declaration.  
+### Underlying type
+
+The *extension_underlying_type* type may not be `dynamic`, a pointer, 
+a ref struct type, a ref type or an extension.  
+The underlying may not include an *interface_type_list* (this is part of Phase C).
```
jjonescz, line 120:
```suggestion
The underlying type may not include an *interface_type_list* (this is part of Phase C).  
```

proposals/extensions.md
```diff
@@ -144,7 +161,7 @@ The underlying type must include all the type parameters from the implicit exten
 
 ### Terminology
 
-We'll use "augments" for relationship to underlying type 
+We'll use "extends" for relationship to underlying/extended type
```
jjonescz, line 164: Is underlying and extended type the same thing? Should we rename "underlying type" to "extended type" (seems we are in process of doing that in the implementation)?

jjonescz (approved):

proposals/extensions.md
```diff
@@ -92,48 +101,56 @@ TODO should we have a naming convention like `Extension` suffixes? (`DataObjectE
...
 TODO should the underlying type be inferred when none was specified but
 base extensions are specified?
 
+### Modifiers
+
+It is a compile-time error if the `implicit` or `explicit` modifiers differ amongst
+all the parts of an extension declaration.  
+
+The permitted modifiers on an extension type are `partial`, 
+`unsafe`, `static`, `file` and the accessibility modifiers.
```
KennethHoff, line 143: > `file` and the accessibility modifiers

Is `file` _not_ an accessibility modifier?

(I don't know if I'm overstepping my boundary here, commenting on this PR)

jcouv: It's subtly different than an accessibility modifier. Accessibility modifiers mean that lookup finds the entity, but then it may not be accessible.
With `file`, lookup will not even find the entity that is not visible.
https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/file-local-types.md

proposals/extensions.md
jcouv: Yes, they are the same and I'll align that in the spec first. For the implementation, I'm trying to be strategic to avoid noisy churn. Thanks

PR #7064, opened 2023-03-16
Create inline-arrays.md

AlekseyTs:


proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+A `fixed_size_buffer_declarator` introduces a fixed-size buffers of a given element type.
+
+The buffer element type is the `type` specified in `field_declaration`. A fixed-size buffer declarator introduces a new member and consists of an identifier that names the member, followed by a constant expression enclosed in `[` and `]` tokens. The constant expression denotes the number of elements in the member introduced by that fixed-size buffer declarator. The type of the constant expression shall be implicitly convertible to type `int`, and the value shall be a non-zero positive integer.
+
+The elements of a fixed-size buffer shall be laid out sequentially in memory as though they are elements of an array.
+
+A `volatile` modifier cannot be used with `fixed_size_buffer_declarator`. A System.ThreadStaticAttribute cannot be applied to a field declaration like that.
+
+Depending on the situation (details are specified below), an access to a fixes-size buffer member is classified as a value (never a variable) of either
+`System.ReadOnlySpan<S>` or `System.Span<S>`, where S is the element type of the fixed-size buffer. Both types provide indexers returning refernce to a
```
Joe4evr, line 59: Typo
```diff
- Both types provide indexers returning refernce
+ Both types provide indexers returning reference
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
+Safe Fixed Size Buffers
+=====
+
+## Summary
+
+Provide a general-purpose and safe mechanism for declaring fixed sized buffers within C# classes, structs and interfaces.
```
tannergooding, line 6: For interfaces, it's specifically `static fields` only, is that correct?

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+
+A `fixed_size_buffer_declarator` introduces a fixed-size buffers of a given element type.
+
+The buffer element type is the `type` specified in `field_declaration`. A fixed-size buffer declarator introduces a new member and consists of an identifier that names the member, followed by a constant expression enclosed in `[` and `]` tokens. The constant expression denotes the number of elements in the member introduced by that fixed-size buffer declarator. The type of the constant expression shall be implicitly convertible to type `int`, and the value shall be a non-zero positive integer.
+
+The elements of a fixed-size buffer shall be laid out sequentially in memory as though they are elements of an array.
+
+A `volatile` modifier cannot be used with `fixed_size_buffer_declarator`. A System.ThreadStaticAttribute cannot be applied to a field declaration like that.
+
+Depending on the situation (details are specified below), an access to a fixes-size buffer member is classified as a value (never a variable) of either
```
Joe4evr, line 58:
```diff
- access to a fixes-size buffer
+ access to a fixed-size buffer
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+`System.ReadOnlySpan<S>` or `System.Span<S>`, where S is the element type of the fixed-size buffer. Both types provide indexers returning refernce to a
+specific element with appropriate "readonly-ness". Which prevents direct assignment to the elements when language rules don't permit that.
```
Joe4evr, line 60:
```diff
- appropriate "readonly-ness". Which prevents
+ appropriate "readonly-ness", which prevents
```

Joe4evr (commented):
One more typo and a grammatical nit.

proposals/safe-fixed-sized-buffers.md
Joe4evr: I can assume that's implied since instance fields in interfaces aren't allowed (as far as the language is concerned, at least), but an explicit call-out would be good.

333fred, line 6: Nit: oxford comma.
```suggestion
Provide a general-purpose and safe mechanism for declaring fixed sized buffers within C# classes, structs, and interfaces.
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Summary
+
+Provide a general-purpose and safe mechanism for declaring fixed sized buffers within C# classes, structs and interfaces.
+
+## Motivation
+
+This proposal plans to address the many limitations of https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/unsafe-code.md#228-fixed-size-buffers.
+Specifically it aims to allow the declaration of safe `fixed` buffers for managed and unmanaged types in a `struct`, `class`, or `interface`.
+And provide language safety verification for them.
```
333fred, line 12: The last line is a fragment, not a complete sentence.
```suggestion
Specifically it aims to allow the declaration of safe `fixed` buffers for managed and unmanaged types in a `struct`, `class`, or `interface`, and provide language safety verification for them.
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Detailed Design 
+
+Grammar for `variable_declarator` in https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#145-fields
```
333fred, line 16:
```suggestion
The grammar for *variable_declarator* in https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#145-fields
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+    ;
+    
+fixed_size_buffer_declarator
+    : identifier '[' constant_expression ']'
+    ;    
+```
+
+A `fixed_size_buffer_declarator` introduces a fixed-size buffers of a given element type.
+
+The buffer element type is the `type` specified in `field_declaration`. A fixed-size buffer declarator introduces a new member and consists of an identifier that names the member, followed by a constant expression enclosed in `[` and `]` tokens. The constant expression denotes the number of elements in the member introduced by that fixed-size buffer declarator. The type of the constant expression shall be implicitly convertible to type `int`, and the value shall be a non-zero positive integer.
```
333fred, line 52:
```suggestion
The buffer element type is the *type* specified in `field_declaration`. A fixed-size buffer declarator introduces a new member and consists of an identifier that names the member, followed by a constant expression enclosed in `[` and `]` tokens. The constant expression denotes the number of elements in the member introduced by that fixed-size buffer declarator. The type of the constant expression must be implicitly convertible to type `int`, and the value must be a non-zero positive integer.
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+
+The elements of a fixed-size buffer shall be laid out sequentially in memory as though they are elements of an array.
+
+A `volatile` modifier cannot be used with `fixed_size_buffer_declarator`. A System.ThreadStaticAttribute cannot be applied to a field declaration like that.
```
333fred, line 56:
```suggestion
A *field_declaration* with a *fixed_size_buffer_declarator* cannot have the `volatile` modifier, and it cannot be marked with a `System.ThreadStaticAttribute`.
```

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,115 @@
...
+    : identifier ('=' variable_initializer)?
+    | fixed_size_buffer_declarator
+    ;
+    
+fixed_size_buffer_declarator
+    : identifier '[' constant_expression ']'
+    ;    
+```
+
+A `fixed_size_buffer_declarator` introduces a fixed-size buffers of a given element type.
```
333fred, line 50:
```suggestion
A *fixed_size_buffer_declarator* introduces a fixed-size buffers of a given element type.
```

proposals/safe-fixed-sized-buffers.md
AlekseyTs: > For interfaces, it's specifically `static fields` only, is that correct?

Yes, instance fields cannot be declared in an interface. I'll add an explicit note since they are not called fields in the speclet.

333fred (commented):
Some style and grammar comments.

proposals/safe-fixed-sized-buffers.md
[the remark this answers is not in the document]
AlekseyTs: > I think that disallowing zero is fine, but did want to call out that C/C++ have "zero-length fixed sized buffers" and often use them as a way to indicate that the data is "variable length". 

No I don't want to call this out. The plan is to use https://github.com/dotnet/runtime/issues/61135 for metadata representation and it is said explicitly:
> `Length` must be greater than 0.

Apparently whatever C/C++ is using in not the same concept and is not using the same metadata representation.

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,121 @@
...
+    : identifier ('=' variable_initializer)?
+    | fixed_size_buffer_declarator
+    ;
+    
+fixed_size_buffer_declarator
+    : identifier '[' constant_expression ']'
+    ;    
+```
+
+A *fixed_size_buffer_declarator* introduces a fixed-size buffers of a given element type.
```
cston, line 49: >fixed-size buffers [](http://example.com/codeflow?start=46&length=18)

"fixed-size buffer" (singular)?

AlekseyTs: Yes, it is a typo.

proposals/safe-fixed-sized-buffers.md
```diff
@@ -0,0 +1,121 @@
...
+A fixed-size buffer can be referenced in an expression using a *simple_name* or a *member_access* .
+
+When an instance fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `this.I`, where `I` is the fixed-size buffer member. When a static fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `E.I`, where `I` is the fixed-size buffer member and `E` is the declaring type.
+
+#### Non-readonly fixed-size buffers
+
+In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies an instance fixed-size member,
+then `E.I` is evaluated and classified as follows:
+
+- If `E` is classified as a value, the result of the expression is classified as a value of type `System.ReadOnlySpan<S>`, where S is the element type of `I`.
```
cston, line 80: >If `E` is classified as a value [](http://example.com/codeflow?start=2&length=31)

Does this cover the case of a non-readonly fixed-size field in a `readonly` method?

AlekseyTs: > Does this cover the case of a non-readonly fixed-size field in a `readonly` method?

I think it does, `this` is qualified as a value rather than a variable in a `readonly` method, I think. In any case, my intent is to cover this scenario as well. If I discover that a special spec wording is necessary for that, I will add it.

333fred:
Consider moving https://github.com/dotnet/csharplang/blob/main/proposals/fixed-sized-buffers.md to the Rejected folder.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,140 @@
...
+
+In a member access of the form `E.I`, if `E` is of a class type and a member lookup of `I` in that class type identifies an instance fixed-size member,
+then `E.I` is evaluated and classified as a value of type `System.Span<S>`, where S is the element type of `I`.
+
+In a member access of the form `E.I`, if member lookup of `I` identifies a static fixed-size member,
+then `E.I` is evaluated and classified as a value of type `System.Span<S>`, where S is the element type of `I`.
+
+#### Readonly fixed-size buffers
+
+When a *field_declaration* includes a `readonly` modifier, the member introduced by the fixed_size_buffer_declarator is a ***readony fixed-size buffer***.
```
RikkiGibson, line 93:
```suggestion
When a *field_declaration* includes a `readonly` modifier, the member introduced by the fixed_size_buffer_declarator is a ***readonly fixed-size buffer***.
```

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,140 @@
...
+
+Member lookup of a fixed-size buffer member proceeds exactly like member lookup of a field.
+
+A fixed-size buffer can be referenced in an expression using a *simple_name* or a *member_access* .
+
+When an instance fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `this.I`, where `I` is the fixed-size buffer member. When a static fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `E.I`, where `I` is the fixed-size buffer member and `E` is the declaring type.
+
+#### Non-readonly fixed-size buffers
+
+In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies an instance fixed-size member,
```
RikkiGibson, line 77:
```suggestion
In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies an instance non-readonly fixed-size member,
```

I know the section title says *non-readonly*, but I found this confusing. This sentence by itself makes it sound like the rule applies even if the fixed-size member is readonly.

proposals/safe-fixed-size-buffers.md
```diff
@@ -114,6 +116,97 @@ Fixed-size buffers are not subject to definite assignment-checking, and fixed-si
...
 
+### Metadata
+
+#### Metadata emit and code generation
+
+For metadata encoding compiler will rely on recently added [```System.Runtime.CompilerServices.InlineArrayAttribute```](https://github.com/dotnet/runtime/issues/61135). 
+
+Fixed-size buffers like:
+``` C#
+public struct C
```
cston, line 127: Should this be `class C` to match the equivalent C# code below?

proposals/safe-fixed-size-buffers.md
```diff
@@ -114,6 +116,97 @@ Fixed-size buffers are not subject to definite assignment-checking, and fixed-si
...
+    public readonly System.ReadOnlySpan<T> AsReadOnlySpan()
+    {
+        // Note, an attempt to compile reports error CS1605: Cannot use '_element0' as a ref or out value because it is read-only
+        return System.Runtime.InteropServices.MemoryMarshal.CreateReadOnlySpan(ref _element0, 10);  
+    }
+}
+```
+
+The actual naming conventions for the type and its members are TBD. Framework will likely to include a set of predefined "buffer" types that cover
+limited set of buffer sizes. When predefined type doesn't exist, compiler will synthesize it in the module being built. Names of the generated types
```
cston, line 165: Perhaps:
> Framework will likely include a set of predefined "buffer" types that cover
> a limited set of buffer sizes. When a predefined type doesn't exist, ...

proposals/safe-fixed-size-buffers.md
```diff
@@ -114,6 +116,97 @@ Fixed-size buffers are not subject to definite assignment-checking, and fixed-si
...
+will be "speakable" in order to support consumption from other languages.
```
cston, line 166: Will the synthesized type be generated as a top-level type with the same namespace and name as the expected predefined type? Will the synthesized type be `internal` rather than `public`?

proposals/safe-fixed-size-buffers.md
```diff
@@ -114,8 +116,118 @@ Fixed-size buffers are not subject to definite assignment-checking, and fixed-si
...
+        return buffer2.AsReadOnlySpan()[1];
+    }
+}
+```
+
+#### Metadata import
+
+When compiler imports a field declaration of type *T* and the following conditions are all met:
+- *T* is a struct type decorated with the ```InlineArray``` attribute, and
+- The first instance field declared within *T* has type *F*, and
```
jjonescz, line 204: Shouldn't there be exactly one field in *T*?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,378 @@
...
+specific element with appropriate "readonly-ness", which prevents direct assignment to the elements when language rules don't permit that.
+
+This limits the set of types that can be used as a fixed-size buffer element type to types that can be used as type arguments. For example, a pointer type cannot be used as an element type.
+
+The resulting span instance will have a length equal to the size declared on the fixed-size buffer.
+Indexing into the span with a constant expression outside of the declared fixed-size buffer bounds is a compile time error.
+
+The *safe-to-escape* scope of the value will be equal to the *safe-to-escape* scope of the container, just as it would if the backing data was accessed as a field.
+
+### Fixed-size buffers in expressions
```
jaredpar, line 68: Do we need to spec the behavior of `nameof(C.Field)` where `Field` is a fixed sized buffer? Or does that just fall out from the member access rules?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,378 @@
...
+
+When an instance fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `this.I`, where `I` is the fixed-size buffer member. When a static fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `E.I`, where `I` is the fixed-size buffer member and `E` is the declaring type.
+
+#### Non-readonly fixed-size buffers
+
+In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies a non-readonly instance fixed-size member,
+then `E.I` is evaluated and classified as follows:
+
+- If `E` is classified as a value, the result of the expression is classified as a value of type `System.ReadOnlySpan<S>`, where S is the element type of `I`.
+  The value can be used to access members’ elements.
```
jaredpar, line 82: The value can be used to access memebr's elements but not must be. For example my reading believes the following should work: 

```csharp
S local = default;
Span<int> span = s.Field;

struct S {
  public int[5] Field;
}

```

Is that correct?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,378 @@
...
+    {
+        // Note, an attempt to compile reports error CS1605: Cannot use '_element0' as a ref or out value because it is read-only
+        return System.Runtime.InteropServices.MemoryMarshal.CreateReadOnlySpan(ref _element0, 10);  
+    }
+}
+```
+
+The actual naming conventions for the type and its members are TBD. The framework will likely include a set of predefined "buffer" types that cover
+a limited set of buffer sizes. When a predefined type doesn't exist, compiler will synthesize it in the module being built. Names of the generated types
+will be "speakable" in order to support consumption from other languages.
```
jaredpar, line 165: Agree they should be speakable. I was thinking about what the best namespace would be for these types when the compiler generates them. Ended up thinking that `System.Runtime.CompilerServices` is a good candidate. Mostly cause that is the likely place the libraries team would want to put the predefined types. That would mean 

1. The libraries team could use the compiler to effectively generate the types they need vs. hand coding them since we're putting them in the right place.
2. It's a consistent location for the types to always exist in

My other thought was `Microsoft.CodeAnalysis`.

proposals/safe-fixed-size-buffers.md
AlekseyTs: > The value can be used to access memebr's elements but not must be. For example my reading believes the following should work:

Yes, ```Span<int> span = s.Field``` will work.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+#### Fixed-size buffer element access
+
+For a fixed-size buffer element access, the *primary_no_array_creation_expression* of the *element_access* must be a variable or value of a fixed-size buffer type. Furthermore, the *argument_list* of a fixed-size buffer element access is not allowed to contain named arguments. The *argument_list* must contain a single expression, and the expression must be 
+- either of type `int`, or
+- must be implicitly convertible to `int`, or
+- must be of type ```System.Index```, or 
+- must be of type ```System.Range```.
```
cston, line 332: Consider dropping the "either" and "must be" from the bullets.
> - of type `int`, or
> - implicitly convertible to `int`, or
> - of type `System.Index`, or
> - of type `System.Range`.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+For a fixed-size buffer element access, the *primary_no_array_creation_expression* of the *element_access* must be a variable or value of a fixed-size buffer type. Furthermore, the *argument_list* of a fixed-size buffer element access is not allowed to contain named arguments. The *argument_list* must contain a single expression, and the expression must be 
+- either of type `int`, or
+- must be implicitly convertible to `int`, or
+- must be of type ```System.Index```, or 
+- must be of type ```System.Range```. 
+
+If *primary_no_array_creation_expression* is a variable, result of evaluating a fixed-size buffer element access is result of indexing into a ```System.Span<T>``` returned by *AsSpan* method with the same argument. 
+
+If *primary_no_array_creation_expression* is a value, result of evaluating a fixed-size buffer element access is result of indexing into a ```System.ReadOnlySpan<T>``` returned by *AsReadOnlySpan* with the same argument.
```
cston, line 336: Perhaps "_the_ result of evaluating a fixed-size buffer element access is _the_ result of ..." for the two cases above.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+There is an implicit conversion from expression representing a value of a fixed-size buffer type to ```System.ReadOnlySpan<T>``` type.
+The result of conversion is a span returned by *AsReadOnlySpan* method.
+
+### List patterns
+
+[List patterns](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/list-patterns.md) will not be supported for instances
+of fixed-size buffer types. In order to be able to take advantage of list patterns on fixed-size buffer types, a target expression must be
+explicitly converted either to ```System.Span<T>```, or to ```System.ReadOnlySpan<T>``` type first. The reason is the fact that pattern evaluation 
+liberally makes copies of the target expression, and it would be impossble to slice the copy into ```System.Span<T>```, but intent of the user might
```
cston, line 359:
```suggestion
liberally makes copies of the target expression, and it would be impossible to slice the copy into ```System.Span<T>```, but intent of the user might
```

proposals/safe-fixed-size-buffers.md
cston, line 336:
```suggestion
If *primary_no_array_creation_expression* is a variable, the result of evaluating a fixed-size buffer element access is the result of indexing into a ```System.Span<T>``` returned by *AsSpan* method with the same argument. 

If *primary_no_array_creation_expression* is a value, the result of evaluating a fixed-size buffer element access is the result of indexing into a ```System.ReadOnlySpan<T>``` returned by *AsReadOnlySpan* with the same argument. 
```

proposals/safe-fixed-size-buffers.md
cston, line 332: Consider dropping the "either" and "must be" from the bullets.
```suggestion
- of type `int`, or
- implicitly convertible to `int`, or
- of type ```System.Index```, or 
- of type ```System.Range```. 
```

cston (approved):

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+When compiler imports a field declaration of type *T* and the following conditions are all met:
+- *T* is a struct type decorated with the ```InlineArray``` attribute, and
+- The first instance field declared within *T* has type *F*, and
+- There is a ```public System.Span<F> AsSpan()``` within *T*, and 
+- There is a ```public readonly System.ReadOnlySpan<T> AsReadOnlySpan()``` or ```public System.ReadOnlySpan<T> AsReadOnlySpan()``` within *T*. 
+
+the field will be treated as C# fixed-size buffer with element type *F*. Otherwise, the field will be treated as a regular field of type *T*.
+
+
+## Detailed Design (Option 2)
```
cston, line 210: >(Option 2) [](http://example.com/codeflow?start=19&length=10)

Are the differences between the two options limited to element access? If so, consider removing the common items from the second option, so the differences are clear.

proposals/safe-fixed-size-buffers.md
cston, line 359:
```suggestion
liberally makes copies of the target expression, and it would be impossible to slice the copy into ```System.Span<T>```, but intent of the user might
```

proposals/safe-fixed-size-buffers.md
AlekseyTs: > Are the differences between the two options limited to element access?

No they are not limited to just to element access. The first option doesn't have special treatment of fixed-size buffer types, there is a difference in field type from the language point of view, any occurrences of fixed-size buffer types are  affected (not just fields), etc.

cston (approved):

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+A *field_declaration* with a *fixed_size_buffer_declarator* in an interface must have `static` modifier.
+
+Depending on the situation (details are specified below), an access to a fixed-size buffer member is classified as a value (never a variable) of either
+`System.ReadOnlySpan<S>` or `System.Span<S>`, where S is the element type of the fixed-size buffer. Both types provide indexers returning a reference to a
+specific element with appropriate "readonly-ness", which prevents direct assignment to the elements when language rules don't permit that.
+
+This limits the set of types that can be used as a fixed-size buffer element type to types that can be used as type arguments. For example, a pointer type cannot be used as an element type.
+
+The resulting span instance will have a length equal to the size declared on the fixed-size buffer.
```
jjonescz, line 63: > The resulting span instance will have a length equal to the size declared on the fixed-size buffer.

It seems this might not be always true if the Span is created by user-supplied buffer type (from metadata).

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+## Open design questions
+
+### Initializer
+
+Should we support initialization at declaration site with, perhaps, [collection literals](https://github.com/dotnet/csharplang/blob/main/proposals/collection-literals.md)?
+
+## Alternatives
+
+### Method or property group like approach in the language
```
jjonescz, line 384: This alternative looks like Option 2, what's the difference?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+#### Non-readonly fixed-size buffers
+
+In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies a non-readonly instance fixed-size member,
+then `E.I` is evaluated and classified as follows:
+
+- If `E` is classified as a value, the result of the expression is classified as a value of type `System.ReadOnlySpan<S>`, where S is the element type of `I`.
+  The value can be used to access members’ elements.
+- Otherwise, `E` is classified as a variable and the result of the expression is classified as a value of type `System.Span<S>`, where S is the element type of `I`.
+  The value can be used to access members’ elements.
```
jjonescz, line 84:
```suggestion
- If `E` is classified as a value, the result of the expression is classified as a value of type `System.ReadOnlySpan<S>`, where S is the element type of `I`.
  The value can be used to access the member's elements.
- Otherwise, `E` is classified as a variable and the result of the expression is classified as a value of type `System.Span<S>`, where S is the element type of `I`.
  The value can be used to access the member's elements.
```

jjonescz (approved):

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+variable_declarators
+    : variable_declarator (',' variable_declarator)*
+    ;
+    
+variable_declarator
+    : identifier ('=' variable_initializer)?
+    | fixed_size_buffer_declarator
+    ;
+    
+fixed_size_buffer_declarator
```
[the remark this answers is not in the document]
AlekseyTs: > Was the sharing intentional?

Yes.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+```ANTLR
+element_access
+    : primary_no_array_creation_expression '[' argument_list ']'
+    ;
+```
+
+The *argument_list* of an *element_access* is not allowed to contain `ref` or `out` arguments.
+
+An *element_access* is dynamically bound ([§11.3.3](expressions.md#1133-dynamic-binding)) if at least one of the following holds:
```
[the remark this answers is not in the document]
AlekseyTs: This is a direct quote. It is not my goal to maintain links

proposals/safe-fixed-size-buffers.md
AlekseyTs: > It seems this might not be always true if the Span is created by user-supplied buffer type (from metadata).

I am not sure why that would make the difference. The attribute emitted on the type is supposed to match the size. Malformed metadata isn't something that the language spec is concerned about.

proposals/safe-fixed-size-buffers.md
AlekseyTs: Starting with

> Safe fixed-size buffer accesses have their own classification (just like e.g. method groups and lambdas)

The option 2 didn't introduce any new classification. Values of fixed-size buffer types are classified as values of any other struct type.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,399 @@
...
+
+Member lookup of a fixed-size buffer member proceeds exactly like member lookup of a field.
+
+A fixed-size buffer can be referenced in an expression using a *simple_name* or a *member_access* .
+
+When an instance fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `this.I`, where `I` is the fixed-size buffer member. When a static fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `E.I`, where `I` is the fixed-size buffer member and `E` is the declaring type.
+
+#### Non-readonly fixed-size buffers
+
+In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies a non-readonly instance fixed-size member,
```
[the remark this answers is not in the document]
AlekseyTs: > The terms "fixed-size buffer", "fixed-size member" all seem ambiguous with the existing feature [fixed-size buffers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#fixed-size-buffers). The full unambiguous name is "safe fixed-size buffers", but that's long, should we use a different shorter name?

I will stick to the current terminology, but will make a note at the top to set the context

proposals/safe-fixed-size-buffers.md
jkotas: > Mostly cause that is the likely place the libraries team would want to put the predefined types.

From the BCL point of view, I do not think we want the compiler to ever auto-generate types in the `System.Runtime.CompilerServices` namespace. That's would be a factory for type-name collisions.

Can we make the physicial type of the field a hidden internal implementation detail and only expose these fields publicly via the Span and ReadOnlySpan accessors?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+    int M2()
+    {
+        return buffer2.AsReadOnlySpan()[1];
+    }
+}
+```
+
+#### Metadata import
+
+When compiler imports a field declaration of type *T* and the following conditions are all met:
```
jkotas, line 526: If this going to kick even for types in the current compilation unit? If somebody manually field of the right shape manually, is it going to be treated as C# fixed size buffer?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+- *T* is a struct type decorated with the ```InlineArray``` attribute, and
+- The first instance field declared within *T* has type *F*, and
+- There is a ```public System.Span<F> AsSpan()``` within *T*, and 
+- There is a ```public readonly System.ReadOnlySpan<T> AsReadOnlySpan()``` or ```public System.ReadOnlySpan<T> AsReadOnlySpan()``` within *T*.
```
jkotas, line 530: If the compiler is not going to validate that these AsSpan/AsReadOnlySpan methods have expected implementation, it opens possibility for some very creative hacks. For example, somebody can provide a custom AsReadOnlySpan implementation that returns variable sized Span. (Just an observation.)

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+- If access occurs in a context where direct assignments to an element of readonly fixed-size buffer are permitted, the result of the expression
+  is classified as a value of type `System.Span<S>`, where S is the element type of the fixed-size buffer. The value can be used to access member's elements.
+- Otherwise, the expression is classified as a value of type `System.ReadOnlySpan<S>`, where S is the element type of the fixed-size buffer.
+  The value can be used to access member's elements.
+
+### Definite assignment checking
+
+Fixed-size buffers are not subject to definite assignment-checking, and fixed-size buffer members are ignored for purposes of definite-assignment checking of struct type variables.
+
+When a fixed-size buffer member is static or the outermost containing struct variable of a fixed-size buffer member is a static variable, an instance variable of a class instance, or an array element, the elements of the fixed-size buffer are automatically initialized to their default values. In all other cases, the initial content of a fixed-size buffer is undefined.
```
jkotas, line 441: Does this is mean that the content will be uninitialized by default when the fixed-size buffer is a field of a local variable on stack? Is the program going to be able to see this uninitialized content without unsafe keyword?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+A *field_declaration* with a *fixed_size_buffer_declarator* in an interface must have `static` modifier.
+
+The field has a special struct type which represents a fixed-size buffer of a given element type.
+The element type is the *type* specified in *field_declaration*.  The constant expression
+denotes the number of elements in the buffer. The elements of a fixed-size buffer are laid out sequentially
+in memory as though they are elements of an array.
+
+The struct type will be defined using [```System.Runtime.CompilerServices.InlineArrayAttribute```](https://github.com/dotnet/runtime/issues/61135) feature.
+The framework will likely include a set of predefined "buffer" types that cover a limited set of buffer sizes. When a predefined type doesn't exist,
+compiler will synthesize it in the module being built. Names of the generated types will be "speakable" in order to support consumption from other languages.
```
stephentoub, line 64: I think we should reconsider this and not have them be speakable.  Related to Jan's feedback about System.Runtime.CompilerServices, I think we should instead:
- Have the compiler recognize whatever known public types we choose to expose from System.Runtime.CompilerServices
- Generate anything else it needs with internal unspeakable types.

This avoids conflicts with any types the user might declare, and avoids encouraging users to create types in System.Runtime.CompilerServices, which in general we don't want them doing.  I know and appreciate that one of the motivations for this was the libraries team creating those few built-in public helpers, but at most there will only be a handful of those and we can write them manually (even copy/pasting/tweaking what the compiler generated).

I also want the compiler to be free to add additional surface area to these types in the future.  For one of the same reason we don't publicly expose records, if the compiler is free to add additional APIs in the future, we wouldn't use it to create these public types, anyway.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+
+A *field_declaration* with a *fixed_size_buffer_declarator* in an interface must have `static` modifier.
+
+The field has a special struct type which represents a fixed-size buffer of a given element type.
+The element type is the *type* specified in *field_declaration*.  The constant expression
+denotes the number of elements in the buffer. The elements of a fixed-size buffer are laid out sequentially
+in memory as though they are elements of an array.
+
+The struct type will be defined using [```System.Runtime.CompilerServices.InlineArrayAttribute```](https://github.com/dotnet/runtime/issues/61135) feature.
+The framework will likely include a set of predefined "buffer" types that cover a limited set of buffer sizes. When a predefined type doesn't exist,
```
jkotas, line 63: What is the motivation for exposing predefined "buffer" types from the BCL? Is it just to save a bit of binary size or is it more than that?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+If *primary_no_array_creation_expression* is a value, the result of evaluating a fixed-size buffer element access is a value
+equaivalent to invoking [`public ref readonly T this[int index] { get; }`](https://learn.microsoft.com/en-us/dotnet/api/system.readonlyspan-1.item?view=net-8.0) with
+that integer value on an instance of ```System.ReadOnlySpan<T>``` returned by ```[UnscopedRef] public readonly System.ReadOnlySpan<T> AsReadOnlySpan()```
+method on *primary_no_array_creation_expression*. 
+
+For example:
+``` C#
+void M1(Buffer10<int> x)
+{
+    ref int a = ref x[0]; // Ok, equivalent to `ref int a = ref x.AsSpan()[0]`
```
stephentoub, line 158: The JIT should already be able to turn this into the equivalent of `ref x._element0`, but we should double-check it indeed does always do all the necessary inlining and the like.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+method on *primary_no_array_creation_expression*. 
+
+If *primary_no_array_creation_expression* is a value, the result of evaluating a fixed-size buffer element access is a value
+equaivalent to invoking [`public ref readonly T this[int index] { get; }`](https://learn.microsoft.com/en-us/dotnet/api/system.readonlyspan-1.item?view=net-8.0) with
+that integer value on an instance of ```System.ReadOnlySpan<T>``` returned by ```[UnscopedRef] public readonly System.ReadOnlySpan<T> AsReadOnlySpan()```
+method on *primary_no_array_creation_expression*. 
+
+For example:
+``` C#
+void M1(Buffer10<int> x)
```
stephentoub, line 156: I personally find it strange that this magically-named type appears out of thin air for a field `public int buffer[10];`.  With fixed buffers today, the type of that field is effectively an int*... have we considered doing the same thing here and having the effective type be `Span<T>`?

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,558 @@
...
+
+[List patterns](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/list-patterns.md) will not be supported for instances
+of fixed-size buffer types.
+
+### MemoryMarshal.CreateReadOnlySpan
+
+We might want to specially recognize ```System.Runtime.InteropServices.MemoryMarshal.CreateReadOnlySpan``` method and allow
+passing a reference to a readonly location as its first argument. This should allow users to define their own fixed-size buffer
+types. For example, in order to be able to share a type across multiple assemblies and also include custom APIs into its definition.
+This might also help framework to define our shared fixed-size buffer types.
```
stephentoub, line 319: Developers can do this today with:
```C#
MemoryMarshal.CreateReadOnlySpan(ref Unsafe.AsRef(in location), length);
```
But this is also an example of an API that's motivation for allowing `ref readonly` on parameters; we would change CreateReadOnlySpan to accept a `ref readonly` instead of a `ref`, existing usage would continue to work, but it would also then work for creating a span around readonly fields.

proposals/safe-fixed-size-buffers.md
stephentoub: > Can we make the physicial type of the field a hidden internal implementation detail and only expose these fields publicly via the Span and ReadOnlySpan accessors?

This is my preference as well.  Basically mirror what we do today for unsafe fixed buffers: a developer writes `fixed int buffer[10]`, the compiler generates an unspeakable type for that, and the effective type visible to consuming code is `int*`.  For safe buffers, it could be the same approach: the developer writes `int buffer[10]`, the compiler generates an unspeakable type for that, and the effective type visible to consuming code is either `Span<int>` or `ReadOnlySpan<int>` depending on whether the field is readonly.

> The libraries team could use the compiler to effectively generate the types they need vs. hand coding them since we're putting them in the right place.

Handcoding these is trivial. _If_ we wanted to expose any publicly for size/sharing reasons, it would be easy to do so without the compiler feature.  As noted earlier, I'd also be concerned about using the compiler feature to implement these, as it would either be limiting for what the compiler could emit on them in the future (any additions the compiler made here would implicitly impact core library public surface area), or if we didn't want to constrain the compiler in that fashion (I don't want to), we wouldn't use it for this, anyway.

proposals/safe-fixed-size-buffers.md
AlekseyTs: > What is the motivation for exposing predefined "buffer" types from the BCL? Is it just to save a bit of binary size or is it more than that?

The primary motivation is to avoid synthesizing types for common sizes in many user assemblies. There is also a benefit of interop between different user assemblies. If, for example, a buffer of size 10 type is in BCL, all user assemblies can easily pass instance of the same type around without a need to copy data between different versions of buffers of size 10.

proposals/safe-fixed-size-buffers.md
AlekseyTs: > I personally find it strange that this magically-named type appears out of thin air for a field `public int buffer[10];`.

As I mentioned on another thread, we don't have to use speakable names for types synthesized for field declarations. The purpose of the examples it to demonstrate the difference between a writable variable, a readonly variable, and a value. Note, that fields can be variables in these terms. That is a term that language specification uses for everything that has specific location in memory (locals, fields, by ref returns, array elements, etc.). The name of the type is not important and the examples can use fields without explicitly naming the types, but with the same exact effect. 

At the same time, one of the goals of this proposal is to support consumption of inline array types regardless of their names and regardless of where they locals, parameters, fields, return values, etc. Especially that anyone can declare types like that.

proposals/safe-fixed-size-buffers.md
AlekseyTs: > Does this is mean that the content will be uninitialized by default when the fixed-size buffer is a field of a local variable on stack? Is the program going to be able to see this uninitialized content without unsafe keyword?

This proposal allows that, but we can require the buffers to be always initialized. Though, this is like to require additional changes in the language because this proposal doesn't support direct assignment to the safe fixed-size member in its entirety, and it might not be possible for compiler to verify that all elements in the member are definitely assigned (for example, when non-constant expressions are used to assign to elements).

proposals/safe-fixed-size-buffers.md
AlekseyTs: > Is this going to kick even for types in the current compilation unit? If somebody manually field of the right shape manually, is it going to be treated as C# fixed size buffer?

Yes. In general, we don't make distinction between types in source and  type in metadata for the purpose of deciding if they match a pattern.

proposals/safe-fixed-size-buffers.md
AlekseyTs: > If the compiler is not going to validate that these AsSpan/AsReadOnlySpan methods have expected implementation, it opens possibility for some very creative hacks. For example, somebody can provide a custom AsReadOnlySpan implementation that returns variable sized Span. (Just an observation.)

The validation of method bodies is not proposed. In general, compiler is not concerned about types violating a pattern they pretend to satisfy. There is an option for these helpers to be automatically provided by runtime on all inline array types with robust to hacks implementation. For example, array types have helpers like that to access element references, etc.

333fred:
Partly discussed in LDM: https://github.com/dotnet/csharplang/blob/main/meetings/2023/LDM-2023-04-03.md#fixed-size-buffers

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,580 @@
...
+### Definite assignment checking
+
+Regular definite assignment rules are applicable to variables that have a fixed-size buffer type. 
+
+### An alternative to relying on AsSpan/AsReadOnlySpan helpers
+
+If there is a guarantee that the first element in a fixed-size buffer type is aligned at the beginning of the type (no gap), then compiler can use the
+following code to get a ```Span``` value (instead of relying on presence of ```AsSpan``` helper):
+``` C#
+new Span<element type>(Unsafe.AsPointer(ref <the fixed-size buffer writable variable>), <size of the fixed-size buffer>)
```
jkotas, line 323: Unsafe.AsPointer is only ok to use when the buffer is pinned. The buffer is not guaranteed to be pinned here.

This needs to use `MemoryMarshal.CreateSpan` API.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,580 @@
...
+```
+
+And the following code to get a ```ReadOnlySpan``` value (instead of relying on presence of ```AsReadOnlySpan``` helper):
+``` C#
+new ReadOnlySpan<element type>(Unsafe.AsPointer(ref Unsafe.AsRef(in <the fixed-size buffer writable variable>)), <size of the fixed-size buffer>)
+```
+
+In order to reduce IL size at use sites compiler should be able to add two generic reusable helpers into private implementation detail type and
```
jkotas, line 331: You can just directly emit `MemoryMarshal.CreateSpan` call inline and pass the address of the buffer and the length to it.

If you are emitting IL, you do not need the `Unsafe.AsPointer` or `Unsafe.AsRef` calls. These calls exist to make the code compile in C#. They are not necessary when emitting IL.

proposals/safe-fixed-size-buffers.md
```diff
@@ -0,0 +1,493 @@
...
+    return MemoryMarshal.CreateSpan(ref Unsafe.As<TBuffer, TElement>(ref buffer), size);
+}
+
+public static System.ReadOnlySpan<TElement> AsReadOnlySpan<TBuffer, TElement>(in TBuffer buffer, int size) where TBuffer : struct
+{
+    return MemoryMarshal.CreateReadOnlySpan(ref Unsafe.As<TBuffer, TElement>(ref Unsafe.AsRef(in buffer)), size);
+}
+```
+
+Also, quite possibly compiler will be able to omit the ```Unsafe.As``` and ```Unsafe.AsRef``` calls in IL.
```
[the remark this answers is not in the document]
AlekseyTs: > Once you do that, the `AsSpan` and `AsReadOnlySpan` helpers do not provide any IL size benefits.

After an offline discussion with @VSadov, I decided to not eliminate the calls. I will remove this statement from the spec.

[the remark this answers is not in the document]
jkotas: Can you share the reasoning behind it? It looks be perfectly fine to me to skip these unnecessary `Unsafe.As` calls in the generated IL.

proposals/inline-arrays.md
```diff
@@ -0,0 +1,492 @@
...
+
+This proposal will refer to types like this as "inline array types".
+
+Elements of an inline array type can be accessed through pointers or through span instances returned by
+[System.Runtime.InteropServices.MemoryMarshal.CreateSpan](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createspan?view=net-7.0)/[System.Runtime.InteropServices.MemoryMarshal.CreateReadOnlySpan](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createreadonlyspan?view=net-7.0) APIs. However, neither
+the pointer approach, nor the APIs provide type and bounds checking out of the box.
+
+Language will provide a type-safe/ref-safe way for accessing elements of inline array types. The access will be span based. 
+This limits support to inline array types with element types that can be used as a type argument.
+For example, a pointer type cannot be used as an element type. Other examples the span types.
```
CyrusNajmabadi, line 49: > "Other examples the span types."

Not sure hwat this means :)

proposals/safe-fixed-size-buffers.md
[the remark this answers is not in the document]
stephentoub: I'm curious about this as well. What was the rationale?

PR #7078, opened 2023-03-27
Update implemented proposal checklist

bagatka:
Found that checklist in [documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/static-abstracts-in-interfaces). I don't think it should be there.


bagatka:
@dotnet-policy-service agree

333fred (approved):

333fred:
Thanks @bagatka!

PR #7098, opened 2023-04-06
Add collection literals working group notes 2023-04-05

cston:


CyrusNajmabadi (approved):

meetings/working-groups/collection-literals/CL-2023-04-05.md
```diff
@@ -0,0 +1,94 @@
...
+public static class ImmutableCollections
+{
+    public static ImmutableArray<T> AsImmutableArray<T>(T[] array)
+    {
+        // ...
+    }
+}
+```
+
+A collection type should have at most one _construction method_. Most collection types will have none.
```
CyrusNajmabadi, line 94: hrmm.. not sure how i feel about that, but it's maybe ok.  My concern woudl be something like a collection that says "if you have known count, call the version on me where i give you the span.  but, if you don't, call the version that passes me the array to own".

But, practically, i care about this working for top collections like IA and List, and ImmutableList.  If this strategy works fine there for known/unknown-length literals, then i'm ok.  If it doesn't work out, we may want to relax and define a strict betterness ordering for the different construct-methods.

meetings/working-groups/collection-literals/CL-2023-04-05.md
```diff
@@ -0,0 +1,94 @@
...
+int __i = 0;
+__span[__i++] = x;
+foreach (T __t in e)
+    __span[__i++] = __t;
+__span[__i++] = y;
+```
+
+### Runtime libraries
+
+The BCL could provide _construction methods_ for `ImmutableArray<T>` and `List<T>`, and perhaps a few additional collection types where the underlying storage can be represented as a span.
```
jnm2, line 64: Should we mention that the BCL may do so already, and it sounded likely from how @stephentoub described it?

`ImmutableArray.CreateUnsafe<T>` is waiting for API review (https://github.com/dotnet/runtime/issues/83141)

I didn't find an issue for a single method for the "set count and request span" construction method that he mentioned. That would have to be a new method combining `CollectionsMarshal.SetCount<T>(List<T>, int)` approved at https://github.com/dotnet/runtime/issues/55217 and the shipped [`CollectionsMarshal.AsSpan<T>(List<T>)`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.asspan), and this use case is listed in the SetCount issue.

jnm2 (approved):

meetings/working-groups/collection-literals/CL-2023-04-05.md
```diff
@@ -0,0 +1,94 @@
+# Collection literals working group meeting for April 4, 2023
+
+## Agenda
+
+* Allow efficient construction of collections that are typically represented as arrays, particularly `ImmutableArray<T>`.
```
RikkiGibson, line 5: I felt like our conversation was also applicable to construction of mutable and immutable dictionaries and sets, etc.

meetings/working-groups/collection-literals/CL-2023-04-05.md
```diff
@@ -0,0 +1,94 @@
...
+__array[__i++] = x;
+foreach (T __t in e)
+    __array[__i++] = __t;
+__array[__i++] = y;
+
+ImmutableArray<T> result = AsImmutableArray(__array);
+```
+
+### Write directly to storage span
+An alternative is a _construction method_ that creates a collection of a given size and returns the collection _and_ a span that refers to the underlying storage, and the caller populates the collection using the span.
```
RikkiGibson, line 39: I'm curious to know more about why giving back a Span is going to result in better codegen. Is it because `.Add()` would inevitably add additional checks and indirections that can't be optimized away?

CyrusNajmabadi: yup.  Add has to go through the full public api.  which will also do things like update version counters inside the type.  The span it literally: here is the memory, write into it.  :)

cston: The example below should probably use `ImmutableArray<T>` rather than `List<T>`, since setting the capacity in the constructor and then using `.Add()` should be sufficient for `List<T>`.

meetings/working-groups/collection-literals/CL-2023-04-05.md
cston: We could consider extending this to allow attributing factory methods for constructing mutable and immutable dictionaries, sets, etc. Would that be for efficiency or simply to enable constructing types that are currently constructible such as `ImmutableDictionary<TKey, TValue>`?

meetings/working-groups/collection-literals/CL-2023-04-05.md
cston: I've updated the example to use `ImmutableArray<T>` instead.

meetings/working-groups/collection-literals/CL-2023-04-05.md
RikkiGibson: Both for efficiency and flexibility.

I feel that we should investigate whether a single pattern like the following could address all the construction scenarios we care about now or may care about in the future, while also providing the high performance we want here:
1. `CreateBuilder(capacity)`--maybe with a hint about whether the exact size is known for sure, or if you may receive a different number of elements than the given capacity.
2. `Add()` 0 or more times
3. `BuildAndClear()`

The builder should be a struct to reduce the amount of allocations, but the language doesn't need to require that.

`Add()` should do a minimal amount of work, be inlinable, and not do things like update a counter or whatever else.

`BuildAndClear()` could be designed with the assumption that the builder is in an invalid state afterwards, and doesn't need to account for being called again.

I avoided the `ToSpecificCollectionType()` naming convention here because I want the method names to be fixed on conventions, like the way we do Deconstruct, GetPinnableReference, so on and so forth.

cston: > I feel that we should investigate whether a single pattern like the following could address all the construction scenarios we care about now or may care about in the future, while also providing the high performance we want here:

Sounds good. Let's add this to the agenda for the next meeting.

CyrusNajmabadi: i'd want @stephentoub  to weigh in, as that's basically requiring such a surface area be provided (as opposed to the approach we talked about in the meeting, which would generally just be a constructor, or a single static method with an attribute) :)

My concern is this being a large tax for the rare case.

RikkiGibson (approved):

meetings/working-groups/collection-literals/CL-2023-04-05.md
cston: I'll leave as is, since it looks like those are API changes are only proposed currently.

PR #7116, opened 2023-04-12
Add interceptors working group notes

RikkiGibson:
The latest compiler feature doc for interceptors can be found at https://github.com/RikkiGibson/roslyn/blob/ic-2/docs/features/interceptors.md.


ufcpp:
Is the file name format MM-dd-yyyyy? While other woking groups use yyyy-MM-dd.

PR #7130, opened 2023-04-20
Add "Inline array type syntax" section to inline-arrays.md

AlekseyTs:


AlekseyTs:
@jaredpar FYI

proposals/inline-arrays.md
```diff
@@ -253,13 +253,184 @@ of inline array types.
...
+> *Example*: The type in `T[][,,][,]` is a single-dimensional array of three-dimensional arrays of two-dimensional arrays of `int`. *end example*
+
+At run-time, a value of a regular array type can be `null` or a reference to an instance of that array type.
+
+> *Note*: Following the rules of https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/arrays.md#166-array-covariance,
+> the value may also be a reference to a covariant array type. *end note*
+
+An anonymous inline array type is a compiler synthesized inline array type with internal accessibility. The element type must be a 
+type that can be used as a type argument. Unlike an explicitly declared inline array type, an anonymous inline array type cannot be
+referenced by name, it can be referenced only by *array_type* syntax. In context of the same program, any two *array_type*s
```
cston, line 303: >In context of the same program [](http://example.com/codeflow?start=70&length=30)

Should this be "in the same module" instead? Do we guarantee anonymous inline array types will be unified across modules?

AlekseyTs: > Should this be "in the same assembly" instead? Do we guarantee anonymous inline array types will be unified across assemblies?

The program refers to the binary being compiled. Other assemblies are not "the same program".

AlekseyTs: A program being compiled is not always an assembly. So, the most precise term would be a module, but language specification doesn't use this term other than in reference to an attribute target.

PR #7143, opened 2023-04-25
Updates to collection-literals.md

cston:
Split dictionaries out to a separate section; minor formatting changes.

### DIFF
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -129,8 +129,8 @@ Actual translation of the literal to the corresponding type is defined [below](#
 int[] a = [];                // array
 Span<int> b = [1, 2];        // span
 ImmutableArray<int> c = [3]; // Construct method
-List<int> c = [5, 6];        // collection initializer type
-IEnumerable<int> d = [];     // list interface
+List<int> d = [5, 6];        // collection initializer type
+IEnumerable<int> e = [];     // list interface
 ```
 
 ## `Construct` methods
@@ -231,9 +231,8 @@ This means there is no way for a literal to have a *natural type* of some `List<
 The choice of `List<T>` rather than `T[]` or `ImmutableArray<T>` is to allow mutation of `var` locals after initialization. `List<T>` is preferred over `Span<T>` because `Span<T>` cannot be used in `async` methods.
 
 ```c#
-var list = [1, 2, 3];
-// ...
-list.Add(4);
+var values = [1, 2, 3];
+values.Add(4); // ok
 ```
 
 Should `IEnumerable` contribute an *iteration type* of `object` or no contribution?
@@ -247,9 +246,8 @@ List<string> e3 = [..e1];  // error?
 The compiler is free to use a representation other than the natural type if the difference is not observable.
 
 ```c#
-foreach (var b in [true, false]) // Not necessarily List<bool>.
+foreach (var b in [true, false]) // Not necessarily List<bool>
 {
-    // ...
 }
 ```
 
@@ -258,36 +256,41 @@ foreach (var b in [true, false]) // Not necessarily List<bool>.
 * For example, given:
 
     ```c#
-    string i = ...;
+    string s = ...;
     object[] objects = ...;
-    var x = [i, ..objects];
+    var x = [s, ..objects]; // List<object>
     ```
 
-    The *natural type* of `x` is `List<T>` where `T` is the *best common type* of `i` and the *iteration type* of `objects`.  Respectively, that would be the *best common type* between `string` and `object`, which would be `object`.  As such, the type of `x` would be `List<object>`.
+    The *natural type* of `x` is `List<T>` where `T` is the *best common type* of `s` and the *iteration type* of `objects`.  Respectively, that would be the *best common type* between `string` and `object`, which would be `object`.  As such, the type of `x` would be `List<object>`.
 
 * Given:
 
     ```c#
-    var values = x ? [1, 2, 3] : [];
+    var values = x ? [1, 2, 3] : []; // List<int>
     ```
 
     The *best common type* between `[1, 2, 3]` and `[]` causes `[]` to take on the type `[1, 2, 3]`, which is `List<int>` as per the existing *natural type* rules. As this is a constructible collection type, `[]` is treated as target-typed to that collection type.
 
 ## Type inference
 
-* Should this work?
+Type inference should infer constructible collection types:
 
-    ```c#
-    var a = AsArray([1, 2, 3]);
+```c#
+static void F<T>(T[] arg) { }
+static void F<T>(ImmutableArray<T> arg) { }
 
-    static T[] AsArray<T>(params T[] array) => array;
-    ```
+F([1, 2, 3]); // error: ambiguous between F<T>(T[]) and F<T>(ImmutableArray<T>)
+```
 
-* And this?
+And for extension methods:
 
-    ```c#
-    var b = AsArray([.. b ? [x] : [y]]);
-    ```
+```c#
+static ImmutableArray<T> AsImmutableArray<T>(this ImmutableArray<T> arg) => arg;
+
+var a = [1, 2, 3].AsImmutableArray(); // ok: ImmutableArray<int>
+```
+
+Supporting type inference may require an intermediate *collection type* at compile-time, similar to the [*natural function type*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md#natural-function-type) used for conversions and type inference for lambda expressions and method groups.
 
 ## Span types
 [span-types]: #span-types
@@ -555,9 +558,9 @@ IReadOnlyDictionary<string, int> y = ["one":1, "two":2];
 Dictionary construction uses `this[int index] { set; }` rather than `Add()` when to ensure consistent _overwrite semantics_ rather than _add semantics_.
 
 ```c#
-var d = [x:x1, y:y1]; // {{x, x1}, {y, y1}}
-d = [..d, y:y2];      // {{x, x1}, {y, y2}}
-d = [x:x2, ..d];      // {{x, x2}, {y, y2}}
+var x = [a:1, b:2]; // {{a, 1}, {b, 2}}
+var y = [..x, b:4]; // {{a, 1}, {b, 4}}
+var z = [a:3, ..x]; // {{a, 3}, {b, 2}}
 ```
 
 As stated in [*natural type*](#natural-type):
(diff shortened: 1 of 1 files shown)


proposals/collection-literals.md
```diff
@@ -555,6 +513,107 @@ Not having a *known length* does not prevent any result from being created. Howe
...
+
+The following are _constructible collection types_ that can be used to construct collection literals.
+
+* ...
+* `Dictionary<TKey, TValue>`
+* Interface types `I<TKey, TValue>` implemented by `Dictionary<TKey, TValue>` (e.g. `IDictionary<TKey, TValue>`, `IReadOnlyDictionary<TKey, TValue>`)
+
+_Note that `Dictionary<TKey, TValue>` does not include a public `Add()` instance method with a single parameter, so the type is not constructible as a *collection initializer*._
+
+_Use `this[int index] { set; }` rather than `Add()` when constructing a dictionary for consistent set rather than add semantics._
```
333fred, line 528: I'm not sure what this means. Why would we want collection literals to have different observable behavior than collection initializers today?

proposals/collection-literals.md
```diff
@@ -112,21 +112,18 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
     * The behavior of collection literals with collections that are not well-behaved is undefined.
 
-## Constructible collection types
+## Constructible collection types (target types)
 [constructible-collection-types]: #constructible-collection-types
 
 The following are _constructible collection types_ that can be used to construct collection literals.
-Actual translation of the literal to the corresponding is defined [below](#collection-literal-translation).
+These are the valid _target types_ for collection literals.
+Actual translation of the literal to the corresponding type is defined [below](#collection-literal-translation).
```
CyrusNajmabadi, line 120: can we add here (or elsewhere) that this list is ordered.  So if something supports multiple of these cases (for example, an array is both `T[]` *and* `IList<T>`, that the earlier one is preferred).  THis is also relevant for `types with a Constrct method` vs `collection initializers`.  `List<T>` may likely end up matching both, and we would want the `construct method` to win out over `collection initializer`.

proposals/collection-literals.md
```diff
@@ -135,19 +132,19 @@ While certain types (like arrays and spans) can always be constructed with a col
 
 * the `Construct` method is found on an instance of `T` (including through [extension methods](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11783-extension-method-invocations)), and
 
-* `CollectionType` is some other type known to be a [`constructible`](constructible-collection-types) type.
+* `CollectionType` is some other type known to be a [*constructible*](constructible-collection-types) type.
 
 If found, the collection can be constructed by creating a fresh instance of its type using `new T()`, producing the corresponding argument to pass to `Construct`, and then calling that method on the fresh instance.  `new T()` supports all structs, including those without a `parameterless struct constructor`.
 
 The allowance for extension methods means that collection literal support can be added to a existing API which does not already directly support this.
+_An extension method could be added that would silently change how collections are created in the program._
```
CyrusNajmabadi, line 140: We might want to add text similar to what is the initial part of the doc, that we assume these symbols are well behaved.

proposals/collection-literals.md
```diff
@@ -178,10 +175,9 @@ Through the use of the [`init`](#init-methods) modifier, existing APIs can direc
     ImmutableArray<int> __result = __builder.MoveToImmutable();
     ```
 
-
 ## Empty collection literal
 
-* The empty literal `[]` has no type.  However, similar to the [`null-literal`](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/lexical-structure.md#6457-the-null-literal), this literal can be implicitly converted to any [`constructible`](#constructible-collection-types) collection type.
+* The empty literal `[]` has no type.  However, similar to the [*null-literal*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/lexical-structure.md#6457-the-null-literal), this literal can be implicitly converted to any [*constructible*](#constructible-collection-types) collection type.
```
CyrusNajmabadi, line 180: aside: why all the changes from `foo` to *foo*`?

proposals/collection-literals.md
```diff
@@ -555,6 +513,107 @@ Not having a *known length* does not prevent any result from being created. Howe
...
+    Dictionary<object, string> d2 = ...;
+    var d3 = [..d1, ..d2];
+    ```
+
+    The *natural type* of `d3` is `Dictionary<object, object>`.  This is because the `..d1` will have a *iteration type* of `KeyValuePair<string, object>` and `..d2` will have a *iteration type* of `KeyValuePair<object, string>`. These will contribute `{string, object}` to the determination of the `TKey` type and `{object, string}` to the determination of the `TValue` type.  In both cases, the *best common type* of each of these sets is `object`.
+
+* Given:
+
+    ```c#
+    var d = [null: null, "a": "b"];
```
333fred, line 591: This will throw an exception at runtime. Will we naturally just have a warning when the key type violates the nullable contract of the dictionary type?

proposals/collection-literals.md
```diff
@@ -207,57 +203,37 @@ Through the use of the [`init`](#init-methods) modifier, existing APIs can direc
...
 [natural-type]: #natural-type
 
 In the absence of a *constructible collection target type*, a non-empty literal can have a *natural type*.
 
 The *natural type* is determined using the [*best common type*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#116315-finding-the-best-common-type-of-a-set-of-expressions) algorithm.
 
-A [*natural element type*](#natural-element-type) `T` is first determined.  If that cannot be determined, the literal has no *natural type*.  If `T` can be determined and it is some `KeyValuePair<TKey, TValue>`, then the *natural type* of the collection is `Dictionary<TKey, TValue>`; otherwise, the *natural type* of the collection is `List<T>`.
+A *natural element type* `T` is first determined from the *best common type* of:
+* The *type* of each element `e_n`, and
+* The *iteration type* of each element `..s_n`.
```
CyrusNajmabadi, line 215: i just realized (and this may be ok), that this means the following won't be legal:

```c#
var  v = ["foo", .. x ? [null] : []];
```

I would be fine seeing if this is relevant in practice and if we need to potentially refine how spread interacts with this algorithm.  Specifically if it has no iteration type, we might say that it contributes nothing.  Then the T is figured out, then the T helps target type the spread.  This may be getting too complex and in the weeds though about a non-relevant scenario in practice.

proposals/collection-literals.md
```diff
@@ -207,57 +203,37 @@ Through the use of the [`init`](#init-methods) modifier, existing APIs can direc
...
 
-If the *natural type* of the collection is `List<T>` the literal is not allowed to contain a `dictionary_element`.
+If a *natural element type* `T` cannot be determined, the literal has no *natural type*.  If `T` can be determined, the *natural type* of the collection is `List<T>`.
 
-This means there is no way for a literal to have a *natural type* of some `List<KeyValuePair<TKey, TValue>>` (though it certainly can be *target-typed* to that type).
```
CyrusNajmabadi, line 222: we likely need text somewhere that carves this case out to ensure no breaks when dictionaries are added.

proposals/collection-literals.md
```diff
@@ -207,57 +203,37 @@ Through the use of the [`init`](#init-methods) modifier, existing APIs can direc
...
+_Should `IEnumerable` contribute a natural element type of `object` or no contribution?_
+
+The choice of `List<T>` rather than `T[]` or `ImmutableArray<T>` is to allow mutation of `var` locals after initialization. `List<T>` is preferred over `Span<T>` because `Span<T>` cannot be used in `async` methods.
```
CyrusNajmabadi, line 221:
```suggestion
The choice of `List<T>` rather than `T[]` or `ImmutableArray<T>` is to allow mutation of `var` locals after initialization. `List<T>` is preferred over `Span<T>` because `Span<T>` cannot be used in `async` methods.
```

proposals/collection-literals.md
```diff
@@ -207,57 +203,37 @@ Through the use of the [`init`](#init-methods) modifier, existing APIs can direc
...
-var y = [1, 2]; // ok: List<int>
-object z = [3]; // ok: List<int>
+var list = [1, 2, 3];
+// ...
+list.Add(4);
 ```
 
-### Natural Element Type
-[natural-element-type]: #natural-element-type
+The compiler is free to use a representation other than the natural type if the difference is not observable.
```
CyrusNajmabadi, line 229: not sure what this means.  Is it that we could make our own subtype of List<T>?  Otherwise, i'm not sure what other representations for the natural type are possible that wouldn't be observable to the user.

CyrusNajmabadi: ignore this. :D

proposals/collection-literals.md
cston: I changed the style for cases that were names of spec concepts rather than C# keywords. I believe that is consistent with the style we've used in proposals.

proposals/collection-literals.md
CyrusNajmabadi: primarily for the scenario of:

```c#
var d3 = [..d1, ..d2];
```

Which teh working group found compelling.  The feeling was that being able to merge dictionaries, with "last one wins" semantics is both reasonable and desirable, so we'd prefer 'lenient overwrite' semantics, versus 'restrictive throw' semantics.

proposals/collection-literals.md
```diff
@@ -555,6 +513,107 @@ Not having a *known length* does not prevent any result from being created. Howe
...
+
+```c#
+var d = [x:x1, y:y1]; // {{x, x1}, {y, y1}}
+d = [..d, y:y2];      // {{x, x1}, {y, y2}}
+d = [x:x2, ..d];      // {{x, x2}, {y, y2}}
+```
+
+The [*natural type*](#natural-type) is updated as follows.
+
+If the [*natural element type*](#natural-element-type-updated) `T` can be determined and it is some `KeyValuePair<TKey, TValue>`, then the *natural type* of the collection is `Dictionary<TKey, TValue>`; otherwise, the *natural type* of the collection is `List<T>`.
```
CyrusNajmabadi, line 538: ok. i'm fine with this, though we likely should have a note in the higher up natural type section that mentions that the algorithm will be tweaked later.

CyrusNajmabadi (approved):

proposals/collection-literals.md
cston: Added a note that the `Construct` method is used even if the type a _collection initializer_ type.

I don't think there's an ordering issue with `T[]` and `IList<T>` though because this list represents target types, where for `IList<T>`, the target type must be `IList<T>` rather than an implementation of `IList<T>`.

CyrusNajmabadi: good point.  thanks!

proposals/collection-literals.md
CyrusNajmabadi: yes.  that should def happen.  good call out and need for tests

Issue #6010, opened 2022-04-12
[Proposal]: ref readonly method parameters


Sergio0694:
# `ref readonly` method parameters

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

> **NOTE:** this proposal is a follow up to numerous conversations over the years about how `in` arguments are implemented in C#, both here on GitHub and on Discord (with folks like @stephentoub, @jaredpar, @tannergooding, @AaronRobinsonMSFT). To give context and cite the last example, see https://github.com/dotnet/runtime/issues/67445. Opening this so we can properly review it and hopefully start making some progress.

## Overview

C# 7.2 introduced `in` parameters ([see here](https://docs.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-7.2/readonly-ref)), which allow callers to pass values by references that cannot be mutated. Additionally, values can be passed by reference transparently on the caller side (using `in` is optional), and **allow passing rvalue-s**. This can be convenient in cases where this is fine from the API standpoint, so that callers don't have to worry about this. But, this introduces issues for APIs that would like to disallow rvalue-s where they might indicate a misuse and result in incorrect or unexpected behavior. Consider for example the new constructor on `ReadOnlySpan<T>` is going to have the form: 

```csharp
public ReadOnlySpan(in T value); 
```

This allows for the following code to be written:

```csharp
var span = new ReadOnlySpan<int>(42);
```

This code is correct in the sense that it compiles, is safe and fully covered by the span safety rules. Yet it’s potentially problematic because the API is designed to abstract over a readonly l-value but there is no syntax which allows the developer to express that. 

This proposal is about introducing a new `ref readonly` syntax for method parameters (as suggested by Jared [here](https://github.com/dotnet/runtime/issues/67445#issuecomment-1086180039)), which would behave as follows:
- A `ref readonly` method parameter results in a warning if callers pass an rvalue.
- A `ref readonly` method parameter can be passed implicitly, by `in`, or by `ref`.
- A `ref readonly` method parameter will otherwise behave exactly like an `in` parameter. Notably, a `ref readonly` method parameter will also carry the same `modreq` as `in` when used in a virtual position, just like `in` does today.

[...]

Issue #1308, opened 2018-02-05
`in` usable as `ref` in older compiler versions causes source break if recompiled with v7.2


tannergooding:
This bug actually requires compiling an assembly with v7.2+ of the compiler, referencing the created assembly from v7.1 or prior, and then moving that assembly forward to v7.2+.

**Repro**
1. Using v7.2+ of the compiler, create a new class library
2. Add both a regular and an `in` overload for a method
```C#
public class C
{
    public void M(int i)
    {
    }

    public void M(in int i)
    {
    }
}
```
3. Create another class library and ensure it uses a v7.1 or prior compiler
  A. Downgrading the language version of the compiler will not reproduce this bug
  B. Referencing the Microsoft.Net.Compilers v2.4.0 NuGet package is probably the easiest way
4. Set the code of the library to the following
```C#
public class Test
{
    public void Scenario()
    {
        int i = 0;
        C c = new C();

        c.M(i);
        c.M(ref i);
    }
}
```
5. Observe that the code compiles fine
  A. The first call, `c.M(i)`, calls the `void M(int)` overload
  B. The second call, `c.M(ref i)`, calls the `void M(in int)` overload
6. Move the class library forward to v7.2 or higher
7. Observe that the code will not compile
  A. `CS0121 - The call is ambiguous between the following methods or properties: 'C.M(int)' and 'C.M(in int)'`
  B. `CS1615 - Argument 1 may not be passed with the 'ref' keyword`

tannergooding:
FYI. @VSadov, @jcouv, @jaredpar

tannergooding:
This repro's going all the way back to the v2.0 of the native compiler (non-Roslyn). I assume it also repros on v1.1 and v1.0 of the native compiler, but I don't have those installed locally.

The first issue is already resolved for 7.3, as the `byval` overload will win by default and users will have to specify `in` to call the other overload.

The second issue is much more complicated.

VSadov:
‘in’ parameters are not always “poisoned”. That was a deliberate choice so that libraries and callers could be updated in stages. There is always a workaround.

In the case “B”, once caller switches to 7.2, can he use ‘in’ keyword?

tannergooding:
That probably depends. Some users may expect it to continue working without a workaround required.

It is definitely strange, however, that a new language feature is consumable in a lower version of the compiler and in such a way that it is a breaking change later on (especially given how strict breaking changes normally are).

sharwell:
[...]

PR #7165, opened 2023-05-04
Add speclet for `ref readonly` parameters

jjonescz:
- https://github.com/dotnet/csharplang/issues/6010
- https://github.com/dotnet/roslyn/issues/68056

### DIFF
--- a/proposals/ref-readonly-parameters.md
+++ b/proposals/ref-readonly-parameters.md
@@ -116,6 +116,7 @@ This was discussed in [LDM 2022-05-11](https://github.com/dotnet/csharplang/blob
 [to-review]: #pending-ldm-review
 
 Inverse ordering of modifiers (`readonly ref` instead of `ref readonly`) could be allowed.
+This would be inconsistent with how `readonly ref` returns and fields behave (inverse ordering is disallowed or means something different, respectively) and could clash with readonly parameters if implemented in the future.
 
 Errors could be emitted instead of warnings when passing rvalues to `ref readonly` parameters or mismatching callsite annotations and parameter modifiers.
 Similarly, special `modreq` could be used instead of an attribute to ensure `ref readonly` parameters are distinct from `in` parameters on the binary level.
--- a/proposals/ref-readonly-parameters.md
+++ b/proposals/ref-readonly-parameters.md
@@ -55,6 +55,11 @@ In the opposite direction, changing
 No changes in grammar are necessary.
 Specification will be extended to allow `ref readonly` modifiers for parameters with the same rules as specified for `in` parameters in [their proposal](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/proposals/csharp-7.2/readonly-ref.md), except where explicitly changed in this proposal.
 
+Note that even though `ref` argument modifier is allowed for `ref readonly` parameters, nothing changes w.r.t. value kind checks, i.e.,
+- `ref` can only be used with assignable values;
+- to pass readonly references, one has to use the `in` argument modifier instead;
+- to pass rvalues, one has to use no modifier (which results in a warning for `ref readonly` parameters as described in [the summary of this proposal][summary]).
+
 ### Interchangeability with `in`
 [interchangeability]: #interchangeability-with-in
 
--- a/proposals/ref-readonly-parameters.md
+++ b/proposals/ref-readonly-parameters.md
@@ -115,6 +115,8 @@ This was discussed in [LDM 2022-05-11](https://github.com/dotnet/csharplang/blob
 ### Pending LDM review
 [to-review]: #pending-ldm-review
 
+Inverse ordering of modifiers (`readonly ref` instead of `ref readonly`) could be allowed.
+
 Errors could be emitted instead of warnings when passing rvalues to `ref readonly` parameters or mismatching callsite annotations and parameter modifiers.
 Similarly, special `modreq` could be used instead of an attribute to ensure `ref readonly` parameters are distinct from `in` parameters on the binary level.
 This would provide stronger guarantees, so it would be good for new APIs, but prevent adoption in existing runtime APIs which cannot introduce breaking changes.
--- a/proposals/ref-readonly-parameters.md
+++ b/proposals/ref-readonly-parameters.md
@@ -1,0 +1,104 @@
+# `ref readonly` parameters
+
+## Summary
+[summary]: #summary
+
+Allow parameter declaration-site modifier `ref readonly` and change call-site rules as follows:
+
+| Call site annotation | `ref` parameter | `ref readonly` parameter | `in` parameter |
+|----------------------|-----------------|--------------------------|----------------|
+| `ref`                | Allowed         | Allowed                  | **Warning**    |
+| `in`                 | Error           | Allowed                  | Allowed        |
+| No annotation        | Error           | Warning                  | Allowed        |
+
+- `ref` and `ref readonly` parameters require lvalues at the call site.
+- `in` allows lvalues or rvalues at the callsite.
+
+Note that there is one change (in bold) to the existing rules which are currently as follows:
+
+| Call site annotation | `ref` parameter | `in` parameter |
+|----------------------|-----------------|----------------|
+| `ref`                | Allowed         | **Error**      |
+| `in`                 | Error           | Allowed        |
+| No annotation        | Error           | Allowed        |
+
+## Motivation
+[motivation]: #motivation
+
+C# 7.2 [introduced `in` parameters](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/proposals/csharp-7.2/readonly-ref.md#solution-in-parameters) as a way to pass readonly references.
+`in` parameters allow both lvalues and rvalues and can be used without any annotation at the callsite.
+However, APIs which capture or return references from their parameters would like to disallow rvalues and also enforce some indication at the callsite that a reference is being captured.
+`ref readonly` parameters are ideal in such cases as they allow only lvalues and warn if used without any annotation.
+
+Furthermore, there are APIs that need only read-only references but use
+- `ref` parameters since they were introduced before `in` became available and changing to `in` would be a source and binary breaking change, e.g., `QueryInterface`, or
+- `in` parameters to accept readonly references even though passing rvalues to them doesn't really make sense, e.g., `ReadOnlySpan<T>..ctor(in T value)`, or
+- `ref` parameters to disallow rvalues even though they don't mutate the passed reference, e.g., `Unsafe.IsNullRef`.
+
+These APIs could migrate to `ref readonly` parameters without breaking users.
+For details on binary compatibility, see the proposed [metadata encoding](#metadata).
+Specifically, changing
+- `ref` → `ref readonly` would only be a binary breaking change for virtual methods,
+- `ref` → `in` would also be a binary breaking change for virtual methods, but not a source breaking change (because the rules change to only warn for `ref` arguments passed to `in` parameters),
+- `in` → `ref readonly` would not be a breaking change (but no callsite annotation would result in a warning).
+
+In the opposite direction, changing
+- `ref readonly` → `ref` would be potentially a source breaking change (unless only `ref` callsite annotation was used), and a binary breaking change for virtual methods,
+- `ref readonly` → `in` would not be a breaking change (but `ref` callsite annotation would result in a warning).
+
+## Detailed design
+[design]: #detailed-design
+
+No changes in grammar are necessary.
+Specification will be extended to allow `ref readonly` modifiers for parameters with the same rules as specified for `in` parameters in [their proposal](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/proposals/csharp-7.2/readonly-ref.md), except where explicitly changed in this proposal.
+
+### Metadata encoding
+[metadata]: #metadata-encoding
+
+As a reminder,
+- `ref` parameters are emitted as plain byref types (`T&` in IL),
+- `in` parameters are like `ref` plus they are annotated with `System.Runtime.CompilerServices.IsReadOnlyAttribute`.
+  In C# 7.3 and later, they are also emitted with `[in]` and if virtual, `modreq(System.Runtime.InteropServices.InAttribute)`.
+
+`ref readonly` parameters will be emitted as `[in] T&`, plus annotated with the following attribute:
+
+```cs
+namespace System.Runtime.CompilerServices
+{
+    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+    public sealed class RequiresLocationAttribute : Attribute
+    {
+    }
+}
+```
+
+Furthermore, if virtual, they will be emitted with `modreq(System.Runtime.InteropServices.InAttribute)` to ensure binary compatibility with `in` parameters.
+Note that unlike `in` parameters, no `[IsReadOnly]` is emitted for `ref readonly` parameters to avoid increasing metadata size.
+
+The `RequiresLocationAttribute` will be matched by namespace-qualified name and synthesized by the compiler if not already included in the compilation.
+
+## Alternatives
+[alternatives]: #alternatives
+
+API authors could annotate `in` parameters designed to accept only lvalues with a custom attribute and provide an analyzer to flag incorrect usages.
+This would not allow API authors to change signatures of existing APIs that opted to use `ref` parameters to disallow rvalues.
+Callers of such APIs would still need to perform extra work to get a `ref` if they have access only to a `ref readonly` variable.
+Changing these APIs from `ref` to `[RequiresLocation] in` would be a source breaking change (and in case of virtual methods, also a binary breaking change).
+
+Instead of allowing the modifier `ref readonly`, the compiler could recognize when a special attribute (like `[RequiresLocation]`) is applied to a parameter.
+This was discussed in [LDM 2022-04-25](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/meetings/2022/LDM-2022-04-25.md#ref-readonly-method-parameters), deciding this is a language feature, not an analyzer, so it should look like one.
+
+Passing lvalues without any modifiers to a `ref readonly` parameters could be permitted without any warnings, similarly to C++'s implicit byref parameters.
+This was discussed in [LDM 2022-05-11](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/meetings/2022/LDM-2022-05-11.md#ref-readonly-method-parameters), noting that the primary motivation for `ref readonly` parameters are APIs which capture or return references from these parameters, so marker of some kind is a good thing.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Allow default values for `ref readonly` parameters? They are allowed for `in` parameters, but `ref readonly` parameters shouldn't be used for rvalues (that's normally a warning).
+- Disallow temps for both `in` and `ref readonly` parameters? See https://github.com/dotnet/roslyn/pull/67955#discussion_r1178138561.
+
+## Design meetings
+
+- [LDM 2022-04-25](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/meetings/2022/LDM-2022-04-25.md#ref-readonly-method-parameters): feature accepted
+- [LDM 2022-05-09](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/meetings/2022/LDM-2022-05-09.md#ref-readonly-parameters): discussion split into three parts
+- [LDM 2022-05-11](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/meetings/2022/LDM-2022-05-11.md#ref-readonly-method-parameters): allowed `ref` and no callsite annotation for `ref readonly` parameters
--- a/proposals/ref-readonly-parameters.md
+++ b/proposals/ref-readonly-parameters.md
@@ -93,6 +93,11 @@ The `RequiresLocationAttribute` will be matched by namespace-qualified name and 
 In function pointers, `in` parameters are emitted with `modreq(System.Runtime.InteropServices.InAttribute)` (see [function pointers proposal](https://github.com/dotnet/csharplang/blob/0376b4cc500b1370da86d26be634c9acf9d60b71/proposals/csharp-9.0/function-pointers.md#metadata-representation-of-in-out-and-ref-readonly-parameters-and-return-types)).
 `ref readonly` parameters will be emitted in the same way to ensure binary compatibility.
 
+### Default parameter values
+[default-params]: #default-parameter-values
+
+Default parameter values are allowed for `ref readonly` parameters with a warning since they are equivalent to passing rvalues.
+
 ## Alternatives
 [alternatives]: #alternatives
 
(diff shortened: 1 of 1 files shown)


proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,104 @@
...
+- `ref` and `ref readonly` parameters require lvalues at the call site.
+- `in` allows lvalues or rvalues at the callsite.
+
+Note that there is one change (in bold) to the existing rules which are currently as follows:
+
+| Call site annotation | `ref` parameter | `in` parameter |
+|----------------------|-----------------|----------------|
+| `ref`                | Allowed         | **Error**      |
+| `in`                 | Error           | Allowed        |
+| No annotation        | Error           | Allowed        |
```
jaredpar, line 23: Need to cover how `ref readony` impacts implements, override and inherits scenarios. Essentially is the following legal or not?

```csharp
class C1
{
    protected virtual void M(in int i) { }
}

class C2 : C1
{
    protected override void M(ref readonly int i) { }
}
```

I do not believe we covered this in LDM. My recommendation is that the it be legal for `in` and `ref readonly` to be interchangable in these scenarios. Otherwise it cuts against the motivation of the proposal here.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,104 @@
...
+- `ref` parameters are emitted as plain byref types (`T&` in IL),
+- `in` parameters are like `ref` plus they are annotated with `System.Runtime.CompilerServices.IsReadOnlyAttribute`.
+  In C# 7.3 and later, they are also emitted with `[in]` and if virtual, `modreq(System.Runtime.InteropServices.InAttribute)`.
+
+`ref readonly` parameters will be emitted as `[in] T&`, plus annotated with the following attribute:
+
+```cs
+namespace System.Runtime.CompilerServices
+{
+    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
```
jaredpar, line 68: Should be allow multiple false as well.

jaredpar (approved):

proposals/ref-readonly-parameters.md
hamarb123: Here's my 2 cents: `in` allows both lvalues and rvalues, whereas `ref readonly` only allows lvalues (without warning). I would think that if you override a function which has an `in` parameter, with one that has a `ref readonly` parameter, you should get a warning, since you're saying "I don't want this kind of input", in a way that is invisible to anyone using the base class. I do think either should be overridable in either way though.

Effectively what I'm suggesting:
```csharp
class C1
{
    protected virtual void M(in int i) { }
}

class C2 : C1
{
    //warning (at definition of function) as some inputs (rvalues) are not intended for this function, make sure that the person defining this function knows that people can still give rvalues anyway (without a warning) through the base class.
    protected override void M(ref readonly int i) { }
}

class C3 : C2
{
    //no warning since this expands the possible options for input from just lvalues, to lvalues or rvalues.
    protected override void M(in int i) { }
}
```

proposals/ref-readonly-parameters.md
Sergio0694, line 68: I don't think we mentioned this, but should this also have `[EditorBrowsable(EditorBrowsableState.Never)]`? Eg. `IsReadOnlyAttribute` has it. It would seem the two would basically be equivalent in use? Ie. compiler only.

Sergio0694: Just checked and eg. `[RefSafetyRules]` doesn't have it either, so I take it it's not needed. Nvm 😄

hamarb123: `IsReadOnlyAttribute` does have it:
image
image
image

I don't see why it has it honestly, but if it has it, then `RequiresLocationAttribute` probably should too.

hamarb123 (changes requested):
(Previously had this as a comment, deleted that and put it as a review so it can be replied to easier, and marked as resolved when done - edit: I thought this would achieve something 😅)
How does this interact with function pointers?
It would be nice if we could have `ref readonly` instead of `in` on function pointers, not just actual methods.
e.g. `delegate*<ref int, in int, out int, void>` currently gets emitted as `method void *(int32&, int32& modreq([System.Runtime]System.Runtime.InteropServices.InAttribute), int32& modreq([System.Runtime]System.Runtime.InteropServices.OutAttribute))`.
Presumably we could emit a `ref readonly int` input parameter as `int32& modreq([System.Runtime]System.Runtime.InteropServices.InAttribute) modreq([System.Runtime]System.Runtime.CompilerServices.RequiresLocationAttribute)` (or a modopt).
If we allowed this, we'd probably want to allow `&Method` which has `ref readonly` parameters to be implicitly castable to the same with `in`, and vice versa, to avoid source breaking changes.
Otherwise, we'd have to disallow `ref readonly` for function pointer parameters, or not actually encode it in binary, neither of which are ideal imo.

jaredpar:
@hamarb123 

> How does this interact with function pointers?

Function pointers, method group conversions, method override and interface implementation generally fall into the same categories. The behavior in one generally follows to the other. My expectation is that whatever decisions comes from the override case will apply to the others.

hamarb123:
@jaredpar Function pointers already encode `in` and `out` differently to normal methods; `in` has an extra modreq (which is only present when virtual normally), and out has an extra modreq that is never present normally. Imo function pointers with `ref readonly` parameters warrant a mention in the spec as to how it's encoded, and what the expected behaviour is.

proposals/ref-readonly-parameters.md
jjonescz: Seems that other compiler-synthesized attributes like `ScopedRefAttribute` or `RefSafetyRulesAttribute` don't have it, not sure why `IsReadOnlyAttribute` has it, but it looks like an exception rather than the rule.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+| `ref`                | Allowed         | **Allowed**              | **Warning**    | Error           |
+| `in`                 | Error           | **Allowed**              | Allowed        | Error           |
+| `out`                | Error           | **Error**                | Error          | Allowed         |
+| No annotation        | Error           | **Warning**              | Allowed        | Error           |
+
+Note that there is one change to the existing rules: `in` parameter with `ref` callsite annotation produces a warning instead of an error.
+
+| Value kind | `ref` parameter | `ref readonly` parameter | `in` parameter | `out` parameter |
+|------------|-----------------|--------------------------|----------------|-----------------|
+| rvalue     | Error           | **Warning**              | Allowed        | Error           |
```
AlekseyTs, line 19: A **Warning** for an rvalue passed for a `ref readonly` parameter sounds really surprising to me. I thought `ref readonly` parameter requires an lvalue as an argument.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+
+Furthermore, there are APIs that need only read-only references but use
+- `ref` parameters since they were introduced before `in` became available and changing to `in` would be a source and binary breaking change, e.g., `QueryInterface`, or
+- `in` parameters to accept readonly references even though passing rvalues to them doesn't really make sense, e.g., `ReadOnlySpan<T>..ctor(in T value)`, or
+- `ref` parameters to disallow rvalues even though they don't mutate the passed reference, e.g., `Unsafe.IsNullRef`.
+
+These APIs could migrate to `ref readonly` parameters without breaking users.
+For details on binary compatibility, see the proposed [metadata encoding](#metadata).
+Specifically, changing
+- `ref` → `ref readonly` would only be a binary breaking change for virtual methods,
```
AlekseyTs, line 38: I do not think this is the right design choice. I think changing to `ref readonly` must be a binary breaking change regardless of virtual-ness. Languages that do not understand the semantics around `ref readonly` should be prevented form consuming APIs that use `ref readonly`. For example, it is common for VB to pass ByRef parameters via a temp with copy-back, which would violate the desired semantics around `ref readonly`.

AlekseyTs: The same argument can be made for old versions of compilers (including C# compiler).

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+## Detailed design
+[design]: #detailed-design
+
+No changes in grammar are necessary.
+Specification will be extended to allow `ref readonly` modifiers for parameters with the same rules as specified for `in` parameters in [their proposal](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/proposals/csharp-7.2/readonly-ref.md), except where explicitly changed in this proposal.
+
+### Interchangeability with `in`
+
+Members declared in a single type cannot differ in signature solely by `ref`/`out`/`in`/`ref readonly`.
+For other purposes of signature matching (e.g., hiding or overriding), `ref readonly` can be interchanged with `in` modifier, but that results in a warning at the declaration site [[§7.6](https://github.com/dotnet/csharpstandard/blob/47912d4fdae2bb8c3750e6485bdc6509560ec6bf/standard/basic-concepts.md#76-signatures-and-overloading)].
```
AlekseyTs, line 55: Similarly, here. In my opinion, for the feature to be really useful and to actually provide guarantees to API authors, the interchangeability between `in` and `ref readonly` in signatures should not be allowed.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Metadata encoding
+[metadata]: #metadata-encoding
+
+As a reminder,
+- `ref` parameters are emitted as plain byref types (`T&` in IL),
+- `in` parameters are like `ref` plus they are annotated with `System.Runtime.CompilerServices.IsReadOnlyAttribute`.
+  In C# 7.3 and later, they are also emitted with `[in]` and if virtual, `modreq(System.Runtime.InteropServices.InAttribute)`.
+
+`ref readonly` parameters will be emitted as `[in] T&`, plus annotated with the following attribute:
```
AlekseyTs, line 66: Following my vision for the feature, I think `ref readonly` parameter should always be decorated with a modreq, which is should be different from the one used for `in` parameters

proposals/ref-readonly-parameters.md
AlekseyTs: > This is part of the changes required to allow updating existing signatures in a way that makes more sense, without breaking callers.

I think (and this is just my opinion) this goal renders the feature completely useless for API authors as it doesn't provide for them any real new guarantees. It makes the feature almost useless for consumers as well because warnings can be easily overlooked or quite possibly un-intentionally disabled, or their significance can be misinterpreted. It is quite possible I want a different feature.  

For example, given the current proposal, the following proposed API change doesn't make sense for me:
``` diff
-       public static ReadOnlySpan<T> CreateReadOnlySpan<T>(scoped ref T reference, int length);
+       public static ReadOnlySpan<T> CreateReadOnlySpan<T>(scoped ref readonly T reference, int length);
```

A consumer is now allowed to pass a reference to a clone as the first argument, however, if I was an author of the API, and given its purpose, I would want a situation like that to be disallowed (i.e. impossible). In fact it is disallowed today, because the parameter is a `ref` parameter.

tannergooding: There is always contention between the optimal and practical.

For the majority case, it is undesirable to allow passing non location values in and it would be more desirable to error always.

For APIs that are taking `ref` today, the switch to `ref readonly` will be seamless and neither source nor binary breaking.

For APIs that are taking `in` but where they intended to take `ref readonly` and the feature simply didn't exist, then they have no non-breaking path forward as it will result in a source break due to the error.

We then have a few possible paths to handle this, such as implementing a second feature to cover this scenario or slightly relaxing the new language feature to warn instead of error.

The latter seems the lesser of the two evils and avoids needing to duplicate all the same handling the compiler is already doing for `ref readonly` just to raise a different diagnostic kind.

proposals/ref-readonly-parameters.md
tannergooding: The feature is effectively dead on arrival if it is a breaking change.

The BCL has numerous existing APIs that are desired to switch from `ref` to `ref readonly` and we cannot break the public surface area.

proposals/ref-readonly-parameters.md
tannergooding: This remains a consideration for back-compat. We have 20 years of existing code taking `ref` and 3-5 years of code taking `in` that would prefer to switch to newer features but can't due to it being a breaking change.

`ref->in` is currently source breaking because `ref x` for an `in T` is invalid syntax. Allowing `ref x` to be used for `in T` avoids this break and fixes a weird quirk of an older compiler consuming metadata emitted by a newer compiler (https://github.com/dotnet/csharplang/issues/1308).

Likewise, the BCL has rejected using `in` for several APIs due to the hidden copies it allows and the high chance for accidental incorrect API usage in this scenario. A warning for such hidden copies with the `ref readonly` feature is considered sufficient to allow the transition and allows APIs (including APIs currently taking `in`) to transition without breaking users.

proposals/ref-readonly-parameters.md
AlekseyTs: > The feature is effectively dead on arrival if it is a breaking change.
 
I do not think so

>The BCL has numerous existing (non-virtual) APIs that are desired to switch from ref to ref readonly and we cannot break the existing public surface area.

The we shouldn't switch them.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
+# `ref readonly` parameters
+
+## Summary
+[summary]: #summary
+
+Allow parameter declaration-site modifier `ref readonly` and change callsite rules as follows:
```
CyrusNajmabadi, line 6: trying to wrap my head around this. I though `in` was literally our shorthand for `ref readonly`.  We even have that in our own code:

image

Are we consdiring making it so that `in` vs `ref readonly` on a parameter would have different semantics?

proposals/ref-readonly-parameters.md
tannergooding: > Then we shouldn't switch them.

This feature is largely driven by a desire to do this switch. If the BCL can't make the switch then one of the biggest reasons to do the feature goes away and we would remain in a world where some of the most used APIs that would benefit from this can't use them, where users still have to manually workaround it using `ref T Unsafe.As(in T value)`.

tannergooding: I believe it is a core requirement for the language to consider the 20+ years of existing code when introducing new features. If it doesn't then a large amount of value ends up being lost because the feature cannot correctly integrate and be used with the existing code that cannot make breaking changes due to back-compat.

CyrusNajmabadi: Personally, i'm struggling to see the value here if it woudl cause breaks.  It would likely need a host of new APIs introduced (and potential deprecation/hiding of the old APIs) and feels like it would boil-the-ocean in terms of getting people moved.  It feels more natural that this naturally work in a binary fashion, with warnings on recompile to get the people doing the rare-odd thing to move over.  If people miss the warnings, that's ok.  It's not the end of the world.  Their code was already busted before, so it's no worse.  But people that are paying attention can and who do care can gracefully move forward.

AlekseyTs:
I would like to address a concern that some how the feature is "dead on arrival" if the existing APIs cannot simply use the new nice syntax without introducing a binary breaking change. 

I do recognize that there is a set of existing APIs that is relatively hard to consume today because C# doesn't allow passing readonly lvalues to them. The APIs have this shape because the authors would never want a reference to an rvalue to be passed as an argument, the `in` parameter doesn't provide this guarantee, so a `ref` parameter was used. At the time `in` was introduced, I think we should have added `ref readonly` with the meaning - I want "strict" Lvalues, but readonly lvalues are Ok. Unfortunately, we haven't. And here we are.

On the other hand, since we are proposing to spend a nice language syntax, which is a **very very scarce** "resource", we should spend it very wisely. And I would say, we should spend it on the design that we would arrive to at the time we designed `in` modifier.
In my opinion, what is being proposed is not that design. The proposed design feels to me as a band-aid for existing APIs. It doesn't go beyond that in my opinion, and, in fact, if we switch existing APIs to use the band-aid, some of them would lose the lvalue guarantees that they have today. Spending a language syntax on a band-aid, is extremely unwise.

I think that the issue around consumption of the existing APIs can be addressed slightly differently. For example, the affected ref parameters could be decorated with an special attribute (quite possibly we will be able to reuse some existing attribute), which simply instructs the compiler to let readonly lvalues through because the method guarantees to not write through the reference. Support for this attribute can be implemented under umbrella of this feature as well. The issue with consumption of the APIs is addressed, there is no back-compatibility issue and we can use language syntax for a nice, clean language design with strong very useful guarantees to use for anyone on new APIs.

AlekseyTs:
My point is that the desire to transition the existing APIs without binary breaking change should not be the main driving factor for the design. Let's figure out how do we want the feature to work. Then we can evaluate if that helps addressing issues around existing APIs, and alternatives for cases that aren't addressed.

And, BTW, a warning doesn't guarantee anything, in my opinion.

tannergooding:
> My point is that the desire to transition the existing APIs without binary breaking change should not be the main driving factor for the design. Let's figure out how do we want the feature to work. Then we can evaluate if that helps addressing issues around existing APIs, and alternatives for cases that aren't addressed.

I understand and the primary interested parties heavily weighed in on the initial design proposal (https://github.com/dotnet/csharplang/issues/6010) which has already gone to LDM and had many of these points raised/discussed.

This comes across as rehashing the same points that were already debated fairly extensively.

> And, BTW, a warning doesn't guarantee anything, in my opinion.

Different users have different opinions. It isn't a strong guarantee but for most users they are more than sufficient and the more common scenario to be encountered. This is particularly the case for those that have `WarnAsError`, that turn on `Warning Waves`, that use `Analyzers` (where "errors" can be disabled), etc.

Most things we surface to users about "you might be doing something wrong" are warnings and this is really no different. It's why we had the discussion around whether this could be an analyzer with the reason why not largely being that a language feature was needed anyways (to allow safely changing from `ref` -> `in`) and that an analyzer identifying the hidden copies inserted by the compiler is itself incredibly difficult/error prone and thus the compiler handling it is likely better.

AlekseyTs:
> Different users have different opinions.

Yes, that is the case. And my comments are not blocking for this PR. However, I think it is worth brining this to LDM as soon as possible to confirm the course.

proposals/ref-readonly-parameters.md
hamarb123: You probably need to add a `RefReadOnlyInput = 4`  or something. The `RefReadOnly` one at least specifics in the comments that it is only for returns, let's hope people read that :)

hamarb123:
As a likely user of this feature, if it were made a breaking change from `in` or `ref`, then it would be dead-on-arrival to me, and would take a long time to recover from the death, since I would only be able to use it on new APIs.

proposals/ref-readonly-parameters.md
jaredpar: > I think changing to ref readonly must be a binary breaking change regardless of virtual-ness.

That would defeat the purpose of this feature. The underlying motivation for this feature is allowing customers, particularly the dotnet/runtime team, the ability to migrate existing APIs which take `ref`, and to a lesser extent `in`, to a more correct API form of a non-mutating `ref`. If there is a binary compat break then that prevents existing APIs from moving over and hence it ceases it solve the underlying problem. Don't think there would be much appetite for that version of the feature.

> I believe there is a core need for the language to consider the 20+ years of existing code when introducing new features

Agree. At the time we designed `in` parameters we weren't as focused on adoption struggles as we should've been. It was also at the very start of the `Span<T>` and `readonly` movement so it was hard to forsee how much we'd move in that direction. If we were doing this feature today there are a number of tweaks in the design I believe we would consider.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+
+Furthermore, if virtual, they will be emitted with `modreq(System.Runtime.InteropServices.InAttribute)` to ensure binary compatibility with `in` parameters.
+Note that unlike `in` parameters, no `[IsReadOnly]` is emitted for `ref readonly` parameters to avoid increasing metadata size.
+
+The `RequiresLocationAttribute` will be matched by namespace-qualified name and synthesized by the compiler if not already included in the compilation.
+
+#### Function pointers
+
+In function pointers, `in` parameters are emitted with `modreq(System.Runtime.InteropServices.InAttribute)`.
+`ref readonly` parameters will be emitted in the same way to ensure binary compatibility.
```
hamarb123, line 86: Do we really want this behaviour?
Consider these 2 different options:

1. The one you have here:
This would mean that if I had it in a declaration that was used by another library, it would have to be treated as `in` anyway. A specific example:
```C#
//Library1
public class Class1
{
  public static delegate*<in int, ref readonly int, void> Func1();
}
```
This would be emitted such that it looks like `delegate*<in int, in int, void>`, therefore in Library 2 we would be able to pass an rvalue with no warning to the second parameter.

2. Consider encoding `ref readonly` with an additional `modreq`. This would allow the calling library to see the correct parameter type on the above example. It does mean that changing from `in` to `ref readonly` in a function pointer type is a breaking change, but the alternative is that it does nothing and gets thrown away. I personally think a breaking change is more useful than throwing away the information altogether (and by useful, I mean I wouldn't change it for existing APIs, but at least for new APIs it would provide some value).
With this solution, you would want to only warn on code like
```csharp
//method declared somewhere
static void X(ref readonly int i) => ...;

//code
delegate*<in int, void> abc = &X; //warning
delegate*<ref readonly int, void> def = abc; //warning?
delegate*<ref int, void> = abc; //error without explicit casting, just like currently
```
because otherwise it would be a breaking change to change any method.

Benefits of (1): allows people to change from `in` to `ref readonly` without any breaking changes.
Drawbacks of (1): doesn't provide much value, since the information will be lost by any external library looking at it.

The benefits of (2): having `ref readonly` in as a parameter of a function pointer has a meaning to external libraries (otherwise this information is lost and they can pass an rvalue without a warning).
Drawbacks of (2): have to implement implicit casting with warning from something like `delegate*<ref readonly int, void>` to `delegage*<in int, void>` and vice versa to avoid any breaking changes. Potential modreq issues if we don't use a type that has existed for some period of time.

There is technically a third option, which is to emit complicated attributes that the compiler can use to determine which ones are `ref readonly` vs `in`.

I think if we don't go with (2), then there's no point allowing `ref readonly` parameters in function pointers at all, since they would be exactly the same as `in` parameters; meaning we wouldn't even be able to expose this information in the future without complex attribute tricks or breaking changes.

Also note with (2), the modreq would ideally be something that has existed previously, to avoid issues if were were to make the type declared in the runtime (since that would make it use a different assembly), or it could emit a type forward whenever it doesn't declare its own copy of the attribute, or it could leave it to me to fix in my own code 😆.

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+
+These APIs could migrate to `ref readonly` parameters without breaking users.
+For details on binary compatibility, see the proposed [metadata encoding](#metadata).
+Specifically, changing
+- `ref` → `ref readonly` would only be a binary breaking change for virtual methods,
+- `ref` → `in` would also be a binary breaking change for virtual methods, but not a source breaking change (because the rules change to only warn for `ref` arguments passed to `in` parameters),
+- `in` → `ref readonly` would not be a breaking change (but no callsite annotation or rvalue would result in a warning).
+
+In the opposite direction, changing
+- `ref readonly` → `ref` would be potentially a source breaking change (unless only `ref` callsite annotation was used), and a binary breaking change for virtual methods,
```
hamarb123, line 43: It would be good to mention here that it would be a source breaking change also since it would now disallow readonly refs to be passed to the method.

proposals/ref-readonly-parameters.md
hamarb123: Reading through the rest of the thing, (2) also works well with the `Interchangeability with in` section, since it would also allow for warnings when implicitly casting between something like `delegate*<ref readonly int, void>` and `delegage*<in int, void>`.

Specifically this part:
> For the purpose of delegate/lambda/method group conversions [[§10.7](https://github.com/dotnet/csharpstandard/blob/47912d4fdae2bb8c3750e6485bdc6509560ec6bf/standard/conversions.md#107-anonymous-function-conversions), [§10.8](https://github.com/dotnet/csharpstandard/blob/47912d4fdae2bb8c3750e6485bdc6509560ec6bf/standard/conversions.md#108-method-group-conversions)], `ref readonly` can be interchanged with `in` modifier, but that results in a warning.

And honestly, I'm thinking (3) (which is to use a complicated (to me anyway) attribute) would be a better solution than (2), but I have no idea how that would be encoded. Presumably, it could be encoded similarly to how names for value tuples are encoded, since they also have to encode things in relatively arbitrary locations.

hamarb123 (changes requested):
Mostly lgtm, but I think we/LDM needs to discuss behaviour with function pointers more.

proposals/ref-readonly-parameters.md
jjonescz: > in Library 2 we would be able to pass an rvalue with no warning to the second parameter.

Good point. I think we could emit a `modopt` for `ref readonly` parameters. It wouldn't be an error like emitting `modreq`, but could be used to emit warnings.

proposals/ref-readonly-parameters.md
```diff
@@ -115,6 +115,8 @@ This was discussed in [LDM 2022-05-11](https://github.com/dotnet/csharplang/blob
 ### Pending LDM review
 [to-review]: #pending-ldm-review
 
+Inverse ordering of modifiers (`readonly ref` instead of `ref readonly`) could be allowed.
```
hamarb123, line 118: May want to add a note that this could cause issues if we wanted readonly parameters, and would be inconsistent with ref fields and ref returns.

Sergio0694: Yeah I'd also personally prefer if the order here was strict. It'd also match all other uses of `readonly`, where the exact position matters when applied to a `ref`. Eg. you can't have a `readonly readonly ref` method. Or a `readonly readonly ref` field. Similarly, a `readonly ref` return or a `readonly ref` field have a very different meaning than a `ref readonly` return or field. It would seem weird if the same didn't apply to parameters as well 🤔

proposals/ref-readonly-parameters.md
```diff
@@ -0,0 +1,176 @@
...
+- `ref readonly` → `in` would not be a breaking change (but `ref` callsite annotation would result in a warning).
+
+## Detailed design
+[design]: #detailed-design
+
+No changes in grammar are necessary.
+Specification will be extended to allow `ref readonly` modifiers for parameters with the same rules as specified for `in` parameters in [their proposal](https://github.com/dotnet/csharplang/blob/c8c1615fcad4ca016a97b7260ad497aad53ebc78/proposals/csharp-7.2/readonly-ref.md), except where explicitly changed in this proposal.
+
+Note that even though `ref` argument modifier is allowed for `ref readonly` parameters, nothing changes w.r.t. value kind checks, i.e.,
+- `ref` can only be used with assignable values;
```
hamarb123, line 59: Does this mean I wouldn't be able to do the following? If so, it seems a bit weird:
```csharp
int a = 1;
ref readonly int b = ref a; //note we use ref for this
ref readonly int c = ref b; //and for this

//would the following 2 lines be errors?
X1(ref c);
X2(ref c); //warning here obviously

void X1(ref readonly int x) { }
void X2(in int x) { }
```

I was hoping we would be able to just use `ref` everywhere. It seems a bit odd to disallow `ref` here, since we use `ref` (not `in`) for `ref readonly` locals etc.

hamarb123 (approved):

proposals/ref-readonly-parameters.md
jjonescz: Yes, `X1(ref c)` will be an error no matter whether `X1` is defined with `ref`/`in`/`ref readonly` parameter (the value kind of `c` is checked against the `ref` modifier at the callsite), you will have to use `X1(in c)` when `c` is readonly. The reason is that `ref` at the call site says "passing a mutable reference" whereas `in` says "passing a readonly reference" - `ref readonly` parameters just allow both of them.

Ref assignments are special, they use `ref` always, but it's pretty clear if it's readonly or not, since that's declared right there on the left-hand side.

PR #7179, opened 2023-05-09
Add details to lookup rules for for extensions

jcouv:

### DIFF
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -452,7 +452,7 @@ The *simple_name* with identifier `I` is evaluated and classified as follows:
 
 TL;DR: After doing an unsuccessful member lookup in a type, we'll perform an member lookup
 in the underlying type if we were dealing with an extension, and if that is still unsuccessful,
-we'll perform an extension member lookup.
+we'll perform an extension member lookup for non-invocations or attempt an extension invocation for invocations.
 
 We modify the [member access rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1176-member-access) as follows:
 
@@ -480,11 +480,12 @@ We modify the [member access rules](https://github.com/dotnet/csharpstandard/blo
   - If `I` identifies a constant, then the result is a value, namely the value of that constant.
   - If `I` identifies an enumeration member, then the result is a value, namely the value of that enumeration member.
   - Otherwise, `E.I` is an invalid member reference, and a compile-time error occurs.
+
 - \***If `E` is classified as an extension, 
   and if a member lookup of `I` in underlying type `U` with `K` type parameters produces a match, 
   then `E.I` is evaluated and classified as follows:**  
   ...
-- \***If `E` is classified as a type, if `E` is not a type parameter, 
+- \***If `E.I` is not invoked and `E` is classified as a type, if `E` is not a type parameter, 
   and if an ***extension member lookup*** of `I` in `E` with `K` type parameters produces a match, 
   then `E.I` is evaluated and classified as follows:**  
   ...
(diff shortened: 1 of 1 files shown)


proposals/extensions.md
```diff
@@ -473,58 +493,252 @@ We modify the [member access rules](https://github.com/dotnet/csharpstandard/blo
...
+### Indexer access
+
+TL;DR: If no candidate is applicable, then we attempt extension indexer access instead.
+
+We modify the [element access rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#128111-general) as follows:
+
+/[...]
+An *element_access* is dynamically bound if \[...]
+If the *primary_no_array_creation_expression* of an *element_access* is a value of an *array_type*, the *element_access* is an array access. 
+Otherwise, the *primary_no_array_creation_expression* shall be a variable or value of a class, struct, or interface type
```
jjonescz, line 658: The if/otherwise/otherwise wording doesn't make sense, the first two cases sound like they together already cover everything, should be probably if/else/otherwise.

```suggestion
Else if the *primary_no_array_creation_expression* is a variable or value of a class, struct, or interface type 
```

proposals/extensions.md
```diff
@@ -473,58 +493,252 @@ We modify the [member access rules](https://github.com/dotnet/csharpstandard/blo
...
+
+The binding-time processing of an indexer access of the form `P[A]`, where `P` is a *primary_no_array_creation_expression* of a class, struct, or interface type `T`, and `A` is an *argument_list*, consists of the following steps:
+
+- The set of indexers provided by `T` is constructed. \[...]
+- The set is reduced to those indexers that are applicable and not hidden by other indexers. \[...]
+- \***If the resulting set of candidate indexers is empty, then further processing 
+  along the following steps are abandoned, and instead an attempt is made 
+  to process the indexer access as an extension indexer access. If this fails, 
+  then no applicable indexers exist, and a binding-time error occurs.**
+- If the resulting set of candidate indexers is empty, then no applicable indexers exist, and a binding-time error occurs.
```
jjonescz, line 675: Seems like this step is covered in the previous one; should it be removed?

proposals/extensions.md
```diff
@@ -473,58 +493,252 @@ We modify the [member access rules](https://github.com/dotnet/csharpstandard/blo
...
+a substituted compatible implicit extension type `X` for `Type`
+so that the corresponding element access can take place:
+```csharp
+((X)expr) . «identifier» [ ]
+((X)expr) . «identifier» [ «args» ]
+```
+
+The search proceeds as follows:
+
+- Starting with the closest enclosing type declaration, continuing with each type declaration,
```
jjonescz, line 704:
```suggestion
- Starting with the closest enclosing type declaration, continuing with each enclosing type declaration,
```

proposals/extensions.md
```diff
@@ -535,7 +749,7 @@ For purposes of member lookup, a type `T` is considered to have the following b
 - If `T` is an *interface_type*, the base types of `T` are the base interfaces of `T` and the class type `object`.
 - If `T` is an *array_type*, the base types of `T` are the class types `System.Array` and `object`.
 - If `T` is a *delegate_type*, the base types of `T` are the class types `System.Delegate` and `object`.
-- \***If `T` is an *extension_type*, the base types of `T` are the base extensions of `T`.**
+- \***If `T` is an *extension_type*, the base types of `T` are the base extensions of `T` and the extended type of `T`.**
```
jjonescz, line 752: What about the extended types of the base extensions - they can be different, right?

Also PR https://github.com/dotnet/roslyn/pull/71779 has "and its base types." in the spec quote which isn't here.

proposals/extensions.md
jcouv: This is existing spec language. I agree it's probably not formulated the best...
The changes I'm making are highlighted with `**`. I just provided this as context to show where the insertion goes.

jjonescz: I know, I think it was formulated fine before you added the second "Otherwise" clause - then I would change the existing "Otherwise" to something else.

PR #7182, opened 2023-05-09
Create breaking-change-warnings.md

MadsTorgersen:
Initial proposal based on the idea discussed in #7033. 

Allow very limited breaking changes in C# when this enables significantly simpler feature designs that are easier to learn, understand and use. 

Retroactively add warnings in previous language versions to help identify and fix user code that would be vulnerable to such breaks upon a language version upgrade.

### DIFF
--- a/proposals/breaking-change-warnings.md
+++ b/proposals/breaking-change-warnings.md
@@ -1,0 +1,93 @@
+# Breaking change warnings
+
+* [x] Proposed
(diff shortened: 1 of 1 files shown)


333fred (approved):

RikkiGibson (approved):
I think this version of the proposal is at a level of simplicity where I don't worry about us building stuff we're not gonna need, and I think users will flow through the *expected path* here often enough that we will achieve what we want. I'm looking forward to us trying it out for some upcoming feature which involves a breaking change and seeing how it works out.

PR #7244, opened 2023-06-01
Collection literal conversions

cston:
Describe the types to which a collection literal expression can be implicitly converted.

### DIFF
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -133,7 +133,7 @@ The following collection literal conversions exist from a collection literal:
     * If `Ei` is a *dictionary_element* `Ki:Vi`, the values `Ki` and `Vi`.
     * If `Ei` is a *spread_element* with *iteration type* `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from the *iteration variable*.
 
-* To an *interface type* *`I<K, V>`* where `I<TKey, TValue>` is implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` where for each element `Ei`:
+* To an *interface type* *`I<K, V>`* where `I<TKey, TValue>` is implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` and where for each element `Ei`:
   * If `Ei` is an *expression_element* then the type of `Ei` is `dynamic`, or the type of `Ei` is a _type_ `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
   * If `Ei` is a *dictionary_element* `Ki:Vi` there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
   * If `Ei` is a *spread_element* with *iteration type* `dynamic`, or `System.Collections.Generic.KeyValuePair<Ki, Vi>` where there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
@@ -145,7 +145,7 @@ The following collection literal conversions exist from a collection literal:
     * If `Ei` is an *expression_element*, the expression `Ei`.
     * If `Ei` is a *spread_element*, the enumerated item.
 
-* To an *interface type* *`I<T0>`* where `I<T>` is implemented by `System.Collections.Generic.List<T>` where for each element `Ei`:
+* To an *interface type* *`I<T0>`* where `I<T>` is implemented by `System.Collections.Generic.List<T>` and where for each element `Ei`:
   * If `Ei` is an *expression_element* there is an implicit conversion from `Ei` to `T0`.
   * If `Ei` is a *dictionary_element* `Ki:Vi`, then `T0` is a _type_ `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
   * If `Ei` is a *spread_element* with *iteration type* `Si` there is an implicit conversion from `Si` to `T0`.
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -88,7 +88,7 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
     - Literals with no `spread_element` in them.
     - Literals with arbitrary ordering of any element type.
 
-* In the following sections, examples of literals without a `k: v` element should assumed to not have any `dictionary_element` in them. Any usages of `..s` should be assumed to be a spread of a non-dictionary value.  Sections that refer to dictionary behavior will call that out.
+* In the following sections, examples of literals without a `k: v` element should be assumed to not have any `dictionary_element` in them. Any usages of `..s` should be assumed to be a spread of a non-dictionary value.  Sections that refer to dictionary behavior will call that out.
 
 * The *iteration type* of `..s_n` is the type of the *iteration variable* determined as if `s_n` were used as the expression being iterated over in a [`foreach_statement`](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/statements.md#1295-the-foreach-statement).
 
@@ -98,7 +98,7 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
 
 * The specification defines a [translation](#collection-literal-translation) of the literal to existing C# constructs.  Similar to the [*query expression translation*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11173-query-expression-translation), the literal is itself only legal if the translation would result in legal code.  The purpose of this rule is to avoid having to repeat other rules of the language that are implied (for example, about convertibility of expressions when assigned to storage locations).
 
-* An implementation is not required to translate literals exactly as specified below.  Any translation is legal as long as the same result is produced and there are no observable differences (outside of timing) in the production of the result.
+* An implementation is not required to translate literals exactly as specified below.  Any translation is legal if the same result is produced and there are no observable differences in the production of the result.
 
     * For example, an implementation could translate literals like `[1, 2, 3]` directly to a `new int[] { 1, 2, 3 }` expression that itself bakes the raw data into the assembly, eliding the need for `__index` or a sequence of instructions to assign each value. Importantly, this does mean if any step of the translation might cause an exception at runtime that the program state is still left in the state indicated by the translation.
 
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
 
     * The behavior of collection literals with collections that are not well-behaved is undefined.
 
-## Constructible collection types (target types)
-[constructible-collection-types]: #constructible-collection-types
+## Conversions
+A collection literal expression is classified as a value.
 
-The following are _constructible collection types_ that can be used to construct collection literals.
-These are the valid _target types_ for collection literals.
+A _collection literal conversion_ permits a collection literal expression to be implicitly converted to a type.
 
-The list is in priority order. If a collection type belongs in multiple categories, the first is used.
+The following collection literal conversions exist from a collection literal:
 
-Actual translation of the literal to the corresponding type is defined [below](#collection-literal-translation).
+* To a single dimensional *array type* `T[]`, or a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>`, where for each element `Ei`:
+  * If `Ei` is an *expression_element* there is an implicit conversion from `Ei` to `T`.
+  * If `Ei` is a *dictionary_element* `Ki:Vi`, then `T` is a _type_ `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread_element* with enumerated element type `Si` there is an implicit conversion from `Si` to `T`.
 
-* Single dimensional arrays (e.g. `T[]`)
-* [*Span types*](#span-types)
-* Types with a suitable [`Construct` method](#construct-methods)
-* Types that implement `System.Collections.IDictionary` - *dictionary [collection initializers](#collection-initializers)*
-* Interface types _`I<TKey, TValue>`_ implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` (e.g. `IDictionary<TKey, TValue>`, `IReadOnlyDictionary<TKey, TValue>`)
-* Types that implement `System.Collections.IEnumerable` - [*collection initializers*](#collection-initializers)
-* Interface types _`I<T>`_ implemented by `System.Collections.Generic.List<T>` (e.g. `IEnumerable<T>`, `IList<T>`, `IReadOnlyList<T>`)
+* To a *type* with an associated *[builder](#construct-methods)* where there is an implicit collection literal conversion from the collection literal to the *builder argument type*.
+
+* To a *type* that implements `System.Collections.IDictionary` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter `int capacity`.
+  * For each element `Ei` there is an applicable instance indexer for the *key_value_pair* from:
+    * If `Ei` is an *expression_element* with type `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from `Ei`.
+    * If `Ei` is a *dictionary_element* `Ki:Vi`, the values `Ki` and `Vi`.
+    * If `Ei` is a *spread_element* with enumerated element `Si` with type `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from `Si`.
+
+* To an *interface type* *`I<K, V>`* implemented by `System.Collections.Generic.Dictionary<K, V>` where for each element `Ei`:
+  * If `Ei` is an *expression_element* then the type of `Ei` is `dynamic`, or the type of `Ei` is a _type_ `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *dictionary_element* `Ki:Vi` there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread_element* with enumerated element type `Si` then the type of `Si` is `dynamic`, or the type of `Si` is a _type_ `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+
+* To a *type* that implements `System.Collections.IEnumerable` where for each element `Ei`:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter `int capacity`.
+  * No element `Ei` is a *dictionary_element*.
+  * For each element `Ei` there is an applicable instance or extension method `Add` for the single argument from:
+    * If `Ei` is an *expression_element*, the expression `Ei`.
+    * If `Ei` is a *spread_element*, the enumerated item.
+
+* To an *interface type* *`I<T>`* implemented by `System.Collections.Generic.List<T>` where for each element `Ei`:
+  * If `Ei` is an *expression_element* there is an implicit conversion from `Ei` to `T`.
+  * If `Ei` is a *dictionary_element* `Ki:Vi`, then `T` is a _type_ `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread_element* with enumerated element type `Si` there is an implicit conversion from `Si` to `T`.
+
+Types for which there is an implicit collection literal conversion from a collection literal are the valid *target types* for that collection literal.
+
+## Construction
+_Give specific ordering for determining how to construct the constructible collection types._
 
 ## `Construct` methods
 [construct-methods]: #construct-methods
@@ -141,7 +166,7 @@ While certain types (like arrays and spans) can always be constructed with a col
 
 If found, the collection can be constructed by creating a fresh instance of its type using `new T()`, producing the corresponding argument to pass to `Construct`, and then calling that method on the fresh instance.  `new T()` supports all structs, including those without a `parameterless struct constructor`.
 
-The allowance for extension methods means that collection literal support can be added to a existing API which does not already directly support this.
+The allowance for extension methods means that collection literal support can be added to an existing API which does not already directly support this.
 
 The `Construct` method is used for construction of the collection even if the type supports *collection initializers*.
 _This means an extension method could be added that would silently change how collection literals for an existing collection initializer type are constructed in the program._
@@ -280,12 +305,27 @@ The existing rules for type inference (see [§11.6.3](https://github.com/dotnet/
 >   - **`V` is a single-dimensional array type `V₁[]` and `U` is a collection literal with element type `U₁`**
 >   - `V` is the type `V₁?` and `U` is the type `U₁`
 >   - `V` is a constructed type `C<V₁...Vₑ>` and `U` is a constructed type `C<U₁...Uₑ>`
->   - **`V` is a constructed type `C<V₁>` that is a constructible collection with element type `V₁`, and `U` is a collection literal with element type `U₁`**
+>   - **`V` is a constructible collection type `C<V₁>` with element type `V₁`, and `U` is a collection literal with element type `U₁`**
+>
 >   If any of these cases apply then an *exact inference* is made from each `Uᵢ` to the corresponding `Vᵢ`.
 
-_How do we recognize that `C<V₁>` is a constructible collection with element type `V₁`?_
+_How do we recognize that `C<V₁>` has element type `V₁`?_
 
 _Update to include inference from dictionary types._
+```c#
+var d = ["Alice": 42, "Bob": 43].AsDictionary(comparer);
+
+static Dictionary<TKey, TValue> AsDictionary<TKey, TValue>(
+    this List<KeyValuePair<TKey, TValue>> list,
+    IEqualityComparer<TKey> comparer = null) { ... }
+```
+
+_What are the implications of type inference of spread elements?_
+```c#
+F([..[1, 2]]); // ok: F<int>(int[])
+
+static void F<T>(T[] arg) { }
+```
 
 ### Interaction with natural type
 
@@ -297,13 +337,16 @@ var y = First(new int[0], [1, 2, 3]);    // ok: int[]
 static T First<T>(T x, T y) => x;
 ```
 
-To allow conversions to other collection types, type inference may need to infer exact, lower-bound, and upper-bound inferences from the **expression `U`** to a type `V` when an expression is given, and *fixing* §11.6.3.12 updates the sets of *candidate types* based on implicit conversion from the **expression** bound to the each *candidate type*.
+To allow conversions to other collection types, type inference may need to infer exact, lower-bound, and upper-bound inferences from the **expression `U`** (rather than the type of `U`) to a type `V` when an expression is given, and *fixing* §11.6.3.12 updates the sets of *candidate types* based on implicit conversion from the **expression** bound to the each *candidate type*.
 
 _Should nested cases infer collection types successfully?_
 ```c#
 var x = new[] { [ulong.MaxValue], [1, 2, 3] };  // ok: List<ulong>[]
 var y = First([[ulong.MaxValue]], [[1, 2, 3]]); // ok: List<List<ulong>>
 ```
+
+## Overload resolution
+_Include betterness order - perhaps the following, from best to worst: spans; arrays and constructed types; interface types._
 
 ## Span types
 [span-types]: #span-types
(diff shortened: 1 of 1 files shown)


proposals/collection-literals.md
```diff
@@ -98,7 +98,7 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
 
 * The specification defines a [translation](#collection-literal-translation) of the literal to existing C# constructs.  Similar to the [*query expression translation*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11173-query-expression-translation), the literal is itself only legal if the translation would result in legal code.  The purpose of this rule is to avoid having to repeat other rules of the language that are implied (for example, about convertibility of expressions when assigned to storage locations).
 
-* An implementation is not required to translate literals exactly as specified below.  Any translation is legal as long as the same result is produced and there are no observable differences (outside of timing) in the production of the result.
+* An implementation is not required to translate literals exactly as specified below.  Any translation is legal if the same result is produced and there are no observable differences in the production of the result.
```
CyrusNajmabadi, line 101: isn't 'outside of timing' still useful to say?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+A _collection literal conversion_ permits a collection literal expression to be implicitly converted to a type.
 
-The list is in priority order. If a collection type belongs in multiple categories, the first is used.
+The following collection literal conversions exist from a collection literal:
 
-Actual translation of the literal to the corresponding type is defined [below](#collection-literal-translation).
+* To a single dimensional *array type* `T[]`, or a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>`, where for each element `Ei`:
+  * If `Ei` is an *expression_element* there is an implicit conversion from `Ei` to `T`.
+  * If `Ei` is a *dictionary_element* `Ki:Vi`, then `T` is a _type_ `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread_element* with enumerated element type `Si` there is an implicit conversion from `Si` to `T`.
```
CyrusNajmabadi, line 125: nice and clear.  do we need a link/description of what "enumerated element type" means?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
 
-* Single dimensional arrays (e.g. `T[]`)
-* [*Span types*](#span-types)
-* Types with a suitable [`Construct` method](#construct-methods)
-* Types that implement `System.Collections.IDictionary` - *dictionary [collection initializers](#collection-initializers)*
-* Interface types _`I<TKey, TValue>`_ implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` (e.g. `IDictionary<TKey, TValue>`, `IReadOnlyDictionary<TKey, TValue>`)
-* Types that implement `System.Collections.IEnumerable` - [*collection initializers*](#collection-initializers)
-* Interface types _`I<T>`_ implemented by `System.Collections.Generic.List<T>` (e.g. `IEnumerable<T>`, `IList<T>`, `IReadOnlyList<T>`)
+* To a *type* with an associated *[builder](#construct-methods)* where there is an implicit collection literal conversion from the collection literal to the *builder argument type*.
```
CyrusNajmabadi, line 127: nice.  this means that teh builder method could take an array, right?  not just a span?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+* To a *type* that implements `System.Collections.IDictionary` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter `int capacity`.
```
CyrusNajmabadi, line 130: do we need to say that the name `capacity` is required? or is saying `int capacity` sufficient to imply that?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
-The following are _constructible collection types_ that can be used to construct collection literals.
-These are the valid _target types_ for collection literals.
+A _collection literal conversion_ permits a collection literal expression to be implicitly converted to a type.
 
-The list is in priority order. If a collection type belongs in multiple categories, the first is used.
+The following collection literal conversions exist from a collection literal:
 
-Actual translation of the literal to the corresponding type is defined [below](#collection-literal-translation).
+* To a single dimensional *array type* `T[]`, or a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>`, where for each element `Ei`:
+  * If `Ei` is an *expression_element* there is an implicit conversion from `Ei` to `T`.
```
CyrusNajmabadi, line 123: does thsi cover dynamic?  does dynamic need to be specified for the lower cases as well?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
-* Interface types _`I<TKey, TValue>`_ implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` (e.g. `IDictionary<TKey, TValue>`, `IReadOnlyDictionary<TKey, TValue>`)
-* Types that implement `System.Collections.IEnumerable` - [*collection initializers*](#collection-initializers)
-* Interface types _`I<T>`_ implemented by `System.Collections.Generic.List<T>` (e.g. `IEnumerable<T>`, `IList<T>`, `IReadOnlyList<T>`)
+* To a *type* with an associated *[builder](#construct-methods)* where there is an implicit collection literal conversion from the collection literal to the *builder argument type*.
+
+* To a *type* that implements `System.Collections.IDictionary` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter `int capacity`.
+  * For each element `Ei` there is an applicable instance indexer for the *key_value_pair* from:
+    * If `Ei` is an *expression_element* with type `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from `Ei`.
```
CyrusNajmabadi, line 132: you mention 'dynamic' here but not for the array/span cases.  intentional/necessary?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+    * If `Ei` is a *dictionary_element* `Ki:Vi`, the values `Ki` and `Vi`.
+    * If `Ei` is a *spread_element* with enumerated element `Si` with type `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from `Si`.
```
CyrusNajmabadi, line 134: in array/span, you used the term "enumerated element type", but here say "enumerate element `Si` with type ...", can/should these be consistent?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+* To an *interface type* *`I<K, V>`* where `I<TKey, TValue>` is implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` where for each element `Ei`:
+  * If `Ei` is an *expression_element* then the type of `Ei` is `dynamic`, or the type of `Ei` is a _type_ `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *dictionary_element* `Ki:Vi` there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread_element* with enumerated element type `Si` then the type of `Si` is `dynamic`, or the type of `Si` is a _type_ `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
```
CyrusNajmabadi, line 139: skipped this section.  i assume it's the same as the concrete dictionary case.

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+* To a *type* that implements `System.Collections.IEnumerable` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter `int capacity`.
```
CyrusNajmabadi, line 142: i assume thsi wording means we would work with constructrs with optional parameters.

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+  * No element `Ei` is a *dictionary_element*.
+  * For each element `Ei` there is an applicable instance or extension method `Add` for the single argument from:
+    * If `Ei` is an *expression_element*, the expression `Ei`.
+    * If `Ei` is a *spread_element*, the enumerated item.
```
CyrusNajmabadi, line 146: nice, i'm glad extension methods would work.

CyrusNajmabadi, line 146: do we need to have a section that explained "enumerated item" or "enumerated item type" etc?

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+* To an *interface type* *`I<T0>`* where `I<T>` is implemented by `System.Collections.Generic.List<T>` where for each element `Ei`:
+  * If `Ei` is an *expression_element* there is an implicit conversion from `Ei` to `T0`.
+  * If `Ei` is a *dictionary_element* `Ki:Vi`, then `T0` is a _type_ `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread_element* with enumerated element type `Si` there is an implicit conversion from `Si` to `T0`.
+
+Types for which there is an implicit collection literal conversion from a collection literal are the valid *target types* for that collection literal.
```
CyrusNajmabadi, line 153: love it.

proposals/collection-literals.md
```diff
@@ -112,23 +112,48 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+## Construction
+_Give specific ordering for determining how to construct the constructible collection types._
```
CyrusNajmabadi, line 156: nice.  you could also take the wording about the priority order and place here.

proposals/collection-literals.md
```diff
@@ -297,14 +337,17 @@ var y = First(new int[0], [1, 2, 3]);    // ok: int[]
...
+To allow conversions to other collection types, type inference may need to infer exact, lower-bound, and upper-bound inferences from the **expression `U`** (rather than the type of `U`) to a type `V` when an expression is given, and *fixing* §11.6.3.12 updates the sets of *candidate types* based on implicit conversion from the **expression** bound to the each *candidate type*.
 
 _Should nested cases infer collection types successfully?_
 ```c#
 var x = new[] { [ulong.MaxValue], [1, 2, 3] };  // ok: List<ulong>[]
 var y = First([[ulong.MaxValue]], [[1, 2, 3]]); // ok: List<List<ulong>>
 ```
 
+## Overload resolution
+_Include betterness order - perhaps the following, from best to worst: spans; arrays and constructed types; interface types._
```
CyrusNajmabadi, line 349: sounds good.  i think one good thing to think-about/design/test is:

```c#
void Foo(IEnumerable<KVP<X,Y>> values);
void Foo(IDictionary<X,Y> values);
```

if we do `Foo([1:2])`, which does that pick?  Waht should it pick?  Should it error?

CyrusNajmabadi (approved):

proposals/collection-literals.md
cston: Perhaps, although I thought there might be other observable differences if we're considering more than program correctness such as number of garbage collections.

proposals/collection-literals.md
cston: I realized we've defined *iteration type* earlier in the spec. I've switched the references to that term.

proposals/collection-literals.md
cston: > this means that teh builder method could take an array, right?  not just a span?

Yes, an array, or even another collection type for which there is a collection literal conversion. Do we want to allow the latter or should we limit this to array and spans?

proposals/collection-literals.md
cston: The reason for mentioning `dynamic` here is we'll use `Key` and `Value` properties, and we want to limit the set of types that are considered *key value pairs*. It feels like `dynamic` might be too much because that would allow types other than `KeyValuePair<,>`.

proposals/collection-literals.md
cston: Yes, this wording was to allow optional parameters and trailing `params` parameter.

proposals/collection-literals.md
cston: Added example, thanks.

proposals/collection-literals.md
cston: Changed to use the defined term "iteration type".

proposals/collection-literals.md
CyrusNajmabadi: honestly, this is pretty cool.  but i'm worried about circularities.  say a collection  type A could be consturcted from a B, and a B can be constructed from an A... what happens?  Limiting to array/span to begin with seems safe, and could be relaxed later.  Those two act as the building blocks we build everything else up out of.

proposals/collection-literals.md
RikkiGibson: I wasn't sure what was meant by "timing" here.

proposals/collection-literals.md
RikkiGibson: [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQAYACdbAFgG4CD0BmMzEgYRIG8CTuyKSBZIgAoAlBwFCYAUwDuJACJQAxgBcoAexgQATgE8APOSJwSsFQD5RIqiQC+XHg+4NeE8vX0BpKboBqEABsAVykABWhtQ2xjUxgLcxIABwiAZzFOfB47JzJGdD5+IQBJRVUNLT0omLME5KhtNI4c+3xbIA)

I haven't reviewed the overload resolution spec to confirm, but I think more-derived interfaces are *better*. It feels reasonable to apply the same rule here, even if different construction strategies would potentially be used for each target type.

proposals/collection-literals.md
cston: I've limited this to span for now since the expected patterns use `Span<T>` and `ReadOnlySpan<T>`.

proposals/collection-literals.md
```diff
@@ -280,12 +303,27 @@ The existing rules for type inference (see [§11.6.3](https://github.com/dotnet/
...
+>   - **`V` is a constructible collection type `C<V₁>` with element type `V₁`, and `U` is a collection literal with element type `U₁`**
+>
 >   If any of these cases apply then an *exact inference* is made from each `Uᵢ` to the corresponding `Vᵢ`.
 
-_How do we recognize that `C<V₁>` is a constructible collection with element type `V₁`?_
+_How do we recognize that `C<V₁>` has element type `V₁`?_
 
 _Update to include inference from dictionary types._
+```c#
+var d = ["Alice": 42, "Bob": 43].AsDictionary(comparer);
```
RikkiGibson, line 314: fwiw, I'm quite confused by this pattern as it has been proposed so far. We don't really do target typing of the receiver argument. The receiver in this example, I expect, has natural type `Dictionary<string, int>`. So I don't expect us to find the below `AsDictionary` method when we lookup members on it.

We could try to think of some way to *lookup members on types the CL could be target-typed to*, and finally choose a conversion and a member to access based on this. But it seems hard, and I also don't feel convinced it's worth it or that it will feel natural to users in the end.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter where the parameter has type `System.Int32` and name `capacity`.
+  * For each element `Ei` there is an applicable instance indexer for the key value pair from:
+    * If `Ei` is an *expression element* with type `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from `Ei`.
+    * If `Ei` is a *dictionary element* `Ki:Vi`, the values `Ki` and `Vi`.
+    * If `Ei` is a *spread element* with *iteration type* `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from the *iteration variable*.
+
+* To an *interface type* *`I<K, V>`* where `I<TKey, TValue>` is implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` and where for each element `Ei`:
+  * If `Ei` is an *expression element* then the type of `Ei` is `dynamic`, or the type of `Ei` is a type `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
```
333fred, line 135: I do not understand this part. If `Ei` is an expression, then its type is `dynamic`?

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
-[constructible-collection-types]: #constructible-collection-types
+## Conversions
+A *collection literal conversion* allows a collection literal expression to be converted to a type.
 
-The following are _constructible collection types_ that can be used to construct collection literals.
-These are the valid _target types_ for collection literals.
+The following implicit *collection literal conversions* exist from a collection literal expression:
 
-The list is in priority order. If a collection type belongs in multiple categories, the first is used.
+* To a single dimensional *array type* `T[]`, or a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>`, where for each element `Ei`:
```
333fred, line 120: Specese clarification: you define `Ei` for this section, then use it for all the rest of the sections. You probably want to define it above.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+  * For each element `Ei` there is an applicable instance indexer for the key value pair from:
+    * If `Ei` is an *expression element* with type `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from `Ei`.
+    * If `Ei` is a *dictionary element* `Ki:Vi`, the values `Ki` and `Vi`.
+    * If `Ei` is a *spread element* with *iteration type* `dynamic` or `System.Collections.Generic.KeyValuePair<Ki, Vi>`, the values of applicable properties `Key` and `Value` from the *iteration variable*.
+
+* To an *interface type* *`I<K, V>`* where `I<TKey, TValue>` is implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` and where for each element `Ei`:
+  * If `Ei` is an *expression element* then the type of `Ei` is `dynamic`, or the type of `Ei` is a type `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *dictionary element* `Ki:Vi` there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
```
333fred, line 136: This phrasing isn't quite right either. You want to say something more like "If ei is a dictionary element ki:vi and there exists implicit conversions from ki to k and vi to v". As written right now, you're bringing those conversions into existence, not querying if they exist.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+  * If `Ei` is a *spread element* with *iteration type* `dynamic`, or `System.Collections.Generic.KeyValuePair<Ki, Vi>` where there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+
+* To a *type* that implements `System.Collections.IEnumerable` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter where the parameter has type `System.Int32` and name `capacity`.
+  * No element `Ei` is a *dictionary element*.
+  * For each element `Ei` there is an applicable instance or extension method `Add` for the single argument from:
+    * If `Ei` is an *expression element*, the expression `Ei`.
+    * If `Ei` is a *spread element*, the *iteration variable* of the spread element.
```
333fred, line 144: Iteration type?

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+* To an *interface type* *`I<T0>`* where `I<T>` is implemented by `System.Collections.Generic.List<T>` and where for each element `Ei`:
```
333fred, line 146: `T0` is normally a nullability thing, why not `I<T>`?

proposals/collection-literals.md
```diff
@@ -297,14 +335,25 @@ var y = First(new int[0], [1, 2, 3]);    // ok: int[]
...
 var x = new[] { [ulong.MaxValue], [1, 2, 3] };  // ok: List<ulong>[]
 var y = First([[ulong.MaxValue]], [[1, 2, 3]]); // ok: List<List<ulong>>
 ```
 
+## Overload resolution
+_Include betterness order - perhaps the following, from best to worst: spans; arrays and constructed types; interface types._
+
+_Which overload should we prefer in the following?_
+```c#
+F([1:2]); // ambiguous?
```
333fred, line 351: I don't see why this would be ambiguous if it's not ambiguous today. `IDictionary<K, V>` is the more specific type. What would be more interesting to talk about is something like `IEnumerable<KVP<K, V>>` vs `IDictionary` (no generic), as I expect that would be ambiguous today.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+* To a *type* with an associated *[builder](#construct-methods)* where there is an implicit collection literal conversion from the collection literal to the *span type* of the builder argument.
 
-* Single dimensional arrays (e.g. `T[]`)
-* [*Span types*](#span-types)
-* Types with a suitable [`Construct` method](#construct-methods)
-* Types that implement `System.Collections.IDictionary` - *dictionary [collection initializers](#collection-initializers)*
-* Interface types _`I<TKey, TValue>`_ implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` (e.g. `IDictionary<TKey, TValue>`, `IReadOnlyDictionary<TKey, TValue>`)
-* Types that implement `System.Collections.IEnumerable` - [*collection initializers*](#collection-initializers)
-* Interface types _`I<T>`_ implemented by `System.Collections.Generic.List<T>` (e.g. `IEnumerable<T>`, `IList<T>`, `IReadOnlyList<T>`)
+* To a *type* that implements `System.Collections.IDictionary` where:
```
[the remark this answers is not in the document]
cston: Yes.

proposals/collection-literals.md
cston: Yes. This clause is part of "and where for each element `Ei`:" above.

proposals/collection-literals.md
cston: This clause is part of "and where for each element `Ei`:" above. We could use different phrasing, but with "If `Ei` is a dictionary element `Ki:Vi` and there exists implicit conversions from `Ki` to K and `Vi` to `V` ..." it seems like we need to complete the sentence.

cston: Updated based on our offline discussion.

proposals/collection-literals.md
cston: Updated based on our offline discussion.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+
+* To an *interface type* *`I<K, V>`* where `I<TKey, TValue>` is implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` and where for each element `Ei`:
+  * If `Ei` is an *expression element* then the type of `Ei` is `dynamic`, or the type of `Ei` is a type `System.Collections.Generic.KeyValuePair<Ki, Vi>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *dictionary element* `Ki:Vi` there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+  * If `Ei` is a *spread element* with *iteration type* `dynamic`, or `System.Collections.Generic.KeyValuePair<Ki, Vi>` where there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
+
+* To a *type* that implements `System.Collections.IEnumerable` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter where the parameter has type `System.Int32` and name `capacity`.
```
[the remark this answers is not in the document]
cston: Added an open issue for now.

proposals/collection-literals.md
cston: This is a specific substitution for `T` rather than the type parameter `T`.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+  * No element `Ei` is a *dictionary element*.
+  * For each element `Ei` there is an applicable instance or extension method `Add` for the single argument from:
+    * If `Ei` is an *expression element*, the expression `Ei`.
+    * If `Ei` is a *spread element*, the *iteration variable* of the spread element.
+
+* To an *interface type* *`I<T0>`* where `I<T>` is implemented by `System.Collections.Generic.List<T>` and where for each element `Ei`:
+  * If `Ei` is an *expression element* there is an implicit conversion from `Ei` to `T0`.
+  * If `Ei` is a *dictionary element* `Ki:Vi`, then `T0` is a type `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
```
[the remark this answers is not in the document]
cston: For the `IEnumerable` case, we haven't defined a mechanism for determining the element type of the target collection type. Added an open issue.

proposals/collection-literals.md
cston: Thanks. Changed to `IDictionary`.

proposals/collection-literals.md
```diff
@@ -112,23 +112,54 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
-* [*Span types*](#span-types)
-* Types with a suitable [`Construct` method](#construct-methods)
-* Types that implement `System.Collections.IDictionary` - *dictionary [collection initializers](#collection-initializers)*
-* Interface types _`I<TKey, TValue>`_ implemented by `System.Collections.Generic.Dictionary<TKey, TValue>` (e.g. `IDictionary<TKey, TValue>`, `IReadOnlyDictionary<TKey, TValue>`)
-* Types that implement `System.Collections.IEnumerable` - [*collection initializers*](#collection-initializers)
-* Interface types _`I<T>`_ implemented by `System.Collections.Generic.List<T>` (e.g. `IEnumerable<T>`, `IList<T>`, `IReadOnlyList<T>`)
+* To a *type* that implements `System.Collections.IDictionary` where:
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter where the parameter has type `System.Int32` and name `capacity`.
+  * For each *expression element* `Ei`:
+    * the type of `Ei` is `dynamic` and there is an applicable indexer setter that can be invoked with two `dynamic` arguments, or
```
333fred, line 130: I'm not sure the last part can actually be required? We don't check to see if an indexer exists with `dynamic`, we let the runtime binder figure that out.

proposals/collection-literals.md
```diff
@@ -112,23 +112,46 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
 
     * The behavior of collection literals with collections that are not well-behaved is undefined.
 
-## Constructible collection types (target types)
-[constructible-collection-types]: #constructible-collection-types
+## Conversions
```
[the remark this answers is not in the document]
333fred: Doesn't look addressed

proposals/collection-literals.md
cston: I believe the compiler can still check for a property group: see [sharplab.io](EYLgZgpghgLgrgJwgZwLQEkB2AbAlpiAQQQSgE9kAaAExAGoAfAAQCYBGAWAChuA3KBAAJqggLyCCAd0EBZMgBFcAYxi4A9pgFkAFAEoA3N2plNAW2WCIYiRGlrgAKwgq9hnlwD616gG0IAXTcvcV8ISkEABkDublZZBWVVDS1uAG9uQUzBJgBmQXsnFUEYAAtcZB98GEEAD3CqwTJ).

cston: I believe the compiler can still check for a property group: see [sharplab.io](EYLgZgpghgLgrgJwgZwLQEkB2AbAlpiAQQQSgE9kAaAExAGoAfAAQCYBGAWAChuA3KBAAJqggLyCCAd0EBZMgBFcAYxi4A9pgFkAFAEoA3N2plNAW2WCIYiRGlrgAKwgq9hnlwD616gG0IAXTcvcV8ISkEABkDublZZBWVVDS1uAG9uQUzBJgBmQXsnFUEYAAtcZB98GEEAD3CqwTJ).

333fred: Huh, TIL. It might be good to reference that applicability is defined by [12.6.5](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1265-compile-time-checking-of-dynamic-member-invocation) of the spec, so it's not ambiguous whether you mean that the function must itself take a `dynamic` argument.

PR #7263, opened 2023-06-08
Publish notes for extensions WG 2023-06-07

jcouv:
FYI @AlekseyTs @MadsTorgersen @333fred 

### DIFF
--- a/meetings/working-groups/roles/extensions-wg-2023-06-07.md
+++ b/meetings/working-groups/roles/extensions-wg-2023-06-07.md
@@ -1,0 +1,66 @@
+# Extensions 2023-06-07
+
+## Reviewing the "invocation expression", "member access" and "extension invocation" sections
+
+We reviewed the proposals in PR https://github.com/dotnet/csharplang/pull/7179 and discussed some open issues.  
+
+### Invoking members from extended type
+
+The first issue is that we want to access members from the extended type on a type or value of an extension type.
+
+```csharp
+extension E for C
+{
+  void M()
+  {
+    this.MethodFromUnderlying(); // problem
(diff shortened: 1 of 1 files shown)


meetings/working-groups/roles/extensions-wg-2023-06-07.md
```diff
@@ -0,0 +1,66 @@
...
+{
+   void M()
+   {
+     (int i, int j) = this;
+   }
+}
+```
+
+Decision: given that an instance `Deconstruct` property would not participate in deconstruction,
+we don't want an extension `Deconstruct` property to participate either.
```
333fred, line 52: I would prefer to state the more general "if the scenario allowed instance properties today, then extension properties should work, if not, then not" rule. The way this is written implies that rule, but doesn't explicitly state it.

meetings/working-groups/roles/extensions-wg-2023-06-07.md
```diff
@@ -0,0 +1,66 @@
...
+
+Any scenario that allowed extension methods should allow methods from extension types to participate.  
+This means we'll need to review all pattern-based sections of the spec to come up with some language
+to only bind to methods in the extension type case. Maybe something like "Do a member access and if that resolves to a method, then ..."
+or "if the expression ... resolves at compile-time to a unique instance or extension method or extension type method, that expression is evaluated".  
+
+Also, we'll need to make decisions for
+non-invocation members, such as the `Current` property involved in a `foreach`, as those were not previously
+covered by extension methods.
```
333fred, line 61: While I do think we need to discuss this, it's important to remember that we already have precedent: `MoveNext()` does not allow extensions.

PR #7271, opened 2023-06-13
Create CL-2023-06-12.md

CyrusNajmabadi:


meetings/working-groups/collection-literals/CL-2023-06-12.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Agenda
+
+* API shape for BCL collections to opt-into collection literals support.
+
+## Discussion
+
+LDM and Runtime members met and discussed the shapes we think we'll want in the runtime to allow a type to either opt into collection-literals support entirely, or to opt into an extremely efficient approach to constructing an instance of itself.
+
+Note: for the following discussion, the code shown is a rough approximation of what we think is needed.  Exact naming, order of arguments, `return` vs `out` discussion will happen during bikeshedding later.
```
cston, line 11:
```suggestion
Note: for the following discussion, the code shown is a rough approximation of what we think is needed.  Exact naming, order of arguments, `return` vs `out` discussion will happen later.
```

meetings/working-groups/collection-literals/CL-2023-06-12.md
```diff
@@ -0,0 +1,159 @@
...
+
+```c#
+ImmutableHashSet<int> values = [1, 2, 3];
+
+// translates to:
+ReadOnlySpan<int> __values = [1, 2, 3];
+ImmutableHashSet<int> values = ImmutableHashSet.Create<int>(__values);
+```
+
+This pattern will allow pointing at methods that already exist for this purpose, as well as allowing the runtime to add new methods to some existing types to allow them to be created from spans.  A similar shape that we want to support is practically the same as the above, except that the factory method can be a constructor of the type.  Continuing the example of above, that would be:
```
cston, line 50:
```suggestion
This pattern will allow pointing at methods that already exist for this purpose, as well as allowing the runtime to add new methods to existing types to allow them to be created from spans.  A similar shape that we want to support is practically the same as the above, except that the factory method can be a constructor of the type.  Continuing the example above, that would be:
```

meetings/working-groups/collection-literals/CL-2023-06-12.md
```diff
@@ -0,0 +1,161 @@
...
+
+    // newly added
+    public HashSet(ReadOnlySpan<T> values);
+}
+
+// code:
+new HashSet<int>(new int[] { 1, 2, 3 });
+```
+
+With the rules of the language today, the above is ambiguous.  This is unfortunate as a prime reason to create Span overloads of methods is to allow for efficient options for callers that do not need to use the heap.  We believe this should be fixed, and the WG will present a proposal to the LDM to add overload resolution rules that consider Spans better than items in the inheritance chain for arrays.
```
cston, line 85:
```suggestion
With the rules of the language today, the above is ambiguous.  This is unfortunate as a prime reason to create Span overloads of methods is to allow for efficient options for callers that do not need to use the heap.  We believe this should be fixed, and the WG will present a proposal to the LDM to add overload resolution rules that prefers `ReadOnlySpan<T>` over `IEnumerable<T>` when both are applicable.
```

CyrusNajmabadi: hrmm... i think we'd always to prefer ReadOnlySpan over things like IList/etc.   Or ReadOnlySpan over 'object'. etc.

meetings/working-groups/collection-literals/CL-2023-06-12.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Shape 2: Writable `Span<T>` or `T[]` storage.
+
+While the above pattern is good for collection types that own their own storage arrangement, it still involves producing the sequence one end, and then processing it to produce the file result within the factory method.  This is unnecessary overhead for a couple of very important, widely used, collections in the .Net ecosystem.  Specifically, `List<T>` and `ImmutableArray<T>`.  Both of these types are thin wrappers around a backing array, and it is highly desirable to just be able to write directly into that array without any overhead at all.  To that end, the above attribute can be used to point to a factory method with the following shape:
```
cston, line 89:
```suggestion
While the above pattern is good for collection types that own their own storage arrangement, it still involves producing the sequence one end, and then processing it to produce the final result within the factory method.  This is unnecessary overhead for a couple of very important, widely used, collections in the .NET BCL.  Specifically, `List<T>` and `ImmutableArray<T>`.  Both of these types are thin wrappers around a backing array, and it is highly desirable to just be able to write directly into that array without any overhead at all.  To that end, the above attribute can be used to point to a factory method with the following shape:
```

meetings/working-groups/collection-literals/CL-2023-06-12.md
RikkiGibson: I still feel shaky about this rule for 2 reasons:
1. It suggests making a user-defined conversion better than a standard implicit reference conversion. It feels like that's something we have avoided to date.
2. I don't see how you ship APIs that depend on this, without just saying: you can't use this from old C#, sorry.

meetings/working-groups/collection-literals/CL-2023-06-12.md
```diff
@@ -0,0 +1,161 @@
...
+Note: the runtime already has methods that expose this data for high performance scenarios.  So this does not introduce any new safety concerns.  The methods are intentionally kept out of the way in the `CompilerServices` namespace to help indicate they are not for normal use, and should be treated very carefully.
+
+#### Shape 2 consequences
+
+The above pattern will not work as efficiently in the case where the Span cannot be kept on the stack while evaluating the elements of the collection literal (for example, if there are `await` calls in the literal.  An alternative/complimentary approach that could be offered would be to have the following:
+
+```c#
+public static class CollectionsMarshall
+{
+    public static void Create<T>(int capacity, out ImmutableArray<T> result, out Span<T> storage); // or:
```
cston, line 123: The first overload here is also in the example above. Is that intentional?

CyrusNajmabadi: yes.  was showing them together so the diff could be seen.

CyrusNajmabadi:
I'm going to merge this in :)

PR #7292, opened 2023-06-20
Add include disclaimer for speclets

BillWagner:
See dotnet/docs#34996 for more background

The include file will be added to each article as part of the docs build process. The text explains the relationship between the feature speclets, and the ECMA standardization effort. It points readers to an article on docs that explains the process more fully, including the C# standard, the speclets, and the LDM notes for decisions made while developing the feature.

Note for reviewers: There are a large number of files, but all but one has the single change to add the docs include. One file changed H3s into H2s for better markdown compliance.


proposals/csharp-10.0/GlobalUsingDirective.md
```diff
@@ -1,5 +1,7 @@
 # Global Using Directive
 
+[!INCLUDE[Specletdisclaimer](~/docs/csharp/includes/speclet-disclaimer.md.md)]
```
333fred, line 3: I can't say that I'm a fan of this, as while it makes the experience better on docs.microsoft.com, it makes it worse here: this link does not resolve, and it appears at the top of every speclet.

333fred, line 3: `.md.md`?

BillWagner: That's my mistake. Thanks for catching it.

BillWagner: I could move the include file into this repo, then the link will resolve. That would help somewhat. (It would also move the issue, because the include file has links to other articles on docs.

I'm ambivalent about this as well. But, it does provide clarity for those customers reading the speclets and finding that they may include designs that were modified by later LDM decisions.

docs-includes/speclet-disclaimer.md
333fred, line 1: Sorry to keep leaving nits, but should we just put this in the `proposals` folder?

BillWagner: I can. My first thought was to put it in a different folder to avoid cluttering that folder. Easy to move.

BillWagner: done.

333fred (approved):

proposals/lambda-method-group-defaults.md
```diff
@@ -1,5 +1,7 @@
 # Optional and parameter array parameters for lambdas and method groups
 
+[!INCLUDE[Specletdisclaimer](speclet-disclaimer.md)]
```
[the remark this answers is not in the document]
BillWagner: If we do that, let's add that in a separate PR.

For the moment, I wasn't planning on it. I make the changes in the docs repo when we publish any of the speclets. That's normally done when I've added docs and / or the feature is available in public preview. I was planning to add the include to the speclet at that time.

/cc @333fred

PR #7295, opened 2023-06-26
Create CL-2023-06-16.md

CyrusNajmabadi:
WIP

### DIFF
--- a/meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
+++ b/meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
@@ -49,7 +49,7 @@ Rule '1' will always apply for immutable arrays.  Rule '2' can apply depending o
 
 ### Breaking changes
 
-We would like to update the language so that `(A)[B]` is *always* a cast of a collection literal, not an indexing expression of a trivial identifier expression.  This would apply to `(A)[B]` (trivial identifeir), or `(A.B.C)[D]` (dotted identifier).  In both of these cases, if the user wanted indexing, it would be far more idiomatic to write this out as `A[B]` or `A.B.C[D]` (the parentheses are superfluous).
+We would like to update the language so that `(A)[B]` is *always* a cast of a collection literal, not an indexing expression of a trivial identifier expression.  This would apply to `(A)[B]` (trivial identifier), or `(A.B.C)[D]` (dotted identifier).  In both of these cases, if the user wanted indexing, it would be far more idiomatic to write this out as `A[B]` or `A.B.C[D]` (the parentheses are superfluous).
 
 THe compelling case to want to make this a cast of a collection literal is:
 
--- a/meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
+++ b/meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
@@ -24,3 +24,28 @@ literals when they will be stackalloc'ed we will ensure that the compiler emits 
 to be an error, blocking unwanted span heap allocations.  In those cases, they can manually stackalloc.
 
 Users who want to be able to return those spans can do so by explicit conversion to an array.  e.g. `Span<int> s = (int[])[1, 2, 3];`. 
+
+### BCL Apis
+
+BCL gave a status update on the work they have done.  Specifically: https://github.com/dotnet/runtime/pull/87945
+
+This (merged) PR adds the pattern of an immutable collection type `C<T>` having a factory method `public static C<T> C.Create<T>(ReadOnlySpan<T> values);` to create the collection.  A soon to be added attribute (`CollectionBuilderAttribute`) will be added that will allow the language/compiler to find this factory method from a given collection type.
+
+We do not plan any more BCL apis beyond this.  However, there are two key changes we will make in the compiler in terms of emit, to get even better codegen than what those above methods allow.  Specifically:
+
+1. while there is an `public static ImmutableArray<T> ImmutableArray.Create<T>(ReadOnlySpan<T> values);` method, the compiler will not call it.  Instead, it will sidestep it and use `ImmutableArray<T> ImmutableCollectionsMarshal.ToImmutableArray<T>(T[] values);` method.  This method passes ownership of the array *directly* to `ImmutableArray<T>` ensuring no overhead (no excess allocations *or* copying).
+
+2. while `List<T>` will already work with collection literals (due to it supporting collection-initializers) the compiler will emit new collection literal `List<T>'s` as:
+
+```c#
+List<int> list = [a, b, c]; // translates to:
+
+List<int> list = new List(capacity: 3);
+CollectionsMarshal.SetCount(list, 3);
+Span<int> __storage = CollectionsMarshal.AsSpan(list);
+__storage[0] = a;
+__storage[1] = b;
+__storage[2] = c;
+```
+
+Rule '1' will always apply for immutable arrays.  Rule '2' can apply depending on how complex it is for the compiler.  For example, if the code were `List<int> list] = [await a, await b, await c];` the compiler might choose to just emit this equivalently to `new List<int> { await a, await b, await c }` (i.e. equivalently to a collection initializer expression).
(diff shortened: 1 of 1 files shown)


meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Discussion
+
+### Collection expression lifetimes
+
+Relevant link: https://gist.github.com/jaredpar/a784f8963ce73613a0a696ef89a00d9a
+
+Discussed the rules about spans produced by collection literals, and whether those are 'safe to return' from the 'current method'.
+
+Conclusion.  We will accept the rules in the provided link. Specifically, spans from collection literals are *not* "Safe to return from current method". This has the benefit of allowing the compiler to potentially stackalloc these spans for very high efficiency code (including cases like `G([1, 2, 3])`).
```
jnm2, line 19: I think there was a second point that the very high-efficiency code is only possible when passing to a `scoped` method argument.

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+Rule '1' will always apply for immutable arrays.  Rule '2' can apply depending on how complex it is for the compiler.  For example, if the code were `List<int> list] = [await a, await b, await c];` the compiler might choose to just emit this equivalently to `new List<int> { await a, await b, await c }` (i.e. equivalently to a collection initializer expression).
+
+### Breaking changes
+
+We would like to update the language so that `(A)[B]` is *always* a cast of a collection literal, not an indexing expression of a trivial identifier expression.  This would apply to `(A)[B]` (trivial identifeir), or `(A.B.C)[D]` (dotted identifier).  In both of these cases, if the user wanted indexing, it would be far more idiomatic to write this out as `A[B]` or `A.B.C[D]` (the parentheses are superfluous).
+
+THe compelling case to want to make this a cast of a collection literal is:
+
+```c#
+using CrazyArray = ImmutableArray<List<HashSet<int>>; // i.e. some complex collection type
```
jnm2, line 57: This might be the reason for the markdown source highlighter going wild
```suggestion
using CrazyArray = ImmutableArray<List<HashSet<int>>>; // i.e. some complex collection type
```

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+(ImmutableArray<Complex<Type<Examples[]>>)[complex, type, examples];
+```
+
+In other words, the extension method approach would allow for type inference to make this much simpler and cleaner than having to fully specify all the types.  Working group believes (but needs to verify) that a simple extension (no pun intended) to 'extension method lookup' would make this work.  Specifically, while the collection type has no natural type currently (like `null`, a `lambda` or a `method group`) it could still trivially participate in extension method lookup and would work in cases like this.
+
+We do, however, believe the value here is diminished if, say, the language adopts features like "postfix cast" and "partial generic type inference".  If those features make progress, you could write the above as:  ```[complex, type, examples].(ImmutableArray<_>)```.  This would be simple and clear.  
+
+WG also thinks that if we did add extension method support for the above, that we could consider doing the same for lambdas and method-groups.
+
+### Syntactic diambiguation update
```
jnm2, line 96:
```suggestion
### Syntactic disambiguation update
```

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+
+Case 1:
+```c#
+a ? [b] ...   // is that:
+
+(a?[b])       // a conditional access expression, or:
+a ? [b] : ... // a conditional expression
+```
+
+Conclusion: we will preserve existing parsing for these.  So if it is a case where it would successfully parse as a conditional-access-expression, it will remain that way.  *However*, if trying that approach would end up failing, we will try again as a conditional-expression and accept that if it succeeds.  This means the following work:
```
jnm2, line 108: ?
```suggestion
Conclusion: we will preserve existing parsing for these.  So if it is a case where it would successfully parse as a conditional-access-expression, it will remain that way.  *However*, if trying that approach would end up failing, we will try again as a conditional-expression and accept that if it succeeds.  This means the following will work:
```

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+
+Case2: 
+The collection literals feature *also* introduces its own ambiguity with: `[a ? [b] : c]` is that `[(a?[b]): c]` (a dictionary literal?), or `[a ? ([b]) : (c)]` a collection literal with a conditional-access in it.
+
+Unlike the above case, there is no back compat issue here as none of this code was legal before.  In this case, we choose an interpretation that prefers list-literals first, then dictionary literals.  So the above is a list-literal containing a conditional-access-expression.  If a dictionary-literal is preferred, the initial expression must be parenthesized.
+
+Case3:
+In `[range1, ..x, range2]`, `..x` could be a spread-element or a range-expression.
+
+WG affirms this is always a spread-element.  This should be an extremely unlikely even.  And, If the user wants a range-expression they can write `0..x` or `(..x)`.
```
jnm2, line 125:
```suggestion
WG affirms this is always a spread-element.  This should be an extremely unlikely event.  And, If the user wants a range-expression they can write `0..x` or `(..x)`. 
```

jnm2 (approved):

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+__storage[0] = a;
+__storage[1] = b;
+__storage[2] = c;
+```
+
+Rule '1' will always apply for immutable arrays.  Rule '2' can apply depending on how complex it is for the compiler.  For example, if the code were `List<int> list] = [await a, await b, await c];` the compiler might choose to just emit this equivalently to `new List<int> { await a, await b, await c }` (i.e. equivalently to a collection initializer expression).
+
+### Breaking changes
+
+We would like to update the language so that `(A)[B]` is *always* a cast of a collection literal, not an indexing expression of a trivial identifier expression.  This would apply to `(A)[B]` (trivial identifeir), or `(A.B.C)[D]` (dotted identifier).  In both of these cases, if the user wanted indexing, it would be far more idiomatic to write this out as `A[B]` or `A.B.C[D]` (the parentheses are superfluous).
```
RikkiGibson, line 52:
```suggestion
We would like to update the language so that `(A)[B]` is *always* a cast of a collection literal, not an indexing expression of a trivial identifier expression.  This would apply to `(A)[B]` (trivial identifier), or `(A.B.C)[D]` (dotted identifier).  In both of these cases, if the user wanted indexing, it would be far more idiomatic to write this out as `A[B]` or `A.B.C[D]` (the parentheses are superfluous).
```

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
CyrusNajmabadi: feel free to suggest whatever change you'd like :)

meetings/working-groups/collection-literals/meetings/working-groups/collection-literals/CL-2023-06-16.md
```diff
@@ -0,0 +1,125 @@
...
+
+Relevant link: https://gist.github.com/jaredpar/a784f8963ce73613a0a696ef89a00d9a
+
+Discussed the rules about spans produced by collection literals, and whether those are 'safe to return' from the 'current method'.
+
+Conclusion.  We will accept the rules in the provided link. Specifically, spans from collection literals are *not* "Safe to return from current method". This has the benefit of allowing the compiler to potentially stackalloc these spans for very high efficiency code (including cases like `G([1, 2, 3])`).
+
+While the compiler is free to use stackallocs for these spans, it is not required to do so.  However, as some consumers *may* only want to allow collection literals when they will be stackalloc'ed we will ensure that the compiler emits a hidden-diagnostic if it heap allocs.  These consumers can set that diagnostic to be an error, blocking unwanted span heap allocations.  In those cases, they can manually stackalloc.
+
+Users who want to be able to return those spans can do so by explicit conversion to an array.  e.g. `Span<int> s = (int[])[1, 2, 3];`.
```
RikkiGibson, line 23: I think at some point Chuck and I will work together to make the relevant revisions to the collection literals spec.

RikkiGibson (approved):

PR #7320, opened 2023-07-05
Collection literals: create methods

cston:
Update section for custom create methods based on recent discussions.

### DIFF
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -176,6 +176,10 @@ The attribute specifies the *builder type* and *method name* of a method to be i
 ```c#
 namespace System.Runtime.CompilerServices
 {
+    [AttributeUsage(
+        AttributeTargets.Class | AttributeTargets.Struct,
+        Inherited = false,
+        AllowMultiple = false)]
     public sealed class CollectionBuilderAttribute : System.Attribute
     {
         public CollectionBuilderAttribute(Type builderType, string methodName);
@@ -200,7 +204,7 @@ Overloads with distinct signatures on the builder type, and overloads on base ty
 
 An error is reported for a collection literal if the `[CollectionBuilder]` attribute on the collection type does not represent an invocable method with the expected signature defined on the *builder type*.
 
-If the parameter is `scoped`, the compiler *may* allocate the storage for the `ReadOnlySpan<T0>` on the stack rather than the heap.
+If the parameter does not escape the *create method* (that is, if the parameter is `scoped` or if the return type is not a `ref struct`), the compiler *may* allocate the storage for the `ReadOnlySpan<T0>` on the stack rather than the heap.
 
 For example, a possible *create method* for `ImmutableArray<T>`:
 ```csharp
@@ -209,7 +213,7 @@ public struct ImmutableArray<T> { ... }
 
 public static class ImmutableArray
 {
-    public ImmutableArray<T> Create<T>(scoped ReadOnlySpan<T> items) { ... }
+    public static ImmutableArray<T> Create<T>(ReadOnlySpan<T> items) { ... }
 }
 ```
 
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -198,7 +198,7 @@ The *create method* must be a `static` method defined on the *builder type*.
 The *create method* must have a single parameter of a type `System.ReadOnlySpan<T0>` passed by value, and a return value of the *collection type* to which the attribute is applied.
 
 The *builder type* cannot be generic.
-The *builder method* can be a generic *method* and the *arity* of the builder method must match the *arity* of the *collection type*.
+The *builder method* can be generic and the *arity* of the builder method must match the *arity* of the *collection type*.
 
 Overloads with distinct signatures on the builder type, and overloads on base types, are ignored.
 
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -177,7 +177,7 @@ The attribute specifies the *builder type* and *method name* of a method to be i
 namespace System.Runtime.CompilerServices
 {
     [AttributeUsage(
-        AttributeTargets.Class | AttributeTargets.Struct,
+        AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface,
         Inherited = false,
         AllowMultiple = false)]
     public sealed class CollectionBuilderAttribute : System.Attribute
@@ -189,7 +189,7 @@ namespace System.Runtime.CompilerServices
 }
 ```
 
-The attribute can be applied to `class` and `struct` types only.
+The attribute can be applied to `class`, `struct`, and `interface` types.
 The attribute is not inherited although the attribute can be applied to a base `class` or an `abstract class`.
 
 An error is reported for a collection literal if there are multiple `[CollectionBuilder]` attributes on the collection type.
--- a/proposals/collection-literals.md
+++ b/proposals/collection-literals.md
@@ -186,17 +186,15 @@ namespace System.Runtime.CompilerServices
 ```
 
 The attribute can be applied to `class` and `struct` types only.
-
 The attribute is not inherited although the attribute can be applied to a base `class` or an `abstract class`.
 
 An error is reported for a collection literal if there are multiple `[CollectionBuilder]` attributes on the collection type.
 
 The *create method* must be a `static` method defined on the *builder type*.
-
 The *create method* must have a single parameter of a type `System.ReadOnlySpan<T0>` passed by value, and a return value of the *collection type* to which the attribute is applied.
 
 The *builder type* cannot be generic.
-The *builder method* must be a generic *method* *iff* the collection type is a generic type, and the type parameters of the builder *method* must correspond in order to the type parameters of *collection type*.
+The *builder method* can be a generic *method* and the *arity* of the builder method must match the *arity* of the *collection type*.
 
 Overloads with distinct signatures on the builder type, and overloads on base types, are ignored.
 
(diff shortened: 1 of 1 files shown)


proposals/collection-literals.md
```diff
@@ -151,70 +156,72 @@ _Open issue: Relying on a parameter named `capacity` seems brittle. Is there an
...
+    {
+        public CollectionBuilderAttribute(Type builderType, string methodName);
+        public Type BuilderType { get; }
+        public string MethodName { get; }
+    }
+}
+```
 
-* Like [*`init` accessors*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/init.md#init-only-setters), an `init` method would be an instance method invocable at the point of object creation but become unavailable once object creation has completed. This facility thus prevents general use of such a marked method outside of known safe compiler scopes where the instance value being constructed cannot be observed until complete.
+The attribute can be applied to `class` and `struct` types only.
```
CyrusNajmabadi, line 188: seems like it would be fine to be on an interface type.  indeed, you could imagine adding this to [`IImutableSet<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.immutable.iimmutableset-1?view=net-7.0)

proposals/collection-literals.md
```diff
@@ -151,70 +156,72 @@ _Open issue: Relying on a parameter named `capacity` seems brittle. Is there an
...
+An error is reported for a collection literal if the `[CollectionBuilder]` attribute on the collection type does not represent an invocable method with the expected signature defined on the *builder type*.
 
-* This formalization is quite beneficial because the only existing mechanism to (safely) create an ImmutableArray with values without copying is both excessively verbose and produces unavoidable garbage:
+If the parameter is `scoped`, the compiler *may* allocate the storage for the `ReadOnlySpan<T0>` on the stack rather than the heap.
```
RikkiGibson, line 203: I would rephrase this as "if the parameter does not escape the Create method". Alternatively we could define a concept of "implicitly scoped" and use it here. All I really want to avoid terminology-wise is having to say in several places "if the method does not return ref struct or have a writable ref parameter of ref struct type etc..."

I'm OK with this limitation as it applies to user-defined ref struct types. I somewhat feel that if you are creating a ref struct which is simply wrapping a span, what are you really accomplishing. I don't know whether that's a real use case.

If such ref structs were a real use case, though, my feeling would be that this is unnecessarily limiting. Why shouldn't my ref-struct be able to hold on to stackalloc'd memory just like a Span would. In that case I would prefer a rule that if a collection-expr is converted to ref-struct, its safe-to-escape is the current local scope. It's a simple rule which remains "intact" even through incidental refactorings. If you don't want that behavior, convert the CL to non-ref-struct type first. This is essentially what we suggested for CLs converted directly to Span, I think.

RikkiGibson (approved):

proposals/collection-literals.md
```diff
@@ -151,70 +156,72 @@ _Open issue: Relying on a parameter named `capacity` seems brittle. Is there an
...
 
-    ```c#
-    var __builder = ImmutableArray.CreateBuilder<int>(initialCapacity: __len);
+For example, a possible *create method* for `ImmutableArray<T>`:
+```csharp
+[CollectionBuilder(typeof(ImmutableArray), "Create")]
```
333fred, line 207: This is striking me as `infoof`. Do we _really_ need to be able to specify member names here, and not just have a well-recognized pattern that the type indicated must follow?

CyrusNajmabadi: @333fred i believe we need this.  But i'll let @stephentoub weigh in.  I think the core issue is that the BCL expects (or already has) differently named methods in a few places, and they'd want these attributes to point at them (versus adding a new duplicative method).  Because of that, we need a type+methodname pair to clearly identify things.  

Note: i believe this is going through Runtime review soon (or maybe already has).  So i'm ok deferring to what they want.  ON our side we just need something to go find the appropriate method :)

stephentoub: @333fred, are you suggesting that the compiler would look at all members of the specified type for one with the right shape and use whatever it finds?

proposals/collection-literals.md
```diff
@@ -151,70 +156,76 @@ _Open issue: Relying on a parameter named `capacity` seems brittle. Is there an
...
+An error is reported for a collection literal if the `[CollectionBuilder]` attribute on the collection type does not represent an invocable method with the expected signature defined on the *builder type*.
 
-* This formalization is quite beneficial because the only existing mechanism to (safely) create an ImmutableArray with values without copying is both excessively verbose and produces unavoidable garbage:
+If the parameter does not escape the *create method* (that is, if the parameter is `scoped` or if the return type is not a `ref struct`), the compiler *may* allocate the storage for the `ReadOnlySpan<T0>` on the stack rather than the heap.
```
jaredpar, line 207:
```suggestion
If the parameter does not escape the *create method* (that is, if the parameter is `scoped`, the return type is not a `ref struct` and there are no `out / ref` of type `ref struct`), the compiler will assign a lifetime of _current method_ to the argument. As such it *may* allocate the storage for the `ReadOnlySpan<T0>` on the stack rather than the heap. 
```

proposals/collection-literals.md
```diff
@@ -151,70 +156,76 @@ _Open issue: Relying on a parameter named `capacity` seems brittle. Is there an
...
+```csharp
+[CollectionBuilder(typeof(ImmutableArray), "Create")]
+public struct ImmutableArray<T> { ... }
 
-    __builder.Add(e1);
-    foreach (var __t in s1)
-        __builder.Add(__t);
+public static class ImmutableArray
+{
+    public static ImmutableArray<T> Create<T>(ReadOnlySpan<T> items) { ... }
```
jaredpar, line 216:
```suggestion
    public static ImmutableArray<T> Create<T>(scoped ReadOnlySpan<T> items) { ... }
```

Ideally this parameter is `scoped`

proposals/collection-literals.md
jaredpar: Do not see any discussion about the lifetime of a collection literal that is used in a local declaration.

proposals/collection-literals.md
333fred: > @333fred, are you suggesting that the compiler would look at all members of the specified type for one with the right shape (accepting a `ReadOnlySpan<T>`, returning a `CollectionType`, name doesn't matter) and use whatever it finds?

Yes, but I would expect the name to be part of the pattern it looks for.

> I think the core issue is that the BCL expects (or already has) differently named methods in a few places, and they'd want these attributes to point at them (versus adding a new duplicative method).

We've already done this multiple times for language features. For example, `List<T>` added the `Slice` method which just defers to `GetRange`, to work with the slicing feature. I also wonder, if we have some small list of canonical methods, can we not just codify them now? Is there a list of the existing method names that we would want to consider for the pattern?

proposals/collection-literals.md
cston: I didn't mention `out` or `ref` parameters here because the *create method* pattern is specifically `static CollectionType<T> Create<T>(ReadOnlySpan<T> items)`, with a single parameter passed by value.

proposals/collection-literals.md
stephentoub: > Yes, but I would expect the name to be part of the pattern it looks for.

That'll work for the specific uses planned for the core libraries for .NET 8, which are all in immutable collections.  It won't work when we add ctors to types like `HashSet<T>` and expect those to be usable.  It won't work when we add additional patterns and want multiple named methods on the same type, e.g. a CollectionsMarshal.CreateList and a CollectionsMarshal.CreateQueue.  And it won't work for other collections that already have an appropriately-shaped factory method with a different name, forcing the developer to add another API for it that does the same thing.

I think we should continue having the name be specified.  It's more flexible and specific and it doesn't hurt anything.

> For example, List<T> added the Slice method which just defers to GetRange, to work with the slicing feature

And we didn't like doing so.  We only did because `List<T>` is such a prominent type and it was the least friction-free way to unblock that use.  I do not see us, for example, adding a static `List` class with a static `Create` method.

proposals/collection-literals.md
jaredpar: True but shouldn't we discuss the more general policy here? Specifically that collection literals when passed to a method have such and such rules. The rules for custom create methods aren't special here, they just fall out from those broader rules.

jaredpar: Basically there is no diff between 

```csharp

ImmutableArray<int> array1 = { 1, 2, 3 };
ImmutableArray<int> array2 = MyCreate({1, 2, 3 });
ImmutableArray<int> MyCreate(scoped ReadOnlySpan<int> span) => ...
```

proposals/collection-literals.md
333fred: Ctors seem like something we codify. The multiple methods also seems like something we can also codify, with a rule involving the type name. Every time we've looked at something like `methodof` or `infoof` in C#, we've arrived at all sorts of issues that need to be resolved. I'm not innately opposed to it, but as this as historically been a difficult problem for us to address, we might want to consider a more hardcoded list.

proposals/collection-literals.md
```diff
@@ -118,12 +118,17 @@ A *collection literal conversion* allows a collection literal expression to be c
...
-  * For each *dictionary element* `Ki:Vi`, `T` is a type `System.Collections.Generic.KeyValuePair<K, V>` and there is an implicit conversion from `Ki` to `K` and from `Vi` to `V`.
-  * For each *spread element* `Si` there is an implicit conversion from the *iteration type* of `Si` to `T`.
+* To a single dimensional *array type* `T[]` where:
+  * For each *element* `Ei` there is an *implicit conversion* to `T`.
 
-* To a *type* with an associated *[builder](#construct-methods)* where there is an implicit collection literal conversion from the collection literal to the *span type* of the builder argument.
+* To a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>` where:
+  * For each *element* `Ei` there is an *implicit conversion* to `T`.
+
+* To a *type* with a *[create method](#create-methods)* with *parameter type* `System.ReadOnlySpan<T>` where:
```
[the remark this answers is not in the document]
cston: > Is there a need to lock the create method in to taking ReadOnlySpan?

Restricting the pattern to `ReadOnlySpan<T>` is likely to simplify testing, especially in cases where there are overloads. We can expand this to other types if there are scenarios.

proposals/collection-literals.md
cston: The comment here was to clarify the requirements for this particular case. We could perhaps remove this comment when we add a more general statement about lifetime in a subsequent PR.

proposals/collection-literals.md
cston: I had this as `scoped` initially but removed the modifier from this example since it's redundant.

PR #7322, opened 2023-07-05
Proposal for `readonly` parameters

333fred:
Initial proposal for `readonly` parameters to be discussed in LDM.

### DIFF
--- a/proposals/readonly-parameters.md
+++ b/proposals/readonly-parameters.md
@@ -106,6 +106,6 @@ parameter must be spelled out?
 
 #### Emit consequences
 
-This proposal that `readonly` on parameters has no effect on the emitted code, and that it will not be possible to tell from metadata (including things that read metadata, such as reflection) that a parameter was declared as
+This proposal states that `readonly` on parameters has no effect on the emitted code, and that it will not be possible to tell from metadata (including things that read metadata, such as reflection) that a parameter was declared as
 `readonly`. Are there use cases for reflecting this information in metadata, and if so, what should the emit strategy we use to convey this information be? And, if we do emit to metadata, should that change how overriding
 carries through that information?
--- a/proposals/readonly-parameters.md
+++ b/proposals/readonly-parameters.md
@@ -1,0 +1,111 @@
+# Readonly Parameters
+
+* [x] Proposed
+* [ ] Prototype: [Complete](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
+* [ ] Implementation: [In Progress](https://github.com/dotnet/roslyn/BRANCH_NAME)
+* [ ] Specification: [Not Started](pr/1)
+
+## Summary
+[summary]: #summary
+
+We allow parameters to marked with `readonly`. This disallows them from being assigned to or being passed by `ref` or `out`.
+
+## Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


proposals/readonly-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+    ;
+```
+
+The spec is not yet updated to include the draft C# 12 specification for [`ref readonly` parameters](https://github.com/dotnet/csharplang/blob/main/proposals/ref-readonly-parameters.md), but we will restrict
+`readonly` such that if a parameter is both `readonly` and `ref readonly`, the `readonly` must appear on the left side of the `ref readonly`, so that `readonly ref readonly` is permitted, but `ref readonly readonly`
+is not.
+
+### Semantics
+
+The for a `readonly` parameter, the compiler will issue an error when it is assigned to or taken as a mutable `lvalue`. This means that a `readonly` parameter cannot be passed by `out` or `ref`, but can be passed
```
jjonescz, line 55:
```suggestion
For a `readonly` parameter, the compiler will issue an error when it is assigned to or taken as a mutable `lvalue`. This means that a `readonly` parameter cannot be passed by `out` or `ref`, but can be passed
```

proposals/readonly-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+by value, `ref readonly`, or `in`.
+
+#### `partial` methods
+
+For `partial` methods, we allow the implementing `partial` method declaration to add the `readonly` modifier to a parameter if the defining `partial` method did not. If the defining partial `method` included the
+`readonly` modifier included the `readonly` modifier, the implementing `partial` method must also include it.
```
jjonescz, line 61:
```suggestion
For `partial` methods, we allow the implementing `partial` method declaration to add the `readonly` modifier to a parameter if the defining `partial` method did not. If the defining partial `method` included the
`readonly` modifier, the implementing `partial` method must also include it.
```

proposals/readonly-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Semantics
+
+For a `readonly` parameter, the compiler will issue an error when it is assigned to or taken as a mutable `lvalue`. This means that a `readonly` parameter cannot be passed by `out` or `ref`, but can be passed
+by value, `ref readonly`, or `in`.
+
+#### `partial` methods
+
+For `partial` methods, we allow the implementing `partial` method declaration to add the `readonly` modifier to a parameter if the defining `partial` method did not. If the defining partial `method` included the
+`readonly` modifier, the implementing `partial` method must also include it.
```
CyrusNajmabadi, line 61: as mentioned offline, my personal preference would be to *not* allow this.  But it's not a blocker for me.

proposals/readonly-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+
+* [x] Proposed
+* [ ] Prototype: [Complete](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
+* [ ] Implementation: [In Progress](https://github.com/dotnet/roslyn/BRANCH_NAME)
+* [ ] Specification: [Not Started](pr/1)
+
+## Summary
+[summary]: #summary
+
+We allow parameters to marked with `readonly`. This disallows them from being assigned to or being passed by `ref` or `out`.
```
333fred, line 11:
```suggestion
We allow parameters to be marked with `readonly`. This disallows them from being assigned to or being passed by `ref` or `out`.
```

333fred:
I'm going to merge this as is so that LDM note history has a commit to point at, then open a followup to narrow the scope.

PR #7348, opened 2023-07-17
fix docs build warnings

BillWagner:
Three related changes here:

1. remove `en-us` so links will resolve to the readers' locale.
1. Remove `?view=net-7.0` so API links resolve to the default view. Note that I didn't remove the query string where .NET 8 was used, as it's in preview.
1. Make absolute links to the C# standard spec language.

This PR fixes the build warnings in dotnet/docs#36277

### DIFF
--- a/proposals/inline-arrays.md
+++ b/proposals/inline-arrays.md
@@ -41,7 +41,7 @@ Runtime provides regular GC tracking for all elements in the struct.
 This proposal will refer to types like this as "inline array types".
 
 Elements of an inline array type can be accessed through pointers or through span instances returned by
-[System.Runtime.InteropServices.MemoryMarshal.CreateSpan](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createspan?view=net-7.0)/[System.Runtime.InteropServices.MemoryMarshal.CreateReadOnlySpan](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createreadonlyspan?view=net-7.0) APIs. However, neither
+[System.Runtime.InteropServices.MemoryMarshal.CreateSpan](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.memorymarshal.createspan)/[System.Runtime.InteropServices.MemoryMarshal.CreateReadOnlySpan](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.memorymarshal.createreadonlyspan) APIs. However, neither
 the pointer approach, nor the APIs provide type and bounds checking out of the box.
 
 Language will provide a type-safe/ref-safe way for accessing elements of inline array types. The access will be span based. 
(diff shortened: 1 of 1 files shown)


proposals/inline-arrays.md
```diff
@@ -112,18 +112,18 @@ For an inline array element access, the *primary_no_array_creation_expression* o
 ##### When the expression type is int
 
 If *primary_no_array_creation_expression* is a writable variable, the result of evaluating an inline array element access is a writable variable
-equivalent to invoking [`public ref T this[int index] { get; }`](https://learn.microsoft.com/en-us/dotnet/api/system.span-1.item?view=net-8.0) with
+equivalent to invoking [`public ref T this[int index] { get; }`](https://learn.microsoft.com/dotnet/api/system.span-1.item?view=net-8.0) with
```
333fred, line 115: I don't think this link needs the .NET 8 specifier, there's nothing different about the API.

proposals/inline-arrays.md
```diff
@@ -173,13 +173,13 @@ the *primary_no_array_creation_expression*. Then the element access is interpret
 ##### When the expression implicitly convertible to ```System.Range```
 
 If *primary_no_array_creation_expression* is a writable variable, the result of evaluating an inline array element access is a value
-equivalent to invoking [`public Span<T> Slice (int start, int length)`](https://learn.microsoft.com/en-us/dotnet/api/system.span-1.slice?view=net-8.0)
+equivalent to invoking [`public Span<T> Slice (int start, int length)`](https://learn.microsoft.com/dotnet/api/system.span-1.slice?view=net-8.0)
```
333fred, line 176: Same comment

proposals/inline-arrays.md
```diff
@@ -173,13 +173,13 @@ the *primary_no_array_creation_expression*. Then the element access is interpret
...
 on an instance of ```System.Span<T>``` returned by ```System.Span<T> InlineArrayAsSpan``` method on *primary_no_array_creation_expression*.
 For the purpose of ref-safety analysis the *ref-safe-to-escape*/*safe-to-escape* of the access are equivalent to the same
 for an invocation of a method with the signature ```static System.Span<T> GetSlice(ref InlineArrayType array)```.
 
 If *primary_no_array_creation_expression* is a readonly variable, the result of evaluating an inline array element access is a value
-equivalent to invoking [`public ReadOnlySpan<T> Slice (int start, int length)`](https://learn.microsoft.com/en-us/dotnet/api/system.readonlyspan-1.slice?view=net-8.0)
+equivalent to invoking [`public ReadOnlySpan<T> Slice (int start, int length)`](https://learn.microsoft.com/dotnet/api/system.readonlyspan-1.slice?view=net-8.0)
```
333fred, line 182: Same comment

BillWagner:
All fixed. Thanks @333fred

333fred (approved):

proposals/inline-arrays.md
```diff
@@ -91,15 +91,15 @@ element_access
 
 The *argument_list* of an *element_access* is not allowed to contain `ref` or `out` arguments.
 
-An *element_access* is dynamically bound ([§11.3.3](expressions.md#1133-dynamic-binding)) if at least one of the following holds:
+An *element_access* is dynamically bound ([§11.3.3](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1133-dynamic-binding)) if at least one of the following holds:
```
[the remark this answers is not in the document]
BillWagner: Fixed in #7368

> By the way, is there a feature request for Microsoft Learn to rewrite this kind of link

Not right now. I may make updates to the tool so that we use custom anchors without the clause numbers. That would be more resilient. I don't know if that will work, because of different clauses with the same name. "General" is an obvious example. I don't know if concatenating all parent clauses would make them unique. I'll run some tests.

Issue #3326, opened 2020-04-01
Allow ref assignment for switch expressions


Rabadash8820:
This idea was already brought up in #2507, but the conclusion there was that:

> I think the switch expression should work the same way [as ternary operators] so I would close this for now

However, switch expressions definitely still do not work the same way as ternary operators with respect to ref assignment, as others have noted in the Roslyn repo:
- dotnet/roslyn#42130
- dotnet/roslyn#40198
- dotnet/roslyn#40236

Long story short, suppose I have a `Direction` enum and want to conditionally do a ref assignment by switching on a `Direction` instance, like so:

```cs
 switch (direction)
{
     case Direction.North: return ref NorthField;
     case Direction.South: return ref SouthField;
     case Direction.East: return ref EasstField;
     case Direction.West: return ref WestField;
     default: throw new NotImplementedException();
}
```

Currently, Visual Studio suggests converting the switch statement to an expression, generating the following invalid code:

```cs
return direction switch
{
     Direction.North => ref NorthField;
     Direction.South =>ref SouthField;
     Direction.East => ref EasstField;
     Direction.West => ref WestField;
     _ => throw new NotImplementedException();    
};
```
Each ref expression has a red squiggly saying `Invalid expression term 'ref'`, even if I add `ref` after the `return`. If I keep the `ref` after the `return` and delete all the case-specific `ref`s, then the whole switch block gets squiggled with the message `An expression cannot be used in this context because it may not be passed or returned by reference`.

I won't say "this should be easy to implement", cause I'm sure its not, but I see know reason why a ref assignment/return with a switch expression should be invalid syntax. Based on the conversation in #2507, I'm guessing the most likely implementation is to require ref after the return and in each case expression, but less typing is always better 🙂

333fred:
@CyrusNajmabadi for the bad VS suggestion.

CyrusNajmabadi:
@alrz  :)

Rabadash8820:
@333fred I believe the bad VS suggestion has already been addressed by dotnet/roslyn#40236. @CyrusNajmabadi recommended I open a new issue here for the actual language feature request.

AJosephsen:
It seems inconsistent that you can do this with the ? operator but not with switch expressions

I am able to do the following with the ? operator
`(lowerBound < upperBound ? ref lowerBound : ref upperBound) = mid;
`
But not with the switch expression

```
(lowerBound < upperBound ) switch
{
	(true) => ref lowerBound,
[...]

PR #7352, opened 2023-07-18
Ref assignment switch expressions

Rekkonnect:
Contributes to #3326

This is a somewhat old spec that was missed from being opened as a PR. Might need to take another iteration on it after two years.

### DIFF
--- a/proposals/ref-assignment-switch-expressions.md
+++ b/proposals/ref-assignment-switch-expressions.md
@@ -174,7 +174,45 @@ private void RefSwitchAssignLocal()
 
 > Note: this does not align with the current behavior of ref ternary expressions (`a ? ref b : ref c`). This is intentional, and the handling of this case in ref ternary expressions is a separate concern.
 
-Ref return safety relies on the individual switch arms. If any of the switch arms that does not throw is unsafe to return, the entire switch expression is unsafe to return. Otherwise, the entire `switch` expression's result is safe to return.
+The scope of the returned expression relies on the narrowest scope of each individual arm that is not a throw expression. That means, the scope of ref safety, and the scope of the actual value are both taken into account, and the narrowest of all is applied to the entire switch expression. This rule is similar to ref ternary expressions.
+
+For example, the following is unsafe to return:
+```csharp
+private void RefSwitchNarrowScope(ref int paramRef)
+{
+    int x = 1;
+    int outer = 2;
+
+    // outerRef contains a reference to the narrowest scope,
+    // which is that of 'x' and 'outer'
+    ref int outerRef = ref axis switch
+    {
+        Axis.X => ref x,
+        Axis.Y => ref outer,
+    };
+
+    {
+        int inner = 3;
+
+        // innerRef contains a reference to the narrowest scope,
+        // which is that of 'inner'
+        ref int innerRef = ref axis switch
+        {
+            Axis.X => ref x,
+            Axis.Y => ref inner,
+        };
+        
+        // 'innerRef' holds a reference to a narrower scope than
+        // that of 'outerRef', so the following is illegal
+        outerRef = ref innerRef;
+    }
+
+    // The scope of 'paramRef' is outer than that of the current method
+    // 'outerRef' holds a reference to a scope within the method,
+    // so the following is illegal
+    paramRef = ref outerRef;
+}
+```
 
 ## Drawbacks
 [drawbacks]: #drawbacks


Rekkonnect:
CC @CyrusNajmabadi

CyrusNajmabadi:
I'd recommend including relevant parts from https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/conditional-ref.md.  For example, safe-to-return, as well as this being an LValue.

333fred:
Modified the original comment to not say `closes`

333fred:
A decent example of the type of thing we're looking for in a spec is https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/target-typed-conditional-expression.md or https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/extension-getenumerator.md. It is going to be harder to get to that level of detail, as there's no checked-in version of the spec. You can base it off the proposal specification here: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md#switch-expression.

proposals/ref-assignment-switch-expressions.md
```diff
@@ -0,0 +1,69 @@
...
+[motivation]: #motivation
+
+The `switch` expression is meant to provide a quicker way to return values based on a given expression, but does not support returning `ref` expressions. This limitation doesn't prevent any harm, and can be lifted.
+
+## Detailed design
+[design]: #detailed-design
+
+A `switch` expression may return a `ref` value. The returning expression, if not a `throw` expression, must be a `ref` expression.
+
+A `switch` expression that returns a `ref` value needs an extra `ref` in front of it when assigned/returned to a `ref` local. The examples below cover this case too. This design aligns with the current design in the ternary operator.
```
jaredpar, line 23: One regret I have with conditional `ref` expressions is that the following compiles without a diagnostic: 

```csharp
cond ? ref a : ref b
```

Clearly the user meant to take a `ref` here yet they forgot the `ref` before `cond` hence ended up with a value. Guessing many devs have spent time wondering what was going wrong here. 

Would not like to repeat that mistake here. Suggest diagnostics for the following scenarios:
1. If all arms of the `switch` have `ref` but the overall expression does not.
2. If there is disagreement in the arms about `ref`

proposals/ref-assignment-switch-expressions.md
```diff
@@ -0,0 +1,69 @@
...
+};
+
+private void RefSwitchAssignLocal()
+{
+    ref int dimension = ref axis switch
+    {
+        Axis.X => ref X,
+        Axis.Y => ref Y,
+    };
+}
```
jaredpar, line 47: This design needs to cover the lifetime implications as it deals with `ref` semantics. Essentially how do we calculate the *safe-to-escape* of the return here?

proposals/ref-assignment-switch-expressions.md
KennethHoff: You could(if it doesn't already exist) add an Analyzer for the `ref` expressions as well "while you're at it"(not that that's a good argument for a language decision)

Rekkonnect: Clearly Jared has a point here, 99% of the time you want the ref of the returned expression, so a warning is very reasonable.

Adding that warning for ref ternaries should probably be done in the future, but it's off the scope of this proposal. Probably a warning wave too, to avoid breaking existing code.

proposals/ref-assignment-switch-expressions.md
Rekkonnect: I don't get that concern, is it relevant to ref scoped? I'm out of the loop with that feature.

proposals/ref-assignment-switch-expressions.md
CyrusNajmabadi: i would simply make it a requirement for switches.  WE can retcon ternaries independent of this.

proposals/ref-assignment-switch-expressions.md
RikkiGibson: I would expect all the ref-safety caveats of ref-`?:` to apply here. Including that RSTE/STE is equal to the narrowest of all switch arms's RSTE/STE, and that when the ref is mutable, the STEs of the switch arms must be equal.

CyrusNajmabadi: thanks @RikkiGibson !

Rekkonnect: That still didn't clear it out to me, what exactly are you referring to with RSTE/STE? I can only infer about the level of access on the referenced memory.

bernd5: STE is Safe-To-Escape and RSTE is Ref-Safe-To-Escape.

Rekkonnect: Included details about RSTE/STE

proposals/ref-assignment-switch-expressions.md
Rekkonnect: Added a simple advisory for causing a warning instead of an error that requires that the returned ref is directly used. If we settle on the error, I will update the spec accordingly.

proposals/ref-assignment-switch-expressions.md
```diff
@@ -0,0 +1,110 @@
...
+
+A `switch` expression may return a `ref` value. The returning expression, if not a `throw` expression, must be a `ref` expression.
+
+A `switch` expression that returns a `ref` value needs an extra `ref` in front of it when assigned/returned to a `ref` local. This design aligns with the current design in the ternary operator.
+
+`throw` expressions can still be used normally, and pose no effect on the return type of the `switch` expression.
+
+Once a single switch arm returns a `ref` expression, all other switch arms must return `ref` expressions, unless they are `throw` expressions. Not providing a `ref` expression in a switch arm causes a compiler error.
+
+Since all ref expressions are LValues, and the switch arms are all ref expressions, the entire result of the switch expression is also an LValue. Therefore, the entire switch expression can be passed by reference, assigned to, and returned by reference. Note that for directly assigning to the switch expression's ref result, the `ref` keyword must not be used, like normal ref locals are assigned. For example:
```
jjonescz, line 29: > assigned to

Nit: not if the references are readonly. Consider a ternary operator equivalent:

```cs
void M(bool b, ref int x, in int y)
{
    (b ? ref x : ref x) = x;
    (b ? ref y : ref y) = y; // error
}
```

proposals/ref-assignment-switch-expressions.md
```diff
@@ -0,0 +1,110 @@
...
+
+`item switch { 1 => ref X, 2 => ref Y } = value`
```
jjonescz, line 31: Consider using a code block for better syntax highlighting.

````suggestion
```cs
item switch { 1 => ref X, 2 => ref Y } = value
```
````

File: proposals/ref-assignment-switch-expressions.md (hunk 3 above)
Comment (Rekkonnect) [reply proposals/ref-assignment-switch-expressions.md:29]: Nice catch

Comment (Rekkonnect) [reply proposals/ref-assignment-switch-expressions.md:29]: After examining this further, you cannot use `ref y` anyway since `y` is readonly. You can only assign a `ref y` expression to a `ref readonly` local or field. For clarity I will include this case in the spec, but it is not really part of the scope of ref switch expressions.

Comment (jjonescz) [reply proposals/ref-assignment-switch-expressions.md:29]: > you cannot use `ref y` anyway since `y` is readonly

That's not correct, I think. This is valid:

```cs
void M(bool b, ref int x, in int y)
{
    (b ? ref x : ref x) = x;
    ref readonly int z = ref (b ? ref y : ref y);
}
```

Similarly I would expect that to be valid for switch ref expressions as you say in the proposal "the entire switch expression can be passed by reference" - but in this case it cannot be "assigned to".

Rekkonnect: You are correct, I'm fixing this in a bit

proposals/ref-assignment-switch-expressions.md
Rekkonnect: Update, after some thought and discussion I noticed that:
```csharp
int a = 4;
ref int b = ref a;
int c = ref a;
int d = a == b ? ref a : ref b;
```

- `c` yields error CS8171 - Cannot initialize a by-value variable with a reference
- `d` does not yield any errors and is completely legal

We should definitely strive for consistency with `ref` variables, and designing `ref` switch expressions should return an error if the value is immediately dereferenced.

As for the ref ternary expressions, it could be a breaking change, via either a warning or an error that immediately applies to existing code, hinting that it was definitely a mistake in the first place. Though discussing this one goes out of scope of this PR.

I'll prepare spec changes for this new approach on the error, would love some more input on this decision to make sure we cover this case elegantly.

Issue #6149, opened 2022-05-20
Open issues for `ref` fields


cston:
### 1. Parse `scoped` as a modifier with `-langversion:11`?
What is the effect of parsing `scoped` as a modifier on existing code that uses `scoped` as a type name or identifier?
```csharp
// C#10: 'scoped' type
class scoped { }
scoped s;
static void F(scoped x, ref scoped y) { }
```
```csharp
// C#10: 'scoped' identifier
bool scoped;
scoped = true;
```

### 2. Allow `ref` auto properties?
Support `ref` auto property declarations? If so, allow `set` or `init`?
```csharp
ref T P0 { get; }
ref T P1 { get; set; }
ref T P2 { get; init; }
```

### 3. Allow `ref` assignment in object initializers?
```csharp
ref struct R<T>
{
    public ref T F;
}

int i = 0;
var r = new R<int> { F = ref i };
```

### 4. Report diagnostic for constructor that does not set `ref` field?
Report a diagnostic for an explicit constructor that does not assign to a `ref` field, or that assigns a value?
```csharp
ref struct R<T>
{
    private ref T _t;

    public R() { } // warning: ref is not assigned

    public R(ref T t)
    {
        _t = t; // warning: assigning value not ref
    }
}
```

### 5. `scoped` differences in overrides and interface implementations
Report a diagnostic for mismatches of `scoped` across overrides or implicit or explicit interface implementations? Allow "safe" variance differences?
```csharp
ref struct R<T> { }

abstract class A<T>
{
    public abstract R<T> F1(scoped R<T> r);
    public abstract R<T> F2(R<T> r);
}

class B : A<int>
{
    public override R<int> F1(R<int> r) => r;              // error?
    public override R<int> F2(scoped R<int> r) => default; // ok?
}
```
_Proposed: Report errors for any differences in `scoped`, even "safe" differences._

### 6. `scoped` differences in delegate conversions
Report a diagnostic for mismatches of `scoped` in conversions from lambda expressions or method groups to delegate types? Allow "safe" variance differences?
```csharp
ref struct R { }

delegate R D1(scoped R r);
delegate R D2(R r);

D1 i1 = (R r) => r;              // error?
D2 i2 = (scoped R r) => default; // ok?

D1 e1 = (D1)((R r) => r);              // error?
D2 e2 = (D2)((scoped R r) => default); // ok?
```
_Proposed: Report errors for any differences in `scoped`, even "safe" differences._

[...]

PR #7369, opened 2023-07-22
Create spec for lambda params with modifiers without type name

Rekkonnect:
Creates a spec for #338

This permits the simplified declaration of lambdas with parameters with modifiers, by omitting their type name.

### DIFF
--- a/proposals/ref-out-lambda-params.md
+++ b/proposals/ref-out-lambda-params.md
@@ -3,14 +3,19 @@
 ## Summary  
 
 Allow lambda parameter declarations with `in`/`ref`/`out` to be declared without requiring their type names.
+
+Given this delegate:
 ```cs
-// Given this delegate
 delegate bool TryParse<T>(string text, out T result);
+```
 
-// Allow this simplified parameter declaration
+Allow this simplified parameter declaration:
+```cs
 TryParse<int> parse1 = (text, out result) => Int32.TryParse(text, out result);
+```
 
-// Currently only this is valid
+Currently only this is valid:
+```cs
 TryParse<int> parse2 = (string text, out int result) => Int32.TryParse(text, out result);
 ```
 
@@ -30,3 +35,5 @@ The change in the spec will require that, in [the grammar for lambda expressions
 ```
 
 The type of the parameters matches the type of the parameter in the target delegate type, including the by-reference modifiers.
+
+Attributes on the parameters will not be affected in any way. Similarly, `async` lambdas will also not be affected from this change.
--- a/proposals/ref-out-lambda-params.md
+++ b/proposals/ref-out-lambda-params.md
@@ -1,0 +1,32 @@
+# Declaration of `in`/`ref`/`out` lambda parameters without type name
+
+## Summary  
+
+Allow lambda parameter declarations with `in`/`ref`/`out` to be declared without requiring their type names.
+```cs
+// Given this delegate
+delegate bool TryParse<T>(string text, out T result);
+
+// Allow this simplified parameter declaration
+TryParse<int> parse1 = (text, out result) => Int32.TryParse(text, out result);
+
+// Currently only this is valid
+TryParse<int> parse2 = (string text, out int result) => Int32.TryParse(text, out result);
+```
+
+## Detailed design
+
+### Parameter declaration
+
+Parameter declarations in lambda expressions now permit a single identifier after an `in`/`ref`/`out` modifier on the parameter.
+
+The change in the spec will require that, in [the grammar for lambda expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12191-general), the `implicit_anonymous_function_parameter` rule must be adjusted as follows:
+
+```diff
+  implicit_anonymous_function_parameter
+-    : identifier
++    : anonymous_function_parameter_modifier? identifier
+     ;
+```
+
+The type of the parameters matches the type of the parameter in the target delegate type, including the by-reference modifiers.
--- a/proposals/ref-out-lambda-params.md
+++ b/proposals/ref-out-lambda-params.md
@@ -46,7 +46,15 @@ delegate T SelfReturnerParams<T>(params T[] t);
 
 are all illegal, due to ambiguity with taking the reference of the returned expression in the `ref` case. For consistency, all other modifiers are also left unsupported and illegal.
 
-Using the `scoped` modifier alone is unsupported, because `scoped` is currently parsed as a type identifier, thus resolving to an explicit lambda parameter declaration.
+Using the `scoped` modifier alone is supported, since it was explicitly ruled out as a type name without the presence of `@` before the identifier in C# 11. This means that the following code will work:
+
+```csharp
+SelfReturnerScoped<string> frr = (scoped x) => x;
+
+delegate T SelfReturnerScoped<T>(scoped T t);
+```
+
+with `x` being resolved as an implicitly-typed lambda parameter with the `scoped` modifier.
 
 The change in the spec will require that [the grammar for lambda expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12191-general) be adjusted as follows:
 
--- a/proposals/ref-out-lambda-params.md
+++ b/proposals/ref-out-lambda-params.md
@@ -1,8 +1,8 @@
-# Declaration of lambda parameters with by-reference modifiers without type name
+# Declaration of lambda parameters with modifiers without type name
 
 ## Summary  
 
-Allow lambda parameter declarations with by-reference modifiers (`in` / `ref` / `out` / `ref readonly`) to be declared without requiring their type names.
+Allow lambda parameter declarations with modifiers (`in` / `ref` / `out` / `ref readonly` / `scoped` / `scoped ref` / `params`) to be declared without requiring their type names.
 
 Given this delegate:
 ```cs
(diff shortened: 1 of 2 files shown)


CyrusNajmabadi:
Can you add a line saying this is only for parenthesized lambdas?

proposals/ref-out-lambda-params.md
```diff
@@ -23,17 +23,39 @@ TryParse<int> parse2 = (string text, out int result) => Int32.TryParse(text, out
 
 ### Parameter declaration
 
-Parameter declarations in lambda expressions now permit a single identifier after an `in`/`ref`/`out` modifier on the parameter.
+Parameter declarations in lambda expressions with parenthesized parameters now permit a single identifier after an `in`/`ref`/`out` modifier on the parameter. This does not apply to lambda expressions with a single parameter with omitted parentheses.
```
CyrusNajmabadi, line 26: Note. We may have to also support `ref readonly ` as that feature is currently being worked on. If that impacts delegates, it would impact this.  @jjonescz ?

proposals/ref-out-lambda-params.md
```diff
@@ -23,17 +23,39 @@ TryParse<int> parse2 = (string text, out int result) => Int32.TryParse(text, out
...
--    : identifier
+  implicit_anonymous_function_parameter_list
+-     : implicit_anonymous_function_parameter
+-       (',' implicit_anonymous_function_parameter)*
++     : implicit_parenthesized_anonymous_function_parameter
++       (',' implicit_parenthesized_anonymous_function_parameter)*
+      ;
+
++ implicit_parenthesized_anonymous_function_parameter
++    : identifier
```
CyrusNajmabadi, line 52: This rule seems superfluous given the one below.

proposals/ref-out-lambda-params.md
```diff
@@ -0,0 +1,83 @@
...
++    : anonymous_function_parameter_modifier? identifier
+     ;
+```
+
+The new grammar rule is added so that this change specifically affects parenthesized lambda expression parameters, leaving lambda expressions with an unparenthesized parameter unsupported.
```
CyrusNajmabadi, line 57: Sentence seems superfluous. You already mentioned this above.

proposals/ref-out-lambda-params.md
```diff
@@ -0,0 +1,83 @@
...
+    b = c;
+}
+```
+
+This remains illegal as the current behavior for implicit-typed parameters without modifiers does not infer the type of the parameters through usage inside the body of the lambda expression. For example, the following is illegal:
+```csharp
+// Error: The delegate type could not be inferred
+var dd = (a, b) => Method2(a, b);
+
+int Method2(int a, int b) => a + b;
```
CyrusNajmabadi, line 82: Could just inline this method to simplify the example.

CyrusNajmabadi:
Speclet lgtm.  @333fred for another pair of eyes.

proposals/ref-out-lambda-params.md
Rekkonnect: In this example I'm trying to showcase that, even though the actual parameter types can be inferred from the usage of the parameters inside the body (through `Method2`), they are not being inferred. If that was not clear enough, I can mention the intend of the example.

CyrusNajmabadi: wfm.

proposals/ref-out-lambda-params.md
Rekkonnect: So do we also declare the support for `ref readonly` in the spec before its approval?

CyrusNajmabadi: Ref-readonly is approved and shipping for 12.  Subsequent proposals should incorporate the fallout from them. :-)

Rekkonnect:
An edge case includes the addition of the `scoped` modifier in C# 11. Currently, lambda parameters accept the scoped modifier, therefore those could also be affected from this spec. We want to decide whether those are being supported or not.

Rekkonnect:
Also, in preview, `params` in lambda parameters is being supported. We also need to consider this case.

proposals/ref-out-lambda-params.md
```diff
@@ -0,0 +1,82 @@
...
+
+delegate T SelfReturnerIn<T>(in T t);
+delegate T SelfReturnerRef<T>(ref T t);
+delegate T SelfReturnerOut<T>(out T t);
+delegate T SelfReturnerRefReadonly<T>(ref readonly T t);
+```
+
+are all illegal, due to ambiguity with taking the reference of the returned expression in the `ref` case. For consistency, `in`, `out` and `ref readonly` are also left unsupported and illegal.
+
+The change in the spec will require that, [the grammar for lambda expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12191-general) must be adjusted as follows:
```
333fred, line 43:
```suggestion
The change in the spec will require that [the grammar for lambda expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#12191-general) be adjusted as follows:
```

proposals/ref-out-lambda-params.md
```diff
@@ -0,0 +1,82 @@
...
+```diff
+  implicit_anonymous_function_parameter_list
+-     : implicit_anonymous_function_parameter
+-       (',' implicit_anonymous_function_parameter)*
++     : implicit_parenthesized_anonymous_function_parameter
++       (',' implicit_parenthesized_anonymous_function_parameter)*
+      ;
+
++ implicit_parenthesized_anonymous_function_parameter
++     : anonymous_function_parameter_modifier? identifier
```
333fred, line 54: Nit: I'd prefer to define this in terms of the existing parameter grammar, rather than copying the content of the parameter.

```suggestion
+     : anonymous_function_parameter_modifier? implicit_anonymous_function_parameter
```

proposals/ref-out-lambda-params.md
```diff
@@ -0,0 +1,82 @@
...
+      ;
+```
+
+The type of the parameters matches the type of the parameter in the target delegate type, including the by-reference modifiers.
+
+Attributes on the parameters will not be affected in any way. Similarly, `async` lambdas will also not be affected from this change.
```
333fred, line 60: I'm not sure what "Similarly, `async` lambdas will also not be affected from this change." means here. They will be affected by this change; it still will be illegal to have `RefFunc<int, Task> f = async (ref i) => { ... };`, but not because the grammar forbids the syntax, it's a semantic error just like having async method take ref parameters is.

333fred (commented):
A few small comments, overall this is looking good.

333fred (commented):
@CyrusNajmabadi this looks good enough to go to LDM to me. Want to add it to "schedule when convenient"?

CyrusNajmabadi:
@333fred I was going to check with @MadsTorgersen @KathleenDollard first when teh right time was to bring things since we're so heads down in 12 right now.  Basically, i'd bring this the next time we're open to looking for the next batch of championed constructs to take :)

333fred:
I believe that's what "schedule when convenient" there for 😉.

CyrusNajmabadi:
Ah, i thought those were for meetings where we didn't have anything pressing, but we were still prior to the point that we were trying to decide on new features :)

Rekkonnect:
> As for `scoped` parameters, since `scoped` is contextual, the compiler currently attempts to parse that as a type. Perhaps we will not make a move to support `(scoped x)` to infer the type of `x` and granting it the `scoped` modifier.

As per the [list of breaking changes](https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/Compiler%20Breaking%20Changes%20-%20DotNet%207.md#types-cannot-be-named-scoped) for C# 11, `scoped` no longer refers to a type name, therefore this restriction should not take place. `scoped` can be safely evaluated as a modifier in an implicit type lambda parameter list.

Therefore:
```csharp
private delegate int D(scoped int x, scoped int y);
D d = (scoped x, scoped y) => x + y;
```
is valid and `scoped` reflects the modifier, not a type named `scoped`.

Will update the spec accordingly to reflect this change, citing the above reference.

jjonescz:
> `scoped` no longer refers to a type name

Is that right? Seems the breaking change only disallows declaring new types named `scoped`, but you could reference an existing type named `scope` from a DLL and then this would be a break (not saying the break wouldn't be acceptable):

```cs
var d = (scoped x, scoped y) => x + y;
```

jjonescz:
> ```cs
> ref scoped local; // Error
> ```

Maybe that's an error only when there's `ref`.

> Hints that all available types named `scoped` are now inaccessible. 

This test succeeds for me:

```cs
[Fact]
public void Scoped()
{
    var source1 = """
        public class scoped { }
        """;

    CreateCompilation(source1).VerifyDiagnostics(
        // (1,14): error CS9062: Types and aliases cannot be named 'scoped'.
        // public class scoped { }
        Diagnostic(ErrorCode.ERR_ScopedTypeNameDisallowed, "scoped").WithLocation(1, 14));

    var comp1 = CreateCompilation(source1, parseOptions: TestOptions.Regular10).VerifyDiagnostics(
        // (1,14): warning CS8981: The type name 'scoped' only contains lower-cased ascii characters. Such names may become reserved for the language.
        // public class scoped { }
        Diagnostic(ErrorCode.WRN_LowerCaseTypeName, "scoped").WithArguments("scoped").WithLocation(1, 14));
    var comp1Ref = comp1.EmitToImageReference();

    var source2 = """
        var d = (scoped x) => System.Console.WriteLine(x);
        d(new scoped());
        """;
    CompileAndVerify(source2, new[] { comp1Ref }, expectedOutput: "scoped").VerifyDiagnostics();
}
```

Rekkonnect:
Nice observation, with a little bit more testing, even the listed breaking change does not currently apply:
[SharpLab link](https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQ6DEUIBtcCGARrgKYAEAJgJYDOxZ66tAxgPYAOpl5AZuQF5yOfAG50YUv1aduffkMn9e4tDK6UAdADlStYHKiDheXKqXl13HXoM8oRxVOGr0icnDhXK6AN7pyQI8AZg8Ecl19OV9yAF90WKA===)

Note that the above uses the Default branch which is C# 11 based on the fact that the error about naming a type `scoped` exists.

This is either an undocumented decision, or an omission in the feature. We need a definite answer to how we're going to support `scoped` in implicit lambda parameter lists.

CyrusNajmabadi:
if we're going to support `ref` and the like without an explicit type, then we need to support `scoped` in the same fashoin.

CyrusNajmabadi:
I would put it as a case to consider and allow the ldm to make a determination on if it's ok or not.  

Personally, I would take the break.

Rekkonnect:
Digging through the issues, I also discovered that #6149 is touching the subject of `scoped`, perhaps it would be helpful to mention it here for future reference once the spec is settled.

proposals/ref-out-lambda-params.md
```diff
@@ -0,0 +1,137 @@
...
+TryParse<int> parse2 = (string text, out int result) => Int32.TryParse(text, out result);
+```
+
+## Detailed design
+
+### Parameter declaration
+
+Parameter declarations in lambda expressions with parenthesized parameters now permit a single identifier after a modifier on the parameter. This does not apply to lambda expressions with a single parameter with omitted parentheses.
+
+For example,
```
CyrusNajmabadi, line 28:
```suggestion
For example, these would not be legal.
```

Also, just do like the first 3 examples.

CyrusNajmabadi (approved):

proposals/ref-out-lambda-params.md
```diff
@@ -83,67 +107,5 @@ var dd = (a, b) => Method2(a, b);
...
-  : 'ref'
-  | 'out'
-  | 'in'
-  | 'scoped'
-  | 'readonly'
-  | 'params'
-  ;
-```
 
 Should the above production rule be updated to only reflect the valid combinations of modifiers, the `implicit_parenthesized_anonymous_function_parameter` rule is updated accordingly to reflect the applicable modifiers.
```
Rekkonnect, line 111: I think this note should also be removed

CyrusNajmabadi (approved):

PR #7370, opened 2023-07-22
Collection expressions: updates for builder methods

cston:
Update *create method* section; remove *extension method invocation*.

### DIFF
--- a/proposals/collection-expressions.md
+++ b/proposals/collection-expressions.md
@@ -378,7 +378,6 @@ static class Extensions
 var x = [1].AsImmutableArray();           // error: collection expression has no target type
 var y = [2].AsImmutableArray<int>();      // error: ...
 var z = Extensions.AsImmutableArray([3]); // ok
-
 ```
 
 ## Open questions
--- a/proposals/collection-expressions.md
+++ b/proposals/collection-expressions.md
@@ -197,8 +197,8 @@ For the *create method*:
 - The method must be defined on the *builder type* directly.
 - The method must be `public` and `static`.
 - The *arity* of the method must match the *arity* of the collection type.
-- The method must have a single parameter of type `System.ReadOnlySpan<E>`, passed by value, where `E` is the [*iteration type*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/statements.md#1295-the-foreach-statement) of the corresponding *collection type*.
-- The method return type must be the corresponding *collection type*.
+- The method must have a single parameter of type `System.ReadOnlySpan<E>`, passed by value, and there is an [*identity conversion*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/conversions.md#1022-identity-conversion) between `E` and the [*iteration type*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/statements.md#1295-the-foreach-statement) of the *collection type*.
+- There is an [*identity conversion*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/conversions.md#1022-identity-conversion) between the method return type and the *collection type*.
 
 An error is reported if the `[CollectionBuilder]` attribute does not refer to an invocable method with the expected signature.
 
--- a/proposals/collection-expressions.md
+++ b/proposals/collection-expressions.md
@@ -112,7 +112,22 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
 
     * The behavior of collection literals with collections that are not well-behaved is undefined.
 
+## Element type
+[element-type]: #element-type
+
+Certain uses of *collection expressions* &mdash; [*type inference*](#type-inference) and [*create methods*](#create-methods) &mdash; require a target type with a known *element type*.
+
+The *element type* of a type is defined as:
+
+* For a single dimensional *array type* `T[]`, the *element type* is `T`.
+* For a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>`, the *element type* is `T`.
+* For an *[inline array type](https://github.com/dotnet/csharplang/blob/main/proposals/inline-arrays.md)* with *element type* `T`, the *element type* is `T`.
+* For a *type* that implements `System.Collections.IEnumerable`, the *element type* is `System.Object`.
+* For an *interface type* `I<T0>` where `System.Collections.Generic.List<T>` implements `I<T>`, the *element type* is `T0`.
+* Otherwise the type does not have an *element type*.
+
 ## Conversions
+[conversions]: #conversions
 
 A *collection literal conversion* allows a collection literal expression to be converted to a type.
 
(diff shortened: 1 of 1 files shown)


proposals/collection-expressions.md
```diff
@@ -112,7 +112,22 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
...
+Certain uses of *collection expressions* &mdash; [*type inference*](#type-inference) and [*create methods*](#create-methods) &mdash; require a target type with a known *element type*.
+
+The *element type* of a type is defined as:
+
+* For a single dimensional *array type* `T[]`, the *element type* is `T`.
+* For a *span type* `System.Span<T>` or `System.ReadOnlySpan<T>`, the *element type* is `T`.
+* For an *[inline array type](https://github.com/dotnet/csharplang/blob/main/proposals/inline-arrays.md)* with *element type* `T`, the *element type* is `T`.
+* For a *type* that implements `System.Collections.IEnumerable`, the *element type* is `System.Object`.
+* For an *interface type* `I<T0>` where `System.Collections.Generic.List<T>` implements `I<T>`, the *element type* is `T0`.
+* Otherwise the type does not have an *element type*.
```
CyrusNajmabadi, line 127: i'm actually surrpised it's defined this way.  For example, the above seems like it would mean that a type that is `foreachable` but not any of the above, would not have an element type?   my understanding is that foreach operates directly as a pattern, looking for a `GetEnumerator()` with a certain shape, which returns an type with a certain shape.  I would expect it would the be the type of `X Current { get; }` that would define the element type.

proposals/collection-expressions.md
```diff
@@ -188,23 +201,26 @@ namespace System.Runtime.CompilerServices
...
 The attribute is not inherited although the attribute can be applied to a base `class` or an `abstract class`.
 
-An error is reported for a collection literal if there are multiple `[CollectionBuilder]` attributes on the collection type.
+The collection type must have an [*element type*](#element-type).
 
-The *create method* must be a `static` method defined on the *builder type*.
-The *create method* must have a single parameter of a type `System.ReadOnlySpan<T0>` passed by value, and a return value of the *collection type* to which the attribute is applied.
+For the *create method*:
+- The *builder type* must be a non-generic `class` or `struct`.
+- The method must be defined on the *builder type* directly.
```
CyrusNajmabadi, line 211: maybe clarify that that means it cannot appear in a base type? (just a nit, this is probably sufficient).

CyrusNajmabadi (approved):

proposals/collection-expressions.md
```diff
@@ -188,23 +201,26 @@ namespace System.Runtime.CompilerServices
...
+- The method must be `public` and `static`.
+- The *arity* of the method must match the *arity* of the collection type.
+- The method must have a single parameter of type `System.ReadOnlySpan<E>`, passed by value, where `E` is the [*element type*](#element-type) of the corresponding *collection type*.
+- The method return type must be the corresponding *collection type*.
```
CyrusNajmabadi, line 215: i still don't love this.  i would say that it either has to have a implicit-reference-conversion to the corresponding collection type.  Or just an implicit-conversion to the collection type.

It seems both unnecessarily restrictive *and* actually undesirable given that factory methods might return concrete types, but still be intended for other types (like interfaces).

proposals/collection-expressions.md
```diff
@@ -188,23 +201,26 @@ namespace System.Runtime.CompilerServices
...
 
-The *builder type* cannot be generic.
-The *builder method* can be generic and the *arity* of the builder method must match the *arity* of the *collection type*.
+An error is reported if the `[CollectionBuilder]` attribute does not refer to an invocable method with the expected signature.
 
-Overloads with distinct signatures on the builder type, and overloads on base types, are ignored.
+Method overloads on the *builder type* with distinct signatures are ignored. Methods declared on base types or interfaces are ignored.
```
CyrusNajmabadi, line 219: ah, you clarified the base-type thing here.  ignore my above comment then.

CyrusNajmabadi:
Done with pass.

CyrusNajmabadi (approved):

proposals/collection-expressions.md
CyrusNajmabadi: basically, i woudl expect the element type to link to the algorithm here: https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/statements.md#1295-the-foreach-statement (or some variation thereof).  Basically:

> The compile-time processing of a foreach statement first determines the collection type, enumerator type and iteration type of the expression. This determination proceeds as follows:

Here, the `element type` seems to equivalent to the `iteration type` specified in this other part of the spec (modulo if we want to make any tweaks).  But it seems to serve better than writing out this different set of rules (which may not be as inclusive, or may deviate in unintended ways?).

Also, wrt the impl, are we sharing the code to figure this out with the foreach code?  If so, then unifying in the spec seems very preferable.

cston: I overlooked the `GetEnumerator()` pattern here. As you suggested, if the type implements the `GetEnumerator()` pattern, we should the `Current` property for the element type.

That said, I think the definition of *element type* might differ from *`foreach` iteration type* in two ways: extensions and accessibility.

In short, the element type should not depend on where the *collection expression* is used. Instead, we should be able to determine the element type at the definition of the target type rather than at the use-site. If so, that might mean the `GetEnumerator()` implementation should be a `public` instance method, and should not depend on extension methods or accessibility.

CyrusNajmabadi: > and should not depend on extension methods or accessibility

So I personally disagree. Since it feels like anything foreachable should reasonably both be able to contribute one of these element types, and then work in an these collection-expr scenarios.

cston: Allowing extension method `GetEnumerator()` may lead to surprising behavior when the extension method is not in scope or inaccessible.

For instance, in the following example, a `MyCollection<T>` is constructed successfully in one case, but not the other.

Should we report a diagnostic (perhaps a warning) at the declaration of `class MyCollection<T>` when there is no instance `GetEnumerator()` that use-sites may not be able to construct the collection from a collection expression?

```csharp
using System;

[CollectionBuilder(typeof(MyCollectionBuilder), nameof(MyCollectionBuilder.Create))]
class MyCollection<T> { }

class MyCollectionBuilder
{
    public static MyCollection<T> Create<T>(ReadOnlySpan<T> items) => default;
}

namespace N
{
    static class Extensions
    {
        public static IEnumerator<T> GetEnumerator<T>(this MyCollection<T> c) => default;
        static MyCollection<T> F<T>() => []; // ok
    }
}

class Program
{
    static void Main()
    {
        MyCollection<int> c = []; // error: CollectionBuilderAttribute but no element type
    }
}
```

CyrusNajmabadi: I don't think it's any more surprising than being able to foreach or not over instances based on it the extension is in scope :-)

proposals/collection-expressions.md
```diff
@@ -112,7 +112,22 @@ Collection literals are [target-typed](https://github.com/dotnet/csharplang/blob
 
     * The behavior of collection literals with collections that are not well-behaved is undefined.
 
+## Element type
+[element-type]: #element-type
+
+Certain uses of *collection expressions* &mdash; [*type inference*](#type-inference) and [*create methods*](#create-methods) &mdash; require a target type with a known *element type*.
```
333fred, line 118:
```suggestion
Certain uses of *collection expressions* – [*type inference*](#type-inference) and [*create methods*](#create-methods) – require a target type with a known *element type*.
```

proposals/collection-expressions.md
333fred, line 215: This implies that `object` != `dynamic` for these purposes. Is what you wanted?

proposals/collection-expressions.md
RikkiGibson: Stuff you can add with just extensions:
- ability to 'foreach' (extension GetEnumerator)

Stuff you can add with both declaration changes+extensions:
- ability to use collection initializers. Declaration needs to implement IEnumerable but an extension method can implement Add.

Stuff you can add only with declaration changes
- ability to use a `Create(ROS<T>)` method for a collection expression. The declaration needs to also add CollectionBuilderAttribute. Extension GetEnumerator doesn't allow us to determine the element type. It needs to be declared on the collection type.

On a certain level, I feel like we've decided to just focus on having original declarations onboard to this pattern, and I feel totally fine with that, as long as we have some idea of how they could be added through extensions in the future, which I think we do.

CyrusNajmabadi: > On a certain level, I feel like we've decided to just focus on having original declarations onboard to this pattern, 

I don't recall ever makgin this decision.  And, personally, it goes against the spirit of Collection Exprs.  The point of collection exprs is to be encompassing.  If people have their own collections then they should work naturally with collection exprs.

We're not creating a new, opinionated, system that intends to leave things behind.

proposals/collection-expressions.md
```diff
@@ -188,23 +187,26 @@ namespace System.Runtime.CompilerServices
...
 
-The *builder type* cannot be generic.
-The *builder method* can be generic and the *arity* of the builder method must match the *arity* of the *collection type*.
+An error is reported if the `[CollectionBuilder]` attribute does not refer to an invocable method with the expected signature.
 
-Overloads with distinct signatures on the builder type, and overloads on base types, are ignored.
+Method overloads on the *builder type* with distinct signatures are ignored. Methods declared on base types or interfaces are ignored.
 
-An error is reported for a collection literal if the `[CollectionBuilder]` attribute on the collection type does not represent an invocable method with the expected signature defined on the *builder type*.
+For a *collection expression* with a target type <code>C&lt;S<sub>0</sub>, S<sub>1</sub>, &mldr;&gt;</code> where the *type declaration* <code>C&lt;T<sub>0</sub>, T<sub>1</sub>, &mldr;&gt;</code> has an associated *builder method* <code>B.M&lt;U<sub>0</sub>, U<sub>1</sub>, &mldr;&gt;()</code>, the *generic type arguments* from the target type are applied in order &mdash; and from outermost containing type to innermost &mdash; to the *builder method*.
```
RikkiGibson, line 207: I think this language is fine, but thought it might be useful to also link how the interceptors design described a similar thing. https://github.com/dotnet/roslyn/blob/main/docs/features/interceptors.md#arity

RikkiGibson (approved):

PR #7417, opened 2023-08-03
Create CL-2023-08-03.md

CyrusNajmabadi:


meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,121 @@
+# Collection literals working group meeting for August 3, 2023
+
+## Agenda
+
+* Collection expressions and ref-structs (like `ReadOnlySpan<T>`).
+* Confirm behavior collection expressions and `ReadOnlySpan`s of blittable data.
+
+## Discussion
+
+The purpose of the meeting was to discuss what semantics we would have in the language for a collection literal assigned to a ref-struct.
```
CyrusNajmabadi, line 10:
```suggestion
The purpose of the meeting was to discuss what semantics we would have in the language for a collection literal target-typed to a ref-struct.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
+# Collection literals working group meeting for August 3, 2023
+
+## Agenda
+
+* Collection expressions and ref-structs (like `ReadOnlySpan<T>`).
+* Confirm behavior collection expressions and `ReadOnlySpan`s of blittable data.
```
cston, line 6:
```suggestion
* Collection expression escape scope for ref-structs (like `ReadOnlySpan<T>`).
* Confirm behavior of collection expressions and `ReadOnlySpan`s of blittable data.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Discussion
+
+The purpose of the meeting was to discuss what semantics we would have in the language for a collection literal target-typed to a ref-struct.
```
cston, line 10:
```suggestion
The purpose of the meeting was to discuss what semantics we would have in the language for a collection expression target-typed to a ref-struct.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+```c#
+ReadOnlySpan<int> x1 = [a, b, c];        // heap allocated, global scope.
+scoped ReadonlySpan<int> x2 = [a, b, c]; // stack allocated, local scope.
+```
+
+This has the positive that both heap and local allocations are possible, and choosing between them is as simple as adding `scoped` or not.  Unfortunately it has several major negatives the working group was quite dissatisfied with:
+
+1. We believe the *common* case for using spans with collection expressions will be for people wanting stack allocated data.  Having the default (non-`scoped`) syntax basically flips that and makes the common case more cluttered and less pleasant to work with.
+2. We believe the standard intuition for nearly all users (including experts in the low-level) space would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types, and their belief would be that "of course if i initialized with a collection expression" that "it will do the right thing".
```
cston, line 49:
```suggestion
2. We believe the standard intuition for nearly all users (including experts in the low-level space) would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types, and their belief would be that "of course if I initialized with a collection expression" that "it will do the right thing".  
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+2. We believe the standard intuition for nearly all users (including experts in the low-level) space would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types, and their belief would be that "of course if i initialized with a collection expression" that "it will do the right thing".  
+3. We believe that the first line would actually be a *large* "pit of fail" for users.  It would be very easy and attractive for customers to just take existing code, hear about collection expressions (including that it works for spans) and translate `stackalloc` code over to just using the collection expression.  This would then result in near immediate performance tanking due to potentially extreme allocations happening in hotspots that didn't exist before.  Such problems could even easily lead to this feature being entirely banned in codebases due to the fear of this happening.
```
cston, line 50: Perhaps:
```suggestion
3. We believe that the first line would actually be a *large* "pit of fail" for users.  It would be very easy and attractive for customers to just take existing code, hear about collection expressions (including that it works for spans) and translate `stackalloc` code over to just using the collection expression, resulting in allocations that didn't exist before.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+4. We believe if we had this behavior we would have to have a diagnostic (likely on by default) that then immediately warned the user that they were getting a heap allocation on 'x1' and to either:
```
cston, line 51:
```suggestion
4. We believe if we had this behavior, we would have to have a diagnostic (likely on by default) that warned the user of the heap allocation on 'x1' and to either:
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+    1. suppress the diagnostic if they really wanted that, or
+    2. put an explicit cast (e.g. `(int[])[a, b, c]`) to indicate that they indeed did want this behavior, or
+    3. add the `scoped` modifier.
+
+All in all, while this allowed for both cases to easily be distinguished with a simple modifier, everything about the defaults and behavior was flipped from where we wanted it to be.  This then led us to option 2:
+
+### Option 2. Determine what scoping a target has based on usage.
```
cston, line 58: Perhaps just:
```suggestion
### Option 2. Use target scope.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+A discussion has been going on about allowing the user to just write:
+
+```c#
+ReadOnlySpan<int> x1 = [a, b, c]; // Could be local or global scoped
+scoped ReadOnlySpan<int> x2 = [a, b, c]; // Always local scoped.
+```
+
+and then determining the scoping based on how the variable is used.  For example, if 'x1' does not escape, then it could be stack allocated.  If it does end up escaping though (for example with `return x1;`), the collection expression would be heap allocated.  'x2' would always be local scoped of course and would be prevented from escaping.
```
cston, line 67: I believe the discussion was to use the static scope of the target, independent of how the target is used.
```c#
ReadOnlySpan<int> x1 = stackalloc int[1]; // local scope
...
x1 = [1, 2, 3];  // collection expression has local scope
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+```c#
+ReadOnlySpan<int> x4 = [a, b, c];
+```
+
+The working group considered several options.
+
+### Option 1.  Stack-alloc'ed only if target has local-scope.
+
+The first option considered was that a collection expression used with a `ReadOnlySpan<T>` would itself only be stack-alloc'ed if the span itself had an explicitly stated local scope.   In other words, the following:
```
cston, line 39: The target may have an implicit scope that we'd use too.
```c#
ReadOnlySpan<int> x3 = stackalloc int[1];  // x3 has local scope
...
x3 = [a, b, c]; // stack allocated, local scope
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+The benefit of this approach would be allowing the user to always write the simple form with a collection expression, and then get the semantics they needed to make the code legal.  
+
+This was attractive because of the simplicity of the user technically writing simple code, and then not having any stumbling blocks.  However, we felt this was actually a net negative.   A significant problem we forsee with this approach is that users will have code that is fast and nicely stack-allocating as they expect, and then some innocuous looking change ends up causing the compiler to conservatively think it may escape, causing the collection to now be heap allocated.  This could happen silently, easily tanking program performance.  Or, it could come with a warning.  But in the latter case, users would then have to suppress or introduce explicit casts.
```
cston, line 71:
```suggestion
This was attractive because of the simplicity of the user technically writing simple code, and then not having any stumbling blocks.  However, we felt this was actually a net negative.   A significant problem we foresee with this approach is that users will have code that is fast and nicely stack-allocating as they expect, and then some innocuous looking change ends up causing the compiler to conservatively think it may escape, causing the collection to now be heap allocated.  This could happen silently, easily tanking program performance.  Or, it could come with a warning.  But in the latter case, users would then have to suppress or introduce explicit casts.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+We then moved to a discussion around our belief that when using ref-structs, and collection expressions, users would expect that data to be put on the stack.  Exploring this led to the following semantics:
+
+```c#
+ReadOnlySpan<int> x1 = [a, b, c];        // stack allocated, local scope.
+scoped ReadonlySpan<int> x2 = [a, b, c]; // stack allocated, local scope.
+```
+
+In other words, the above two have the same meaning.  With the latter just being explicit about the local scoping.  Returning either of those from a method would be illegal.  This has the benefit that the simple form matches the intuition nearly all users will want this to have, and will provide the best performance by default.  This heavily aligns with the stated goal of collection expressions (which we will continue to evangelize it with) that it makes the right default choices and by using it you should always (*ideally*) be getting the best performance results.  
+
+This does have some downsides though. First, the aforementioned redundancy if you actually include `scoped`.  We could consider warning or blocking that to prevent confusion and to help convey that the data is always scoped.
```
cston, line 88: We have the redundancy today with `scoped ReadOnlySpan<int> x3 = stackalloc int[1];`.

meetings/working-groups/collection-literals/CL-2023-08-03.md
jnm2:
```suggestion
2. We believe the standard intuition for nearly all users (including experts in the low-level space) would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types. Their belief would be that if they initialize with a collection expression, will do the right thing.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+The above discussion applied to collection expressions, target typed to ref-structs, *without* constant, blittable data in it.  Blittable data is effectively defined as C# builtin of 1-8 bytes in size.  Intuitively, the primitive integers, floating points, and characters.  
+
+Today, the language already allows optimizing the following:
+
+```c#
+private static ReadOnlySpan<char> s_chars = new char[] { 'a', 'b', 'c' };
+```
+
+Despite this explicitly having a 'new' in the code, and explicitly referencing the `char[]` type, the above will be compiled without any allocations or array types.  Instead, the data will be embedded directly into the data segment of the compiled assembly, and the span will be created pointing directly at that raw data.
```
cston, line 125:
```suggestion
Despite this explicitly having a `new` in the code, and explicitly referencing the `char[]` type, the above will be compiled without any allocations or array types.  Instead, the data will be embedded directly into the data segment of the compiled assembly, and the span will be created pointing directly at that raw data. 
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,140 @@
...
+
+```c#
+ReadOnlySpan<int> x1 = [a, b, c];        // heap allocated, global scope.
+scoped ReadonlySpan<int> x2 = [a, b, c]; // stack allocated, local scope.
+```
+
+This has the positive that both heap and local allocations are possible, and choosing between them is as simple as adding `scoped` or not.  Unfortunately it has several major negatives the working group was quite dissatisfied with:
+
+1. We believe the *common* case for using spans with collection expressions will be for people wanting stack allocated data.  Having the default (non-`scoped`) syntax basically flips that and makes the common case more cluttered and less pleasant to work with.
+2. We believe the standard intuition for nearly all users (including experts in the low-level space) would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types, and their belief would be that "of course if I initialized with a collection expression" that "it will do the right thing".
```
jnm2, line 49:
```suggestion
2. We believe the standard intuition for nearly all users (including experts in the low-level space) would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types. Their belief would be that if they initialize with a collection expression, will do the right thing.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
CyrusNajmabadi: fair.  let me try to revise.

meetings/working-groups/collection-literals/CL-2023-08-03.md
CyrusNajmabadi: will flip.

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,146 @@
...
+
+// Less likely, but shown for completeness:
+ReadOnlySpan<int> x3 = stackalloc int[] { d, e, f }; // stack allocated, local scope.
+x3 = [g, h, i];                                      // stack allocated.
+```
+
+This has the positive that both heap and local allocations are possible, and choosing between them is as simple as adding `scoped` or not.  Unfortunately it has several major negatives the working group was quite dissatisfied with:
+
+1. We believe the *common* case for using spans with collection expressions will be for people wanting stack allocated data.  Having the default (non-`scoped`) syntax basically flips that and makes the common case more cluttered and less pleasant to work with.
+2. We believe the standard intuition for nearly all users (including experts in the low-level space) would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types. Their belief would be that if they initialize with a collection expression, will do the right thing.
```
CyrusNajmabadi, line 53:
```suggestion
2. We believe the standard intuition for nearly all users (including experts in the low-level space) would be that the first line produces stack-allocated data.  After all, that's the most common reason to create locals of span types. Their belief would be that if they initialize with a collection expression, it will do the right thing.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,146 @@
...
+
+This does have some downsides though.
+
+First, if the user did want to return the ref struct, they would now be blocked.  The workaround for this would be to instead be explicit that you wanted heap allocation for this span (e.g. `ReadOnlySpan<int> x1 = (int[])[1, 2, 3];`).  However, we see this explicitness as a *good* thing.  When working with spans and collection-expressions, we feel that having code be explicit when it is on the heap is actually a desirable documentary property for it to have.  And, while the cast syntax is not attractive, we could consider a simpler way of expressing that in the future (e.g. `new [1, 2, 3]`), where the syntax was light, but clear about intent.
+
+Second, the aforementioned redundancy if you actually include `scoped`.  We could consider warning or blocking that to prevent confusion and to help convey that the data is always scoped.  That said, we already have this redundancy with `scoped ReadONlySpan<int> x = stackalloc int[] { ... };` and we give no warning.  So we likely can just do the same here.
```
jnm2, line 96:
```suggestion
Second, the aforementioned redundancy if you actually include `scoped`.  We could consider warning or blocking that to prevent confusion and to help convey that the data is always scoped.  That said, we already have this redundancy with `scoped ReadOnlySpan<int> x = stackalloc int[] { ... };` and we give no warning.  So we likely can just do the same here.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,146 @@
...
+
+The benefit of this approach would be allowing the user to always write the simple form with a collection expression, and then get the semantics they needed to make the code legal.  
+
+This was attractive because of the simplicity of the user technically writing simple code, and then not having any stumbling blocks.  However, we felt this was actually a net negative.   A significant problem we foresee with this approach is that users will have code that is fast and nicely stack-allocating as they expect, and then some innocuous looking change ends up causing the compiler to conservatively think it may escape, causing the collection to now be heap allocated.  This could happen silently, easily tanking program performance.  Or, it could come with a warning.  But in the latter case, users would then have to suppress or introduce explicit casts.
+
+Similar to option 1, it felt very strange for us to introduce behavior where the compiler would heap allocate for the user, only to warn since in the majority of times it was believed users would not want that.  After all, in the case of heap allocating, one might as well just choose a heap allocated type *to begin with*.
```
jnm2, line 77:
```suggestion
Similar to option 1, it felt very strange for us to introduce behavior where the compiler would heap allocate for the user only to warn, since in the majority of times it was believed users would not want that.  After all, in the case of heap allocating, one might as well just choose a heap allocated type *to begin with*.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,146 @@
...
+static class FrugalListBuilder
+{
+    public static FrugalList<T> Create(ReadOnlySpan<T> values);
+}
+
+// ...
+FrugalList<int> f = [a, b, c]; // Stack-allocated, local scope.
+```
+
+We believe that this is the semantics ref-struct-collections will want.  And, similar to spans, users would expect to get *without* having to annotate with `scoped` on the local variable.
```
jnm2, line 119:
```suggestion
We believe that this is the semantics ref-struct-collections will want.  And, similar to spans, the semantics users would expect to get *without* having to annotate with `scoped` on the local variable.
```

meetings/working-groups/collection-literals/CL-2023-08-03.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+In other words, if a collection literal has nothing but constant blittable data, and it is target typed to a `ReadOnlySpan<BlittableType>` *only*, then it will actually be considered to have `global` not `local` scope.  This means that the following would be allowed:
+
+```c#
+ReadOnlySpan<int> x = [1, 2, 3]; // pointer to data segment, global scope.
+return x;
+```
+
+No decisions were needed on this.  Working group just affirmed this is the desired semantics and these rules and optimizations will be maintained with collection expressions.
```
jnm2, line 146:
```suggestion
No decisions were needed on this.  Working group just affirmed these are the desired semantics and these rules and optimizations will be maintained with collection expressions.
```

jnm2 (approved):

cston (approved):

PR #7426, opened 2023-08-07
Create Supporting-Interface-TargetTypes.md

CyrusNajmabadi:


meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,70 @@
+## Notes for supporting interfaces.
+
+1.  Interfaces are important. Outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), collections are the next most important set.  Interfaces are used ~90% of the time.  And when used, 90% of those are IEnumerable<T> itself.  (These numbers are real)  If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs.
```
CyrusNajmabadi, line 3:
```suggestion
1.  Interfaces are important. For apis that take collections, outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), the `System.Collections.XXX` collection types are the next most commonly used set.   Of the types in that namespace, the interface collections `IEnumerable<T>/ICollection<T>/IReadOnlyCollection<T>/IList<T>/IReadOnlyList<T>` are used ~90% of the time.  And when used, 90% of those are `IEnumerable<T>` itself.  If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs.  
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,70 @@
+## Notes for supporting interfaces.
+
+1.  Interfaces are important. For apis that take collections, outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), the `System.Collections.XXX` collection types are the next most commonly used set.   Of the types in that namespace, the interface collections `IEnumerable<T>/ICollection<T>/IReadOnlyCollection<T>/IList<T>/IReadOnlyList<T>` are used ~90% of the time.  And when used, 90% of those are `IEnumerable<T>` itself.  If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs.
```
CyrusNajmabadi, line 3:
```suggestion
1.  Interfaces are important. For apis that take collections, outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), the `System.Collections.XXX` collection types are the next most commonly used set.   Of the types in that namespace, the interface collections `IEnumerable<T>/ICollection<T>/IReadOnlyCollection<T>/IList<T>/IReadOnlyList<T>` are used ~90% of the time.  And when used, 90% of those are `IEnumerable<T>` itself.  (If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs).  The large set of these methods (which extends out into the ecosystem as well), motivates supporting interfaces to have collection expressions feel complete at launch.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+   Because they core scenarios differ so wildly on their expectations around mutability and around the problems that arise (for either) by aligning on what the other does, it seems reasonable to state and explain things if we go down a path where there are different types and choices between natural and target types.
+
+4. Furthermore, alignment between the two systems (target and natural typing) could actually be highly negative if we promise a concrete collection type. For example, if we promise a `List<T>` type, then that eliminates the ability to use a singleton for the *very* common and *core* use-case of passing `[]` as an argument to `IEnumerable` apis (or initializing an `IEnumerable` property).  This would force users away from this, and require them to be as verbose as today to be explicit to get the desired savings.  This deeply undercuts are core narrative that you can use collection expressions and they can be the *efficient* type you can depend on.
```
CyrusNajmabadi, line 17:
```suggestion
4. Furthermore, alignment between the two systems (target and natural typing) could actually be highly negative if we promise a concrete collection type. For example, if we promise a `List<T>` type, then that eliminates the ability to use a singleton for the *very* common and *core* use-case of passing `[]` as an argument to `IEnumerable<T>` apis (or initializing an `IEnumerable<T>` property).  This would force users away from this, and require them to be as verbose as today to be explicit to get the desired savings.  This deeply undercuts are core narrative that you can use collection expressions and they can be the *efficient* type you can depend on.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+   Conversely, by not promising a concrete type, we get more opportunities to optimize, even for *non* empty cases.  For example, with:
+
+      ```c#
+      Goo(IEnumerable<int> values);
+      Goo([1, 2, 3])
+      ```
+
+   We can optimize in a way similar to what we do with `yield`ing iterators.  Namely, you would an `IEnumerable<T>` instance that was itself its own `IEnumerator<T>` instance (in the case where it has not been enumerated yet, and was being enumerated on the same thread that created it).  This would again be much more efficient than what you get today with any normal concrete type a user uses here (lists, arrays, etc.)   This again ties heavily into our story that by using collection expressions you can trust you're going to get the best results.
```
CyrusNajmabadi, line 26:
```suggestion
   We can optimize in a way similar to what we do with `yield`ing iterators.  Namely, you would an `IEnumerable<T>` instance that was itself its own `IEnumerator<T>` instance (in the case where it has not been enumerated yet, and was being enumerated on the same thread that created it).  This would again be much more efficient than what you get today with any normal concrete type a user uses here (lists, arrays, etc.), and again ties heavily into our story that by using collection expressions you can trust you're going to get the best results.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+   As before, promising too much can hurt us (and defy user expectations around performance and safety).
+
+5. Certain concrete types (specifically `List<T>` or `T[]`) have highly undesirable downsides for an interface.  Specifically, it would make it a very dangerous footgun for any API to expose an `IEnumerable<T>/IReadOnlyList<T>` property that was instantiated with a collection expression.  A set of users would be very unhappy and perturbed to discover that they switched away from an existing read-collection type to a collection expression and now had things become unsafe.  This would tie into the feeling that collection expressions could not be trusted to make sound decisions that people could trust.
```
CyrusNajmabadi, line 30:
```suggestion
5. Certain concrete types (specifically `List<T>` or `T[]`) have highly undesirable downsides for an interface target.  Specifically, it would make it a very dangerous footgun for any API to expose an `IEnumerable<T>/IReadOnlyList<T>` property that was instantiated with a collection expression.  A set of users would be very unhappy and perturbed to discover that they switched away from an existing read-collection type to a collection expression and now had things become unsafe.  This would tie into the feeling that collection expressions could not be trusted to make sound decisions that people could trust.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+   As before, promising too much can hurt us (and defy user expectations around performance and safety).
+
+5. Certain concrete types (specifically `List<T>` or `T[]`) have highly undesirable downsides for an interface target.  Specifically, it would make it a very dangerous footgun for any API to expose an `IEnumerable<T>/IReadOnlyList<T>` property that was instantiated with a collection expression.  A set of users would be very unhappy and perturbed to discover that they switched away from an existing read-collection type to a collection expression and now had things become unsafe.  This would tie into the feeling that collection expressions could not be trusted to make sound decisions that people could trust.
```
CyrusNajmabadi, line 30:
```suggestion
5. Certain concrete types (specifically `List<T>` or `T[]`) have highly undesirable downsides for an interface target.  Specifically, it would make it a very dangerous footgun for any API to expose an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` property that was instantiated with a collection expression.  A set of users would be very unhappy and perturbed to discover that they switched away from an existing read-collection type to a collection expression and now had things become unsafe.  This would tie into the feeling that collection expressions could not be trusted to make sound decisions that people could trust.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
+## Notes for supporting interfaces.
+
+1.  Interfaces are important. For apis that take collections, outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), the `System.Collections.XXX` collection types are the next most commonly used set.   Of the types in that namespace, the interface collections `IEnumerable<T>/ICollection<T>/IReadOnlyCollection<T>/IList<T>/IReadOnlyList<T>` are used ~90% of the time.  And when used, 90% of those are `IEnumerable<T>` itself.  (If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs).  The large set of these methods (which extends out into the ecosystem as well), motivates supporting interfaces to have collection expressions feel complete at launch.
```
CyrusNajmabadi, line 3:
```suggestion
1.  Interfaces are important. For apis that take collections, outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), the `System.Collections.XXX` collection types are the next most commonly used set.   Of the types in that namespace, the interface collections `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>/ICollection<T>/IList<T>` are used ~90% of the time.  And when used, 90% of those are `IEnumerable<T>` itself.  (If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs).  The large set of these methods (which extends out into the ecosystem as well), motivates supporting interfaces to have collection expressions feel complete at launch.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+   As before, promising too much can hurt us (and defy user expectations around performance and safety).
+
+5. Certain concrete types (specifically `List<T>` or `T[]`) have highly undesirable downsides for an interface target.  Specifically, it would make it a very dangerous footgun for any API to expose an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` property that was instantiated with a collection expression.  A set of users would be very unhappy and perturbed to discover that they switched away from an existing read-collection type to a collection expression and now had things become unsafe.  This would tie into the feeling that collection expressions could not be trusted to make sound decisions that people could trust.
+
+## Options
+
+If we picked an existing concrete collection type to instantiate when targeting an interface type, the choices would likely be the following:
+
+1. `List<T>`.  The most flexible.  But also not very efficient, and definitely unsafe to expose through an `IEnumerable<T>/IReadOnlyList<T>` property.
```
CyrusNajmabadi, line 36:
```suggestion
1. `List<T>`.  The most flexible.  But also not very efficient, and definitely unsafe to expose through an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` property.  
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Options
+
+If we picked an existing concrete collection type to instantiate when targeting an interface type, the choices would likely be the following:
+
+1. `List<T>`.  The most flexible.  But also not very efficient, and definitely unsafe to expose through an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` property.  
+2. `T[]`.  A strange middle ground.  Only partially flexibly.  Can mutate the values in the array themselves, but cannot do anything affecting the length of the array.  Also unsafe to expose through an`IEnumerable<T>/IReadOnlyList<T>` property.
```
CyrusNajmabadi, line 37:
```suggestion
2. `T[]`.  A strange middle ground.  Only partially flexibly.  Can mutate the values in the array themselves, but cannot do anything affecting the length of the array.  Also unsafe to expose through an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` property.  
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+2. State you get an unnamed impl of a type that is guaranteed to implement both `IReadOnlyList<T>` and `IList<T>` *and* that that type returns `true` for `ICollection<T>.IsReadOnly` or not, depending on what you were targeting.  If you were targeting only an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` you would get something read-only.  If you asked for an `IList<T>/ICollection<T>` it would be non-read-only.  This would have the benefit of being entirely safe for the non-mutation cases (because the collection IsReadOnly, consumers could not mutate elements).  It would also be something highly optimizable.  BCL methods would query for `IList<T>` and do fast paths for it.  We could also still fast-path our `GetEnumerator`` impl to be low allocs.
```
CyrusNajmabadi, line 44:
```suggestion
2. State you get an unnamed impl of a type that is guaranteed to implement both `IReadOnlyList<T>` and `IList<T>` *and* that that type returns `true` for `ICollection<T>.IsReadOnly` or not, depending on what you were targeting.  If you were targeting only an `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` you would get something read-only.  If you asked for an `ICollection<T>/IList<T>` it would be non-read-only.  This would have the benefit of being entirely safe for the non-mutation cases (because the collection IsReadOnly, consumers could not mutate elements).  It would also be something highly optimizable.  BCL methods would query for `IList<T>` and do fast paths for it.  We could also still fast-path our `GetEnumerator`` impl to be low allocs.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+1.  Interfaces are important. For apis that take collections, outside of the large set of api the BCL has that deal with contiguous data (and thus take arrays or spans), the `System.Collections.XXX` collection types are the next most commonly used set.   Of the types in that namespace, the interface collections `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>/ICollection<T>/IList<T>` are used ~90% of the time.  And when used, 90% of those are `IEnumerable<T>` itself.  (If we include linq-extensions this number shoots even higher up.  But we've cut those out as *currently* you cannot call extensions on collection-exprs).  The large set of these methods (which extends out into the ecosystem as well), motivates supporting interfaces to have collection expressions feel complete at launch.
+
+2. Our own best practices, which are aped in the ecosystem, are that (if contiguous data segments are not necessary) that apis be maximally permissive in what they take as arguments.  Put another way, our apis "accept the general, but return the specific".  For collections, the way that we "accept the general" is normally as `IEnumerable<T>` (which corresponds to the data above).
```
CyrusNajmabadi, line 5:
```suggestion
2. Our own best practices, which are aped in the ecosystem, are that (if contiguous data segments are not necessary) that apis be maximally permissive in what they take as arguments.  Put another way, our apis "accept the general, but return the specific".  For collections, the way that we "accept the general" is normally as `IEnumerable<T>` (which corresponds to the data above).  This makes `IEnumerable<T>` apis replete in the ecosystem.  Along with the fact that `Linq` effectively pushed this heavily as such a common data interchange format, not supporting interfaces would carve out a substantive number of cases where collection expressions would be desirable.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+
+   Because they core scenarios differ so wildly on their expectations around mutability and around the problems that arise (for either) by aligning on what the other does, it seems reasonable to state and explain things if we go down a path where there are different types and choices between natural and target types.
+
+4. Furthermore, alignment between the two systems (target and natural typing) could actually be highly negative if we promise a concrete collection type. For example, if we promise a `List<T>` type, then that eliminates the ability to use a singleton for the *very* common and *core* use-case of passing `[]` as an argument to `IEnumerable<T>` apis (or initializing an `IEnumerable<T>` property).  This would force users away from this, and require them to be as verbose as today to be explicit to get the desired savings.  This deeply undercuts are core narrative that you can use collection expressions and they can be the *efficient* type you can depend on.
```
CyrusNajmabadi, line 17:
```suggestion
4. Furthermore, alignment between the two systems (target and natural typing) could actually be highly negative if we promise a concrete collection type. For example, if we promise a `List<T>` type, then that eliminates the ability to use a singleton for the *very* common and *core* use-case of passing `[]` as an argument to `IEnumerable<T>/IReadOnlyCollection/IReadOnlyList<T>` apis (or initializing an `IEnumerable<T>/IReadOnlyCollection/IReadOnlyList<T>` property).  This would force users away from this, and require them to be as verbose as today to be explicit to get the desired savings.  This deeply undercuts are core narrative that you can use collection expressions and they can be the *efficient* type you can depend on.
```

meetings/working-groups/collection-literals/Supporting-Interface-TargetTypes.md
```diff
@@ -0,0 +1,72 @@
...
+   Conversely, by not promising a concrete type, we get more opportunities to optimize, even for *non* empty cases.  For example, with:
+
+      ```c#
+      Goo(IEnumerable<int> values);
+      Goo([1, 2, 3])
+      ```
+
+   We can optimize in a way similar to what we do with `yield`ing iterators.  Namely, you would an `IEnumerable<T>` instance that was itself its own `IEnumerator<T>` instance (in the case where it has not been enumerated yet, and was being enumerated on the same thread that created it).  This would again be much more efficient than what you get today with any normal concrete type a user uses here (lists, arrays, etc.), and again ties heavily into our story that by using collection expressions you can trust you're going to get the best results.
+
+   As before, promising too much can hurt us (and defy user expectations around performance and safety).
```
CyrusNajmabadi, line 28:
```suggestion
   As before, promising too much can hurt us and defy user expectations around performance and safety.
```

CyrusNajmabadi:
Closing out in favor of https://github.com/dotnet/csharplang/blob/main/meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md

PR #7428, opened 2023-08-08
Create Core-interface-target-type-proposal.md

CyrusNajmabadi:


meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+While the wins now with 'safety' are definitely welcome, the largest benefits come in the "performance" category.  Specifically, not having to name a particular type opens up many areas of optimization that would simply be unavailable when having to pick an existing generalized type. This would also live up to the promise that we will do an excellent job with perf, and that it would be *very* hard for a user to do the same, and practically impossible for them to do so with simple syntax.  Specifically, all of the following are available as things we could do to heavily optimize here:
```
CyrusNajmabadi, line 218:
```suggestion
While the wins now with 'safety' are definitely welcome, the largest benefits come in the "performance" category.  Specifically, not having to name a particular type for the read-only interfaces opens up many areas of optimization that would simply be unavailable when having to pick an existing generalized type. This would also live up to the promise that we will do an excellent job with perf, and that it would be *very* hard for a user to do the same, and practically impossible for them to do so with simple syntax.  Specifically, all of the following are available as things we could do to heavily optimize here when target-typing `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>:
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+> From a single-dimensional array type `S[]` to `IList<T>`, `IReadOnlyList<T>`, and their base interfaces, provided that there is an implicit identity or reference conversion from S to T.
+
+
+### Context
+
+Collection expressions aim to be a substantively 'complete' replacement for the myriad of ways people create collections today.  They are intended to be used for the *vast majority* of cases where people use collections, and default (as much as possible) to choices that fit the needs of the majority of the user base.  A strong desire here is to not add "just another way to create a collection" alongside everything else, but to provide a singular *replacement* that is then suitable to move almost all code wholesale to.  It aims to replace *all* of the following forms:
+
+1. `new T[] { ... }`
+2. `new[] { ... }`
+3. `stackalloc T[] { ... }` and `stackalloc[] { ... }`
```
CyrusNajmabadi, line 44:
```suggestion
1. `stackalloc T[] { ... }`
1. `stackalloc[] { ... }`
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+Collection expressions aim to be a substantively 'complete' replacement for the myriad of ways people create collections today.  They are intended to be used for the *vast majority* of cases where people use collections, and default (as much as possible) to choices that fit the needs of the majority of the user base.  A strong desire here is to not add "just another way to create a collection" alongside everything else, but to provide a singular *replacement* that is then suitable to move almost all code wholesale to.  It aims to replace *all* of the following forms:
+
+1. `new T[] { ... }`
+2. `new[] { ... }`
+1. `stackalloc T[] { ... }`
+1. `stackalloc[] { ... }`
+4. `new CollectionType<Etc> { ... }`
+5. `ImmutableCollectionType.Create(...)`
+6. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`
```
CyrusNajmabadi, line 48:
```suggestion
1. `new T[] { ... }`
1. `new[] { ... }`
1. `stackalloc T[] { ... }`
1. `stackalloc[] { ... }`
1. `new CollectionType<Etc> { ... }`
1. `ImmutableCollectionType.Create(...)`
1. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`
1. And more... including things like: `ImmutableArray<T>.Empty.Add(x).AddRange(y).etc()`
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,249 @@
...
+
+1. `new T[] { ... }`
+1. `new[] { ... }`
+1. `stackalloc T[] { ... }`
+1. `stackalloc[] { ... }`
+1. `new CollectionType<Etc> { ... }`
+1. `ImmutableCollectionType.Create(...)`
+1. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`
+1. And more... including things like: `ImmutableArray<T>.Empty.Add(x).AddRange(y).etc()`
+
```
CyrusNajmabadi, line 50:
```suggestion
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+An examination of the BCL and the top 20 Nuget packages (Newtonsoft, EF, Azure, Castle, AWS, Automapper), all of which are >400m downloads, reveals very relevant data here.  Methods taking those interface collections account for roughly 28% of all methods taking some collection type (arrays, spans, other BCL collections), and `IEnumerable<T>` alone accounts for 25% of the collection-taking methods.  This is not surprising as our own practices, and general feedback is simply that methods should be extremely permissive in what they accept, and be precise in what they return.  `IEnumerable<T>` (and our other collection interfaces) act as that permissive type that we and our ecosystem have broadly adopted.
```
CyrusNajmabadi, line 51:
```suggestion
An examination of the BCL and the top 20 Nuget packages (Newtonsoft, EF, Azure, Castle, AWS, Automapper, and more), all of which are >400m downloads, reveals very relevant data here.  Methods taking those interface collections account for roughly 28% of all methods taking some collection type (arrays, spans, other BCL collections), and `IEnumerable<T>` alone accounts for 25% of the collection-taking methods.  This is not surprising as our own practices, and general feedback is simply that methods should be extremely permissive in what they accept, and be precise in what they return.  `IEnumerable<T>` (and our other collection interfaces) act as that permissive type that we and our ecosystem have broadly adopted.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+1. `new T[] { ... }`
+1. `new[] { ... }`
+1. `stackalloc T[] { ... }`
+1. `stackalloc[] { ... }`
+1. `new CollectionType<Etc> { ... }`
+1. `ImmutableCollectionType.Create(...)`
+1. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`
+1. And more... including things like: `ImmutableArray<T>.Empty.Add(x).AddRange(y).etc()`
+
+An examination of the BCL and the top 20 Nuget packages (Newtonsoft, EF, Azure, Castle, AWS, Automapper, and more), all of which are >400m downloads, reveals very relevant data here.  Methods taking those interface collections account for roughly 28% of all methods taking some collection type (arrays, spans, other BCL collections), and `IEnumerable<T>` alone accounts for 25% of the collection-taking methods.  This is not surprising as our own practices, and general feedback is simply that methods should be extremely permissive in what they accept, and be precise in what they return.  `IEnumerable<T>` (and our other collection interfaces) act as that permissive type that we and our ecosystem have broadly adopted.
```
CyrusNajmabadi, line 51:
```suggestion
An examination of the BCL and the top 20 Nuget packages (Newtonsoft, EF, Azure, Castle, AWS, Automapper, and more), all of which are >400m downloads, reveals very relevant data here.  Methods taking those interface collections account for roughly 28% of all methods taking some collection type (arrays, spans, other BCL collections), and `IEnumerable<T>` alone accounts for 25% of the collection-taking methods.  This is not surprising as our own practices, and general design guidance we give the community, is simply that methods should be extremely permissive in what they accept, and be precise in what they return.  `IEnumerable<T>` (and our other collection interfaces) act as that permissive type that we and our ecosystem have broadly adopted.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+Indeed, if it were not for `params T[]` (a full 50% of all collection-taking methods), `IEnumerable<T>` would be the most commonly taken collection by far for the ecosystem.
+
+Ideally, we would ship with support for everything, but we've currently made judicious moves from C#12 to C#13 based on complexity, but also based on impact.  For example, `Dictionary expressions` were moved to C#13 to lighten our load, and because data indicates that APIs that consume those dictionary types are only <3% of all apis that take collections in the first place.  `Natural type` support has also been pushed out because the complexity is felt to be substantive enough to warrant more time.
+
+With how important these interface types are, we do not believe pushing out from C# 12 will allow us to ship a viable and coherent story to customers.
```
CyrusNajmabadi, line 57:
```suggestion
With how important these interface types are though, we do not believe pushing out from C# 12 will allow us to ship a viable and coherent story to customers.  
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+2. Universality.  This restates the background context.  Namely that we want to meet the literal 97% case at launch.  This means needing good stories for arrays, spans, BCL concrete collections *and* BCL core interfaces.
+
+3. Brevity.  It is a strong goal of this feature that users be able to just write the simple, idiomatic, collection expression form without the need to do things like add coercive casts to commonly appease the compiler.  Specifically, for casts, once you add them (e.g. `(List<int>)[1, 2, 3]`) then the benefit of the feature as a whole is vastly diminished, and the new form (`new List<int> {1, 2, 3}`) isn't substantively better (just 2 characters saved).  Unlike other features, the cliff here is very steep, often fundamentally negating the idea that this is a valuable feature in the first place.  Many parts of the design (especially broad adoption of target-typing) has been entirely around ensuring users can just write the simple expression form and almost never have to do things to appease the language.
```
CyrusNajmabadi, line 67:
```suggestion
3. Brevity.  It is a strong goal of this feature that users be able to just write the simple, idiomatic, collection expression form without the need to do things like add coercive casts to commonly appease the compiler.  Specifically, for casts, once you add them (e.g. `(List<int>)[1, 2, 3]`) then the benefit of the feature as a whole is vastly diminished. Where the new form isn't substantively better than the existing form (just 2 characters saved over `new List<int> {1, 2, 3}`).  Unlike other features, the cliff here is very steep, often fundamentally negating the idea that this is a valuable feature in the first place.  Many parts of the design (especially broad adoption of target-typing) has been entirely around ensuring users can just write the simple expression form and almost never have to do things to appease the language.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+4. Performance. A core pillar of collection expressions that we are both evangelizing it by, and which we are seeing customers resonate with, is the idea of: "Absent external information unavailable to the compiler, the compiler should almost always do as well or better than the user could.  Sometimes much more so."  Because the user can write a simple `[a, b, c, d, e]` expression, without having to explicitly state what is going on, and because we can provide so much smart understanding to each situation, we can heavily optimize.  For example:
```
CyrusNajmabadi, line 69:
```suggestion
4. Performance. A core pillar of collection expressions that we are both evangelizing it by, and which we are seeing customers resonate with, is the idea of: "Absent external information unavailable to the compiler, the compiler should almost always do as well or better than the user could.  Often much more so.  And almost certainly with clearer code for the user to write."  Because the user can write a simple `[a, b, c, d, e]` expression, without having to explicitly state what is going on, and because we can provide so much smart understanding to each situation, we can heavily optimize.  For example:
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+class Order
+{
+    // Legal, is equivalent to: `Tags { get; } = new List<OrderTag>();
+    public IReadOnlyList<OrderTag> Tags { get; } = [];
+}
+```
+
+This option *nails* the "simplicity", "universality, and "brevity" aspects we are trying to solve.  Explaining to users what happens in these cases is trivial: "It makes a `List<T>`.  The part-and-parcel type of .Net that you've know about and have used for nearly 20 years now".  Similarly, it can be used for *all* these apis that take in one of these interfaces.  Finally, it always allows the nice short syntactic-form that really sells people on why collection-expressions are a superior choice to use over practically every other collection-creation constructs they can use today.
+
+However, this also falls short in both the 'performance' and 'safety' domains.  Indeed, it does so so egregiously, that our own analyzers will flag this as inappropriate and push both the  users who run these analyzers (and the users who just care about these facets of development) away from feeling they can trust this feature to live up to its recommendation as making the "right, smart, best choices" for them.
```
CyrusNajmabadi, line 148:
```suggestion
However, this also falls short in both the 'performance' and 'safety' domains.  Indeed, it does so so egregiously, that our own analyzers will flag this as inappropriate and push both the users who run these analyzers, and the users who just care about these facets of development, away from feeling they can trust this feature to live up to its recommendation as making the "right, smart, best choices" for them.   It will also likely lead to negative-evangelization, where voices in the community steer people away from the feature as a whole.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+However, this also falls short in both the 'performance' and 'safety' domains.  Indeed, it does so so egregiously, that our own analyzers will flag this as inappropriate and push both the users who run these analyzers, and the users who just care about these facets of development, away from feeling they can trust this feature to live up to its recommendation as making the "right, smart, best choices" for them.   It will also likely lead to negative-evangelization, where voices in the community steer people away from the feature as a whole.
+
+Specifically, for "performance", `List<T>` is actually a fairly undesirable type:
```
CyrusNajmabadi, line 150:
```suggestion
Next, for "performance", `List<T>` is actually a fairly undesirable type: 
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+However, this also falls short in both the 'performance' and 'safety' domains.  Indeed, it does so so egregiously, that our own analyzers will flag this as inappropriate and push both the users who run these analyzers, and the users who just care about these facets of development, away from feeling they can trust this feature to live up to its recommendation as making the "right, smart, best choices" for them.   It will also likely lead to negative-evangelization, where voices in the community steer people away from the feature as a whole.
+
+Next, for "performance", `List<T>` is actually a fairly undesirable type: 
+
+1. It *already* comes with two allocations for itself.
+2. Getting an iterator for it (through the `IEnumerable<T>.GetEnumerator` path) will produce another allocation.
+3. It has excess overhead internally to support both being able to grow, and has overhead internally to ensure it is not mutated while it is being iterated.
+
+`List<T>` is a great type when you need flexibility and permissiveness.  It is not a good choice when you know precisely what you are producing, and you have no need to access the flexible behaviors it provides.
```
CyrusNajmabadi, line 156:
```suggestion
`List<T>` is a great type when you need flexibility and permissiveness.  It is not a good choice when you know precisely what you are producing, and you have no need to access the flexible, mutation-oriented, capabilities it provides.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+For "safety" `List<T>` is also an unacceptable choice for many (and our own analyzers will push you away from it).  Users who expose data currently (and safely) through the readonly interfaces `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>` today will find that they cannot effectively move to collection expressions.  They will have to maintain the code the highly verbose and clunky code they have today, to use types like `ImmutableArray`, or things like `new List<int> { a, b, c }>.AsReadOnly()`.  This will make collection-expressions feel half baked for these users, again undercutting the story that they make "right, smart, best choices" for the user, and forcing them to lose the brevity and consistency of being able to use collection-expressions with confidence everwhere.
+
+### Option 3: Be smart. :)
+
+Deep discussion around the problems with the above approach led the WG to come up with a tweak to the core idea it has, to ideally leverage nearly all its benefits, while curtailing its drawbacks.
```
CyrusNajmabadi, line 162:
```suggestion
Deep discussion around the problems with Option 2 led the WG to come up with a tweak to the core idea it has, to ideally leverage nearly all its benefits, while curtailing its drawbacks.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+```c#
+{
+    public IEnumerator<T> GetEnumerator(); // and the non-generic equivalent.
+    public int Count { get; }
+    public T this[int index] { get; }
+}
+```
+
+That's it.  Just *three* members.  Now, given the above, the rules would be as follows:
```
CyrusNajmabadi, line 174:
```suggestion
That's it.  Just *three* members.  With that in mind the rules for target-typing an interface would be as follows:
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,248 @@
...
+
+    - Importantly, both of the above are possible, and we could even do one, then migrate to the other over different releases.  Because all that user could could ever know about is the target-interface type, there would be no source, binary, or runtime breaking changes here.  This means that in C# 12 we could use a compiler-synthesized type.  But also then migrate over to a BCL api if the BCL would like to own a canonical API for this purpose.
+
+2. For a target-type of `ICollection<T>/IList<T>`, the type actually created would be guaranteed to be `List<T>`.
+
+Effectively, this *intentionally* bifurcates the interfaces into two important categories.  The 'read only' interfaces, which support no mutation, and the 'mutable interfaces' which do.  Our feeling is that for the latter, there is no real better choice than `List<T>`.  It is the 'bread and butter' type the BCL has for this purpose that the entire ecosystem understands and feels comfortable with.  For the domain of mutable-sequences, it is extremely good and does its job well.
+
```
CyrusNajmabadi, line 187:
```suggestion
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+Performance and safety though would be mostly neutral here.  Users who did not care about either would likely just use `List<T>`, and users who did would pick something appropriate for their domain.  However, this would still be a small tick in the 'negative' category as our claims about us making the "right, smart, best choices" for users would be undercut by then forcing the user to have to make those choices themselves.
+
+### Option 2: Use `List<T>` for *all* of the possible interface targets.
+
+This is the next simplest option.  It says that for any of those five interface cases we are target-typing to, we should *always* use `List<T>` as the concrete type actually created.  Specifically:
+
+```c#
+void DoSomething(IEnumerable<int> values) { ... }
+
+// Legal, is equivalent to `DoSomething(new List<int> { 1, 2, 3 })`
```
CyrusNajmabadi, line 132:
```suggestion
// Legal. Is equivalent to `DoSomething(new List<int> { 1, 2, 3 })`
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+DoSomething([1, 2, 3]);
+```
+
+and
+
+```c#
+class Order
+{
+    // Legal, is equivalent to: `Tags { get; } = new List<OrderTag>();
```
CyrusNajmabadi, line 141:
```suggestion
    // Legal. Is equivalent to: `Tags { get; } = new List<OrderTag>();
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+// to take a dependency on it.
+DoSomething([1, 2, 3]);
+```
+
+and
+
+```c#
+class Order
+{
+    // Allowed, an efficient 'empty' instance will be used.
```
CyrusNajmabadi, line 204:
```suggestion
    // Legal. an efficient 'empty' instance will be used.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+    public IReadOnlyList<OrderTag> Tags { get; } = [];
+
+    // Allowed.  Equivalent to: `{ get; } = new List<int>()`
```
CyrusNajmabadi, line 207:
```suggestion
    // Legtal. Equivalent to: `{ get; } = new List<int>()`
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+}
+```
+
+With respect to the factors we care about, here's how the above falls out:
+
+For 'brevity/universality', this still strongly satisfies our goals.  Users will be able to use the succinct collection-expression form for all these APIs with ease.
+
+However, compared to option 2, we now see *large* wins in both safety and performance.  Starting with safety, we now have a good option for everyone exposing data through any of those three interfaces.  Data is safe by default (a good choice for all users), and analyzers do not trigger, pushing people away from using our cohesive story here.  Users who do not care about safety continue not to care, while users for whom it is very important are extremely satisfied that they can trust this feature and that it lives up to the stated goal that it will make the smart choices they can depend on, and they do not have to ban this feature, or recommend others steer away from it.  This approach also naturally extends out to when we do dictionary-literals.  There, it will also be the case that for `IReadOnlyDictionary<TKey, Value>` we will not want to expose a mutable type (like `Dictionary<TKey, TValue>`) for safety reasons.
```
CyrusNajmabadi, line 216:
```suggestion
For 'brevity/universality', this still strongly satisfies our goals.  Users will be able to use the succinct collection-expression form for all these APIs with ease. However, compared to option 2, we now see *large* wins in both safety and performance. 

Starting with safety, we now have a good option for everyone exposing data through any of those three interfaces.  Data is safe by default (a good choice for all users), and analyzers do not trigger, pushing people away from using our cohesive story here.  Users who do not care about safety continue not to care, while users for whom it is very important are extremely satisfied that they can trust this feature and that it lives up to the stated goal that it will make the smart choices they can depend on, and they do not have to ban this feature, or recommend others steer away from it.  This approach also naturally extends out to when we do dictionary-literals.  There, it will also be the case that for `IReadOnlyDictionary<TKey, Value>` we will not want to expose a mutable type (like `Dictionary<TKey, TValue>`) for safety reasons.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+}
+```
+
+With respect to the factors we care about, here's how the above falls out:
+
+For 'brevity/universality', this still strongly satisfies our goals.  Users will be able to use the succinct collection-expression form for all these APIs with ease. However, compared to option 2, we now see *large* wins in both safety and performance. 
+
+Starting with safety, we now have a good option for everyone exposing data through any of those three interfaces.  Data is safe by default (a good choice for all users), and analyzers do not trigger, pushing people away from using our cohesive story here.  Users who do not care about safety continue not to care, while users for whom it is very important are extremely satisfied that they can trust this feature and that it lives up to the stated goal that it will make the smart choices they can depend on, and they do not have to ban this feature, or recommend others steer away from it.  This approach also naturally extends out to when we do dictionary-literals.  There, it will also be the case that for `IReadOnlyDictionary<TKey, Value>` we will not want to expose a mutable type (like `Dictionary<TKey, TValue>`) for safety reasons.
```
CyrusNajmabadi, line 216:
```suggestion
For 'brevity/universality', this still strongly satisfies our goals.  Users will be able to use the succinct collection-expression form for all these APIs with ease. However, compared to option 2, we now see *large* wins in both safety and performance. 

Starting with safety, we now have a good option for everyone exposing data through any of those three interfaces.  Data is safe by default (a good choice for all users).  Analyzers do not trigger, preventing our own tooling from pushing people away from using our cohesive story here.  Users who do not care about safety continue not to care, while users for whom it is very important are extremely satisfied that they can trust this feature and that it lives up to the stated goal that it will make the smart choices they can depend on, and they do not have to ban this feature, or recommend others steer away from it.  This approach also naturally extends out to when we do dictionary-literals.  There, it will also be the case that for `IReadOnlyDictionary<TKey, Value>` we will not want to expose a mutable type (like `Dictionary<TKey, TValue>`) for safety reasons.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+1. `new T[] { ... }`
+1. `new[] { ... }`
+1. `stackalloc T[] { ... }`
+1. `stackalloc[] { ... }`
+1. `new CollectionType<Etc> { ... }`
+1. `ImmutableCollectionType.Create(...)`
+1. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`
+1. And more... including things like: `ImmutableArray<T>.Empty.Add(x).AddRange(y).etc()`
+
+An examination of the BCL and the top 20 Nuget packages (Newtonsoft, EF, Azure, Castle, AWS, Automapper, and more), all of which are >400m downloads, reveals very relevant data here.  Methods taking those interface collections account for roughly 28% of all methods taking some collection type (arrays, spans, other BCL collections), and `IEnumerable<T>` alone accounts for 25% of the collection-taking methods.  This is not surprising as our own practices, and general design guidance we give the community, is simply that methods should be extremely permissive in what they accept, and be precise in what they return.  `IEnumerable<T>` (and our other collection interfaces) act as that permissive type that we and our ecosystem have broadly adopted.
```
jnm2, line 51:
```suggestion
An examination of the BCL and the top 20 NuGet packages (Newtonsoft, EF, Azure, Castle, AWS, AutoMapper, and more), all of which are >400m downloads, reveals very relevant data here.  Methods taking those interface collections account for roughly 28% of all methods taking some collection type (arrays, spans, other BCL collections), and `IEnumerable<T>` alone accounts for 25% of the collection-taking methods.  This is not surprising as our own practices, and general design guidance we give the community, are simply that methods should be extremely permissive in what they accept, and be precise in what they return.  `IEnumerable<T>` (and our other collection interfaces) act as that permissive type that we and our ecosystem have broadly adopted.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+Indeed, if it were not for `params T[]` (a full 50% of all collection-taking methods), `IEnumerable<T>` would be the most commonly taken collection by far for the ecosystem.
+
+Ideally, we would ship with support for everything, but we've currently made judicious moves from C#12 to C#13 based on complexity, but also based on impact.  For example, `Dictionary expressions` were moved to C#13 to lighten our load, and because data indicates that APIs that consume those dictionary types are only <3% of all apis that take collections in the first place.  `Natural type` support has also been pushed out because the complexity is felt to be substantive enough to warrant more time.
+
+With how important these interface types are though, we do not believe pushing out from C# 12 will allow us to ship a viable and coherent story to customers.  
+
+### What factors are at play?
+
+In any interesting design, there are many factors that must be assessed and considered as a holistic whole.  For collection expressions, the broad areas we want the land the feature are include, but are not limited to:
```
jnm2, line 61:
```suggestion
In any interesting design, there are many factors that must be assessed and considered as a whole.  For collection expressions, these include but are not limited to:
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+1. Simplicity. Ideally the feature works in a fashion that is both simple to explain and simple for users to understand. Using some of our modern terminology, we'd like to avoid 'decoder rings' when people use it.
+
+2. Universality.  This restates the background context.  Namely that we want to meet the literal 97% case at launch.  This means needing good stories for arrays, spans, BCL concrete collections *and* BCL core interfaces.
```
jnm2, line 65: People may wonder... 97% case, as opposed to what percent if interfaces aren't allowed as targets? 95%? 50%? Can we ballpark a number for this?

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+4. Performance. A core pillar of collection expressions that we are both evangelizing it by, and which we are seeing customers resonate with, is the idea of: "Absent external information unavailable to the compiler, the compiler should almost always do as well or better than the user could.  Often much more so.  And almost certainly with clearer code for the user to write."  Because the user can write a simple `[a, b, c, d, e]` expression, without having to explicitly state what is going on, and because we can provide so much smart understanding to each situation, we can heavily optimize.  For example:
+
+    - if the above 5-element collection were converted to an `ImmutableArray<T>` our emitted could would practically always be better than users using normal construction patterns.  We can also greatly leverage extremely fast and efficient systems under hte covers (like synthesizing `Inline-Arrays` types) that would generally be extremely ugly and painful for users to do themselves.
```
jnm2, line 71:
```suggestion
    - If the above 5-element collection were converted to an `ImmutableArray<T>`, our emitted code could would practically always be better than users using normal construction patterns.  We can also greatly leverage extremely fast and efficient systems under the covers (like synthesizing `Inline-Arrays` types) that would generally be extremely ugly and painful for users to do themselves.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+    - what we emit can adopt ecosystem best practices around performance.  `[]` can emit as efficient singletons, not causing undesirable allocations.  etc. etc.
```
jnm2, line 73:
```suggestion
    - What we emit can adopt ecosystem best practices around performance.  For instance, `[]` can emit as efficient singletons, not causing undesirable allocations.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+    While many customers will not care about performance to this level, we still want customers that do to have confidence in this feature (and we definitely do not want to see the feature immediately banned).
+
+    It would also be highly unfortunate if we shipped and our own analyzers immediately flagged the usage as being a problem.
+
+    Finally, part of performance means being a good .Net citizen.  So the collections we produce should be able to pick up the optimizations the BCL has today for collection types.
```
jnm2, line 79:
```suggestion
    Finally, part of performance means being a good .NET citizen.  So the collections we produce should be able to pick up the optimizations the BCL has today for collection types.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+5. Safety.  Specifically, keeping data safe from undesirable mutation. We broadly think of users as being in two categories.  The first category generally doesn't think about safety, and thus would suffice with any solution on our part.  If mutable data is exposed or not isn't relevant to them as they do not care about keeping their data safe.  However, the second group (which is likely smaller, but present, vocal, and influential) absolutely wants to keep their exposed data safe from mutation.  Similar to the perf concerns, we believe we need a solution for this group that fits their expectations, puts them at ease, and isn't immediately banned (or flagged by analyzers) for doing unsafe things.
```
jnm2, line 81:
```suggestion
5. Safety.  Specifically, keeping data safe from undesirable mutation. We broadly think of users as being in two categories.  The first category generally doesn't consider it to be a safety concern to return mutable instances via read-only interfaces, and thus would suffice with any solution on our part.  However, the second group (which is likely smaller, but present, vocal, and influential) absolutely wants a roadblock at runtime to keep their exposed data safe from mutation.  (It won't prevent a determined user from mutating, but it prevents accidents.)  Similar to the perf concerns, we believe we need a solution for this group that fits their expectations, puts them at ease, and isn't immediately banned (or flagged by analyzers) for doing unsafe things.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Options
+
+Note: there are truly a plethora of options available here.  However, based on feedback from LDM and the working group meetings, we tried to whittle things down to a reasonable few that we feel warrant discussion and comparison.
```
jnm2, line 85:
```suggestion
Based on feedback from LDM and the working group meetings, we tried to whittle a large number of options down to a reasonable few that we feel warrant discussion and comparison.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+    public IReadOnlyList<OrderTag> Tags { get; } = (ImmutableArray<OrderTag>)[];
+
+    // Which is worse than just:
+    public IReadOnlyList<OrderTag> Tags { get; } = Array.Empty<OrderTag>();
+}
+```
+
+First, this immediately fails the simplicity and universality concerns. From our own experience, and the data about the ecosystem, we know that users will need to interface with, well, interfaces :).  This will be an immediate hurdle that will deeply undercut the value of this feature space, potentially even deeply tainting it for the future.
+
+Second, collection expressions would not provide any sort of actual useful syntax or brevity for users in this case.  Because an explicit type would have to be provided, users would be left with syntax with nearly the same complexity and verbosity as what they would have to write today.  Users seeing this would rightfully ask "why would i pick this new form, that is really just the same as the old form, just with some tokens tweaked?"
```
jnm2, line 121:
```suggestion
Second, collection expressions would not provide any sort of actual useful syntax or brevity for users in this case.  Because an explicit type would have to be provided, users would be left with syntax with nearly the same complexity and verbosity as what they would have to write today.  Users seeing this would rightfully ask "why would I pick this new form, that is really just the same as the old form, just with some tokens tweaked?"
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+```c#
+class Order
+{
+    // Legal. Is equivalent to: `Tags { get; } = new List<OrderTag>();
+    public IReadOnlyList<OrderTag> Tags { get; } = [];
+}
+```
+
+This option *nails* the "simplicity", "universality, and "brevity" aspects we are trying to solve.  Explaining to users what happens in these cases is trivial: "It makes a `List<T>`.  The part-and-parcel type of .Net that you've know about and have used for nearly 20 years now".  Similarly, it can be used for *all* these apis that take in one of these interfaces.  Finally, it always allows the nice short syntactic-form that really sells people on why collection-expressions are a superior choice to use over practically every other collection-creation constructs they can use today.
```
jnm2, line 146:
```suggestion
This option *nails* the "simplicity", "universality, and "brevity" aspects we are trying to solve.  Explaining to users what happens in these cases is trivial: "It makes a `List<T>`.  The part-and-parcel type of .NET that you've known about and have used for nearly 20 years now".  Similarly, it can be used for *all* these APIs that take in one of these interfaces.  Finally, it always allows the nice short syntactic form that really sells people on why collection expressions are a superior choice to use over practically every other collection creation construct they can use today.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+and
+
+```c#
+class Order
+{
+    // Legal. an efficient 'empty' instance will be used.
+    public IReadOnlyList<OrderTag> Tags { get; } = [];
+
+    // Legtal. Equivalent to: `{ get; } = new List<int>()`
```
jnm2, line 207:
```suggestion
    // Legal. Equivalent to: `{ get; } = new List<int>()`
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+// to take a dependency on it.
+DoSomething([1, 2, 3]);
+```
+
+and
+
+```c#
+class Order
+{
+    // Legal. an efficient 'empty' instance will be used.
```
jnm2, line 204:
```suggestion
    // Legal. An efficient 'empty' instance will be used.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+While the wins now with 'safety' are definitely welcome, the largest benefits come in the "performance" category.  Specifically, not having to name a particular type for the read-only interfaces opens up many areas of optimization that would simply be unavailable when having to pick an existing generalized type. This would also live up to the promise that we will do an excellent job with perf, and that it would be *very* hard for a user to do the same, and practically impossible for them to do so with simple syntax.  Specifically, all of the following are available as things we could do to heavily optimize here when target-typing `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>:
```
jnm2, line 218:
```suggestion
While the wins now with 'safety' are definitely welcome, the largest benefits come in the "performance" category.  Specifically, not having to name a particular type for the read-only interfaces opens up many areas of optimization that would simply be unavailable when having to pick an existing generalized type. This would also live up to the promise that we will do an excellent job with perf, and that it would be *very* hard for a user to do the same, and practically impossible for them to do so with simple syntax.  Specifically, all of the following are available as things we could do to heavily optimize here when target-typing `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>`:
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+1. Empty collection expressions can always be collapsed down to a single empty singleton (per `T` element type) across the entire program.
+
+1. We can use features like inline-arrays to generate instances that inline all their elements are literally no larger than space for the elements themselves.  A count would not even need to be stored as it could just be hard-coded as the value returned by `int Count { get; }`.  While this might generate many types in the assembly (one type per unique count), it's worth noting that this is what we *already* do for arrays *today*, and what we will be doing for the inline-array types we generate for spans/params-spans.
```
jnm2, line 222:
```suggestion
1. We can use features like inline-arrays to generate instances that inline all their elements and are literally no larger than space for the elements themselves.  A count would not even need to be stored as it could just be hard-coded as the value returned by `int Count { get; }`.  While this might generate many types in the assembly (one type per unique count), it's worth noting that this is what we *already* do for arrays *today*, and what we will be doing for the inline-array types we generate for spans/params-spans.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+1. We can borrow a trick from `yield iterators` where the iterator returned from `IEnumerator<T> GetEnumerator()` literally *reuses* the `this` instance to be its own iterator *if* it has never been iterated yet, and the calling thread is the same one that generated the `IEnumerable<T>`.  This means that if you have:
+
+    ```c#
+    DoSomething([1, 2, 3]);
+    ```
+
+    Then this may be literally a *single* allocation, *including* for when it is likely almost immediately iterated.
```
jnm2, line 230: I'm not entirely sure why the Markdown syntax highlighting broke so badly from this point forward (not ending the code syntax highlighting) but maybe it's due to not having the expected number of spaces of indentation?

````suggestion
   ```c#
   DoSomething([1, 2, 3]);
   ```

   Then this may be literally a *single* allocation, *including* for when it is likely almost immediately iterated.
````

File: meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md (hunk 24 above)
Comment (CyrusNajmabadi) [reply meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md:65]: Yes.  If we take interfaces out, we go down to hitting 69% of all apis.  Importantly though, these APIs are modal.  So it's not like you're limited from calling the interface form, but then have another form (like the array/span) that is available to you.  Instead, it is just hte case that your only option is to fall back to explicitly instantiating a concrete type.

Also, The 97% comes from the remaining 3% being things in the dictionary family.  While present is reasonable numbers, it is dwarfed by the number of sequence-like collections everywhere.  This can also help explain why we felt ok (though not great) punting on dictionaries for 12.  It only limited you that much, as opposed to dropping 28% in the case of not supporting interfaces.

File: meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md [hunk 37]
```diff
@@ -0,0 +1,247 @@
...
+
+    It would also be highly unfortunate if we shipped and our own analyzers immediately flagged the usage as being a problem.
+
+    Finally, part of performance means being a good .NET citizen.  So the collections we produce should be able to pick up the optimizations the BCL has today for collection types.
+
+5. Safety.  Specifically, keeping data safe from undesirable mutation. We broadly think of users as being in two categories.  The first category generally doesn't consider it to be a safety concern to return mutable instances via read-only interfaces, and thus would suffice with any solution on our part.  However, the second group (which is likely smaller, but present, vocal, and influential) absolutely wants a roadblock at runtime to keep their exposed data safe from mutation.  (It won't prevent a determined user from mutating, but it prevents accidents.)  Similar to the perf concerns, we believe we need a solution for this group that fits their expectations, puts them at ease, and isn't immediately banned (or flagged by analyzers) for doing unsafe things.
```
CyrusNajmabadi, line 81:
```suggestion
5. Safety.  Specifically, keeping data safe from undesirable mutation. We broadly think of users as being in two categories.  The first category generally doesn't consider it to be a safety concern to return mutable instances via read-only interfaces, and thus would suffice with any solution on our part.  However, the second group (which is likely smaller, but present, vocal, and influential) absolutely wants a roadblock at runtime to keep their exposed data safe from mutation.  Similar to the perf concerns, we believe we need a solution for this group that fits their expectations, puts them at ease, and isn't immediately banned (or flagged by analyzers) for doing unsafe things.
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
```diff
@@ -0,0 +1,247 @@
...
+
+Collection expressions aim to be a substantively 'complete' replacement for the myriad of ways people create collections today.  They are intended to be used for the *vast majority* of cases where people use collections, and default (as much as possible) to choices that fit the needs of the majority of the user base.  A strong desire here is to not add "just another way to create a collection" alongside everything else, but to provide a singular *replacement* that is then suitable to move almost all code wholesale to.  It aims to replace *all* of the following forms:
+
+1. `new T[] { ... }`
+1. `new[] { ... }`
+1. `stackalloc T[] { ... }`
+1. `stackalloc[] { ... }`
+1. `new CollectionType<Etc> { ... }`
+1. `ImmutableCollectionType.Create(...)`
+1. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`
```
CyrusNajmabadi, line 48:
```suggestion
1. `ImmutableCollection.CreateBuilder(); builder.Add(...); ...; builder.ToImmutable()`.  (Wasteful, and doesn't work for base cases like a field initializer).
```

meetings/working-groups/collection-literals/Core-interface-target-type-proposal.md
jnm2:
```suggestion
2. Universality.  This restates the background context: we want to meet the literal 97% case at launch, not 69% which would be the case if interfaces can't be targeted.  This means needing good stories for arrays, spans, BCL concrete collections *and* BCL core interfaces.
```

PR #7432, opened 2023-08-08
Record C# 12

jcouv:
TODO:
- [x] Add one-liner description for each feature
- [ ] Once this PR is merged, update links from championed proposals and test plans to the new speclet locations


Language-Version-History.md
```diff
@@ -1,6 +1,16 @@
...
+- [Primary Constructors](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/primary-constructors.md): helps reduce field and constructor boilerplate (`class Point(int x, int y);`)
+- [Inline Arrays](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/inline-arrays.md): provides a general-purpose and safe mechanism for declaring arrays using the `[InlineArray(size)]` attribute.
+- [Using aliases for any type](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/using-alias-types.md): relaxes many restrictions on `using` alias declarations, allowing built-in types, tuple types, pointer types, array types (`using Point = (int x, int y);`)
+- [Ref readonly parameters](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/ref-readonly-parameters.md): `ref readonly` parameters mandate that arguments are passed by reference instead of optionally copied, can’t be modified, and warn if a temporary variable must be created.
```
jjonescz, line 10: It's not really an option - if passed by-value, structs would be copied while classes wouldn't - so the argument is *potentially* copied, I think, depending on its type.

```suggestion
- [Ref readonly parameters](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/ref-readonly-parameters.md): `ref readonly` parameters mandate that arguments are passed by reference instead of potentially copied, can’t be modified, and warn if a temporary variable must be created.
```

PR #7463, opened 2023-08-13
Update LDM notes wording to match recollection of working group

jnm2:
Several of us in the working group were surprised to see this paragraph. I was on vacation for this LDM, but I'm relying on the recommendation @CyrusNajmabadi made, which is that these notes likely don't capture what the LDM actually decided on.

(The working group recommends not allowing collection expressions to *target* the nongeneric interfaces. This is because it raises additional design questions and the situation seems too rare to be worth the investigation now. At the same time, we stress the importance of *implementing* at least the nongeneric ICollection and possibly IList when the *generic* interfaces are targeted. This enables nongeneric code, such as data binding, to recognize our synthesized implementations as collections.)

@MadsTorgersen Does this sound right?


meetings/2023/LDM-2023-08-09.md
```diff
@@ -71,7 +71,7 @@ Looking to the future, there aren't good existing dictionary types to implement
...
+We can't ignore the non-generic core interfaces (`IEnumerable`, `ICollection` and `IList`). They probably don't change the overall decision, but they should be considered. Many APIs and frameworks, especially older ones, rely on these, but not actually often as a targetable type. More commonly they come in as `object` and are type-tested against these interfaces at runtime. Regardless of whether target typing works for the non-generic interfaces, it is in fact more important that the concrete types we provide for the *generic* interfaces also implement the appropriate corresponding *non*-generic interfaces. That way, those type-discovery scenarios would work well over collections generated from collection expressions.
```
jnm2, line 74: If desired, this code sample could be included:

```cs
void InitDataSource(IList<Person> people)
{
    // This is the 'object DataSource' property on System.Windows.Forms.BindingSource
    peopleBindingSource.DataSource = people;
}
```

CyrusNajmabadi (approved):

cston (approved):

jnm2:
This was cleared up in https://github.com/dotnet/csharplang/blob/main/meetings/2023/LDM-2023-10-02.md:

> These rules also prompted a discussion on support for non-generic `IEnumerable`. The WG was unclear on the exact outcome from a
> [prior LDM](LDM-2023-08-09.md#target-typing-of-collection-expressions-to-core-interfaces), and whether the non-generic `IEnumerable` and descendent interfaces (such as `ICollection`)
> need to be supported for target-typing as well. We clarified that we don't think we can make a decision about whether to consider the non-generic `IEnumerable` "deprecated" by
> ourselves; it would need to be a larger decision among all the .NET teams. We also don't think supporting the non-generic types will cause any new ambiguities that didn't already
> exists: for example, if an API is overloaded on `ICollection` vs `ICollection<object>` today, it's already ambiguous, because `ICollection<T>` doesn't inherit from `ICollection`.
> 
> As part of this consideration, we also confirmed that we do want to be able to target the non-generic interfaces with collection expressions. IE, this should work:
> 
> ```cs
> ICollection c = ["a", 2, null];
> ```
> However, we think we may not be able to get this in C# 12, and it may have to wait for C# 13 and be considered in tandem with natural types for collection expressions.

PR #7470, opened 2023-08-16
Create LDM-questions-2023-08-15.md

CyrusNajmabadi:


meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,224 @@
...
+We know we want a `Span<T>` overload to be preferred over a `T[]` overload or an interface overload ( `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>/ICollection<T>/IList<T>`).
+
+However, we have options on how to specify things to get the above outcome.
+
+1. Option A: Span/ReadOnlySpan is better than arrays and those specific interfaces.
+
+    The general intuition here is Spans/Arrays/Interfaces have a natural ordering we can make an airtight case around.  Specifically:
+
+    Spans are better than arrays as they're the fast form
+    that can subsume arrays and also be on the stack.
```
CyrusNajmabadi, line 136:
```suggestion
    Spans are better than arrays as they're the fast form that can subsume arrays and also be on the stack.
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+            // 's' has 'local scope'
+            scoped Span<int> s = [x, y, z];
+            foreach (var v in s)
+                Console.WriteLine(v);
+        }
+        ```
+
+        This feels punishing to the common/default case that users will want.  Importantly, it means that if a user wants to switch off from `stackalloc` to collection-expressions they get less pleasant code. 
+
+    - Option B: Ref-struct local always has local scope.
```
cston, line 86: Should this be *collection expression* instead of just locals (for return values, method arguments, etc.)?
```suggestion
    - Option B: Ref struct collection expression always has local scope.
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+            
+            return x ? s1 : s2;
+        }
+        ```
+
+        However, WG believe this is a good thing.  This will be the rare case, and having *it* be slightly more verbose is unlikely to be a burden.  In other words, we want the common, fast, case to sing with collection-expressions, while the less common, slower, case is ok to pay a small price.
+
+Working group suggestion:
+
+Option 2B. Collection-expressions of ref-struct type have local-scope.
```
cston, line 123: Updating to match text above.
```suggestion
Option 2B. Collection expressions of `ref struct` type have local scope.
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+1. Option B: Same as option A, except that it applies to all ref-structs, not just Span/ReadOnlySpan.
+
+    The general intuition is that anything in the ref-struct realm should be considered fast/span-like and then fit into this bucket.  
+
+    With this approach the following would also hold:
+
+    ```c#
+    ref struct ValueList<T> { ... }
+
+    void M(T[] list);
```
cston, line 161:
```suggestion
    void M(int[] list);
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+
+Option 2B. Collection-expressions of ref-struct type have local-scope.
+
+### Overload resolution priority
+
+We know we want a `Span<T>` overload to be preferred over a `T[]` overload or an interface overload ( `IEnumerable<T>/IReadOnlyCollection<T>/IReadOnlyList<T>/ICollection<T>/IList<T>`).
+
+However, we have options on how to specify things to get the above outcome.
+
+1. Option A: Span/ReadOnlySpan is better than arrays and those specific interfaces.
```
cston, line 131: I think we need to be clear that the *iteration types* must be compatible. Perhaps:

```suggestion
1. Option A: Span/ReadOnlySpan is preferred over an array or specific interface type when the iteration type of the span is implicitly convertible to the iteration type of the array or interface.
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+    ```c#
+    ref struct ValueList<T> { ... }
+
+    void M(List<int> list);
+    void M(ValueList<int> list);
+
+    M([1, 2, 3]); // Ambiguous.
+    ```
+
+1. Option C: Ref-structs are always considered better than all other types.
```
cston, line 178:
```suggestion
1. Option C: `ref struct` collection types are preferred over non-`ref struct` collection types (when there is an implicit conversion from the iteration type of the `ref struct` type to the iteration type of the non-`ref struct` type).
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+
+```c#
+Memory<int> m = [1, 2, 3];
+```
+
+Based on last LDM meeting, we are deciding "We will do if we have time, but it is low pri and can be cut if we do not have the time/resources"
+
+1. We made a similar decision for inline-array types.  These types are not in the core majority of apis/use-cases, and can come later.
+
+1. Adding support for a new type *is* a breaking change.  However, this is unavoidable in this space as any type can become a collection type in the future (through usage of the `[CollectionBuilder]` attribute).  As such, we do not think it critical to front load support to avoid this problem.
```
cston, line 209: Adding a `[CollectionBuilder]` builder attribute to a core type would require analyzing the breaking change impact as well. Perhaps just combine this bullet with the following, and remove the attribute example.

```suggestion
1. Adding support for a new type *is* a breaking change.  However, this is unavoidable in this space as any type can become a collection type in the future (through usage of the `[CollectionBuilder]` attribute).  As such, we do not think it critical to front load support to avoid this problem.
```

meetings/working-groups/collection-literals/LDM-questions-2023-08-15.md
```diff
@@ -0,0 +1,223 @@
...
+
+1. In practice `Memory<T>` is for async versions of `Span<T>` methods.  These are not generally overloads as one method is `Foo(Span<T>)` while the other is `FooAsync`.  So the concern about overload ambiguity in the future is low.
```
cston, line 211: Adding a `[CollectionBuilder]` builder attribute to a core type would require analyzing the breaking change impact as well. Perhaps just combine this bullet with the following, and remove the attribute example.

```suggestion
1. Adding support for a new type *is* a breaking change.  In practice though, `Memory<T>` is for async versions of `Span<T>` methods.  These are not generally overloads as one method is `DoWork(Span<T>)` while the other is `DoWorkAsync`.  So the concern about overload ambiguity in the future is low.
```

PR #7473, opened 2023-08-16
add anchor

BillWagner:
The anchor `#constructible-collection-types` was referenced in this document, but not defined.

I think this is where it was intended to resolve.

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -205,13 +205,13 @@ ImmutableArray<int> ia =
 ```
 
 ## Construction
-[constructible-collection-types]: #constructible-collection-types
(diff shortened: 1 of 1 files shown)


cston:
The reference should probably link to <code>#conversions</code> for now. Thanks.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -205,6 +205,7 @@ ImmutableArray<int> ia =
 ```
 
 ## Construction
+[constructible-collection-types]: #constructible-collection-types
```
cston, line 208: Let's use <code>[construction]: #construction</code> here instead, and there shouldn't be any references to this section currently.

BillWagner:
2nd commit addresses feedback.

PR #7498, opened 2023-08-28
Collection expressions builder method: public vs. accessible.

BillWagner:
The speclet states that the builder method must be `public` and `static`.

While creating samples, I found that the actual requirement isn't `public` but rather that the builder method must be *accessible*.

This is one proposed fix. If the design intent was that builder methods must be `public`, I'll close this PR and create an issue in the roslyn repo.

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -170,7 +170,7 @@ For the *create method*:
 * The *builder type* must be a non-generic `class` or `struct`.
 * The method must be defined on the *builder type* directly.
 * The method must be `static`.
-* The method must be accessible.
+* The method must be accessible where the collection expression is used.
 * The *arity* of the method must match the *arity* of the collection type.
 * The method must have a single parameter of type `System.ReadOnlySpan<E>`, passed by value, and there is an [*identity conversion*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/conversions.md#1022-identity-conversion) between `E` and the [*iteration type*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/statements.md#1295-the-foreach-statement) of the *collection type*.
 * There is an [*identity conversion*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/conversions.md#1022-identity-conversion), [*implicit reference conversion*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/conversions.md#1028-implicit-reference-conversions), or [*boxing conversion*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/conversions.md#1029-boxing-conversions)  between the method return type and the *collection type*.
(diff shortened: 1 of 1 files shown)


BillWagner:
cc @CyrusNajmabadi @jnm2 to check if this is the correct fix.

BillWagner:
>  I have no personal preference here. I think simplest is just requiring public.

@CyrusNajmabadi The current feature, as implemented, doesn't require `public`.I could create an `internal` collection class, and an `internal` builder class & method, and that compiles with the current build. To enforce `public` would require a compiler change.

jnm2:
Some folks who don't like putting public methods on internal classes could be forced out of their comfort zone with this.

I'd like it if the accessibility of the Create method was required to be no less than the accessibility of the collection type itself. So the collection type and Create method could both be internal, but you couldn't have an internal Create method with a public collection type. That seems like a strictly accidental situation where you can use collection literals only within the same assembly. (I assume it doesn't fall back to the collection initializer strategy if Create is inaccessible, but if it did, that would be even weirder.)

jnm2:
Some folks who don't like putting public methods on internal classes could be forced out of their comfort zone with this.

I'd like it if the accessibility of the Create method was required to be no less than the accessibility of the collection type itself. So the collection type and Create method could both be internal, but a public collection type couldn't point to an internal Create method. That seems like a strictly accidental situation where you can use collection literals only within the same assembly. (I assume it doesn't fall back to the collection initializer strategy if Create is inaccessible, but if it did, that would be even weirder.)

I like the principle we decided on, when deciding whether extension GetEnumerator should be considered for [CollectionBuilder] types, that it should be possible to put all diagnostics at the declaration site rather than the collection expression sites. If we were to allow the Create method to be less accessible than the collection type which points at it, we would fall foul of that principle.

```diff
+The method must be at least as accessible as the collection type which it creates.
```

This is different from saying "the method must have the same accessibility as the collection type which it creates," by the way, because the Create method could be more accessible than the collection type that points to it. The Create method's return type could be a public interface but be pointed to from an internal concrete collection type.

jnm2:
I am a bit worried that we need to make it clear we'd be talking about total accessibility, not just the immediate accessibility modifier on the method itself and the collection type itself.

Maybe this is better:
```diff
+The method must be accessible at all locations in which the collection type is accessible.
```

333fred:
As a dissenting voice here: I don't think we should have any requirements on the visibility of the builder method. The closest analogy to this, that I can see, is controlling the visibility of a constructor. And it seems perfectly reasonable to me that a user might want to make a collection type externally visible, while not allowing it to be constructed manually.

In other words, I think the spec change proposed by Bill here is the one we should make.

KathleenDollard:
I strongly feel we should not make a change for .NET 8. If _accessible_ is insufficient, we can handle it as a feature change later.

Overall, it seems reasonable to me that people will want to limit scope as much as practical to get their job done, and thus think this is also the best route going forward. 

So, I am in favor of Bill's suggested change to the spec.

cston (approved):

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -169,7 +169,8 @@ For the *create method*:
 
 * The *builder type* must be a non-generic `class` or `struct`.
 * The method must be defined on the *builder type* directly.
-* The method must be `public` and `static`.
+* The method must be `static`.
+* The method must be accessible.
```
333fred, line 173:
```suggestion
* The method must be accessible where the collection expression is used.
```

PR #7541, opened 2023-09-15
Collection expressions: construction

cston:
Describe evaluation order and methods used in collection instance construction.

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -216,7 +216,7 @@ ImmutableArray<int> ia =
 The elements of a collection expression are *evaluated* in order, left to right.
 Each element is evaluated exactly once, and any further references to the elements refer to the results of this initial evaluation.
 
-Spread elements are *iterated* in the order the elements appear in the collection expression, left to right. A spread element may be *iterated* before or after the subsequent elements in the collection expression are *evaluated*.
+A spread element may be *iterated* before or after the subsequent elements in the collection expression are *evaluated*.
 
 An unhandled exception thrown from any of the methods used during construction will be uncaught and will prevent further steps in the construction.
 
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -261,9 +261,9 @@ If the target type is an *array*, a *span*, a type with a *[create method](#crea
     * An applicable `GetEnumerator` instance or extension method is invoked on the *spread element expression* and for each item from the enumerator, the collection *indexer* is invoked to add the item at the current index. If the enumerator implements `IDisposable`, then `Dispose` will be called after enumeration, regardless of exceptions.
     * An applicable `CopyTo` instance or extension method is invoked on the *spread element expression* with the collection instance and `int` index as arguments.
 
-* If the target type has a *create method*, the create method is invoked with the span instance created initially or a span of the expected length from intermediate storage.
+* If intermediate storage was allocated for the collection, a collection instance is allocated with the actual collection length and the values from the intermediate storage are copied to the target collection instance, or if a span is required the compiler *may* use a span of the actual collection length from the intermediate storage.
 
-* Otherwise if intermediate storage was allocated for the collection, a collection instance is allocated with the actual length, and the values from the intermediate storage are copied to the target collection instance.
+* If the target type has a *create method*, the create method is invoked with the span instance.
 
 > *Note:*
 > The compiler may *delay* adding elements to the collection &mdash; or *delay* iterating through spread elements &mdash; until after evaluating subsequent elements. (When subsequent spread elements have *countable* properties that would allow calculating the expected length of the collection before allocating the collection.) Conversely, the compiler may *eagerly* add elements to the collection &mdash; and *eagerly* iterate through spread elements &mdash; when there is no advantage to delaying.
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -213,13 +213,11 @@ ImmutableArray<int> ia =
 ## Construction
 [construction]: #construction
 
-A collection expression has a *known length* if the compile-time type of each *spread element* in the collection expression is [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types).
-The *countable* properties, `Length` and `Count`, are assumed to have no side effects.
-The *countable* properties do not include *explicit implementations* of `ICollection<T>.Count` or `ICollection.Count`.
+`Length` and `Count` properties, and `GetEnumerator` methods, are assumed to have no side effects.
 
 If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance as follows:
 
-* If the collection expression does not contain *spread elements*, and the collection type contains an applicable constructor with a single `int capacity` parameter, the compiler *may* invoke that constructor with a non-zero capacity value. Otherwise, the constructor that is applicable with no arguments is invoked.
+* The constructor that is applicable with no arguments is invoked.
 
 * For each element in order:
   * If the element is an *expression element*, the applicable `Add` instance or extension method is invoked with the element *expression* as the argument. (Interleaving `Add` calls matches classic [*collection initializer behavior*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#117154-collection-initializers).)
@@ -231,16 +229,21 @@ If the target type is an *array*, a *span*, a type with a *[create method](#crea
 
 * The elements expressions are evaluated in order, exactly once each. Any further references to the element expressions refer to the results of this initial evaluation. All element expressions are evaluated before any references to the evaluated expressions.
 
-* If the target type is an *array*, a *span*, or a type with a *create method*, and the collection expression has a *known length*, the compiler will allocate the underlying storage directly with the expected length. If the collection length is not known, an intermediate buffer *may be* used to construct the collection before copying to the target collection instance.
-
-* *The compiler *may* attempt to determine the length *at runtime* for otherwise unknown length spread elements, by using `ICollection<T>` or `ICollection`, or using alternatives when the behavior difference is not observable &mdash; for well-known collection types in particular.*
+* A collection instance is created as follows:
+  * If the target type is an *array* and the collection expression has no *spread elements*, an array is allocated with the expected length.
+  * If the target type is a *span* or a type with a *create method*, and the collection has no *spread elements*, a span with the expected length is created referring to allocated storage.
+  * Otherwise intermediate storage is allocated.
+  * *The compiler may determine the expected length of a collection expression that contains spread elements, at compile-time or at runtime, by checking for [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties or well-known interfaces or types on each spread element. In those cases, the compiler may allocate the target *array* or *span* with the expected length directly rather than allocating intermediate storage.*
 
 * For each element in order:
   * If the element is an *expression element*, the evaluated expression is assigned to the collection.
   * If the element is a *spread element* then:
-    * If the spread element *expression* type implements `IEnumerable<T>`, the interface implementation *may be* used to iterate the items, with each item added to the collection in order.
-    * Otherwise, the spread element *expression* is iterated using the applicable `GetEnumerator` instance or extension method, and each item from the enumerator is assigned to the collection in order.
-    * *The compiler *may* use alternatives for copying items from the spread element *expression* when the behavior difference is not observable &mdash; when copying between well-known collection types in particular.*
+    * The spread element *expression* is iterated using the applicable `GetEnumerator` instance or extension method, and each item from the enumerator is assigned to the collection in order.
+    * *The compiler *may* copy items from the spread element *expression* by using well-known interfaces or types directly when the behavior difference is not otherwise observable.*
+
+* If intermediate storage was allocated for the collection, a collection instance is allocated with the expected type and length, and the values from the intermediate storage are copied to the target collection instance.
+
+* If the target type has a *create method*, the create method is invoked with the span instance.
 
 ## Empty collection literal
 
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -214,8 +214,8 @@ ImmutableArray<int> ia =
 [construction]: #construction
 
 A collection expression has a *known length* if the compile-time type of each *spread element* in the collection expression is [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types).
-
-*Does *countable* include explicit implementations of `ICollection<T>.Count` and `ICollection.Count`? We should check for those interfaces as well.*
+The *countable* properties, `Length` and `Count`, are assumed to have no side effects.
+The *countable* properties do not include *explicit implementations* of `ICollection<T>.Count` or `ICollection.Count`.
 
 If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance as follows:
 
@@ -230,8 +230,6 @@ If the target type is a *struct* or *class type* that implements `System.Collect
 If the target type is an *array*, a *span*, a type with a *[create method](#create-methods)*, or an *interface*, the construction of the collection instance is as follows:
 
 * The elements expressions are evaluated in order, exactly once each. Any further references to the element expressions refer to the results of this initial evaluation. All element expressions are evaluated before any references to the evaluated expressions.
-
-* If the target type is an *array*, a *span*, or a type with a *create method*, and the collection expression has a *known length*, the compiler will allocate the underlying storage directly using the 
 
 * If the target type is an *array*, a *span*, or a type with a *create method*, and the collection expression has a *known length*, the compiler will allocate the underlying storage directly with the expected length. If the collection length is not known, an intermediate buffer *may be* used to construct the collection before copying to the target collection instance.
 
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -213,6 +213,9 @@ ImmutableArray<int> ia =
 ## Construction
 [construction]: #construction
 
+The element expressions of a collection expression are evaluated in order, left to right.
+The element expressions are evaluated exactly once each &mdash; assuming no exceptions are thrown evaluating preceding elements &mdash; and any further references to the element expressions refer to the results of this initial evaluation.
+
 `Length` and `Count` properties, and `GetEnumerator` methods, are assumed to have no side effects.
 
 If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance as follows:
@@ -221,9 +224,12 @@ If the target type is a *struct* or *class type* that implements `System.Collect
 
 * For each element in order:
   * If the element is an *expression element*, the applicable `Add` instance or extension method is invoked with the element *expression* as the argument. (Interleaving `Add` calls matches classic [*collection initializer behavior*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#117154-collection-initializers).)
-  * If the element is a *spread element* then:
-    * If the target type has an `AddRange` instance or extension method that is applicable when invoked with the spread element *expression*, the `AddRange` method *may be* invoked with expression as the argument.
-    * Otherwise, the spread element *expression* is iterated using the applicable `GetEnumerator` instance or extension method, and the applicable `Add` instance or extension method is invoked for each item from the enumerator in order.
+  * If the element is a *spread element* then one of the following is used:
+    * An applicable `AddRange` instance or extension method is invoked on the *collection instance* with the spread element *expression* as the argument.
+    * An applicable `CopyTo` instance or extension method is invoked on the *spread element expression* with the collection instance and `int` index as arguments.
+    * The spread element *expression* is iterated using the applicable `GetEnumerator` instance or extension method, and the applicable `Add` instance or extension method is invoked for each item from the enumerator in order.
+
+* During the construction steps above, an applicable `EnsureCapacity` instance or extension method *may* be invoked one or more times on the *collection instance* with an `int` capacity argument.
 
 If the target type is an *array*, a *span*, a type with a *[create method](#create-methods)*, or an *interface*, the construction of the collection instance is as follows:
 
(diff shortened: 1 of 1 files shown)


proposals/csharp-12.0/collection-expressions.md
```diff
@@ -116,28 +116,22 @@ The following implicit *collection expression conversions* exist from a collecti
...
-  * For each *element* `Ei` there is an *implicit conversion* to `T`.
-
-* To a *type* that implements `System.Collections.IEnumerable` where:
-
-  * The *type* contains an applicable instance constructor that can be invoked with no arguments or invoked with a single argument for the 0-th parameter where the parameter has type `System.Int32` and name `capacity`.
+  * The *type* contains an applicable instance constructor that can be invoked with no arguments.
   * For each *expression element* `Ei` there is an applicable instance or extension method `Add` for a single argument `Ei`.
   * For each *spread element* `Si` there is an applicable instance or extension method `Add` for a single argument of the *iteration type* of `Si`.
 
-  *Open issue: Relying on a parameter named `capacity` seems brittle. Is there an alternative?*
```
CyrusNajmabadi, line 129: i don't like that we got rid of capacity.  IMO this is an optimization that important for our perf story.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,36 @@ ImmutableArray<int> ia =
...
 [construction]: #construction
 
-*Give specific ordering for determining how to construct the constructible collection types.*
+A collection expression has a *known length* if the compile-time type of each *spread element* in the collection expression is [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types).
+
+*Does *countable* include explicit implementations of `ICollection<T>.Count` and `ICollection.Count`? We should check for those interfaces as well.*
+
+If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance as follows:
+
+* If the collection expression does not contain *spread elements*, and the collection type contains an applicable constructor with a single `int capacity` parameter, the compiler *may* invoke that constructor with a non-zero capacity value. Otherwise, the constructor that is applicable with no arguments is invoked.
```
CyrusNajmabadi, line 222: this doesn't seem good to me either.  even if we have spread elements, we can call the capacity constructor.  The primary differentiation is if something has sttaically known length, or runtime known length.

proposals/csharp-12.0/collection-expressions.md
cston: We haven't removed the potential use of `.ctor(int capacity)`, it still might be used during construction, but the availability of that constructor shouldn't affect convertibility.

proposals/csharp-12.0/collection-expressions.md
cston: If we need to interleave element evaluation and `Add` calls, I don't think we can determine the length of all spread elements before the collection instance is constructed.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,36 @@ ImmutableArray<int> ia =
...
+
+* *The compiler *may* attempt to determine the length *at runtime* for otherwise unknown length spread elements, by using `ICollection<T>` or `ICollection`, or using alternatives when the behavior difference is not observable &mdash; for well-known collection types in particular.*
+
+* For each element in order:
+  * If the element is an *expression element*, the evaluated expression is assigned to the collection.
+  * If the element is a *spread element* then:
+    * If the spread element *expression* type implements `IEnumerable<T>`, the interface implementation *may be* used to iterate the items, with each item added to the collection in order.
+    * Otherwise, the spread element *expression* is iterated using the applicable `GetEnumerator` instance or extension method, and each item from the enumerator is assigned to the collection in order.
+    * *The compiler *may* use alternatives for copying items from the spread element *expression* when the behavior difference is not observable &mdash; when copying between well-known collection types in particular.*
```
CyrusNajmabadi, line 245: should we mention TryGetNonEnumeratedCount at all? or is it implied here?

proposals/csharp-12.0/collection-expressions.md
CyrusNajmabadi: > If we need to interleave element evaluation and Add calls

the question is whether we need to do that or not.  I'd say this makes sense for a collection *without* spreads (since we know syntactically the number of elemnets).  But if there's a computation of element-count, it probably makes sense to determine that first, create the collection with the right capacity, then add to that :)

cston: We should probably be consistent with interleaving element evaluation and `Add` calls regardless of whether there are spread elements. We could decide to evaluate all elements before adding any, although that would be a change from the classic *collection initializer* behavior.

proposals/csharp-12.0/collection-expressions.md
cston: I think `TryGetNonEnumeratedCount` is implied here.

The intent was to be specific about which methods and interfaces the compiler might use, so it's clear when a spread expression type is conforming, without necessarily referencing specific helpers that might use those methods or interfaces.

proposals/csharp-12.0/collection-expressions.md
CyrusNajmabadi: Yeah, let's bring to LDM.  there's pros/cons each way.  :)

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,36 @@ ImmutableArray<int> ia =
 ## Construction
 [construction]: #construction
 
-*Give specific ordering for determining how to construct the constructible collection types.*
+A collection expression has a *known length* if the compile-time type of each *spread element* in the collection expression is [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types).
+
+*Does *countable* include explicit implementations of `ICollection<T>.Count` and `ICollection.Count`? We should check for those interfaces as well.*
```
RikkiGibson, line 218: Precedent from list patterns is to ignore the count if it's only provided in an explicit implementation

https://sharplab.io/#v2:EYLgtghglgdgNAFxFANnAJiA1AHwAIBMAjALABQeADAAR5EB0AwgPYooCmAxglMzAM4BuclVpEALMLLkAbhABO1TtQC81GOwDu1RgAoAlFKgAzaruVR+1ANpE4BALr7qAb2oBfciIDMtAjuoQagBJFjYuHj4AHlgEAD5yF3JqFOpYkLCObl4YGJh4pmYAV3zVOOoEAAt5Zm0NbQA5ZgRgsAAHDjB2fPZ0AFEAD052NsiYAylk1OBmVgzWLLG8guD+ACV2CHQAeRgUAE8yiura9S1qJpb2zu6EXsHh0ZyJrzJU2nF58Ozo2Lj6ACC6HQunSUDuYH0UxSSTe71SVRqdXOl1aHXYXR6/SGIzGLzhKU80gJHy+ixyy3+jA4CgM0Nc9PeiNO9QuzTRNyxD1xz0M9KJ9Jmc1CCwiFL+hXy0AEoNK4IxUJJsPhCJOyMa7OuGNu9xxTz4+PeApJeE+Iu+SwlLDa+wAKsxZQhrA5qAp5BB9nA0qU3R7gjB0OwBor3sqVcckWcNVd0Zi7tjHni+STje9gn0YEUuu6EMx5JSxN5KfQAOLsBAZrPsHN5ulKxmqyOs1FauO6xO8qRG15pyvZiC5xR0byl8t96sD2sh1JhlXM9VsmOc+Pc/XjZPd4nvIUoMli375f4bMDMGTsR1pCHTmENlLzqOLjnarl6pNd1JE9xAA==

proposals/csharp-12.0/collection-expressions.md
BillWagner: When I parsed the new text, I interpreted it to mean the parameterless ctor is mandatory, and the `(int capacity)` ctor is optional. Meaning that if a type had a ctor with one parameter named `capacity`, but did not have an accessible parameterless ctor, that type couldn't participate in collection expressions. Is that the intent? 

I can see a case for both choices, so mostly just curious.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,34 @@ ImmutableArray<int> ia =
...
+
+* For each element in order:
+  * If the element is an *expression element*, the applicable `Add` instance or extension method is invoked with the element *expression* as the argument. (Interleaving `Add` calls matches classic [*collection initializer behavior*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#117154-collection-initializers).)
+  * If the element is a *spread element* then:
+    * If the target type has an `AddRange` instance or extension method that is applicable when invoked with the spread element *expression*, the `AddRange` method *may be* invoked with expression as the argument.
+    * Otherwise, the spread element *expression* is iterated using the applicable `GetEnumerator` instance or extension method, and the applicable `Add` instance or extension method is invoked for each item from the enumerator in order.
+
+If the target type is an *array*, a *span*, a type with a *[create method](#create-methods)*, or an *interface*, the construction of the collection instance is as follows:
```
BillWagner, line 230: Is the intent that any other `ref struct` type other than `Span<T>` and `ReadOnlySpan<T>` use a *create_method*? Or should the language be modified so *span* means any `ref struct` type that matches the proper shape?

BillWagner (approved):
This looks good @cston 

I had a couple clarifying questions, but that's it.

proposals/csharp-12.0/collection-expressions.md
cston: Yes, a constructor that can be invoked with no arguments is required for the *collection expression conversion*, and the constructor with `int capacity` parameter was optional.

proposals/csharp-12.0/collection-expressions.md
cston: A `ref struct` type could be implemented using the *collection initializer type* pattern or it could have a *create method*. The intent here was to be explicit about the expected types other than *collection initializer types*.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,37 @@ ImmutableArray<int> ia =
 ## Construction
 [construction]: #construction
 
-*Give specific ordering for determining how to construct the constructible collection types.*
+`Length` and `Count` properties, and `GetEnumerator` methods, are assumed to have no side effects.
+
+If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance as follows:
```
333fred, line 218: Missing a word at the end of the sentence; either a "proceeds as follows", or "occurs", or some other phrasing of your choice.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,37 @@ ImmutableArray<int> ia =
...
+
+* A collection instance is created as follows:
+  * If the target type is an *array* and the collection expression has no *spread elements*, an array is allocated with the expected length.
+  * If the target type is a *span* or a type with a *create method*, and the collection has no *spread elements*, a span with the expected length is created referring to allocated storage.
+  * Otherwise intermediate storage is allocated.
+  * *The compiler may determine the expected length of a collection expression that contains spread elements, at compile-time or at runtime, by checking for [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties or well-known interfaces or types on each spread element. In those cases, the compiler may allocate the target *array* or *span* with the expected length directly rather than allocating intermediate storage.*
+
+* For each element in order:
+  * If the element is an *expression element*, the evaluated expression is assigned to the collection.
```
333fred, line 239: I have no idea what this means. What is "assigned" here?

333fred (commented):
I don't see anything in this doc that specifies when enumeration occurs, and clarifies that it can occur before or after evaluating the rest of the expressions in the collection.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+
+`Length` and `Count` properties, and `GetEnumerator` methods, are assumed to have no side effects.
+
+---
+
+If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance is as follows:
+
+* The elements are evaluated in order. Some or all elements may be evaluated *during* the steps below rather than before.
+
+* The compiler *may* determine the *known length* of the collection expression by invoking [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties &mdash; or equivalent properties from well-known interfaces or types &mdash; on each *spread element expression*.
```
CyrusNajmabadi, line 231: this doesn't preclude us using TryGetNonEnumeratedCount in the future right?

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+  * Otherwise intermediate storage is allocated.
+
+* For each element in order:
+  * If the element is an *expression element*, the collection *indexer* is invoked to add the evaluated expression at the current index.
+  * If the element is a *spread element* then one of the following is used:
+    * A member of a well-known interface or type is invoked to copy items from the spread element expression to the collection instance.
+    * An applicable `GetEnumerator` instance or extension method is invoked on the *spread element expression* and for each item from the enumerator, the collection *indexer* is invoked to add the item at the current index. If the enumerator implements `IDisposable`, then `Dispose` will be called after enumeration, regardless of exceptions.
+    * An applicable `CopyTo` instance or extension method is invoked on the *spread element expression* with the collection instance and `int` index as arguments.
+
+* If intermediate storage was allocated for the collection, a collection instance is allocated with the actual length, and the values from the intermediate storage are copied to the target collection instance.
```
CyrusNajmabadi, line 264: 1. perhaps mention that if the length is correct, no copy is needed.
2. for array, i get why we need to copy, so that part is fine.  However, for span, it seems like we could just point the span at the portion of the array we filled.  Given that the span will be short lived (Since it can only be on the stack), that might be nice to do instead.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+
+* If the target type has a *create method*, the create method is invoked with the span instance.
```
CyrusNajmabadi, line 266: this is a case where the copy would be pointless, and would be good to avoid.  We're about to call into the Create method, so instead of copying, and getting a span to the copy, we should just make a subspan around the intermediate storage we used.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+
+If the target type is an *array*, a *span*, a type with a *[create method](#create-methods)*, or an *interface*, the construction of the collection instance is as follows:
+
+* The elements are evaluated in order. Some or all elements may be evaluated *during* the steps below rather than before.
+
+* The compiler *may* determine the *known length* of the collection expression by invoking [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties &mdash; or equivalent properties from well-known interfaces or types &mdash; on each *spread element expression*.
+
+* A collection instance is created as follows:
+  * If the target type is an *array* and the collection expression has a *known length*, an array is allocated with the expected length.
+  * If the target type is a *span* or a type with a *create method*, and the collection has a *known length*, a span with the expected length is created referring to allocated storage.
```
CyrusNajmabadi, line 254: for the span case, this allows us to stack-allocate, or heap-allocate depending on what the compiler wants, right?  this doesn't imply that it's always heap allocated, right?

proposals/csharp-12.0/collection-expressions.md
cston: It shouldn't and I didn't intend it to.

proposals/csharp-12.0/collection-expressions.md
cston: The "intermediate storage" case is only for when the collection does not have a known length (see line 255).

CyrusNajmabadi: note: even for the case with a span destination, we likely still don't want a final copy of the array.  INstead, we should just make a span over the subset of the array containing elements.  

In other words, say we're using an array, starting at 8 elements large, and growing it by doubling.

If we add a total of 10 elements, we should double once (to 16), then get a span pointing to the first 10 elements.  We should *not* take those 16-element-array and copy out the 10 elements into a 10-element array that is then wrapped by a span.

Does that make sense?

proposals/csharp-12.0/collection-expressions.md
cston: Good catch. We should use a span from intermediate storage if possible. Updated here and we should fix in the implementation.

proposals/csharp-12.0/collection-expressions.md
cston: It wasn't meant to imply heap allocation. Changed to "contiguous storage".

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+
+* The compiler *may* determine the *known length* of the collection expression by invoking [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties &mdash; or equivalent properties from well-known interfaces or types &mdash; on each *spread element expression*.
+
+* A collection instance is created as follows:
+  * If the target type is an *array* and the collection expression has a *known length*, an array is allocated with the expected length.
+  * If the target type is a *span* or a type with a *create method*, and the collection has a *known length*, a span with the expected length is created referring to contiguous storage.
+  * Otherwise intermediate storage is allocated.
+
+* For each element in order:
+  * If the element is an *expression element*, the collection *indexer* is invoked to add the evaluated expression at the current index.
```
333fred, line 258: This isn't the collection indexer, it's the intermediate storage indexer.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+  * If the element is a *spread element* then one of the following is used:
+    * A member of a well-known interface or type is invoked to copy items from the spread element expression to the collection instance.
+    * An applicable `GetEnumerator` instance or extension method is invoked on the *spread element expression* and for each item from the enumerator, the collection *indexer* is invoked to add the item at the current index. If the enumerator implements `IDisposable`, then `Dispose` will be called after enumeration, regardless of exceptions.
+    * An applicable `CopyTo` instance or extension method is invoked on the *spread element expression* with the collection instance and `int` index as arguments.
+
+* If intermediate storage was allocated for the collection, a collection instance is allocated with the actual collection length and the values from the intermediate storage are copied to the target collection instance, or if a span is required the compiler *may* use a span of the actual collection length from the intermediate storage.
+
+* If the target type has a *create method*, the create method is invoked with the span instance.
+
+> *Note:*
```
333fred, line 268: The way this is structured it feels like this note is only applicable to `Create` methods, when it's actually applicable to both of these sections.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+
+---
+
+If the target type is a *struct* or *class type* that implements `System.Collections.IEnumerable`, and the target type does not have a *[create method](#create-methods)*, the construction of the collection instance is as follows:
+
+* The elements are evaluated in order. Some or all elements may be evaluated *during* the steps below rather than before.
+
+* The compiler *may* determine the *known length* of the collection expression by invoking [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties &mdash; or equivalent properties from well-known interfaces or types &mdash; on each *spread element expression*.
+
+* The constructor that is applicable with no arguments is invoked.
```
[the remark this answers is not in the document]
cston: Added a link to the definition in the *Conversions* section. And if there is no applicable constructor, we wouldn't have reached *Construction*.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
+
+If the target type is an *array*, a *span*, a type with a *[create method](#create-methods)*, or an *interface*, the construction of the collection instance is as follows:
+
+* The elements are evaluated in order. Some or all elements may be evaluated *during* the steps below rather than before.
+
+* The compiler *may* determine the *known length* of the collection expression by invoking [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties &mdash; or equivalent properties from well-known interfaces or types &mdash; on each *spread element expression*.
+
+* A collection instance is created as follows:
+  * If the target type is an *array* and the collection expression has a *known length*, an array is allocated with the expected length.
+  * If the target type is a *span* or a type with a *create method*, and the collection has a *known length*, a span with the expected length is created referring to contiguous storage.
```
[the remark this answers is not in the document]
cston: Lifetime is covered in the *Ref safety*. I think I'd prefer to leave it unspecified here.

proposals/csharp-12.0/collection-expressions.md
cston: It's the indexer on the *collection instance* created in the previous top-level bullet. I've added "instance" here to clarify.

proposals/csharp-12.0/collection-expressions.md
cston: Open to suggestions on how to clarify this.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,79 @@ ImmutableArray<int> ia =
...
 
-*Give specific ordering for determining how to construct the constructible collection types.*
+The elements of a collection expression are *evaluated* in order, left to right.
+Each element is evaluated exactly once, and any further references to the elements refer to the results of this initial evaluation.
+
+A spread element may be *iterated* before or after the subsequent elements in the collection expression are *evaluated*.
+
+An unhandled exception thrown from any of the methods used during construction will be uncaught and will prevent further steps in the construction.
+
+`Length` and `Count` properties, and `GetEnumerator` methods, are assumed to have no side effects.
```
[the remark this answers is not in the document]
cston: I think I'd prefer to leave this terse.

proposals/csharp-12.0/collection-expressions.md
333fred: The way this is currently written, it says that the type under construction must have an indexer. That isn't the case. The intermediate storage, which, in a few cases, is the target type itself, must have an indexer.

proposals/csharp-12.0/collection-expressions.md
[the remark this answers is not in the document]
333fred: All you need to do to make it not a sentence fragment is remove the last comma :).

333fred (approved):

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,81 @@ ImmutableArray<int> ia =
...
+
+If the target type is an *array*, a *span*, a type with a *[create method](#create-methods)*, or an *interface*, the construction of the collection instance is as follows:
+
+* The elements are evaluated in order. Some or all elements may be evaluated *during* the steps below rather than before.
+
+* The compiler *may* determine the *known length* of the collection expression by invoking [*countable*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/ranges.md#adding-index-and-range-support-to-existing-library-types) properties &mdash; or equivalent properties from well-known interfaces or types &mdash; on each *spread element expression*.
+
+* An *initialization instance* is created as follows:
+  * If the target type is an *array* and the collection expression has a *known length*, an array is allocated with the expected length.
+  * If the target type is a *span* or a type with a *create method*, and the collection has a *known length*, a span with the expected length is created referring to contiguous storage.
```
CyrusNajmabadi, line 254: this reads fine to em.   it might be ok to mention that such contiguous storage migth be on the stack or on the heap.  but it's ok to not mention as well.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -219,7 +213,81 @@ ImmutableArray<int> ia =
...
+  * If the element is a *spread element* then one of the following is used:
+    * A member of a well-known interface or type is invoked to copy items from the spread element expression to the initialization instance.
+    * An applicable `GetEnumerator` instance or extension method is invoked on the *spread element expression* and for each item from the enumerator, the initialization instance *indexer* is invoked to add the item at the current index. If the enumerator implements `IDisposable`, then `Dispose` will be called after enumeration, regardless of exceptions.
+    * An applicable `CopyTo` instance or extension method is invoked on the *spread element expression* with the initialization instance and `int` index as arguments.
+
+* If intermediate storage was allocated for the collection, a collection instance is allocated with the actual collection length and the values from the initialization instance are copied to the collection instance, or if a span is required the compiler *may* use a span of the actual collection length from the intermediate storage. Otherwise the initialization instance is the collection instance.
```
CyrusNajmabadi, line 264: nice.

CyrusNajmabadi (approved):

PR #7547, opened 2023-09-20
Collection expressions: Update better conversion from expression rule

cston:
Update *better conversion from expression* rule following LDM 2023-09-20.

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -356,13 +356,16 @@ var z = Extensions.AsImmutableArray([3]); // ok
 
 [*Better conversion from expression*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11644-better-conversion-from-expression) is updated to prefer certain target types in collection expression conversions.
 
+In the following rules, an *array_or_array_interface_type* is a type that is either an *array type* or one of the following *interface types* implemented by an array type: `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`.
+
 > Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if one of the following holds:
 >
 > * `E` exactly matches `T₁` and `E` does not exactly match `T₂`
-> * `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a [*better conversion target*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11646-better-conversion-target) than `T₂` **and one of the following holds:**
->   * **`C₁` and `C₂` are *not* collection expression conversions or**
->   * **`T₁` and `T₂` are both *ref struct types* or neither are *ref struct types***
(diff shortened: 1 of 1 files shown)


333fred (approved):

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -356,13 +356,17 @@ var z = Extensions.AsImmutableArray([3]); // ok
 
 [*Better conversion from expression*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11644-better-conversion-from-expression) is updated to prefer certain target types in collection expression conversions.
 
+In the updated rules:
+* A *span_type* is `System.Span<T>` or `System.ReadOnlySpan<T>`.
+* An *array_or_array_interface_type* is an *array type* or one of the following *interface types* from `System.Collections.Generic` implemented by an *array type*: `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, `IList<T>`.
```
CyrusNajmabadi, line 361: i believe the spec already states this elsewhere (i think in the conversions section).  Let's link there instead?  That way if that list updates, it applies here?

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -356,13 +356,17 @@ var z = Extensions.AsImmutableArray([3]); // ok
...
 >
 > * `E` exactly matches `T₁` and `E` does not exactly match `T₂`
-> * `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a [*better conversion target*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11646-better-conversion-target) than `T₂` **and one of the following holds:**
->   * **`C₁` and `C₂` are *not* collection expression conversions or**
->   * **`T₁` and `T₂` are both *ref struct types* or neither are *ref struct types***
-> * **`C₁` and `C₂` are collection expression conversions, and `T₁` is a *ref struct type* with *iteration type* `E₁`, and `T₂` is a *non&dash; ref struct type* with *iteration type* `E₂`, and `E₁` is implicitly convertible to `E₂`**
+> * `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a [*better conversion target*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11646-better-conversion-target) than `T₂` **and the following is *not true*:**
```
CyrusNajmabadi, line 366:
```suggestion
> * `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a [*better conversion target*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11646-better-conversion-target) than `T₂` **and the following are all *not true*:**
```

? 

acutally, i don't know how to read it.  Do all need to be not-true? can one be not true?

can we invert this so that it's more `and all of hte following are true`, and change the inners?  

(i don't feel strongly here, i'm just mostly confused on how to read this :))

proposals/csharp-12.0/collection-expressions.md
cston: I will leave the interface types here explicitly for now since the two sections are otherwise unrelated.

proposals/csharp-12.0/collection-expressions.md
cston: The two clauses below are joined with "and" so both need to be true.

I tried several different forms for composing these conditions including inverting the cases and this was the most readable to me, but it's not easy to read.

CyrusNajmabadi (approved):

PR #7555, opened 2023-09-26
Expanding ref support

jaredpar:
This proposal expands the `ref` support in the compiler to include:

1. `ref struct` as `ref` fields
2. `ref scoped` parameters and fields
3. Fully sunsetting restricted types

At the moment I'm not proposing that LDM move forward with the design. I'm still deciding if the complexity / cost is worth the gain it provides. Yet recent discussions crystalized in my mind how this would all work hence I wanted to get it written down and out for review.

The intent here is to socialize these ideas, get feedback on where this would or would not help and then make a decision on whether to push this through LDM / future .NET release.

### DIFF
--- a/proposals/csharp-11.0/low-level-struct-improvements.md
+++ b/proposals/csharp-11.0/low-level-struct-improvements.md
@@ -1,5 +1,5 @@
 Low Level Struct Improvements
-=====
+====
 
 [!INCLUDE[Specletdisclaimer](../speclet-disclaimer.md)]
 
@@ -904,6 +904,12 @@ Lifetime variables when defined on types can be invariant or covariant. These ar
 ref struct S<out $this, $a> 
 ```
 
+The lifetime parameter `$this` on type definitions is _not_ predefined but it does have a few rules associated with it when it is defined:
+- It must be the first lifetime parameter.
+- It must be covariant: `out $this`. 
+- The lifetime of `ref` fields must be convertible to `$this`
+- The `$this` lifetime of all non-ref fields must `$heap` or `$this`.
+
 The lifetime of a ref is expressed by providing a lifetime argument to the ref. For example a `ref` that refers to the heap is expressed as `ref<$heap>`.
 
 When defining a constructor in the model the name `new` will be used for the method. It is necessary to have a parameter list for the returned value as well as the constructor arguments. This is necessary to express the relationship between constructor inputs and the constructed value. Rather than having `Span<$a><$ro>` the model will use `Span<$a> new<$ro>` instead. The type of `this` in the constructor, including lifetimes, will be the defined return value.
@@ -912,10 +918,12 @@ The basic rules for the lifetime are defined as:
 
 - All lifetimes are expressed syntactically as generic arguments, coming before type arguments. This is true for predefined lifetimes except `$heap` and `$local`. 
 - All types `T` that are not a `ref struct` implicitly have lifetime of `T<$heap>`. This is implicit, there is no need to write `int<$heap>` in every sample.
-- For a ref field defined as `ref T<$l1, $l2, ... $ln>` all lifetimes `$l1` through `$ln` must be invariant. 
-- For a ref defined as `ref<$a> T<$b, ...>`, `$b` must convertible to `$a`
-- The `ref` of a variable has a lifetime of the 
-    - For a ref local, parameter, field or return of type `ref<$a> T` the lifetime is `$a`
+- For a `ref` field defined as `ref<$l0> T<$l1, $l2, ... $ln>`:
+    - All lifetimes `$l1` through `$ln` must be invariant. 
+    - The lifetime of `$l0` must be convertible to `$this`
+- For a `ref`` defined as `ref<$a> T<$b, ...>`, `$b` must convertible to `$a`
+- The `ref` of a variable has a lifetime defined by:
+    - For a `ref` local, parameter, field or return of type `ref<$a> T` the lifetime is `$a`
     - `$heap` for all reference types and fields of reference types
     - `$local` for everything else
 - An assignment or return is legal when the underlying type conversion is legal
@@ -926,11 +934,6 @@ The basic rules for the lifetime are defined as:
 For the purpose of lifetime rules a `ref` is considered part of the type of the expression for purposes of conversions. It is logically represented by converting `ref<$a> T<...>` to `ref<$a, T<...>>` where `$a` is covariant and `T` is invariant. 
 
 Next let's define the rules that allow us to map C# syntax to the underlying model.
-
-The lifetime parameter `$this` on type definitions is _not_ predefined but it does have a few rules associated with it when it is defined:
-- It must be the first lifetime parameter.
-- It must be covariant: `out $this`. 
-- The lifetime parameters of all non-ref fields, and the ref lifetime of ref fields, must be `$this`
 
 For brevity sake a type which has no explicit lifetime parameters treated as if there is `out $this` defined and applied to all fields of the type. A type with a `ref` field must define explicit lifetime parameters.
 
--- a/proposals/csharp-11.0/low-level-struct-improvements.md
+++ b/proposals/csharp-11.0/low-level-struct-improvements.md
@@ -908,7 +908,7 @@ The lifetime parameter `$this` on type definitions is _not_ predefined but it do
 - It must be the first lifetime parameter.
 - It must be covariant: `out $this`. 
 - The lifetime of `ref` fields must be convertible to `$this`
-- The `$this` lifetime of all non-ref fields must `$heap` or `$this`.
+- The `$this` lifetime of all non-ref fields must be `$heap` or `$this`.
 
 The lifetime of a ref is expressed by providing a lifetime argument to the ref. For example a `ref` that refers to the heap is expressed as `ref<$heap>`.
 
@@ -921,7 +921,7 @@ The basic rules for the lifetime are defined as:
 - For a `ref` field defined as `ref<$l0> T<$l1, $l2, ... $ln>`:
     - All lifetimes `$l1` through `$ln` must be invariant. 
     - The lifetime of `$l0` must be convertible to `$this`
-- For a `ref`` defined as `ref<$a> T<$b, ...>`, `$b` must convertible to `$a`
+- For a `ref` defined as `ref<$a> T<$b, ...>`, `$b` must be convertible to `$a`
 - The `ref` of a variable has a lifetime defined by:
     - For a `ref` local, parameter, field or return of type `ref<$a> T` the lifetime is `$a`
     - `$heap` for all reference types and fields of reference types
--- a/proposals/expand-ref.md
+++ b/proposals/expand-ref.md
@@ -209,10 +209,10 @@ void M(ref scoped S s)
 // maps to 
 
 void M<$param1>(ref<$local> S<$param1> s)
-    where $param1 : $cm
+    where $param1 : $local
 ```
 
-This definition prevents the value from escaping from the method as the lifetime is not returnable. It also prevents local data from escaping from the current method through the parameter as the lifetime is wider that `$local` but not equivalent.
+This definition prevents the value from escaping from the method as the lifetime is not returnable. It also prevents local data from escaping from the current method through the parameter as the lifetime is wider than `$local` but not equivalent.
 
 ```csharp
 void M<$param1>(ref<$local> S<$param1> p)
--- a/proposals/expand-ref.md
+++ b/proposals/expand-ref.md
@@ -1,0 +1,254 @@
+Expanding ref support
+===
+
+## Summary
+This proposal expands the capabilities of `ref` and `scoped` in the language. The goal being to leverage the existing types of rules in the model to allow `ref struct` usage in more locations and provide more lifetime expressiveness for APIs.
+
+## Motivation
+There are still a number of scenarios around `ref` which cannot be safely expressed in the language. These are generally when using multiple mutable `ref struct` parameters where many are passed by `ref` or when trying to use `ref struct` in `ref` fields.
+
+To _fully_ satisfy all of these scenarios would require us to introduce explicit lifetime parameters and relationships into the language. That is a _huge_ investment that is not yet motivated by need. Instead this proposal takes our existing lifetime annotation, `scoped`, and sees how much further `ref` safety can be taken without introducing any other annotations or keywords. 
+
+This doesn't solve all scenarios but does remove several known friction points in the language. It also serves to show us exactly where the limits are without introducing explicit lifetime parameters.
+
+## Detailed Design
+The rules for `ref struct` safety are defined in the following documents:
+
+- [ref safety proposal](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md).
+- [ref fields proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md)
+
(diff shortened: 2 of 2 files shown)


jaredpar:
@cston, @RikkiGibson, @agocke, @AaronRobinsonMSFT

agocke:
Confirmed that I see this, but currently a bit underwater and it may take me some time to sit down and digest it

jaredpar:
@agocke 

> Confirmed that I see this, but currently a bit underwater and it may take me some time to sit down and digest it

No rush on this. Was just paged in from other discussions so I wanted to get it written out.

proposals/expand-ref.md
```diff
@@ -0,0 +1,254 @@
...
+    p2 = p1;
+
+    // Error: can't assign current parameter 1 to current parameter 2
+    p1 = p2;
+
+    // Okay: as current parameter 1 and 2 can be converted to current method
+    scoped Span<int> local1 = p1; 
+    scoped Span<int> local2 = p2; 
+
+    // Okay: however the safe-to-escape here is current parameter X, not
```
AaronRobinsonMSFT, line 82:
```suggestion
    // Okay: however the safe-to-escape here is current parameter N, not 
```

?

proposals/expand-ref.md
```diff
@@ -0,0 +1,254 @@
...
+    Span<int> local4 = p2; 
+
+    // Okay: the safe-to-escape of the value is inferred in this case as it is 
+    // done for ref locals today.
+    ref Span<int> refLocal1 = ref p1;
+    ref Span<int> refLocal2 = ref p2;
+}
+```
+
+A `ref scoped` parameter is also implicitly `scoped ref`. That means neither the value nor it's `ref` can be returned from the method. Both `ref` and `in` parameters can have their values modified with `scoped`. An `out` parameter cannot have its value modified with `scoped` as such a declaration is non-sensical.
```
AaronRobinsonMSFT, line 94:
```suggestion
A `ref scoped` parameter is also implicitly `scoped ref`. That means neither the value nor its `ref` can be returned from the method. Both `ref` and `in` parameters can have their values modified with `scoped`. An `out` parameter cannot have its value modified with `scoped` as such a declaration is non-sensical. 
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,254 @@
...
+    in scoped Span<int> p2,     // Okay
+    out scoped Span<int> p2,    // Error
+)
+```
+
+The [method arguments must match](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#rules-method-arguments-must-match) rules will be updated to take `ref scoped` into account. Values passed to such parameters do not need to be considered when calculating the return scopes.
+
+Detailed notes:
+- A `ref scoped` parameter is implicitly `scoped ref`
+- A `out scoped` parameter declaration is an error
```
AaronRobinsonMSFT, line 108:
```suggestion
- An `out scoped` parameter declaration is an error
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,254 @@
...
+        // okay
+        reader.Skip();
+
+        // Error: can't escape the ref data the ref scoped field refers to
+        return reader.ValueSpan;
+    }
+}
+```
+
+This is accomplished by giving every `ref scoped` field a new two new escape scopes named _current field N_ and _current ref field N_ where _N_ is the numeric order of the field. For example the first field has a _safe-to-escape_ of _current field 1_ and a _ref-safe-to-escape_ of _current ref field N_. Both escape scopes can be converted to _current method_, and _current field N_ can be converted to _current ref field N_, but no other defined relationships exist. That serves to restrict their usage to the current method where the containing value is used. This escape scope applies to both
```
AaronRobinsonMSFT, line 132:
```suggestion
This is accomplished by giving every `ref scoped` field two new escape scopes named _current field N_ and _current ref field N_ where _N_ is the numeric order of the field. For example, the first field has a _safe-to-escape_ of _current field 1_ and a _ref-safe-to-escape_ of _current ref field N_. Both escape scopes can be converted to _current method_, and _current field N_ can be converted to _current ref field N_, but no other defined relationships exist. That serves to restrict their usage to the current method where the containing value is used. This escape scope applies to both.
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,254 @@
...
+Detailed notes:
+- A `ref` field where the type is a `ref struct` must be `ref scoped`
+- A `ref` field may be marked `scoped ref`
+
+### Sunset restricted types
+The ability for any type to be a `ref` field allows us to fully sunset the notion of restricted types.  The compiler has a concept of a set of _restricted types_ which is largely undocumented. These types were given a special status because in C# 1.0 there was no general purpose way to express their behavior. Most notably the fact that the types can contain references to the execution stack. Instead the compiler had special knowledge of them and restricted their use to ways that would always be safe: disallowed returns, cannot use as array elements, cannot use in generics, etc ...
+
+Once `ref` fields are available and extended to support `ref struct` these types can be fully rationalized within those rules. As such the compiler will no longer have the notion of restricted types when using a language version that supports `ref` fields of `ref struct`.
+
+To support this our `ref` safety rules will be updated as follows:
```
AaronRobinsonMSFT, line 187: Can we assume updates to the following don't indicate anything other than clarity on their semantics rather than changing anything about their "officially supported" state.

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+    {
+        // Okay
+        field = new(); 
+
+        // Error: calling-method is not convertible to current-field-1 as they have 
+        // no relationship
+        field = s;
+
+        // Error: safe-to-escape is current-field-1 which isn't returnable 
+        return field;
```
jaredpar, line 157: Actually I think these returns can be allowed. I need to tweak the rules slightly so that `current-param-N` is related to `this` not `local`.

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+}
+S M(S p) { }
+
+// maps to 
+ref struct S<out $this, $refField1>
+{
+    ref<$refField1> int i;
+}
+
+S<$cm> M<$cm, $l1>M(S<$cm, $l1> p)
```
AaronRobinsonMSFT, line 245:
```suggestion
S<$cm> M<$cm, $l1>(S<$cm, $l1> p)
```

?

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+    where $l1 : $local
+{
+
+}
+```
+
+Every field marked as `ref scoped` will have a new lifetime parameter defined. Th name will be `$fieldN` where _N_ is the numerical order of the field. That lifetime will have the relationship `where $fieldN : $refFieldN` defined on the type. It will also have the relationship `where $fieldN : $local` in all method that use the type.
```
AaronRobinsonMSFT, line 252:
```suggestion
Every field marked as `ref scoped` will have a new lifetime parameter defined. The name will be `$fieldN` where _N_ is the numerical order of the field. That lifetime will have the relationship `where $fieldN : $refFieldN` defined on the type. It will also have the relationship `where $fieldN : $local` in all method that use the type.
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+The more detailed rules will rely on the [annotation syntax](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#annotations) to describe the detailed rules. This is the most direct way to discuss how syntax behaves in the greater model. Readers interested in the very low level details should familiarize themselves with that syntax before digesting this proposal. 
+
+### ref scoped parameters
+The language will allow for parameters to be declared as `ref scoped`. This will serve to constrain the _safe-to-escape_ of the value such that it cannot be returned from the current method. 
+
+```csharp
+Span<int> M(Span<T> p1, ref scoped Span<int> p2)
+{
+    // Error: cannot returned scoped value
+    return span;
```
jjonescz, line 31:
```suggestion
    // Error: cannot return scoped value
    return p2;
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+
+// maps to 
+
+void M<$param1>(ref<$local> S<$param1> s)
+    where $param1 : $cm
+```
+
+This definition prevents the value from escaping from the method as the lifetime is not returnable. It also prevents local data from escaping from the current method through the parameter as the lifetime is wider that `$local` but not equivalent.
+
+```
```
jjonescz, line 217:
````suggestion
```csharp
````

File: proposals/expand-ref.md [hunk 11]
```diff
@@ -0,0 +1,294 @@
...
+
+    S<$local> s = new S<$local>(stackalloc int[42]);
+
+    // error: cannot convert S<$local> to S<$param1>
+    p = s;
+}
+```
+
+<a name="annotations-field"></a>
+At an annotation level every field marked `scoped ref` (explicitly or implicitly via `ref scoped`)will have a new lifetime parameter defined. The name will be `$refFieldN` where _N_ is the numerical order of the field. That lifetime will have the relationship `where $refFieldN : $local` in all methods that use the type.
```
jjonescz, line 230:
```suggestion
At an annotation level every field marked `scoped ref` (explicitly or implicitly via `ref scoped`) will have a new lifetime parameter defined. The name will be `$refFieldN` where _N_ is the numerical order of the field. That lifetime will have the relationship `where $refFieldN : $local` in all methods that use the type.
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+        // okay
+        reader.Skip();
+
+        // Error: can't escape the ref data the ref scoped field refers to
+        return reader.ValueSpan;
+    }
+}
+```
+
+This is accomplished by giving every `ref scoped` field a new two new escape scopes named _current field N_ and _current ref field N_ where _N_ is the numeric order of the field. For example the first field has a _safe-to-escape_ of _current field 1_ and a _ref-safe-to-escape_ of _current ref field N_. Both escape scopes can be converted to _current method_, and _current field N_ can be converted to _current ref field N_, but no other defined relationships exist. That serves to restrict their usage to the current method where the containing value is used. This escape scope applies to both
```
RikkiGibson, line 132: I think that all lifetimes `$current_field_N` and `$current_ref_field_N` must also be convertible to lifetime `$this`. In the absence of this constraint, it would be possible for the `this` variable to hold references to fields with narrower lifetimes.

RikkiGibson: Or, depending on how you look at it, it simply wouldn't be possible to declare any 'ref scoped' fields in ref structs, because the field's lifetime would not be convertible to the lifetime of the containing 'this' variable.

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -912,10 +918,12 @@ The basic rules for the lifetime are defined as:
 
 - All lifetimes are expressed syntactically as generic arguments, coming before type arguments. This is true for predefined lifetimes except `$heap` and `$local`. 
 - All types `T` that are not a `ref struct` implicitly have lifetime of `T<$heap>`. This is implicit, there is no need to write `int<$heap>` in every sample.
-- For a ref field defined as `ref T<$l1, $l2, ... $ln>` all lifetimes `$l1` through `$ln` must be invariant. 
-- For a ref defined as `ref<$a> T<$b, ...>`, `$b` must convertible to `$a`
-- The `ref` of a variable has a lifetime of the 
-    - For a ref local, parameter, field or return of type `ref<$a> T` the lifetime is `$a`
+- For a `ref` field defined as `ref<$l0> T<$l1, $l2, ... $ln>`:
+    - All lifetimes `$l1` through `$ln` must be invariant.
```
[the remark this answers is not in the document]
jaredpar: It's possible that `readonly` could allow us to relax the restriction a bit here. At the same time it's something we'd have to really map out to make sure we weren't missing anything.

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+There are still a number of scenarios around `ref` which cannot be safely expressed in the language. These are generally when using multiple mutable `ref struct` parameters where many are passed by `ref` or when trying to use `ref struct` in `ref` fields.
+
+To _fully_ satisfy all of these scenarios would require us to introduce explicit lifetime parameters and relationships into the language. That is a _huge_ investment that is not yet motivated by need. Instead this proposal takes our existing lifetime annotation, `scoped`, and sees how much further `ref` safety can be taken without introducing any other annotations or keywords. 
+
+This doesn't solve all scenarios but does remove several known friction points in the language. It also serves to show us exactly where the limits are without introducing explicit lifetime parameters.
+
+## Detailed Design
+The rules for `ref struct` safety are defined in the following documents:
+
+- [ref safety proposal](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md).
```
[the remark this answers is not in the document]
jaredpar: At some point I do need to put together a unified doc that we can reference.

proposals/expand-ref.md
jaredpar: Never mind, they cannot I was wrong.

proposals/expand-ref.md
jaredpar: Correct. This is just about getting clarity on the semantics. Particularly I know the runtime team has wanted to leverage `TypedReference` more and this would provide clarity on what can and cannot be done safely.

proposals/expand-ref.md
jaredpar: I disagree. Logically yes I agree that the lifetimes must be convertible to `$this`. The moment you allow that though in all contexts, suddenly it's not `scoped` anymore, it's returnable 

```csharp
struct S<$this, $refField1, $field1>
  where $refField : $this
  where $field1: $this, $refField1 {
  ref<$refField1> Span<$field1, byte> span;
}

Span<$cm> M<$cm, $l1, $l2>(S<$cm, $l1, $l2> s) => 
  // works because $l2 is convertible to $cm
  s.span;
}
```

The way this works is that at construction time the rules ensure that the lifetimes of both the `ref` and the value are convertible to `$this`. That just falls out from our existing method args rules. Given that the fields lifetimes are invariant there is no need to keep that relationship alive for correctness after that. The only thing we have to worry about is how they are interpreted in the context of a method. 

Note: if these were `ref unscoped` fields then you are 100% correct we'd put that relationship in place at the decl of the type.

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+At an annotation level every parameter marked `ref scoped` will have a new lifetime parameter defined. The name will be `$paramN` where _N_ is the numerical order of the parameter. That lifetime will only have the relationship `where $paramN : $local`. 
+
+```csharp
+ref struct S { } 
+void M(ref scoped S s) 
+
+// maps to 
+
+void M<$param1>(ref<$local> S<$param1> s)
+    where $param1 : $cm
```
cston, line 212:
```suggestion
void M<$param1>(ref<$local> S<$param1> s)
    where $param1 : $local
```

proposals/expand-ref.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+This definition prevents the value from escaping from the method as the lifetime is not returnable. It also prevents local data from escaping from the current method through the parameter as the lifetime is wider that `$local` but not equivalent.
```
cston, line 215:
```suggestion
This definition prevents the value from escaping from the method as the lifetime is not returnable. It also prevents local data from escaping from the current method through the parameter as the lifetime is wider than `$local` but not equivalent.
```

PR #7570, opened 2023-10-02
Collection expressions: replace conversion rule for types that implement IEnumerable

cston:
Replace the collection expression conversion rule for types that implement `System.Collections.IEnumerable`, based on LDM-2023-10-02.


333fred (approved):

BillWagner (approved):
Thanks @cston 

This LGTM.

PR #7574, opened 2023-10-04
Fix typos

jnm2:


meetings/2023/LDM-2023-10-02.md
```diff
@@ -54,7 +54,7 @@ M(['a', 'b', 'c']);
 ```
 
 Scenario 1 is extremely realistic, and in fact already exists in the BCL. We'd much prefer that it call the `ReadOnlySpan` API, which does not need to allocate; the `string` API
-needs an instance of a `string`, which could very well allocate if the items were not not constant. To accomplish this, we need to update the betterness rule from a
```
333fred, line 57: No no, it's the rare double negative. I swear

333fred (approved):

PR #7577, opened 2023-10-05
Spec LDM decision that collection expressions targeting IEnumerable<T> implement nongeneric IList

jnm2:
See details at: https://github.com/dotnet/roslyn/issues/70253

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -515,11 +515,14 @@ A collection expression has a *known length* if the compile-time type of each *s
 ### Interface translation
 [interface-translation]: #interface-translation
 
-Given a target type `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`, a compliant implementation is required to produce a value that implements that interface. It is recommended that any type that is synthesized implement all these interfaces. This ensures maximal compatibility with existing libraries, including those that introspect the interfaces implemented by a value in order to light up performance optimizations.
+#### Non-mutable interface translation
+[non-mutable-interface-translation]: #non-mutable-interface-translation
+
+Given a target type which does not contain mutating members, namely `IEnumerable<T>`, `IReadOnlyCollection<T>`, and `IReadOnlyList<T>`, a compliant implementation is required to produce a value that implements that interface. If a type is synthesized, it is recommended the synthesized type implements all these interfaces, as well as `ICollection<T>` and `IList<T>`, regardless of which interface type was targeted. This ensures maximal compatibility with existing libraries, including those that introspect the interfaces implemented by a value in order to light up performance optimizations.
 
 In addition, the value must implement the nongeneric `ICollection` and `IList` interfaces. This enables collection expressions to support dynamic introspection in scenarios such as data binding.
 
-A compliant implementation is free to: 
+A compliant implementation is free to:
 
 1. Use an existing type that implements the required interfaces.
 1. Synthesize a type that implements the required interfaces.
@@ -528,22 +531,15 @@ In either case, the type used is allowed to implement a larger set of interfaces
 
 Synthesized types are free to employ any strategy they want to implement the required interfaces properly.  For example, a synthesized type might inline the elements directly within itself, avoiding the need for additional internal collection allocations.  A synthesized type could also not use any storage whatsoever, opting to compute the values directly.  For example, returning `index + 1` for `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`.
 
-#### Non-mutable interface translation
-[non-mutable-interface-translation]: #non-mutable-interface-translation
-
 1. The value must return `true` when queried for `ICollection<T>.IsReadOnly` (if implemented) and nongeneric `IList.IsReadOnly` and `IList.IsFixedSize`.  This ensures consumers can appropriately tell that the collection is non-mutable, despite implementing the mutable views.
 1. The value must throw on any call to a mutation method (like `IList<T>.Add`).  This ensures safety, preventing a non-mutable collection from being accidentally mutated.
 
 #### Mutable interface translation
-[non-mutable-interface-translation]: #non-mutable-interface-translation
+[mutable-interface-translation]: #mutable-interface-translation
 
-Given a target type or `ICollection<T>` or `IList<T>`:
+Given target type that contains mutating members, namely `ICollection<T>` or `IList<T>`:
 
-1. The value must return `false` when queried for `ICollection<T>.IsReadOnly` and nongeneric `IList.IsReadOnly` and `IList.IsFixedSize`. 
-
-The value generated is allowed to implement more interfaces than required.  Specifically, implementing `IList<T>` even when only targeting `ICollection<T>`.  However, in that case:
-
-1. The value must support all mutation methods (like `IList<T>.RemoveAt`).
+1. The value must be an instance of `List<T>`.
 
 ### Known length translation
 [known-length-translation]: #known-length-translation
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -515,31 +515,31 @@ A collection expression has a *known length* if the compile-time type of each *s
 ### Interface translation
 [interface-translation]: #interface-translation
 
-Given a target type `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`a compliant implementation is only required to produce a value that implements that interface.  A compliant implementation is free to: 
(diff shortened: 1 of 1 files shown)


proposals/csharp-12.0/collection-expressions.md
```diff
@@ -515,31 +515,31 @@ A collection expression has a *known length* if the compile-time type of each *s
...
-
-1. The value must return `true` when queried for `ICollection<T>.IsReadOnly`.   This ensures consumers can appropriately tell that the collection is non-mutable, despite implementing the mutable views.
+1. The value must return `true` when queried for `ICollection<T>.IsReadOnly` (if implemented) and nongeneric `IList.IsReadOnly` and `IList.IsFixedSize`.  This ensures consumers can appropriately tell that the collection is non-mutable, despite implementing the mutable views.
 1. The value must throw on any call to a mutation method (like `IList<T>.Add`).  This ensures safety, preventing a non-mutable collection from being accidentally mutated.
 
-It is recommended that any type that is synthesized implement all these interfaces. This ensures that maximal compatibility with existing libraries, including those that introspect the interfaces implemented by a value in order to light up performance optimizations.
```
jnm2, line 535: I moved this text directly under the `#interface-translation` section since it doesn't fall foul of the mutable interface translation side, nor would we ever want it to.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -515,31 +515,31 @@ A collection expression has a *known length* if the compile-time type of each *s
...
+Synthesized types are free to employ any strategy they want to implement the required interfaces properly.  For example, a synthesized type might inline the elements directly within itself, avoiding the need for additional internal collection allocations.  A synthesized type could also not use any storage whatsoever, opting to compute the values directly.  For example, returning `index + 1` for `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`.
 
 #### Non-mutable interface translation
 [non-mutable-interface-translation]: #non-mutable-interface-translation
 
-Given a target type or `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, the value generated is allowed to implement more interfaces than required.  For example, implementing the mutable interfaces as well (specifically, implementing `ICollection<T>` or `IList<T>`).  However, in that case:
```
jnm2, line 530: This was already said in the outer `#interface-translation` section, but the "in that case" no longer applies to the following bullet points, given that nongeneric `IList` is always implemented.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -515,31 +515,31 @@ A collection expression has a *known length* if the compile-time type of each *s
 ### Interface translation
 [interface-translation]: #interface-translation
 
-Given a target type `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`a compliant implementation is only required to produce a value that implements that interface.  A compliant implementation is free to: 
+Given a target type `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`, a compliant implementation is required to produce a value that implements that interface. It is recommended that any type that is synthesized implement all these interfaces. This ensures maximal compatibility with existing libraries, including those that introspect the interfaces implemented by a value in order to light up performance optimizations.
 
-1. Use an existing type that implements that interface.
-1. Synthesize a type that implements the interface.
+In addition, the value must implement the nongeneric `ICollection` and `IList` interfaces. This enables collection expressions to support dynamic introspection in scenarios such as data binding.
```
jnm2, line 520: This line is the addition. The rest of the changes were required to be consistent with this line.

cston (approved):

proposals/csharp-12.0/collection-expressions.md
cston, line 520: Should this be "must implement the nongeneric ... interfaces *corresponding to the generic target type*"?

From the LDM notes below, it wasn't clear to me whether "appropriate corresponding non-generic interfaces" meant corresponding to the specific generic target interface or all of `IEnumerable`, `ICollection`, and `IList`.

> it is in fact more important that the concrete types we provide for the generic interfaces also implement the appropriate corresponding non-generic interfaces.

RikkiGibson: Collection interface hierarchy

Given the collection interface hierarchy illustrated here I think it's fine if the value we create when targeting to interface types, is simply spec'd to implement IList, as the others will come along for the ride.

RikkiGibson: I think I might have talked around your original point.. I guess what I'm really trying to say is, I think it's fine for the value we produce to implement IList in all cases where we are converting collection-expr to generic collection interface type.

jnm2: Functionality-wise, there is very little correspondence between the generic and nongeneric hierarchies in how they define 'collection' versus 'list.'

I do think targeting IEnumerable<T> should implement IList for the data binding reason.

CyrusNajmabadi: Agreed. You either get List<>, which implements IList, or we synthesize the immutable collection, which should do the same.

cston: I think I agree, the synthesized type generated by the compiler should probably implement `IList` in addition to the generic interfaces. But it wasn't clear to me from the LDM notes whether the spec should guarantee that.

CyrusNajmabadi: that was my understanding of hte discussion and voting.  If we don't guarantee this, then what's the point?  Removing it would just break anyone who now takes a dependency on this.  And we want people to take a dependency on this so that they can use collection literals in these scenarios.

jnm2: Would it help to do a clarifying LDM email? I think we shouldn't conditionally light up IList depending on which of the interfaces is targeted. It's an interface dynamically discovered from `object`-typed instances, or sometimes `IEnumerable`-typed instances. The generic target interface type doesn't really have any bearing on the intended data binding scenario.

jnm2:
@cston @CyrusNajmabadi @RikkiGibson I believe this documents an LDM decision that was implemented when the feature originally shipped. I'm not sure where to take this next, can anyone help?

CyrusNajmabadi:
@cston Can you review this and merge in please.  Thanks!

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -515,31 +515,31 @@ A collection expression has a *known length* if the compile-time type of each *s
 ### Interface translation
 [interface-translation]: #interface-translation
 
-Given a target type `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`a compliant implementation is only required to produce a value that implements that interface.  A compliant implementation is free to: 
+Given a target type `IEnumerable<T>`, `IReadOnlyCollection<T>`, `IReadOnlyList<T>`, `ICollection<T>`, or `IList<T>`, a compliant implementation is required to produce a value that implements that interface. It is recommended that any type that is synthesized implement all these interfaces. This ensures maximal compatibility with existing libraries, including those that introspect the interfaces implemented by a value in order to light up performance optimizations.
```
CyrusNajmabadi, line 518: change to just be the readonly interfaces.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -515,31 +515,31 @@ A collection expression has a *known length* if the compile-time type of each *s
...
-
 #### Mutable interface translation
 [non-mutable-interface-translation]: #non-mutable-interface-translation
 
 Given a target type or `ICollection<T>` or `IList<T>`:
 
-1. The value must return `false` when queried for `ICollection<T>.IsReadOnly`. 
+1. The value must return `false` when queried for `ICollection<T>.IsReadOnly` and nongeneric `IList.IsReadOnly` and `IList.IsFixedSize`. 
 
 The value generated is allowed to implement more interfaces than required.  Specifically, implementing `IList<T>` even when only targeting `ICollection<T>`.  However, in that case:
```
CyrusNajmabadi, line 544: remove entirely and just say that the final value must be a `List<T>`.

proposals/csharp-12.0/collection-expressions.md
CyrusNajmabadi, line 518: or say: `It is recommended that if a type is synthesized that is implements all these interfaces`

jnm2:
Updated!

CyrusNajmabadi (approved):

proposals/csharp-12.0/collection-expressions.md
mungojam: Is supporting IList an existing requirement for regular types that users want to pass into data binding scenarios that are accepting IEnumerable<T>? Seems odd since you might be passing in a read-only collection that doesn't support adding. And you'll presumably end up with a less performance synthetic version

Maybe I've misunderstood and it's just saying to do that for synthetic versions of IList<T> and not IEnumerable<T>

jnm2: `IList` (and `IList<T>`) have `IsReadOnly` which is set to true in these scenarios. Part of consuming `IList`/`IList<T>` is knowing that that they will throw NotSupportedException from mutation methods if `IsReadOnly` is true.

In the case of Windows Forms, data binding doesn't recognize IEnumerable as a collection unless it finds `System.Collections.ICollection` and for other things, `System.Collections.IList`.

Issue #7189, opened 2023-05-11
[Proposal]: Breaking change warnings


MadsTorgersen:
# Breaking change warnings

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [x] Specification: https://github.com/dotnet/csharplang/blob/main/proposals/breaking-change-warnings.md

Please see the proposal [here](https://github.com/dotnet/csharplang/blob/main/proposals/breaking-change-warnings.md).

eyalalonn:
Personally I don't like the two far extreme alternatives and I think that users shouldn't be able to disable the warnings and the only way forward is to fix the code be it manually, through fixers or this hypothetical upgrade tool that does this automatically.

DavidArno:
As it stands, this feature is extremely worrying as it will warn that silent breaking changes have been introduced to the language, but only if you do the right combination of update the SDK, then compile, then change the target framework. If you don't perform that compile step at the right point, there will be no warning of the change.

The limited alternatives offered ignore two obvious - and in my opinion, far superior - options:

1. Always warn. For the code example:
```cs
public class Entity
{
    string field;
    public string Field
    {
        get { return field; }         
        set { field = value.Trim(); } 
    }
    ...
}
```
if I target C# 11 or lower, I get a warning that this code will introduce a breaking change in C# 12 as described in the specification. If I target C# 12, then I now get a warning on the `string field;` line that this field is not used by the `Field` property. The breaking change is now no longer silent. 

2. Avoid new contextual keywords. This approach is a radical one for a formally conservative language like C#, but it creates the cleanest break. For C# 11 or lower, I get a warning that `field` is being promoted to a keyword in C# 12. In C# 12, the `string field;` line becomes a compilation error. I have clear choices this way: rename to `@field`, rename completely to eg `name` or  delete it and use the new `field` feature.

jnm2:
@DavidArno "2 - Avoid new contextual keywords" doesn't catch a silent breaking change where the field was already declared as `string @field;` and referenced inside the accessor as either `field` or `@field`, though this break is much less likely since there would have likely been no pressure from tooling to declare using `@` in the first place.

aradalvand:
Despite the possible (and likely) resistance that many people might show against this, I believe it's an direlt needed policy shift, that's becoming increasingly relevant as more and more features like that are added to the language.

It's a great way mitigating the bloat problem, by significantly simplifying the language and the amount of gotchas and "special rules".

[...]

Issue #1349, opened 2018-03-01
Champion: "Partial Type Inference"


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

See #1348 

Currently, in an invocation of a generic method, you either have to specify all of the type arguments or none of them, in the latter case depending on the compiler to infer them for you. The proposal is to permit the programmer to provide *some* of the type arguments, and have the compiler infer only those that were not provided.

There are (at least) three possible forms this could take:
1. Named type arguments (#280, #1348), e.g. `M<TElement: int>(args)`
2. Omitted type arguments separated by commas, e.g. `M<int, >(args)`
3. Types to be inferred specified by `var` (#1348), e.g. `M<int, var>(args)`

bondsbw:
Pros/cons for 2 vs 3?

Both are DRY but I prefer 2... you don't buy much visually with 3.

jnm2:
1 is okay but verbose. I prefer 2 because I would like this potential progression in levels of needed disambiguation for `Foo<TKey, TValue>`:
`Foo<TKey,>`, `Foo<,>`, `Foo`.

Logerfo:
I'm pretty sure there was an issue regarding recursive generic, but I can't find it. Without recursive type parameters, I think 2 is better because is less verbose, since verbosity is part of the issue here. But with recursive type parameters, I think 1 would fit better.

TheUnlocked:
I think that either 2 or 3 is the best one. 2 is certainly the least verbose (which is nice), but 3 both looks more complete and full. I don't think the slight verbosity of 3 would really be an issue though because it still requires no thinking of what the type name is, and is only three keystrokes.

bondsbw:
I support 1 even with the inclusion of 2 or 3.  It helps to document the parameters.

vladd:
I believe, F# is using `_`:

```
> open System.Collections.Generic;;
> let s = [ 1; 2; 3 ] :> IEnumerable<_>;;
val s : IEnumerable<int> = [1; 2; 3]
> let l = List<_>([1; 2; 3]);;
val l : List<int>
```

qrli:
Isn't 2 already used in other constructs?
```cs
typeof(IDictionary<,>)
```

zippec:
@qrli 
[...]

PR #7582, opened 2023-10-07
Add proposal for partial type inference

TomatorCZ:


jcouv:
Sorry for the delay. We've focused on wrapping up C# 12 bugs. We'll look in a few weeks. Thanks

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
+# Partial type inference
+
+* [x] Proposed
```
jcouv, line 3: Those bullets are best kept in the championed proposal ([https://github.com/dotnet/csharplang/issues/1349](https://github.com/dotnet/csharplang/issues/1349)) so they can be updated fluidly

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+* [ ] Prototype
+* [ ] Implementation
+* [ ] Specification
+
+> Note: This proposal was created because of championed [Partial type inference](https://github.com/dotnet/csharplang/issues/1349). It is a continuation of the proposed second version published in [csharplang/discussions/7467](https://github.com/dotnet/csharplang/discussions/7467) and the first version published in [csharplang/discussions/7286](https://github.com/dotnet/csharplang/discussions/7286)
+
+## Summary
+[summary]: #summary
+
+Partial type inference introduces a syntax skipping obvious type arguments in the argument list of
```
jcouv, line 13: >obvious [](http://example.com/codeflow?start=52&length=7)

nit: instead of "obvious" consider "inferrable" (also applies below)

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+  > log<_, object>(new Message( ... ), null); // We can use _ to mark type arguments which should be inferred by the compiler.
+  >
+  > public void log<T, U>(T message, U appendix) { ... }
+  > ```
+
+* The next motivation is constructor type inference. 
+  Method type inference is not defined on *object_creation_expression*, prohibiting taking advantage of type inference.
+  We divide use cases into the following categories, where type inference would help the programmer. 
+
+  1. Cases where the method type inference would success.
```
jcouv, line 93:
```suggestion
  1. Cases where the method type inference would succeed.
```

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+   
+  > Example
+  > 
+  > ```csharp
+  > var wrappedData = Create(new MyData());
```
jcouv, line 98: nit: Do we need this line to illustrate the point? 
I think the line below (`return new Wrapper<T>(item);`) is central and sufficient

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+  > 
+  > public static Wrapper<T> Create<T>(T item) { return new Wrapper<T>(item); }
+  > 
+  > class Wrapper<T> { public Wrapper(T item) { ... } }
+  > ```
+  
+  2. Cases where the method type inference would be weak. (Using type info from target type, or type arguments' constrains)
```
jcouv, line 105:
```suggestion
  2. Cases where the method type inference would be weak. (Using type info from target type, or type constraints)
```

typo: constrains -> constraints, also applies in other locations below

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+So will want to look ahead to other potential directions, which can be done after this feature.
+
+## Detailed design
+[design]: #detailed-design
+
+### Grammar
+
+> Specification: Original section changed in the following way
+
+The following changes are made in [tokens](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#64-tokens) located in the [grammar](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#62-grammars) section.
```
jcouv, line 177: nit: Can be simplified
```suggestion
We modify [Identifiers](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#643-identifiers) as follows:
```
(then the "Identifiers" link below becomes redundant)

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+
+> [Identifiers](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#643-identifiers)
+
+- The semantics of an identifier named `_` depends on the context in which it appears:
+  - It can denote a named program element, such as a variable, class, or method, or
+  - It can denote a discard ([§9.2.9.1](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/variables.md#9291-discards)).
+  - **It can denote an inferred type argument avoiding specifying type arguments which can be inferred by the compiler.**
```
jcouv, line 184: Can be simplified
```suggestion
  - **It can denote a type argument to be inferred.**
```

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,743 @@
...
+  -   | 'var'    | 'when'       | 'where'     | 'yield'
+    ;
+  ```
+
+### Type arguments
+
+> Specification: Original section changed in the following way
+
+* We change the meaning of the content of *type_argument_list* in two contexts.
+  * [Constructed types](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/types.md#84-constructed-types) occuring in [*object_creation_expression*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#128162-object-creation-expressions)
```
jcouv, line 207: As a general comment, I appreciate the references back to the standard/spec. The format you adopted in other sections (copy a paragraph from original spec and add some highlighted edits) is really nice. But when I look at "Type arguments" section, it's not clear which section of the spec is being edited. 

Maybe you mean to edit section [8.4.2](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/types.md#842-type-arguments) ("type arguments")? That seems to be the right place to introduce ***inferred_type_argument*** into the ***type_argument_list*** grammar.
(If so, then the references to "constructed types" (8.4) and  "object_creation_expression" could be removed)

Below is a possible grammar change to introduce `inferred_type_argument`:
  ```antlr
type_argument_list
    : '<' type_arguments '>'
    ;
    
type_arguments // modified
    : type_argument (',' type_arguments)
    | inferred_type_argument (',' type_arguments)
    ;   

type_argument
    : type
    ;

inferred_type_argument // new
    : '_'
    ;
  ```
Then we can say that ***inferred_type_argument*** is only allowed in the contexts that you listed.

We still need to resolve the issue of back compat:
```
public class _ { }
public class C<T> 
{
    public void M(C<_> x) { } // allowed today
}
```
We'll need to decide whether the presence of an existing type named "_"  in scope affects the interpretation of "_" in type argument lists, or whether "_" always means "inferred type argument" (ie. we take a compat break, maybe warn on types named "_").

proposals/partial-type-inference.md
jcouv: I made some edits directly to save time. I hope that's okay with you.
I removed these bullets and simplified the links in the copied sections of the spec (makes it easier to review the diff).

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,738 @@
...
+* If a type is a *generic_inferred*, then we resolve the identifier in the same manner except ignoring the arity of type parameters (Types of arity 0 is ignored). 
+If there is an ambiguity in the current scope, a compilation-time error occurs.
+  > Example
+  >
+  > ```csharp
+  > class P1
+  > {
+  >     void M() 
+  >     {
+  >         new C1<>( ... ); // Refers generic_inferred type C1<T>
```
jcouv, line 251: `T`, `T1` and `T2` don't exist here.
Reading ahead to the "Object creation expressions" section, I believe the comments in this snippet are not what you intended and some actual types should be inferred from inspecting candidate constructors.

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,738 @@
...
+  - **Once the *inferred_type_arguments* are inferred and together with remaining type arguments are substituted for the corresponding method type parameters, all constructed types in the parameter list of `F` satisfy their constraints ([§8.4.5](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/types.md#845-satisfying-constraints)), and the parameter list of `F` is applicable with respect to `A` ([§12.6.4.2](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#12642-applicable-function-member))**
+- If `F` is generic and `M` includes a type argument list, `F` is a candidate when:
+  - `F` has the same number of method type parameters as were supplied in the type argument list, and
+  - Once the type arguments are substituted for the corresponding method type parameters, all constructed types in the parameter list of `F` satisfy their constraints (§8.4.5), and the parameter list of `F` is applicable with respect to `A` (§12.6.4.2).
+
+### Object creation expressions
```
jcouv, line 289: I stopped here for today. Will resume tomorrow

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,738 @@
...
+  The second improvement regards the "all or nothing" principle, where the method type inference infers either all of the type arguments or nothing.
+
+  > Example
+  > ```csharp
+  > log<, object>(new Message( ... ), null); // Error, impossible to specify just one type argument. We have to specify all of them. 
+  >
+  > public void log<T, U>(T message, U appendix) { ... }
+  > ```
+
+  The first improvement, which would improve the method type inference algorithm, has a significant disadvantage of the breaking change.
```
jaredpar, line 67:
```suggestion
  The first improvement, which would improve the method type inference algorithm, has a significant disadvantage of introducing a breaking change.
```

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,738 @@
...
+  On the other hand, the second improvement, which would enable specifying some of the method's type arguments, does not influence old code, solves problems regarding the "all or nothing" principle, and reduces the first weakness.
```
jaredpar, line 68: I'm unclear what the "all or nothing" principle means here.

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,738 @@
...
+> Specification: Original section changed in the following way
+
+The following changes are made in [tokens](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#64-tokens) located in the [grammar](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#62-grammars) section.
+
+> [Identifiers](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#643-identifiers)
+
+- The semantics of an identifier named `_` depends on the context in which it appears:
+  - It can denote a named program element, such as a variable, class, or method, or
+  - It can denote a discard (§9.2.9.1).
+  - \***It can denote an inferred type argument avoiding specifying type arguments which can be inferred by the compiler.**
```
jaredpar, line 179: Rather than _it can_ I'd prefer the language _it will_. Essentially when flipping to `/langversion:X` where `X` is the release that introduces this feature, I'd prefer the meaning to be unambiguous. That is for the case `M<_, object>(42, null)`, the `_` unambiguously means an inferred type parameter. 

Yes this would be breaking change to developers who defined `class _ { }` but that seems like a minority case. It in the same ballpark as `class record` which we broke when introducing records.

TomatorCZ:
Thank you for all comments. I will revise the proposal during the weekend.

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,738 @@
+# Partial type inference
+
+> Note: This proposal was created because of championed [Partial type inference](https://github.com/dotnet/csharplang/issues/1349). It is a continuation of the proposed second version published in [csharplang/discussions/7467](https://github.com/dotnet/csharplang/discussions/7467) and the first version published in [csharplang/discussions/7286](https://github.com/dotnet/csharplang/discussions/7286)
+
+## Summary
```
jcouv, line 5: Chatted with @jaredpar about the whole proposal. There's three parts that I can see:
1. allowing free/floating type arguments in method invocation (`M<int, _>(...)`)
2. allowing free/floating type arguments in object creation **with floating arity** (`new C<>(...)`)
3. allowing object initializers contribute to the inference in object creation (`new C<>(...) { ...initializer... }`)

The second and third part seem much more complicated as they introduce bidirectional type inference.

What we would propose instead as core design proposal is:
1. allowing free/floating type arguments in method invocation [KEEP]
2. allowing free/floating type arguments in object creation **with fixed arity** using the same syntax: `new C<_, _>()`
The benefit of restricting object creation to fixed/known arity is that the flow for object creation remains more like today: pick a single type and apply overload resolution to its constructors (and overload resolution can now deal with fixed vs. floating type arguments).

Those two core parts would be very useful, straightforward and not very controversial with LDM.

We would recommend either removing or clearly separating out (potentially in a separate doc) the two parts that are controversial and complex:
1. allowing floating arity in object creation
2. allowing initializers to contribute to type inference in object creation

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by commit [33b71f0](https://github.com/dotnet/csharplang/pull/7582/commits/33b71f0c489eaa03018d18a9cabf8b2cdf74a43a)

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by commit [33b71f0](https://github.com/dotnet/csharplang/pull/7582/commits/33b71f0c489eaa03018d18a9cabf8b2cdf74a43a)

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by commit [33b71f0](https://github.com/dotnet/csharplang/pull/7582/commits/33b71f0c489eaa03018d18a9cabf8b2cdf74a43a)

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by commit [33b71f0](https://github.com/dotnet/csharplang/pull/7582/commits/33b71f0c489eaa03018d18a9cabf8b2cdf74a43a)

proposals/partial-type-inference.md
TomatorCZ: Line [58](https://github.com/dotnet/csharplang/pull/7582/files#diff-98ed61744ec680992e9779b8a70429db7e853f6942a0cb4e94b43b8e7ae96b91R58) explains "all or nothing" principle. The method type inference infers either all of the type arguments or nothing (e.g. we can't hint the compiler some of the type arguments).

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by commit [5e59e96](https://github.com/dotnet/csharplang/pull/7582/commits/5e59e96d98c7c7c12678c293b04d5e20d8b8f037)

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by commit [b6081e3](https://github.com/dotnet/csharplang/pull/7582/commits/b6081e37e9a7e3b6892d3b0164e5d91b3a61cfc3)

proposals/partial-type-inference.md
TomatorCZ: Should be fixed by [6039b9f](https://github.com/dotnet/csharplang/pull/7582/commits/6039b9fcc71f26a254f65ce4261200d04ba90e17).

proposals/partial-type-inference.md
TomatorCZ: Should by fixed by commit [ea87f2f](https://github.com/dotnet/csharplang/pull/7582/commits/ea87f2f973aff156f9467c56d4e1861e3c7306bb). I moved the mentioned parts (initializers and floating arity) to alternatives.

TomatorCZ:
Done with the first round of revisions.

TomatorCZ:
Hi,

I implemented a prototype which passes `Test.cmd -testCompilerOnly` tests(except the ones, which fail in `main` on which the change is rebased on). See [fork](https://github.com/TomatorCZ/roslyn/tree/dev/PartialTypeInference). The implementation is rather a proof of concept than ready to production code.

Probably the most interesting [file](https://github.com/TomatorCZ/roslyn/blob/dee1c346d00b43874ddc991b6f7a04a6129a7a1c/src/Compilers/CSharp/Test/Semantic/Semantics/PartialTypeInferenceTests.cs) contains tests for the partial type inference.

The implementation contains the mentioned design in the proposal without `2.` and `3.` parts mentioned [here](https://github.com/dotnet/csharplang/pull/7582#discussion_r1382110305)

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,794 @@
...
+
+* The current method type inference has two possible areas for improvement.
+  The first improvement regards the strength of the method type inference, which uses only arguments' types to deduce the method's type arguments.
+  Concretely, we can see the weakness in cases where type arguments only depend on target type or type parameter restrictions.
+
+  > Example
+  >
+  > ```csharp
+  > object data = database.fetch();
+  > int count = data.Field("count"); // Error, method type inference can't infer the return type. We have to specify the type argument.
```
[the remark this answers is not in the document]
TomatorCZ: Should be fixed by commit [e973d80](https://github.com/dotnet/csharplang/pull/7582/commits/e973d80d053e19997a8dde270eb8579e8d167a82)

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,794 @@
...
+type_argument
+  : type
+  ;
+
++inferred_type_argument
++  : '_'
++  ;
+```
+
+* When a type with `_` identifier is presented in the scope where **inferred_type_argument** is used, a warning should appear since the **inferred_type_argument** hides the type's name causing a breaking change.
```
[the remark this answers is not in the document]
TomatorCZ: Should be fixed by commit [4485ab9](https://github.com/dotnet/csharplang/pull/7582/commits/4485ab9ea0bdb6a4c0b04c3575728caf54e1a19d)

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,794 @@
...
+    - **The set of candidate instance constructors is empty, or if a single best instance constructor cannot be identified, and there is no chance to know the target type now or later.**
+  - **If the set of candidate instance constructors is still empty, or if a single best instance constructor cannot be identified, repeat the binding of the *object_creation_expression* to the time, when target type will be known and add it to inputs of type inference.**
+  - The result of the *object_creation_expression* is a value of type `T`, namely the value produced by invoking the instance constructor determined in the two steps above.
+  - Otherwise, the *object_creation_expression* is invalid, and a binding-time error occurs.
+
+### Type inference
+
+> Specification: Original section changed in the following way
+
+We change the [type inference](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1263-type-inference) as follows.
```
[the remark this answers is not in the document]
TomatorCZ: Should by fixed by commit [9dfa24f](https://github.com/dotnet/csharplang/pull/7582/commits/9dfa24f073118a81555a11db2edda8de4cc6846c)

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,794 @@
...
+  * If the expression contains *where* clauses defining type constraints of type parameters of the type containing constructor, for each constraint not representing *constructor* constraint, *reference type constraint*, *value type constraint* and *unmanaged type constraint* perform *lower-bound inference* from the constraint to the corresponding type parameter.
+
+* Arguments binding
+  * It can happen that an argument of an expression will be *object_creation_expression*, which needs a target type to be successful binded. 
+  * In these situations, we behave like the type of the argument is unknown and bind it when we will know the target type.
+  * We treat it in the same manner as an unconverted *new()* operator.
+
+#### Type inference algorithm change
+
+> Specification: Original section changed in the following way.
```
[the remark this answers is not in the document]
TomatorCZ: Should be fixed by commit [7aa277e](https://github.com/dotnet/csharplang/pull/7582/commits/7aa277e63f69c12f3bd27cae6c98a7a965fa6958)

TomatorCZ:
Done with the second round of revisions.

proposals/partial-type-inference.md
```diff
@@ -0,0 +1,824 @@
...
+- **If the method group contains inferred type arguments, a warning should appear since partial type inference is not supported by runtime.** 
+
+We add the following
+- If an object creation expression is inferred and the argument list contains a dynamic value, an error should appear since constructor type inference is not supported by runtime.  
+
+### Nullability
+
+We add the following section about nullability. 
+
+* We can use an examination mark `?` to say that the inferred type argument should be a nullable type (e.g. `F<_?>(...)`).
```
RikkiGibson, line 475:
```suggestion
* We can use a question mark `?` to say that the inferred type argument should be a nullable type (e.g. `F<_?>(...)`).
```
Also, if this is intended to update a particular section of the language standard, let's link to that section here.

proposals/partial-type-inference.md
OJacot-Descombes: In line [68](https://github.com/dotnet/csharplang/pull/7582/files/7aa277e63f69c12f3bd27cae6c98a7a965fa6958#diff-98ed61744ec680992e9779b8a70429db7e853f6942a0cb4e94b43b8e7ae96b91R68), wouldn't it even be possible to infer twice: `test<TestCaseDefault<_>, _>(new MyData());` ?

proposals/partial-type-inference.md
TomatorCZ: Solved by commit [98d835e](https://github.com/dotnet/csharplang/pull/7582/commits/98d835e3dcbd67efbb7ae1c87cf54ac19578d68f)

proposals/partial-type-inference-v2.md
```diff
@@ -160,7 +160,7 @@ The initial set of candidate methods for is changed by adding new condition.
 - If `F` is generic and `M` has no type argument list, `F` is a candidate when:
   - Type inference (§12.6.3) succeeds, inferring a list of type arguments for the call, and
   - Once the inferred type arguments are substituted for the corresponding method type parameters, all constructed types in the parameter list of `F` satisfy their constraints ([§8.4.5](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/types.md#845-satisfying-constraints)), and the parameter list of `F` is applicable with respect to `A` (§12.6.4.2)
-- **If `F` is generic and `M` has a type argument list containing at least one *inferred_type* or *partial_inferred_type*, `F` is a candidate when:**
+- **If `F` is generic and `M` is *partial_inferred*, `F` is a candidate when:**
   - **Method type inference (See the [Type Inference] section) succeeds, inferring the type arguments list for the call, and**
```
RikkiGibson, line 164: Was Type Inference meant to be a link here?

TomatorCZ: https://github.com/dotnet/csharplang/pull/7582/files#diff-79921a867e573503f54d5a13d465b08a27cf745e8ed47b914438e95fd17a4edcR139

Its just a name for these types of expressions (`M<_, List<_>>(...)`)

proposals/partial-type-inference-v2.md
```diff
@@ -0,0 +1,398 @@
...
+
+The binding-time processing of an *object_creation_expression* of the form new `T(A)`, where `T` is a *class_type*, or a *value_type*, and `A` is an optional *argument_list*, consists of the following steps:
+
+- If `T` is a *value_type* and `A` is not present:
+  - **The *object_creation_expression* is a default constructor invocation.**
+    - **If the type is *partial_inferred*, Constructor type inference (See the [Type Inference] section) of the default constructor occurs to determine closed type. If it succeeded, construct the type using the inferred type arguments. If it failed the binding-time error occurs.**
+    - **If the type inference above succeeded or the type is not inferred, the result of the *object_creation_expression* is a value of (constructed) type `T`, namely the default value for `T` as defined in [§8.3.3](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/types.md#833-default-constructors).**
```
RikkiGibson: I'm conflicted about having these lines here because:
- they are out of date (`new ValueType()` can be invocation of a parameterless (explicitly declared) struct constructor).
- the described inference will never succeed. Perhaps in future if target type contributes to inference, but not with this version of the feature.

I think for the purposes of this feature speclet we could delete these lines and just remove "Otherwise," from the "Otherwise, if `T` is a *type parameter*"... below.

proposals/partial-type-inference-v2.md
```diff
@@ -0,0 +1,389 @@
...
+_[] //partial_inferred_type, but not inferred_type
+```
+
+* We can use a question mark `?` to say that the inferred type should be a nullable type (e.g. `F<_?>(...)`).
+
+* A method group is said to be *partial_inferred* if it contains at least one *inferred_type_placeholder* or *partial_inferred_type* in its type argument list. 
+
+* A type is said to be *partial_inferred* if it is *inferred_type_placeholder* or *partial_inferred_type*. 
+
+* When a type with the `_` identifier is presented in the scope where *inferred_type_placeholder* is used, a warning should appear since the *inferred_type_placeholder* hides the type's name causing a breaking change.
```
RikkiGibson, line 143: If we can, we should start issuing a warning on declaring or using a type named `_` in the previous language version to let people know this change is coming. (where changing to `@_` suppresses the warning and doesn't get flagged in analyzers for detecting unnecessary `@`.)

Hopefully we can separate this question of whether/how to adjust behavior of `_` for discards, while leaving space for them to eventually both work in a coherent manner. (IIRC, fixing the quirks around `_` is one of the more often discussed breaking changes, see #7189)

proposals/partial-type-inference-v2.md
```diff
@@ -0,0 +1,389 @@
...
+    An intention is to keep the exact form of the given hint(`IList<_>`).
+
+  * Example: Given `IList<_>` as a type argument, when we treat nullability, we want the hinted type parameter to be non-nullable(not `IList<_>?`).
+    It can happen, other bounds would infer the nullable version, and although `IList<_>` can be converted to `IList<_>?`, it is not the user's intention.
+
+### Compile-time checking of dynamic member invocation
+
+We change the [compile-time checking](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1265-compile-time-checking-of-dynamic-member-invocation).
+
+- First, if `F` is a generic method and type arguments were provided, then if the method group is *partial_inferred*, an binding error occurs since it is not supported in the runtime binding.
```
RikkiGibson, line 374: We will likely decide to perform a partial inference at compile time when a single candidate is found for the dynamic invocation. It's not urgent to revise this proposal to get this precisely right. Certain details of this may be subject to change. See https://github.com/dotnet/csharplang/blob/main/proposals/params-collections.md#dynamic-vs-static-binding.

Issue #6926, opened 2023-01-28
`is` expression evaluating `const` expression should be considered constant


333fred:
### Discussed in https://github.com/dotnet/csharplang/discussions/6925

<div type='discussions-op-text'>

<sup>Originally posted by **Rekkonnect** January 28, 2023</sup>
# Summary

Despite being provided a constant expression, using an `is` pattern is not considered constant. This can be safely adjusted. As an added bonus, a warning is generated by code flow analysis saying that the expression always evaluates to `true` or `false`.

# Motivation

Conditional compilation symbols, reliance on runtime constants deriving from conditional statements

# Description

You cannot assign an `is` expression on constant LHS value to a constant field. For example,
```csharp
public const int A = 4;
public const bool B = A is 4; // illegal, despite the result being available for constant evaluation
public const bool C = A is not 3 and < 1 or 5; // illegal too
```

In the example above, `A is 4` is `A == 4`, but is not treated as such. `A == 4` passes, `A is 4` does not. My personal preference is using `is` wherever available, especially on enum constants, which also allows for extensibility of the pattern.

Since patterns on the RHS of the pattern matching expression are always constant (as of now), the only limitation is the LHS also being a constant expression.

# Example

[...]

PR #7589, opened 2023-10-09
Create spec for const expressions for is patterns

Rekkonnect:
Spec for #6926

### DIFF
--- a/proposals/pattern-const-expressions.md
+++ b/proposals/pattern-const-expressions.md
@@ -1,0 +1,97 @@
+# Constant `is` pattern expressions
+
+* [x] Proposed
+* [ ] Prototype: None
+* [ ] Implementation: None
+* [ ] Specification: Started, below
+
+## Summary
+[summary]: #summary
+
+Consider `is` pattern expressions as constant if the LHS is constant.
+
+## Motivation
+[motivation]: #motivation
+
+When the pattern expressions are converted into simpler boolean expressions using equality and comparison operators, they are considered constant expressions if all parts of the expression are considered constant and can be evaluated during compilation. Also, the expressions returned from the ternary operator can also be considered constant if all sides are constant (condition, consequence and alternative).
+
+## Detailed design
+[design]: #detailed-design
+
+### Spec
+The [section](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1223-constant-expressions) in the spec needs to be updated accordingly, in the following segments:
+
+> Only the following constructs are permitted in constant expressions:
+> - Literals (including the null literal).
+> - References to const members of class and struct types.
+> - References to members of enumeration types.
+> - References to local constants.
+> - Parenthesized subexpressions, which are themselves constant expressions.
+> - Cast expressions.
+> - checked and unchecked expressions.
+> - nameof expressions.
+> - The predefined +, –, !, and ~ unary operators.
+> - The predefined +, –, *, /, %, <<, >>, &, |, ^, &&, ||, ==, !=, <, >, <=, and >= binary operators.
+> - The ?: conditional operator.
+> - **Pattern expressions.**
+> - sizeof expressions, provided the unmanaged-type is one of the types specified in §23.6.9 for which sizeof returns a constant value.
+> - Default value expressions, provided the type is one of the types listed above.
+
+### Grammar
+The grammar remains untouched, as nothing changes syntactically for this change.
+
+### Semantics
+A pattern expression with a constant LHS may be evaluated at compile time currently, but it cannot be assigned to a constant symbol. With this change, constant fields and locals of type `bool` may be assigned pattern expressions, as already evaluated during compilation.
+
+### Examples
+The below example shows a pattern expression being assigned to a constant field.
+```csharp
+public const int A = 4;
+public const bool B = A is 4;
+public const bool C = A is not 3 and < 1 or 5; 
+```
+
+Since `A` is constant, and pattern expressions require that the operands on the right are constant, all operands are constant, and the expression is thus evaluated during compilation. This result will then be assigned to the constant field.
+
+Another example, using type and `null` value checks:
+```csharp
+const int a = 4;
+const bool b = false;
+const long c = 4;
+const string d = "hello";
+
+const bool x = a is int; // always true
+const bool y = a is long; // always false
+const bool z = d is not null; // always true
+const bool p = b is null; // always false
+```
+
+All the above are currently valid pattern matching expressions, that also emit warnings about their constant evaluation results, about them being always true or false.
+
+When assigning those expressions to a constant symbol, it would be preferrable to not report these warnings about the constant result of the expression.
+
+Expressions accessing properties and comparing their values are illegal, as property evaluation occurs at runtime:
+
+```csharp
+const bool q = d is string { Length: 5 }; // Error: not a constant expression
+```
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+None.
+
+## Alternatives
+[alternatives]: #alternatives
+
+Currently, equality and comparison operators can be used to compare against other literals, including nullability of the objects (e.g. `x != null`, or `y == 4 || y < 3`). However, type checking cannot be currently peformed at compile time and assigned to a constant.
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- [ ] Requires LDM review
+
+## Design meetings
+[meetings]: #design-meetings
+
+None.


proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,97 @@
...
+> - References to members of enumeration types.
+> - References to local constants.
+> - Parenthesized subexpressions, which are themselves constant expressions.
+> - Cast expressions.
+> - checked and unchecked expressions.
+> - nameof expressions.
+> - The predefined +, –, !, and ~ unary operators.
+> - The predefined +, –, *, /, %, <<, >>, &, |, ^, &&, ||, ==, !=, <, >, <=, and >= binary operators.
+> - The ?: conditional operator.
+> - **Pattern expressions.**
```
333fred, line 36: This is nowhere close to enough info. What types of patterns are allowed? What conversions, if any, are allowed? This needs to be significantly expanded.

Rekkonnect: As I see it, we should include:
- numeric literal patterns
- null literal patterns
- default literal patterns
- relative numeric literal patterns
- char literal patterns
- string literal patterns
- type check patterns without member (field/property) evaluators

Excluded are:
- patterns with member evaluators (including type check or not)
- `var` patterns
- list patterns

On all of the above included patterns, the LHS must always be a constant.

As for the constant patterns, we would not mind the ability to corrupt the memory via unsafe methods, invalidating the constant expression's result, resulting in an inconsistency (i.e. changing the value of a const field that causes an expression to return true instead of false).

333fred: I generally agree with the list, but not the type pattern check. We don't have any analogous constant check to that today, and it doesn't seem useful to me.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Summary
+[summary]: #summary
+
+Consider `is` pattern expressions as constant if the LHS is constant.
+
+## Motivation
+[motivation]: #motivation
+
+When the pattern expressions are converted into simpler boolean expressions using equality and comparison operators, they are considered constant expressions if all parts of the expression are considered constant and can be evaluated during compilation. Also, the expressions returned from the ternary operator can also be considered constant if all sides are constant (condition, consequence and alternative).
```
CyrusNajmabadi, line 16: Not sure what the ternary-operator clause means here.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,97 @@
...
+const long c = 4;
+const string d = "hello";
+
+const bool x = a is int; // always true
+const bool y = a is long; // always false
+const bool z = d is not null; // always true
+const bool p = b is null; // always false
+```
+
+All the above are currently valid pattern matching expressions, that also emit warnings about their constant evaluation results, about them being always true or false.
```
CyrusNajmabadi, line 69: note: `is` checks worry me.  because language and runtime views may differ.  For example, language thinks that `int[]` and `uint[]` are different types.  But runtime succeeds on `is` checks.

Rekkonnect: You mean type checks in general. Aligning with the comment above by @333fred I'm leaning towards not including them in the spec.

proposals/pattern-const-expressions.md
Rekkonnect: The `?:` conditional operator, I'm used to calling it the ternary operator and forget its actual name sometimes. Will adjust the doc accordingly.

CyrusNajmabadi: that's not what i mean.  What i'm saying is: i don't see how that sentence helps here.  WHy are we bringing up conditional expressions?  Are you intending to change how they work?

Rekkonnect: I see now, the only reason why this is being mentioned is that now certain `is` expressions will be able to be used in the condition of a conditional operator, and the result would then be able to be considered constant. In other words, `a is 3 ? 3 : 4` will now be considered constant, since the underlying `a is 3` expression is also constant. More of a side effect without explicitly touching that area.

CyrusNajmabadi: i would leave it out from here.  You can show it as an example later on :)

proposals/pattern-const-expressions.md
333fred: The `int[]` vs `uint[]` thing definitely occurred to me as well. I think we should just avoid the complication and say you can't type test in a constant `is` expr.

proposals/pattern-const-expressions.md
```diff
@@ -84,12 +107,13 @@ None.
...
 [alternatives]: #alternatives
 
-Currently, equality and comparison operators can be used to compare against other literals, including nullability of the objects (e.g. `x != null`, or `y == 4 || y < 3`). However, type checking cannot be currently peformed at compile time and assigned to a constant.
+Currently, equality and comparison operators can be used to compare against other literals, including nullability of the objects (e.g. `x != null`, or `y == 4 || y < 3`).
 
 ## Unresolved questions
 [unresolved]: #unresolved-questions
 
 - [ ] Requires LDM review
+- [ ] Should we introduce a new error for non-constant subpatterns in order to isolate the root cause of the inability to consider the expression constant?
```
333fred, line 116: This isn't a language concern. The compiler is free to make most-specific errors if it can and it would be helpful. It doesn't need to be specified.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,120 @@
...
+
+The allowed subpatterns as shown in the list above are called "constant subpatterns", as they will be eligible for compile-time evaluation.
+
+### Grammar
+[grammar]: #grammar
+The grammar remains untouched, as nothing changes syntactically for this change.
+
+### Semantics
+[semantics]: #semantics
+A pattern expression only consisting of the above subpatterns with a constant LHS may be evaluated at compile time currently, but it cannot be assigned to a constant symbol. With this change, constant fields and locals of type `bool` may be assigned pattern expressions, as already evaluated during compilation.
```
333fred, line 58: This is overly specific. For example, it would prevent this code, which I would expect to be valid:

```cs
const int i = (1 is 2) ? 3 : 4;
```

Rekkonnect: I didn't intend to exclude support for this valid expression, and I think the document doesn't imply the inability to use constant pattern expressions in the evaluated expression of a conditional operator

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Summary
+[summary]: #summary
+
+Consider `is` pattern expressions as constant if the LHS is constant.
+
+## Motivation
+[motivation]: #motivation
+
+Certain pattern expressions that can be converted into simpler boolean expressions using equality and comparison operators are never considered constant expressions, even when all parts of the expression are considered constant and can be evaluated during compilation. However, the lowered versions of those expressions (that only involve equality and comparison operators) are always considered constant. This prohibits the ability to utilize `is` expressions for operations like comparing against a range (e.g. `x is >= 'a' and 'z'`), or checking against distinct values (e.g. `x is Values.A or Values.B or Values.C`).
```
CyrusNajmabadi, line 16: > like comparing against a range (e.g. `x is >= 'a' and 'z'`)

i'm confused by this statement.  `'is'` is not prohibited in either of these cases.  Both are legal.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,120 @@
...
+>   - **Character literal patterns.**
+>   - **String literal patterns.**
+>   - **Relative numeric literal patterns.**
+>   - **Relative character literal patterns.**
+>   - **Null literal patterns.**
+>   - **Default literal patterns.**
+> - sizeof expressions, provided the unmanaged-type is one of the types specified in §23.6.9 for which sizeof returns a constant value.
+> - Default value expressions, provided the type is one of the types listed above.
+
+The allowed subpatterns as shown in the list above are called "constant subpatterns", as they will be eligible for compile-time evaluation.
```
CyrusNajmabadi, line 50: presumably the `not`, `and` and `or` patterns should also work.

proposals/pattern-const-expressions.md
CyrusNajmabadi: the wording says: 

> With this change, constant fields and locals of type `bool` may be assigned pattern expressions,

Which indicates, only fields/locals could benefit from this.   My recommendation would be to go to the part of the spec on constant evaluation and just update it as necessary.  The cases you want will fall out.  By adding these extra clauses, it seems to be adding more restrictive requirements.

proposals/pattern-const-expressions.md
CyrusNajmabadi: i'd enumerate the exact set of patterns we expect this to work for (same as how we enumerate the exact set of unary and binary operators, etc.).  note: based on how the constant-expressions spec works out, i think everything else will likely fall out.

You can then give examples for each case.  I don't htink you have to specify how it evaluates as the constant expression section already says this: 

> Whenever an expression fulfills the requirements listed above, the expression is evaluated at compile-time. This is true even if the expression is a subexpression of a larger expression that contains non-constant constructs.

> The compile-time evaluation of constant expressions uses the same rules as run-time evaluation of non-constant expressions, except that where run-time evaluation would have thrown an exception, compile-time evaluation causes a compile-time error to occur.

--

Note: @333fred if we're doing this spec, would it makes sense to roll in `switch` expressions at the same time, given that the lang already supports `?:`?  Or would you prefer to keep that excluded?  Thanks!

333fred: LDM did express interest in it, but didn't directly approve it. Probably best to leave it as an open question for now.

CyrusNajmabadi: That works for me.  We could bring for an LDM read and easily add if desired.

Rekkonnect: Note, there is a discussion for the equivalent extension to support switch expressions in #7489, which will also be noted in the spec doc

proposals/pattern-const-expressions.md
huoyaoyuan: Another case: `SomeType` didn't implemented `SomeInterface` in referenced version of assembly, but implements it in the newer version used at run time. This is practical, consider the collections that implements readonly interfaces later.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+All the above are currently valid pattern matching expressions, that also emit warnings about their constant evaluation results, about them being always true or false.
+
+When assigning those expressions to a constant symbol, these warnings about the constant result of the expression will **not** be reported, as the user intends to capture the constant value of the expression.
+
+Pattern expressions containing non-constant subpatterns, like accessing properties, list patterns and var patterns, are **not** constant. In the below examples, all expressions will report compiler errors:
+
+```csharp
+const bool q = d is string { Length: 5 }; // Error: not a constant expression
```
huoyaoyuan, line 121: This is an interesting and special case for string. Can we treat length of string literal constant in some way? It can also be useful elsewhere.

Rekkonnect: That falls as a separate concern and would suit for another proposal I believe.

333fred: Yeah, _way_ out of scope of this proposal.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,145 @@
...
+>   - **Numeric literal patterns.**
+>   - **Character literal patterns.**
+>   - **String literal patterns.**
+>   - **Relative numeric literal patterns.**
+>   - **Relative character literal patterns.**
+>   - **Null literal patterns.**
+>   - **Default literal patterns.**
+>   - **And, or and not patterns.**
+>   - **Parenthesized patterns.**
+>   - **References to constant fields or locals inside constant patterns.**
```
333fred, line 72: Can we remove the word `literal` from these? It doesn't seem to be necessary, given the recursive nature of "constant expressions must be composed of other constant expressions", and it will reduce the brittleness here for future changes.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,145 @@
...
+## Alternatives
+[alternatives]: #alternatives
+
+Currently, equality and comparison operators can be used to compare against other literals, including nullability of the objects (e.g. `x != null`, or `y == 4 || y < 3`).
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+- [ ] Requires LDM review
+- [ ] Should this be extended to `switch` expressions too? (Discussion [#7489](https://github.com/dotnet/csharplang/discussions/7489))
```
333fred, line 140: LDM did seem in favor of this, but I would hold off on fully specifying until we work through the subsumption warnings and errors issues.

proposals/pattern-const-expressions.md
```diff
@@ -111,9 +111,54 @@ const bool z = d is not null; // always true
...
 
-When assigning those expressions to a constant symbol, these warnings about the constant result of the expression will **not** be reported, as the user intends to capture the constant value of the expression.
+All the above are currently valid pattern matching expressions, that also emit warnings about their constant evaluation results, being always true or false.
+
+When assigning those expressions in a constant context, these warnings about the constant result of the expression will **not** be reported, as the user intends to capture the constant value of the expression. Constant context involves:
+- initializers for const symbols (fields or locals)
+- attribute arguments
+- parameter default value
+- switch statement case labels
+- switch expression arms
```
333fred, line 123: We need to see the actual spec changes here. How do we have to redefine the sections that specify these constructs to introduce a constant context? What is the definition of a constant context itself?

Rekkonnect:
I believe we should also support tuple pattern matching expressions of the form `(x, y) is (a, b)`, when `x` and `y` are also constants. This tuple syntax does not create a tuple value, but instead deconstructs the expression into two comparisons of the two values, equivalent to `x is a && y is b`. This can be evaluated in compile time, and thus is eligible for being supported from this change.

I did not add the above to the spec before discussing the idea, since it deviates from the classical concept of a constant expression by being primarily driven by syntactical simplification.

333fred:
> I believe we should also support tuple pattern matching expressions of the form (x, y) is (a, b), when x and y are also constants.

I don't think I'm willing to champion this. These are not constants today, and will need a good deal of language and compiler changes to implement. Let's keep the proposal to just the things that have been championed.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,246 @@
...
+```csharp
+const bool q = d is string { Length: 5 }; // Error: not a constant expression
+const bool r = d is [.. var prefix, 'l', 'o']; // Error: not a constant expression
+const bool s = d is var someString; // Error: not a constant expression
+```
+
+### Switch expressions
+[switch-expressions]: #switch-expressions
+
+When evaluating a constant value on a switch expression, we adjust the reported diagnostics based on the matching arms:
```
333fred, line 197: This doesn't seem quite right, and LDM was concerned by the specific behavior that you show examples of below. We wanted to limit the subsumption changes to only constant contexts.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,246 @@
...
+- If no subpattern is matched, we report an **error** instead of a warning about the unmatched subpattern, asking the user to either handle the specific value, or the default case.
+
+For example:
+```csharp
+const int a = 1;
+const int b = 2;
+const int c = 3;
+
+// no warning about the missing default arm,
+// the expression always returns the value of b
```
333fred, line 208: LDM was specifically interested in making this _not_ change behavior.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,246 @@
...
+
+    default:
+        // always false, no warning/error
+        goto case a is not c;
+}
+
+// switch expression arm
+var p = b switch
+{
+    // always true, no warning/error
```
333fred, line 176: 2 problems with this:
1. This is actually missing a case. `b` is `false`, and `a is c` is `true`.
2. The overall switch expression should not change behavior, which is to say it should still warn about missing cases.

proposals/pattern-const-expressions.md
Rekkonnect: Is it fine if we make raise the warning into an error if we fail to match any subpattern in the constant switch? The constant value cannot be assigned without being provided a value from the default case, should all others fail to match.

For example,

```csharp
const int a = 4;
// ERROR: No subpattern is matched, and the assigned value is a constant
const int c = a switch
{
    1 => 2,
    2 => 3,
    3 => 5,
};

// WARNING: No subpattern is matched, the default case will be handled and it will throw at runtime.
int v = a switch
{
    1 => 2,
    2 => 3,
    3 => 5,
};
```

I believe that both of the above cases should be an error, but I'd like your input about those.

333fred: The second case should have unchanged behavior from today. The first case would indeed be an error.

proposals/pattern-const-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+// the expression always returns the value of b
+int x = a switch
+{
+    a => b,
+    b => a,
+    c => a + b,
+};
+
+// we get a warning about the missing default arm,
+// the expression results in xc being assigned the value of b
```
333fred, line 218: I would not expect to get a warning here.

proposals/pattern-const-expressions.md
333fred: In general, the wording itself is not what I expected here. I don't see any mention of constant contexts, which is what I am expecting to drive changes in warning behavior.

PR #7591, opened 2023-10-10
Collection expressions: prefer ReadOnlySpan<T> over Span<T> in overload resolution

cston:
See [LDM-2023-10-11.md](https://github.com/dotnet/csharplang/blob/main/meetings/2023/LDM-2023-10-11.md#conclusion)

See https://github.com/dotnet/roslyn/issues/70318

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -441,8 +441,8 @@ In the updated rules:
 > Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if one of the following holds:
 >
 > * **`E` is a *collection expression* and one of the following holds:**
->   * **`T₁` is `System.ReadOnlySpan<E₁>`, and `T₂` is `System.ReadOnlySpan<E₂>`, and an implicit conversion exists from `E₁` to `E₂`**
->   * **`T₁` is `System.ReadOnlySpan<E₁>` or `System.Span<E₁>`, and `T₂` is `System.Span<E₂>` or an *array_or_array_interface_or_string_type* with *iteration type* `E₂`, and an implicit conversion exists from `E₁` to `E₂`**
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi (approved):

333fred (approved):

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -440,12 +440,14 @@ In the updated rules:
...
->   * **`C₁` and `C₂` are collection expression conversions *and***
->   * **`T₁` is a *span_type* and `T₂` is an *array_or_array_interface_or_string_type*, *or* `T₂` is a *span_type* and `T₁` is an *array_or_array_interface_or_string_type***
-> * **`C₁` and `C₂` are collection expression conversions, and `T₁` is a *span_type* with *iteration type* `E₁`, and `T₂` is an *array_or_array_interface_or_string_type* with *iteration type* `E₂`, and `E₁` is implicitly convertible to `E₂`**
-> * ...
+> * **`E` is a *collection expression* and one of the following holds:**
+>   * **`T₁` is `System.ReadOnlySpan<E₁>`, and `T₂` is `System.Span<E₂>`, and an implicit conversion exists from `E₁` to `E₂`**
+>   * **`T₁` is `System.ReadOnlySpan<E₁>` or `System.Span<E₁>`, and `T₂` is an *array_or_array_interface_or_string_type* with *iteration type* `E₂`, and an implicit conversion exists from `E₁` to `E₂`**
+>   * **`T₁` is not a *span_type*, and `T₂` is not a *span_type*, and an implicit conversion exists from `T₁` to `T₂`**
```
RikkiGibson, line 446: nit: I wasn't sure why the qualifier "not a *span_type*" was needed here. Maybe without it, we would consider array to be "more derived" than span and therefore better?

Have we fully tested the difference in the betterness rules for a user-defined span type versus System.Span? e.g. if a user-defined ref struct collection were implicitly convertible from array, then array would be a better conversion from collection-expression, right?

cston: Yes, "not a *span_type*" is needed to avoid allowing an implicit conversion from span to array.

Added tests to https://github.com/dotnet/roslyn/pull/70328 for a ref struct collection that is implicitly convertible to array. Thanks.

PR #7593, opened 2023-10-11
Add docs template for spec issues

BillWagner:
Once this is merged I'll add the config to direct customer issues to this template.

The "View file" tab on changes will give a reasonable facsimile of the experience a user will see when they are routed to this form. The fields at the bottom will be filled it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/docs-feedback.yml
+++ b/.github/ISSUE_TEMPLATE/docs-feedback.yml
@@ -3,6 +3,9 @@ description: |
   ⛔ This template is hooked into the feedback control on the bottom of every page on the learn.microsoft.com site. It automatically fills in several fields for you. Don't use for other purposes. ⛔
 assignees:
   - BillWagner
+labels:
+  - Area-Speclets
+  - untriaged
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/docs-feedback.yml
+++ b/.github/ISSUE_TEMPLATE/docs-feedback.yml
@@ -1,0 +1,47 @@
+name: Learn feedback control.
+description: |
+  ⛔ This template is hooked into the feedback control on the bottom of every page on the learn.microsoft.com site. It automatically fills in several fields for you. Don't use for other purposes. ⛔
+assignees:
+  - BillWagner
+body:
+  - type: markdown
+    attributes:
+      value: "## Issue information"
+  - type: markdown
+    attributes:
+      value: Select the issue type, and describe the issue in the text box below. Add as much detail as needed to help us resolve the issue.
+  - type: dropdown
+    id: issue-type
+    attributes:
+      label: Type of issue
+      options:
+        - Typo
+        - Spec incorrect
+        - Spec incomplete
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/docs-feedback.yml
```diff
@@ -0,0 +1,47 @@
+name: Learn feedback control.
+description: |
+  ⛔ This template is hooked into the feedback control on the bottom of every page on the learn.microsoft.com site. It automatically fills in several fields for you. Don't use for other purposes. ⛔
+assignees:
+  - BillWagner
```
BillWagner, line 5: Who else should I add here?

I think I can add a team, as long as the org isn't needed (meaning its a team in the dotnet org). The GitHub docs are unclear.

BillWagner:
The labels will be specified in the PR that updates the docs config. We've learned that those fields aren't obeyed when a URL routes directly to the template.

.github/ISSUE_TEMPLATE/docs-feedback.yml
jcouv: I'd rather not assign a team. It's preferable to have a single owner for each issue.
If you're okay routing/triaging incoming issues by default, that'd be great, but we can discuss some options if the load is too great.

jcouv (approved):

.github/ISSUE_TEMPLATE/docs-feedback.yml
BillWagner: Triaging these hasn't been a major time sink in the past. I don't expect that to change, but I'll let everyone know if it is.

.github/ISSUE_TEMPLATE/docs-feedback.yml
```diff
@@ -0,0 +1,47 @@
...
+body:
```
333fred, line 6: Can we add a `labels` section and apply some label like `speclet issue` or something like that?

333fred: We'll need to make the label as well.

BillWagner: The labels are added as a config in the docs repo. What we found during initial testing is that when the URL targets a specific issue template, some of the items in the template aren't honored (title and labels). So, we add add those in the docfx.json file in the docs repo, and the labels get passed on the URL.

I was thinking of adding the "not triaged" label, and a new "area-speclet" label.

333fred: > The labels are added as a config in the docs repo.

The problem, though, is that this template will show up when people create a new issue from this repo as well. If they do, the right labels won't be applied.

BillWagner: Good point. I'll add them in both.

BillWagner:
I added an "Area-Speclets" label to match the other "Area-*" labels.

I updated the template to apply that label, and the "untriaged" label.

333fred (approved):

PR #7606, opened 2023-10-17
replace speclet terms with ECMA standard terms

BillWagner:
Add a note at the top of each file that explains the migration from "safe to escape" to "safe context".

I didn't add a key for the terms "enclosing scope", "entire method", "calling method" and "safe to return" because they weren't used consistently.

Namely, "entire method" was sometimes called "current method". There were also multiple spellings, capitalizations, and word casing involved.


BillWagner:
Note to @jcouv , @333fred 

I think this is complete, but I haven't done a proofread yet. I'll do that in the morning, and then mark it ready for review.

333fred (approved):

Issue #7608, opened 2023-10-18
[Proposal]: Let ref structs implement interfaces and substitute into type parameters


agocke:
# Ref structs implementing interfaces

* [x] Proposed
* [ ] Prototype: Not Started
* [ ] Implementation: Not Started
* [ ] Specification: Not Started

## Summary
[summary]: #summary

There's no particular problem with letting `ref structs` implement interfaces today. The problem is that there's no way to call them. I propose we allow ref structs to substitute type parameters, which would let users call the interface methods safely.

This proposal has two parts. The first part is a constraint, called `box`. This constraint allows boxing the constrained type parameter. It would look something like,

```C#
void M<T>(T t) where T : box
{
   object o = t; // box
}
```

The trick is that this `box` constraint already implicitly exists on all type parameters, even unconstrained type parameters. So we need a way to remove it. The second part of this proposal is a new language feature, called "anti-constraints." An anti-constraint is a constraint that removes, or cancels out, an implicit constraint. We could use the `~` syntax to represent the complement operation which cancels out the implicit constraint. If C# users want to pass ref structs through generic parameters they would now have to declare the parameter as

```C#
void M<T>(T t) where T : ~box
{
   object o = t; // would now produce an error
}
```

With this restriction in place, ref structs would be able to implement interfaces **and** users would be able to call those interface implementations.

## Motivation
[motivation]: #motivation

The motivation here is simple: this is an onerous restriction that is there primarily to ensure ref struct safety. We can ensure ref struct safety with a small new feature that will still let users implement interfaces. Spans should be able to implement IEnumerable, wrappers should be able to implement IDisposable, etc.

The current restriction is a side effect of a rule to preserve `ref struct` safety. Ref structs can never appear in the heap, meaning that they can never be boxed. To invoke an interface implementation on a struct in C#, the user must either box the struct instance to the interface and then call the method, or pass the instance to a method with a generic parameter that is constrained to the interface. Ref structs cannot be boxed, and therefore the first option is never viable. The second option is not _currently_ viable because C# does not place any restrictions on boxing generic parameters inside generic methods. This is easily fixable.

## Detailed design
[design]: #detailed-design

We don't need much more than the above. We would have to go through and ensure that ref struct boxing is prohibited in all the right places, but this is already implemented for boxing to `object`.

## Drawbacks
[drawbacks]: #drawbacks

[...]

PR #7625, opened 2023-10-24
Allow ref struct to implement interfaces

jaredpar:
This proposal sketches out what it would look like if `ref struct` were allowed to implement interfaces and participate as generic parameters.  

Related #7608

### DIFF
--- a/proposals/ref-struct-interfaces.md
+++ b/proposals/ref-struct-interfaces.md
@@ -32,7 +32,7 @@ IDisposable d = f;
 
 The ability to implement interfaces is only useful when combined with the ability for `ref struct` to participate in generic arguments (as [laid out later](#ref-struct-generic)).
 
-To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when the `class` implements the interface. 
+To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when a `class` or non-ref `struct` implements the interface. 
 
 Default interface methods pose a problem for `ref struct` as there are no protections against the default implementation boxing the `this` member. 
 
@@ -67,7 +67,7 @@ T Identity<T>(T p)
     => p;
 
 // Okay
-Span<T> local = Identity(new Span<int>(new int[10]));
+Span<int> local = Identity(new Span<int>(new int[10]));
 ```
 
 This is similar to a `where` in that it specifies the capabilities of the generic parameter. The difference is `where` limits the set of types that can fulfill a generic parameter while the behavior defined here expands the set of types. This is effectively an anti-constraint as it removes the implicit constraint that `ref struct` cannot satisfy a generic parameter. As such this is given a new syntax, `allow`, to make that clearer.
@@ -168,7 +168,7 @@ This feature requires several pieces of support from the runtime / libraries tea
 
 Most of this support is likely already in place. The general `ref struct` as generic parameter support is already implemented as described [here](https://github.com/dotnet/runtime/blob/main/docs/design/features/byreflike-generics.md). It's possible the DIM implementation already account for `ref struct`. But each of these items needs to be tracked down.
 
-### Span<Span<T>>
+### Span&lt;Span&lt;T&gt;&gt;
 This combination of features does not allow for constructs such as `Span<Span<T>>`. This is made a bit clearer by looking at the definition of `Span<T>`: 
 
 ```csharp
--- a/proposals/ref-struct-interfaces.md
+++ b/proposals/ref-struct-interfaces.md
@@ -107,6 +107,7 @@ T M2<T>(T p)
 }
 
 T M3<T>(Span<int> span)
+    allow T : ref struct
 {
     return default;
 }
--- a/proposals/ref-struct-interfaces.md
+++ b/proposals/ref-struct-interfaces.md
@@ -41,6 +41,7 @@ interface I1
 {
     void M()
     {
+        // Error: both of these box if I1 is implemented by a ref struct
         I1 local1 = this;
         object local2 = this;
     }
@@ -50,6 +51,7 @@ ref struct S = I1 { }
 ```
 
 To handle this a `ref struct` will be forced to implement all members of an interface, even if they have default implementations. The runtime will also be updated to throw an exception if a default interface member is called on a `ref struct` type.
+
 Detailed Notes:
 - A `ref struct` can implement an interface
 - A `ref struct` cannot participate in default interface members
@@ -68,13 +70,55 @@ T Identity<T>(T p)
 Span<T> local = Identity(new Span<int>(new int[10]));
 ```
 
-**TO COVER**
-- lifetime rules
+This is similar to a `where` in that it specifies the capabilities of the generic parameter. The difference is `where` limits the set of types that can fullfill a generic parameter while the behavior defined here expands the set of types. This is effectively an anti-constraint as it removes the implicit constraint that `ref struct` cannot satisfy a generic parameter. As such this is given a new syntax, `allow`, to make that clearer.
+
+A type parameter bound by `allow T: ref struct` has all of the behaviors of a `ref struct` type: 
+
+1. Instances of it cannot be boxed
+2. Instances participate in lifetime rules like a normal `ref struct`
+3. The type parameter cannot be used in `static` fields, elements of an array, etc ...
+4. Instances can be marked with `scoped`
+
+Examples of these rules in action: 
+
+```csharp
+interface I1 { }
+I1 M1<T>(T p)
+    allow T : ref struct, I1
+{
+    // Error: cannot box potential ref struct
+    return p;
+}
+
+T M2<T>(T p)
+{
+    Span<int> span = stackalloc int[42];
+
+    // The safe-to-escape of the return is current method because one of the inputs is 
+    // current method
+    T t = M3<T>(span);
+
+    // Error: the safe-to-escape is current method.
+    return t;
+
+    // Okay
+    return default;
+    return p;
+}
+
+T M3<T>(Span<int> span)
+{
+    return default;
+}
+```
+
+These parameters will be encoded in metadata as described in the [byref-like generics doc](https://github.com/dotnet/runtime/blob/main/docs/design/features/byreflike-generics.md). Specifically by using the `gpAcceptByRefLike(0x0020)` attribute value.
 
 Detailed notes: 
 - A `allow T : ref struct` generic parameter cannot 
     - Have `where T : U` where `U` is a known reference type
     - Have `where T : class` constraint
+    - Cannot be used as a generic argument unless the corresponding parameter is also `allow T: ref struct`
 - A type parameter `T` which has `allow T: ref struct` has all the same limitations as a `ref struct` type.
 
 ## Open Issues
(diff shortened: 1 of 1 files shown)


jaredpar:
@agocke, @AaronRobinsonMSFT, @stephentoub, @RikkiGibson, @cston

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+}
+
+File f = ...;
+// Error: cannot box `ref struct` type `File`
+IDisposable d = f;
+```
+
+The ability to implement interfaces is only useful when combined with the ability for `ref struct` to participate in generic arguments (as [laid out later](#ref-struct-generic)).
+
+To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when the a `class` implements the interface.
```
AaronRobinsonMSFT, line 35:
```suggestion
To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when a `class` implements the interface. 
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
+Ref Struct Interfaces
+=====
+
+## Summary
+This proposal will expand the capabilities of `ref struct` such that they can implement interfaces and participate as generic type arguments.
+
+## Motivation
+The inability for `ref struct` to implement interfaces means they cannot partipcate in fairly fundamental abstraction techniques of .NET. A `Span<T>`, even though it has all the attributes of a sequential list cannot participate in methods that take `IReadOnlyList<T>`, `IEnumerable<T>`, etc ... Instead specific methods must be coded for `Span<T>` that have virtually the same implementation. Allowing `ref struct` to implement interfaces will allow operations to be abstracted over them as they are for other types.
```
cston, line 8:
```suggestion
The inability for `ref struct` to implement interfaces means they cannot participate in fairly fundamental abstraction techniques of .NET. A `Span<T>`, even though it has all the attributes of a sequential list cannot participate in methods that take `IReadOnlyList<T>`, `IEnumerable<T>`, etc. ... Instead specific methods must be coded for `Span<T>` that have virtually the same implementation. Allowing `ref struct` to implement interfaces will allow operations to be abstracted over them as they are for other types. 
```

proposals/ref-struct-interfaces.md
cston, line 35:
```suggestion
To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when the `class` implements the interface. 
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+```csharp
+T Identity<T>(T p) 
+    allow T : ref struct
+    => p;
+
+// Okay
+Span<T> local = Identity(new Span<int>(new int[10]));
+```
+
+This is similar to a `where` in that it specifies the capabilities of the generic parameter. The difference is `where` limits the set of types that can fullfill a generic parameter while the behavior defined here expands the set of types. This is effectively an anti-constraint as it removes the implicit constraint that `ref struct` cannot satisfy a generic parameter. As such this is given a new syntax, `allow`, to make that clearer.
```
cston, line 73:
```suggestion
This is similar to a `where` in that it specifies the capabilities of the generic parameter. The difference is `where` limits the set of types that can fulfill a generic parameter while the behavior defined here expands the set of types. This is effectively an anti-constraint as it removes the implicit constraint that `ref struct` cannot satisfy a generic parameter. As such this is given a new syntax, `allow`, to make that clearer.
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+To be maximally useful type parameters that are `allow T : ref struct` must be compatible with generic variance. Specifically it must be legal for a parameter to be both co/contravariant and also `allow T: ref struct`. Lacking that they would not be usable in many of the most popular `delegate` and `interface` types in .NET like `Func<T>`, `Action<T>`, `IEnumerable<T>`, etc ...
+
+Given there is no actual variance when `struct` are involved these should be compatible. There is still some concern that I'm missing deeply generic variance cases. Need to sit down with @agocke to work out if this is truly safe or if there are deeply generic scenarios that need to be worked out.
+
+### Auto-applying to delegate members
+**Decision**: do not auto-apply
+
+For many generic `delegate` members the language could automatically apply `allow T: ref struct` as it's purely an upside change. Consider that for `Func<> / Action<>` style delegetes there is no downside to expanding to allowing `ref struct`. The language can outline rules where it is safe to automatically apply this anti-constraint. This removes the manual process and would speed up the adoption of this feature.
```
cston, line 154:
```suggestion
For many generic `delegate` members the language could automatically apply `allow T: ref struct` as it's purely an upside change. Consider that for `Func<> / Action<>` style delegates there is no downside to expanding to allowing `ref struct`. The language can outline rules where it is safe to automatically apply this anti-constraint. This removes the manual process and would speed up the adoption of this feature.
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+
+While that is true it can present a problem in multi-targeted scenarios. Code would compile in one target framework but fail in another. This could lead to confusion with customers and result in a desire for a more explicit opt-in.
+
+### Binary breaking change
+Adding `allow T: ref struct` to an existing API is not a source breaking change. It is purely expanding the set of allowed types for an API. Need to track down if this is a binary breaking change or not. Unclear if updating the attributes of a generic parameter constitute a binary breaking change.
+
+## Considerations
+
+### Runtime support
+This feature requires several pieces of support from the runtime / libaries team:
```
cston, line 164:
```suggestion
This feature requires several pieces of support from the runtime / libraries team: 
```

proposals/ref-struct-interfaces.md
jaredpar: I thought that was wrong but VS code didn't highlight it so I went along with it.

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Runtime support
+This feature requires several pieces of support from the runtime / libraries team: 
+- Preventing default interface methods from applying to `ref struct`
+- API in `System.Reflection.Metadata` for encoding the `gpAcceptByRefLike` value.
+- Support for generic parameters being a `ref struct`
+
+Most of this support is likely already in place. The general `ref struct` as generic parameter support is already implemented as described [here](https://github.com/dotnet/runtime/blob/main/docs/design/features/byreflike-generics.md). It's possible the DIM implementation already account for `ref struct`. But each of these items needs to be tracked down.
+
+### Span<Span<T>>
```
cston, line 171:
```suggestion
### Span&lt;Span&lt;T&gt;&gt;
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+<a name="ref-struct-generic"></a>
+The language will allow for generic parameters to opt into supporting `ref struct` as arguments by using the `allow T : ref struct` syntax: 
+
+```csharp
+T Identity<T>(T p) 
+    allow T : ref struct
+    => p;
+
+// Okay
+Span<T> local = Identity(new Span<int>(new int[10]));
```
cston, line 70:
```suggestion
Span<int> local = Identity(new Span<int>(new int[10]));
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+
+    // Error: the safe-to-escape is current method.
+    return t;
+
+    // Okay
+    return default;
+    return p;
+}
+
+T M3<T>(Span<int> span)
```
cston, line 109: Should `M3<T>` have a `Span<T>` parameter? If so, perhaps:

```suggestion
T M2<T>(T p)
    where T : unmanaged
{
    Span<T> span = stackalloc T[42];

    // The safe-to-escape of the return is current method because one of the inputs is 
    // current method
    T t = M3<T>(span);

    // Error: the safe-to-escape is current method.
    return t;

    // Okay
    return default;
    return p;
}

T M3<T>(Span<T> span)
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+Span<T> local = Identity(new Span<int>(new int[10]));
+```
+
+This is similar to a `where` in that it specifies the capabilities of the generic parameter. The difference is `where` limits the set of types that can fulfill a generic parameter while the behavior defined here expands the set of types. This is effectively an anti-constraint as it removes the implicit constraint that `ref struct` cannot satisfy a generic parameter. As such this is given a new syntax, `allow`, to make that clearer.
+
+A type parameter bound by `allow T: ref struct` has all of the behaviors of a `ref struct` type: 
+
+1. Instances of it cannot be boxed
+2. Instances participate in lifetime rules like a normal `ref struct`
+3. The type parameter cannot be used in `static` fields, elements of an array, etc ...
```
agocke, line 79: In short, `ref struct` cannot appear on the heap, and the rules are meant to prevent that.

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+// Okay
+Span<T> local = Identity(new Span<int>(new int[10]));
+```
+
+This is similar to a `where` in that it specifies the capabilities of the generic parameter. The difference is `where` limits the set of types that can fulfill a generic parameter while the behavior defined here expands the set of types. This is effectively an anti-constraint as it removes the implicit constraint that `ref struct` cannot satisfy a generic parameter. As such this is given a new syntax, `allow`, to make that clearer.
+
+A type parameter bound by `allow T: ref struct` has all of the behaviors of a `ref struct` type: 
+
+1. Instances of it cannot be boxed
+2. Instances participate in lifetime rules like a normal `ref struct`
```
agocke, line 78: This is perfectly fine, but I also think we could slip some of this support to a second release if necessary. If we don't allow scoped/unscoped/etc then I think the default ref-struct rules are very useful and the cases where you need different lifetimes are pretty narrow.

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+While that is true it can present a problem in multi-targeted scenarios. Code would compile in one target framework but fail in another. This could lead to confusion with customers and result in a desire for a more explicit opt-in.
+
+### Binary breaking change
+Adding `allow T: ref struct` to an existing API is not a source breaking change. It is purely expanding the set of allowed types for an API. Need to track down if this is a binary breaking change or not. Unclear if updating the attributes of a generic parameter constitute a binary breaking change.
+
+## Considerations
+
+### Runtime support
+This feature requires several pieces of support from the runtime / libraries team: 
+- Preventing default interface methods from applying to `ref struct`
```
agocke, line 165: Yeah, this is an important one that will likely need runtime work.

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+
+Default interface methods pose a problem for `ref struct` as there are no protections against the default implementation boxing the `this` member. 
+
+```csharp
+interface I1
+{
+    void M()
+    {
+        // Error: both of these box if I1 is implemented by a ref struct
+        I1 local1 = this;
```
agocke, line 45: The runtime will demand a `this` pointer as a boxed interface even without any of these explicit usages. Simply calling this method will break the runtime. So you're right that this needs a runtime exception.

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+- A `ref struct` cannot participate in default interface members
+- A `ref struct` cannot be cast to interfaces it implements as that is a boxing operation
+
+### ref struct Generic Parameters
+<a name="ref-struct-generic"></a>
+The language will allow for generic parameters to opt into supporting `ref struct` as arguments by using the `allow T : ref struct` syntax: 
+
+```csharp
+T Identity<T>(T p) 
+    allow T : ref struct
```
agocke, line 66: \*standard handwaves around syntax\*

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,195 @@
...
+}
+
+File f = ...;
+// Error: cannot box `ref struct` type `File`
+IDisposable d = f;
+```
+
+The ability to implement interfaces is only useful when combined with the ability for `ref struct` to participate in generic arguments (as [laid out later](#ref-struct-generic)).
+
+To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when the `class` implements the interface.
```
Joe4evr, line 35:
```suggestion
To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when a `class` implements the interface. 
```

proposals/ref-struct-interfaces.md
jaredpar: This should be `allow T : ref struct`

proposals/ref-struct-interfaces.md
jaredpar: Not sure what you want to slip here. Instances of the type parameter need to behave like `ref struct` from a lifetime perspective else it's easy to create memory safety issues. That support is fairly easy to add. Conceptually it's just applying lifetime rules to any type which could be a `ref struct` but really nothing else changes.

proposals/ref-struct-interfaces.md
jaredpar: What does this change do? Today it renders as I would expect so not sure what this change brings.

proposals/ref-struct-interfaces.md
agocke: > Not sure what you want to slip here

Just the lifetime modifier syntax like `scoped`. Agreed it may not be worth it to skip.

stephentoub:
Thanks, @jaredpar. I read this with an eye towards https://github.com/dotnet/runtime/issues/27229#issuecomment-1537274804 and really wanting to get a solution there in for .NET 9. Nothing in this PR jumped out at me as being problematic for that, but wanted to double-check.

proposals/ref-struct-interfaces.md
cston: The title is currently rendered as `Span<Span>` where the nested `<T>` is missing.

jaredpar:
@stephentoub Looked through the `IMappingComparer<TMappedKey, TTarget>` and that seems compatible with the design proposal here.

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+I1 M1<T>(T p)
+    allow T : ref struct, I1
+{
+    // Error: cannot box potential ref struct
+    return p;
+}
+
+T M2<T>(T p)
+{
+    Span<int> span = stackalloc int[42];
```
cston, line 95:
```suggestion
T M2<T>(T p)
    where T : unmanaged
{
    Span<T> span = stackalloc T[42];
```

proposals/ref-struct-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+
+    // Error: the safe-to-escape is current method.
+    return t;
+
+    // Okay
+    return default;
+    return p;
+}
+
+T M3<T>(Span<int> span)
```
cston, line 109:
```suggestion
T M3<T>(Span<T> span)
```

agocke (approved):
LGTM, I think we can fill in details as needed

proposals/ref-struct-interfaces.md
jaredpar: corrected to `allow T : ref struct`

PR #7635, opened 2023-10-30
Mention nullability annotations

jcouv:
@KathleenDollard @CyrusNajmabadi for review. Thanks

### DIFF
--- a/proposals/csharp-12.0/using-alias-types.md
+++ b/proposals/csharp-12.0/using-alias-types.md
@@ -22,6 +22,8 @@ using_alias_directive
     ;
 ```
 
+Note: this grammar excludes top-level reference type nullability annotations.
+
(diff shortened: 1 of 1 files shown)


RikkiGibson (approved):

proposals/csharp-12.0/using-alias-types.md
```diff
@@ -22,6 +22,8 @@ using_alias_directive
     ;
 ```
 
+Note: this grammar excludes top-level reference type nullability annotations.
```
RikkiGibson, line 25: It's not obvious to me how the grammar can do this, when nullable value types and reference types are syntactically the same.

jcouv: You're right. We need a special rule to exclude nullable reference types. Thanks

Issue #179, opened 2017-02-24
Champion "Permit IEnumerable<T> as the type of a params parameter"


gafter:
- [ ] Proposal added
- [ ] Discussed in LDM
- [ ] Decision in LDM
- [ ] Finalized (done, rejected, inactive)
- [ ] Spec'ed

See also
- github.com/dotnet/roslyn/issues/36
- https://github.com/dotnet/csharplang/issues/178

### Design meetings

https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-08-31.md#params-ienumerable
https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-09-26.md#params-improvements

Richiban:
I would assume that params was originally introduced so that the programmer didn't have the duality of having to write two methods just so that a piece of functionality could work with either a T or a collection of T, such as below:

    public void MyMethod(string item) 
    {
        MyMethod(new string[] {item});
    }

    public void MyMethod(string[] items)
    {
        // ... Do something with items
    }

While this was great, unfortunately as soon as IEnumerable<T> was introduced and it became the de facto standard base collection type we're back to the duality again:

    public void MyMethod(params string[] item) => MyMethod(items);

    public void MyMethod(IEnumerable<string> items)
    {
        // ... Do something with items
    }

Actually, since we're on the subject, I'd like to propose that not only is `params IEnumerable<T>` allowed, but so should `params IReadOnlyList<T>`, `params IList<T>` etc. These types are all allowed because they are all implemented by `T[]`. So, more formally:

`params M<U> identifier` is allowed if `U[] : M<U>`.

This works because at the callsite of method

    public void MyMethod(params T<U> items)

the call:

    MyMethod(item1, item2, item3)

gets converted to:

    MyMethod(new[] {item1, item2, item3})

So the `params` keyword should be allowed on any type that is assignable from an array.

This is desirable because, as an API designer, it can sometimes feel dangerous to accept `IEnumerable<T>` as a parameter when you expect an in-memory collection because you don't know that the given argument is actually a collection at runtime. It could be lazily-loaded like a database query, or it might even be infinite, in which case you cannot safely call `Count()` or even `foreach`.

aluanhaddad:
[...]

PR #7661, opened 2023-11-06
Create params-collections.md proposal

AlekseyTs:


333fred:
This proposal needs a corresponding Champion issue, and it seems broader than either https://github.com/dotnet/csharplang/issues/1757 or https://github.com/dotnet/csharplang/issues/179. Perhaps we should create a new one, and close the other two as superseded by the new one?

proposals/params-collections.md
```diff
@@ -0,0 +1,171 @@
...
+    ;
+
+-parameter_array
++parameter_collection
+-    : attributes? 'params' array_type identifier
++    : attributes? 'params' 'scoped'? type identifier
+    ;
+```
+
+A *parameter_collection* consists of an optional set of *attributes*, a `params` modifier, an optional `scoped` modifier,
```
333fred, line 41: Feels like we should have an open question on whether to permit `scoped` for `params`, or if we should always infer it if the type is a ref struct, similar to how we do for `this` and `out` parameters.

proposals/params-collections.md
```diff
@@ -0,0 +1,171 @@
...
+a *type*, and an *identifier*. A parameter collection declares a single parameter of the given type with the given name.
+The *type* of a parameter collection shall be one of the following valid target types for a collection expression
```
333fred, line 43: Could we not simplify this list by just specifying that there must exist a `collection expression conversion` from a collection expression to the target type?

proposals/params-collections.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Alternatives 
+
+There is an alternative [proposal](https://github.com/dotnet/csharplang/blob/main/proposals/params-span.md) that extends
+`params` only for `ReadOnlySpan<T>`.
+
+Also, one might say, that with [collection expressions](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md)
+now in the language, there is no need to extend `params` support at all. Foe any collection type. To consume an API with collection type, a developer
+simply needs to add two characters, `[` before the expanded list of arguments, and `]` after it. Given that, extending `params` support might be an overkill,
+especially that other languages are unlikely to support consumption of non-array `params` parameters any time soon.
```
CyrusNajmabadi, line 167: right.  this was the POV of a few people the last time we discussed this.  namely, that it's totally fine to just have the user write `[...]` around teh arguments.  THe critical case as `params Span` because in that case we want the API developer to be able to add the overload and have it take precedence over the array version, just by recompiling (so no change to the callsite). 

--

Note: i'm personally fine going either way.  `params collection` seems the most consistent approach to take.  but `params span` also seems sufficient.

333fred: It's probably worth expanding the "we at least need to do `params Span` so the BCL can offer better overloads" part explicitly here.

proposals/params-collections.md
jaredpar: When `params` modifies a `ref struct` parameter I think `scoped` should be implicitly enabled. The number of cases where you want that is virtually 100% when looking through the BCL cases. If it turns out that there is need / demand for it to be not `scoped` in a few cases then we could augment the design to support `[UnscopedRef]` here but I wouldn't want to do that without a strong supporting scenario.

proposals/params-collections.md
jaredpar: The better overloads part is definitely a big motivator for this feature.

proposals/params-collections.md
```diff
@@ -0,0 +1,171 @@
...
+- A *type* with a valid [create method](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md#create-methods)
+  with *parameter type* `System.ReadOnlySpan<T>` and the method is at least as accessible as the *type*.
+- A *struct* or *class type* that implements `System.Collections.Generic.IEnumerable<T>`
+- A *struct* or *class type* that implements `System.Collections.IEnumerable` and *does not implement* `System.Collections.Generic.IEnumerable<T>`.
+- An *interface type* `System.Collections.Generic.IEnumerable<T>`, `System.Collections.Generic.IReadOnlyCollection<T>`,
+  `System.Collections.Generic.IReadOnlyList<T>`, `System.Collections.Generic.ICollection<T>`, or `System.Collections.Generic.IList<T>`
+
+In a method invocation, a parameter collection permits either a single argument of the given parameter type to be specified, or
+it permits zero or more arguments of the collection [iteration type](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/statements.md#1395-the-foreach-statement)
+to be specified. Parameter collections are described further in *[Parameter collections](#parameter-collections)*.
```
jjonescz, line 56: Or it permits a single argument of the parameter collection type.

proposals/params-collections.md
```diff
@@ -0,0 +1,171 @@
...
+We should follow the spec for [`params Span<T>`](https://github.com/dotnet/csharplang/issues/1757)." It sounds like we have to answer
+the questions in context of this proposal.
+
+## Alternatives 
+
+There is an alternative [proposal](https://github.com/dotnet/csharplang/blob/main/proposals/params-span.md) that extends
+`params` only for `ReadOnlySpan<T>`.
+
+Also, one might say, that with [collection expressions](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md)
+now in the language, there is no need to extend `params` support at all. Foe any collection type. To consume an API with collection type, a developer
```
jjonescz, line 165:
```suggestion
now in the language, there is no need to extend `params` support at all. For any collection type. To consume an API with collection type, a developer
```

proposals/params-collections.md
AlekseyTs: > Feels like we should have an open question on whether to permit scoped for params, or if we should always infer it if the type is a ref struct, similar to how we do for this and out parameters.

I captured what I am proposing. You should be able to raise concerns, if any, during LDM discussion.

proposals/params-collections.md
AlekseyTs: > Could we not simplify this list by just specifying that there must exist a collection expression conversion from a collection expression to the target type?

First, there is no collection expression. Second, I intentionally exclude the nullable value type case.

proposals/params-collections.md
AlekseyTs: > Or it permits a single argument of the parameter collection type.

Doesn't "a parameter collection permits either a single argument of the ..." part capture that?

jjonescz: You're right, missed that.

proposals/params-collections.md
333fred: I think it is good to capture any known open questions before LDM. This is one such question.

PR #4396, opened 2021-02-05
Update Language-Version-History.md

leandromoh:
Added Source Generators on C# 9.

I know it is not a language feature but a compiler feature, but since the page also contains items like `Draft Specification online` and `Compiler-as-a-service (Roslyn)` I think this is ok

PR #7701, opened 2023-11-16
Update Language-Version-History.md

leandromoh:


333fred:
This is not a language change, it should not be documented here.

leandromoh:
@333fred the file is not just about language features.. checkout [this comment](https://github.com/dotnet/csharplang/pull/4396#issuecomment-789089907) of a similar situation...

another example of items that are present in the file and are not a language change.
- Source Generators
- Caller info attributes
- Draft Specification online 
- Compiler-as-a-service (Roslyn)
- Improved overload candidates
- etc

333fred:
Those are all either compiler or language features. This is neither.

PR #7734, opened 2023-11-30
Collection expressions: clarify multi-dimensional arrays are not valid target types

cston:


CyrusNajmabadi (approved):

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -120,7 +120,7 @@ The following implicit *collection expression conversions* exist from a collecti
 
   * For each *element* `Ei` there is an *implicit conversion* to `T`.
 
-* To a *struct* or *class type* that implements `System.Collections.IEnumerable` and *does not implement* `System.Collections.Generic.IEnumerable<T>`.
+* To a *struct* or *class type*, other than a multi dimensional *array type*, that implements `System.Collections.IEnumerable` and *does not implement* `System.Collections.Generic.IEnumerable<T>`.
```
AlekseyTs, line 123: I think this exception applies to other bullets as well. Instead of spreading this around, consider if it would be simpler to state the fact that MD arrays are not supported as a separate note, as a clarification for the intent.

AlekseyTs (approved):
LGTM (commit 2)

333fred (approved):

PR #7749, opened 2023-12-08
Add a design question to collection-expressions.md

AlekseyTs:


proposals/csharp-12.0/collection-expressions.md
```diff
@@ -874,6 +874,67 @@ However, given the breadth and consistency brought by the new literal syntax, we
...
+    IEnumerator<char> IEnumerable<char>.GetEnumerator() => throw null;
+    IEnumerator IEnumerable.GetEnumerator() => throw null;
+}
+```
+
+It looks like the situation is very similar to what we used to have with method group to delegate conversions. I.e.
+there were scenarios where the conversion existed, but was erroneus. We decided to improve that by ensuring that,
+if conversion is erroneous, then it doesn't exist.
+
+Note, that with "Params Coolections" feature we will be running into a similar issue. It might be good to disallow
```
jnm2, line 933:
```suggestion
Note, that with "Params Collections" feature we will be running into a similar issue. It might be good to disallow
```

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -874,6 +874,67 @@ However, given the breadth and consistency brought by the new literal syntax, we
...
+    }
+
+    public static implicit operator char[](C1 x) => throw null;
+    IEnumerator<char> IEnumerable<char>.GetEnumerator() => throw null;
+    IEnumerator IEnumerable.GetEnumerator() => throw null;
+}
+```
+
+It looks like the situation is very similar to what we used to have with method group to delegate conversions. I.e.
+there were scenarios where the conversion existed, but was erroneus. We decided to improve that by ensuring that,
```
jnm2, line 930:
```suggestion
there were scenarios where the conversion existed, but was erroneous. We decided to improve that by ensuring that,
```

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -874,6 +874,67 @@ However, given the breadth and consistency brought by the new literal syntax, we
   Span<int> span = __result;
   ```
 
+### Should *collection expression conversion* require availability of a minimal set of APIs for construction?
```
jcouv, line 877: It may be good to include a link to the recent LDM discussion on that topic, for context: https://github.com/dotnet/csharplang/blob/main/meetings/2023/LDM-2023-10-02.md#collection-expressions

AlekseyTs: > It may be good to include a link to the recent LDM discussion on that topic

Added

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -874,6 +874,114 @@ However, given the breadth and consistency brought by the new literal syntax, we
...
+    public static void M2(params ushort[] x)
+    {
+    }
+
+    IEnumerator<char> IEnumerable<char>.GetEnumerator() => throw null;
+    IEnumerator IEnumerable.GetEnumerator() => throw null;
+}
+```
+
+It looks like the issue was somewhat discussed previously, see https://github.com/dotnet/csharplang/blob/main/meetings/2023/LDM-2023-10-02.md#collection-expressions.
```
333fred, line 965: I feel like, for the example you included at the start here, it's important to include another part of that same meeting:

> The second API, though, is not ambiguous by today's rules, as string is not constructable and therefore no conversion exists. By the new rules though, string is a valid target type for collection expressions, and will simply fail at a later stage. This makes the M2 API ambiguous. After some more consideration, we think that there's good reasoning for making strings constructable with collection expressions. List patterns already work on them, and we think that syntax could be useful for combining multiple string slices together. Support for this, though, is likely not going to make C# 12, and will be considered for future C# versions.

LDM intentionally considered that `string` and `char[]` overloads will be ambiguous with the current rules, and intentionally proceeded with the intention that we will at some point allow creation of strings with collection expressions.

AlekseyTs: > I feel like, for the example you included at the start here, it's important to include another part of that same meeting

I do not think it makes sense to copy the meeting notes here. Especially that the issue is not string specific and I do not want to make it appear this way. If during discussion we decide that strings are special, that would be fine. However, that is not my proposal, which I am still working on formulating.

PR #7768, opened 2023-12-14
Update "Better function member" section in params-collections.md

AlekseyTs:


333fred (approved):

proposals/params-collections.md
```diff
@@ -144,8 +142,105 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂,
...
+
+it is applicable to arrays and, therefore, performing the tie-break earlier will introduce a behavior change for existing scenarios.
+
+For example:
+``` C#
+class Program
+{
+    static void Main()
+    {
+        Test(1);
```
RikkiGibson, line 166: Would it also be useful to show which overload is chosen for `Test(1, [])`?

proposals/params-collections.md
```diff
@@ -144,8 +142,105 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂,
...
+    {
+        Test(x: 1, y: 2); // Ambiguous
+    }
+
+    static void Test(int x, params System.ReadOnlySpan<int> y) {}
+    static void Test(int y, params System.Span<int> x) {}
+}
+```
+
+It doesn't feel reasonable to "compare" collections that are built from different elements.
```
RikkiGibson, line 241: I think I can imagine this occurring when the methods have different number of parameters but both have a `params` parameter. Still, an example may help here.

AlekseyTs: > I think I can imagine this occurring when the methods have different number of parameters but both have a `params` parameter. Still, an example may help here.

The different number of parameters is handled by a different pre-existing rule.

PR #7784, opened 2023-12-22
Remove invalid collection literal examples

333fred:
There is no anonymous inline array syntax.


333fred:
@AlekseyTs please review

AlekseyTs (approved):

PR #7835, opened 2024-01-10
Interceptors open issues for C# 13

RikkiGibson:
Related to #7009


proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
+# Open issues for interceptors
+
```
jaredpar, line 2: Would add a blurb here about how this is about closing the gap on open issues that would get us from preview to productivity. Locking down the existing feature set vs. opening the door for new ones.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+## File path portability
+
+One of our design goals with interceptors was to have a particular form of "portability": essentially, we should be able to run the source generators for a project, commit the results to disk, then remove the source generators from the project and be able to build in a variety of environments. Note that this is different from *determinism*--we don't mind if the resulting assembly differs in things like layout, etc. in each build environment, but want it to be the case that if an interceptor declaration functions in one environment, then it also functions in a different environment, assuming all the sources and dependencies are fully available in each environment.
```
jaredpar, line 5:
```suggestion
One of our design goals with generators was to have a particular form of "portability": essentially, we should be able to run the source generators for a project, commit the results to disk, then remove the source generators from the project and be able to build in a variety of environments. Note that this is different from *determinism*--we don't mind if the resulting assembly differs in things like layout, etc. in each build environment, but want it to be the case that if an interceptor declaration functions in one environment, then it also functions in a different environment, assuming all the sources and dependencies are fully available in each environment.
```

This is a design goal of generators in general. Essentially the "you can always walk away from them" guarantee. Interceptors, due to the file path issues, are not fully meeting that guarantee.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+However, we currently require absolute paths for `InterceptsLocationAttribute`, with pathmap substitution performed if present. This doesn't work across local builds and CI because only CI will actually pass `/pathmap` during the build. The local build prefers to preserve local paths (e.g. in the PDB) so that the local versions of the sources can be discovered easily during debugging. The local and CI environments will have different answers for what they expect the paths to be, and the InterceptsLocation will have errors in one of the two environments.
```
jaredpar, line 7: Think we should be more general here. This just doesn't work when the code is built on a different path. That can happen in CI or just when another developer builds the code. Also the `/pathmap` part is not necessarily tied to CI: generally yes but not always.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+Having looked at several strategies to address here, including adjusting when the build system passes `/pathmap` to the compiler, it seems the most straightforward way is to permit relative paths. This is a strategy which is already proven out by `#line` directives.
```
jaredpar, line 9: I would end here with the reason we didn't permit relative paths in preview is that generated files don't have a known path. It's an undefined value. Hence relative paths didn't work as a solution.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+When a relative path appears in an `[InterceptsLocation]`, we will resolve it relative to the path of the containing file.
```
jaredpar, line 11: Need to add a note that we will normalize `/` and `/` during this process that way paths are portable across operating system builds. Effectively the resolution for relative paths here will match what happens in `#line` today. What works there will work in interceptors

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+Clearly this means there is a need for a generated file to know what path it "would be written out" to, while the text is being generated, before calling `AddSource` with a `hintName`. We may want to add a new public API like the following for this:
+
+```diff
+ namespace Microsoft.CodeAnalysis;
+
+ public readonly struct SourceProductionContext
+ {
+     public void AddSource(string hintName, string source);
++    public string GetFilePath(string hintName);
```
jaredpar, line 43: Think we need to add a bit more about what this does. In my mind this returns the path the file would be written out to _if_ `$(EmitCompilerGeneratedFiles)` is `true`. Also I think we need to firmly state that `SyntaxTree` produced from generated files will effectively, or actually, have this set as the path.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+ }
+```
+
+This solution will not work when the intercepted call and the interceptor method are in completely "disjunct" parts of the file system, for example, if one of the source files is included through a NuGet package which is outside the project folder, and the other is within the project folder. This limitation does not seem especially problematic in practice, but a solution in this case could be to use an absolute path and give up "portability".
```
jaredpar, line 47: I would add that while this is a problem, it is the exact same problem you would have if you wanted to use `#line` to refer to a file on another drive. Effectively yes there are issues here but these issues are now identical to the limitations we have around `#line` and those haven't proven to be a real world problem.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+ public static class CSharpExtensions
+ {
++    public static IMethodSymbol? GetInterceptorMethod(this SemanticModel model, InvocationExpressionSyntax invocation, CancellationToken cancellationToken = default);
+ }
+```
+
+Note that if we do something like this, there will be a deep need to discover interceptors, and incrementally update compilations containing interceptors, with as little binding work as possible. It will be non-tenable perf-wise if, for example, we had to bind all the attributes in a compilation in order to decide what the interceptor for a given invocation is. We may need to store some additional information in the declaration tree, which can be incrementally updated, in order to address this.
+
+# Future (possibly post-C# 13)
```
jaredpar, line 78:
```suggestion
## Future (possibly post-C# 13)
```

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Brittleness of line/column approaches
```
jaredpar, line 80:
```suggestion
### Brittleness of line/column approaches
```

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+Clearly this means there is a need for a generated file to know what path it "would be written out" to, while the text is being generated, before calling `AddSource` with a `hintName`. We may want to add a new public API like the following for this:
+
+```diff
+ namespace Microsoft.CodeAnalysis;
+
+ public readonly struct SourceProductionContext
+ {
+     public void AddSource(string hintName, string source);
++    public string GetFilePath(string hintName);
+ }
```
jaredpar, line 44: Was expecting to see another API here: one that made it easy for generators to get the file path to emit into the attribute. Given only the API in the doc so for every generator will have to write the same basic code: 

a. Get the file path of the `SyntaxNode` you're trying to intercept 
b. Get the file path of the generator 
c. Write the code to make (b) relative to (a) making sure to handle all edge cases like unix paths, different drive, etc ...

Rather than have every generator out there repeat this code I suggest we have an API where by generators can just get the right answer. Something along the lines of

```csharp
string GetInterceptsFilePath(SyntaxTree tree, string generatorHintName); 
```

Something in that rough shape.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+
+The desire for less-frequent source generation is at odds with the desire for interceptor info in IDE components, as interceptors are currently designed.
+
+For example, we want the ILLinker analyzer's results to be as useful as possible, and ideally to not include *spurious transient warnings* while the user is making edits. For example, if a user inserts some white spaces in a method body which contains intercepted calls, we don't want the ILLinker analyzer to report that some non-intercepted call is using an unsupported method. But this requires source generators to run. This *strongly* pushes the ILLinker analyzer to run *only as frequently as interceptor source generators do*.
+
+This would be an undesirable degradation in the user experience. Instead of being able to give users confidence that their mistakes "Native AOT-wise" will be caught very quickly by the tooling and shown to them as they type, they instead are looking at longer delays, or even not being told something is wrong until they save or hit build.
```
jaredpar, line 88: This section feels a bit off to me. This problem is not specific to interceptors or the ILLinker analyzers. It is a more general tension between when we run generators (any generator) and analyzers. If generators are run less frequently then analyzers that depend on their outputs will suffer. 

Think this may be better as a note of concern on the IDE discussion on strategies to delay running of generators

https://github.com/dotnet/roslyn/issues/67441

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,147 @@
...
+## Intercepting more member kinds
+
+Currently the interceptors feature only supports intercepting ordinary methods. Much like with partial methods and the upcoming partial properties, users want to be able to intercept usages of more kinds of members, such as properties and constructors.
+
+For all the member kinds we want to support, we will need to decide how usages of each member kind are denoted syntactically.
+
+For method invocations like `receiver.M(args)`, we use the location of the `M` token to intercept the call. Perhaps for property accesses like `receiver.P` we can use the `P` token. Things may get ambiguous if the property returns a delegate which is immediately invoked, e.g. `receiver.P()`. (TODO: do we support intercepting a delegate invocation? I think no, but check.)
+
+For constructors we may use either the `new` token or the type name. It's possible the type name would be more consistent in some sense, but it's not clear.
+
```
jaredpar, line 146: The other item that needs to be decided is what should be passed for the `name` portion for constructors, property get / set, etc ... The good news is that there is prior art we can draw from here. The `UnsafeAccessor` feature has the same problems and has already come up with a naming scheme that we could re-use. This would likely be appreciated by customers as some, like EF, use both `UnsafeAccessor` and interceptors in the same generator. Similarity here would be appreciated. 


https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-8.0

proposals/interceptors-issues-2024-01.md
RikkiGibson: Since source generators are already creating the `#line` directives with relative paths, even before they were run as Roslyn extensions, I'm wondering how they are doing that. Perhaps this is something generators already get wrong regularly, but isn't noticed as much, as the only impact is on debugging in a different environment than we generated the code in, rather than completely failing to build.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,153 @@
...
+Currently the interceptors feature only supports intercepting ordinary methods. Much like with partial methods and the upcoming partial properties, users want to be able to intercept usages of more kinds of members, such as properties and constructors.
+
+For all the member kinds we want to support, we will need to decide how usages of each member kind are denoted syntactically. Where possible, we should try to be consistent with [UnsafeAccessorAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-8.0), which also works by referencing members in attribute arguments.
+
+For method invocations like `receiver.M(args)`, we use the location of the `M` token to intercept the call. Perhaps for property accesses like `receiver.P` we can use the `P` token. Things may get ambiguous if the property returns a delegate which is immediately invoked, e.g. `receiver.P()`. (TODO: do we support intercepting a delegate invocation? I think no, but check.)
+
+For constructors we may use either the `new` token or the type name. It's possible the type name would be more consistent in some sense, but it's not clear.
```
svick, line 151: Note that type name wouldn't work for target-typed `new`, so I think always using `new` would be more consistent.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,153 @@
...
+
+However, we currently require absolute paths for `InterceptsLocationAttribute`, with pathmap substitution performed if present. This doesn't work between local builds on different paths, or across local builds and CI, because only CI will actually pass `/pathmap` during the build. The local build prefers to preserve local paths (e.g. in the PDB) so that the local versions of the sources can be discovered easily during debugging. The different environments will have different answers for what they expect the paths to be, and the InterceptsLocation will have errors in environments which differ from the "generation-time" environment.
+
+Having looked at several strategies to address here, including adjusting when the build system passes `/pathmap` to the compiler, it seems the most straightforward way is to permit relative paths. This is a strategy which is already proven out by `#line` directives.
+
+When a relative path appears in an `[InterceptsLocation]`, we will resolve it relative to the path of the containing file. The relative path used in the attribute must have `\` normalized to `/`, again, for portability.
```
jaredpar, line 15: The `#line` directive doesn't require `/` for portability so don't think we should have it for this attribute either. Would also just mention that paths in this attribute function exactly like paths in `#line`.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,153 @@
...
+
+// /project/obj/generated/file2.g.cs
+class Interceptors
+{
+    [InterceptsLocation("../../src/file1.cs", ...)] // resolves to "/project/src/file1.cs"
+    public static void Interceptor() { }
+}
+```
+
+This means that generator authors will be best advised to produce relative paths using e.g. `Path.GetRelativePath("/project/src/file1.cs", "/project/obj/generated/file2.g.cs")`.
```
jaredpar, line 37: `Path.GetRelativePath` is not a `netstandard2.0` method hence not available to source generators. This is part of the reason I suggested the `GetInterceptsFilePath` below. Reduce the burden on generator authors to produce the correct path, let's give them the easy button here.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,153 @@
...
+ namespace Microsoft.CodeAnalysis;
+
+ public readonly struct SourceProductionContext
+ {
+     public void AddSource(string hintName, string source);
++    public string GetFilePath(string hintName);
+ }
+```
+
+`GetFilePath()` will return the file path which the generated file *would be written to* if `$(EmitCompilerGeneratedFiles)` is true in the project. Additionally, the syntax tree for the file added by a subsequent call to `AddSource` using the same `hintName` will have a `FilePath` which exactly matches the result of `GetFilePath()`.
```
jaredpar, line 51: Feel like this is burying the lead. To me the most important item is that the `SyntaxTree` is going to be given a file path with these characteristics. But we talk about how to get the information before we talk about what it means. 

I would consider reorganize this paragraph and the one above the sample a bit.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,153 @@
...
+class Interceptors
+{
+    [InterceptsLocation("../../src/file1.cs", ...)] // resolves to "/project/src/file1.cs"
+    public static void Interceptor() { }
+}
+```
+
+This means that generator authors will be best advised to produce relative paths using e.g. `Path.GetRelativePath("/project/src/file1.cs", "/project/obj/generated/file2.g.cs")`.
+
+One reason why we haven't added relative path support already is that source generated files haven't had clearly defined paths. Now, there is a need for a generated file to know what path it "would be written out" to, while the text is being generated, before calling `AddSource` with a `hintName`. We may want to add a new public API like the following for this:
```
jaredpar, line 39: > Now, there is a need for a generated file to know what path it "would be written out" to, while the text is being generate

This feels a bit off to me. It's not that theer is a need to know what the path would be if it's written out, it's that there is a need to have certainty about the path. Essentially that we need to define it such that relative paths will work. To do this we're choosing to define it as the path it would be written out to. It may seem nit picky but I think it's an important distinction.

proposals/interceptors-issues-2024-01.md
RikkiGibson: Have adjusted the language here. Thanks

proposals/interceptors-issues-2024-01.md
RikkiGibson: Have adjusted the language here.

proposals/interceptors-issues-2024-01.md
RikkiGibson: I can ease it from a requirement to merely a recommendation.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,152 @@
...
+Ideally if an analyzer needs to know if an interceptor is being used, it could simply make a public API call to ask. This adds an extra decision that component authors need to make. For some use cases, such as rename-refactoring, it doesn't make sense to rename the interceptor method--instead the original method is what needs to be renamed. We think that in most cases, the original method is what should be analyzed, but it should still be noted as a pit that people may fall into.
+
+That public API could look something like the following:
+
+```diff
+ namespace Microsoft.CodeAnalysis;
+
+ public static class CSharpExtensions
+ {
++    public static IMethodSymbol? GetInterceptorMethod(this SemanticModel model, InvocationExpressionSyntax invocation, CancellationToken cancellationToken = default);
```
RikkiGibson, line 79: Some feedback I got includes:

1. What about IOperation? Do we have to get the .Syntax out of the operation node and cast it? Or can we just add a property to IInvocationOperation?
2. What about future additional member kinds? Will we add additional public API for additional kinds?

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,152 @@
...
+Since we want generated files to refer relatively to "user" files, we now need to define the paths of generated files. Also, for source generators to be able to insert the correct relative paths into the generated source, a source generator needs to know what the file path will be before the file is actually added to the compilation. We may want to add new public API like the following for this:
+
+```diff
+ namespace Microsoft.CodeAnalysis;
+
+ public readonly struct SourceProductionContext
+ {
+     public void AddSource(string hintName, string source);
++    public string GetFilePath(string hintName);
++    public string GetInterceptsFilePath(SyntaxNode intercepted, string hintName);
```
RikkiGibson, line 46: Perhaps the public API should just give "the stuff to stick in the InterceptsLocationAttribute". It's possible that if we move toward an "opaque identifier" approach, and the analyzer author is using such an API, that the compiler could "transparently" update the behavior of the API, in order to automatically start emitting new "versions" of the opaque representation, which might have benefits such as aiding incrementality/delayed generator runs in the IDE.

We would then have to also discuss the concept of versioning the opaque representation, probably with some kind of simple prefix, and the stipulation that older "versions" would work in newer compilers, but older compilers won't be able to decode newer "versions".

proposals/interceptors-issues-2024-01.md
captainsafia: Do we intend for generated files to have a known path moving forward? Further down, I read:

> When a relative path appears in an `[InterceptsLocation]`, we will resolve it relative to the path of the containing file,

Does "containing file" here refer to the generated file or the file containing the original invocation?

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,152 @@
...
+ }
+```
+
+`GetFilePath()` will return the absolute file path which the generated file *would be written to* if `$(EmitCompilerGeneratedFiles)` is true in the project. Additionally, the syntax tree for the file added by a subsequent call to `AddSource` using the same `hintName` will have a `FilePath` which exactly matches the result of `GetFilePath()`.
```
captainsafia, line 50: This is assuming that as a generator author my `AddSource` calls look like this:

```cs
context.AddSource(context.GetFilePath("GeneratedRouteHandlers.g.cs"), mySourceText)
```

If so, can we make it implicit so that I can just write:

 ```cs
context.AddSource("GeneratedRouteHandlers.g.cs", mySourceText)
```

Comment (RikkiGibson) [reply proposals/interceptors-issues-2024-01.md:50]: Yes my intent here matches what you're saying. AddSource will always take a hintName, not an absolute path. Also, it's feeling unclear to me now if `GetFilePath()` API is worth it, if we also have to add `GetInterceptsFilePath()` or similar API anyway.

File: proposals/interceptors-issues-2024-01.md [hunk 21]
```diff
@@ -0,0 +1,152 @@
...
+
+### Intercepting more member kinds
+
+Currently the interceptors feature only supports intercepting ordinary methods. Much like with partial methods and the upcoming partial properties, users want to be able to intercept usages of more kinds of members, such as properties and constructors.
+
+For all the member kinds we want to support, we will need to decide how usages of each member kind are denoted syntactically. Where possible, we should try to be consistent with [UnsafeAccessorAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-8.0), which also works by referencing members in attribute arguments.
+
+For method invocations like `receiver.M(args)`, we use the location of the `M` token to intercept the call. Perhaps for property accesses like `receiver.P` we can use location of the `P` token. For constructors should use the location of the `new` token.
+
+For member usages such as user-defined implicit conversions which do not have any specific corresponding syntax, we expect that interception will never be possible.
```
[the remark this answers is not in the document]
RikkiGibson: Yes, and it will also be useful to get a sense of how urgent it is to get it done in C# 13.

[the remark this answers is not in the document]
RikkiGibson: BTW, it feels like the real kicker here is not just that we want generics support (through something more sophisticated like a unification approach), but that we actually want the interceptor to be able to have a more specific signature than the original method. e.g. in this case where in order for type parameters to propagate through the system, the interceptor parameter needs to be `Func<TypeParam>` even though the original parameter type is `Delegate`.

proposals/interceptors-mvp.md
```diff
@@ -0,0 +1,32 @@
+Interceptors MVP:
```
jaredpar, line 1: Why is this a separate doc vs. being part of the other doc?

proposals/interceptors-mvp.md
```diff
@@ -0,0 +1,32 @@
...
+
+- generated SyntaxTree.FilePath
```
jaredpar, line 3:
```suggestion
- generated `SyntaxTree.FilePath`
```

Is this referring to the file path for generated `SyntaxTree` being the "what we would write" generated path.

proposals/interceptors-mvp.md
RikkiGibson: This basically was a way for me to quickly get some thoughts out, I didn't actually think about the fact that I was pushing to a branch which was already being PR'ed, I'm intending to "collapse" this doc back into the single doc and then open back up for review.

proposals/interceptors-issues-2024-01.md
```diff
@@ -0,0 +1,84 @@
...
+### Intercepting properties
+Currently the interceptors feature only supports intercepting ordinary methods. Much like with partial methods and the upcoming partial properties, users want to be able to intercept usages of more kinds of members, such as properties and constructors.
+
+  - UnsafeAccessor does not support properties directly. Instead, a given usage of UnsafeAccessor decorates a method which calls through to a single property accessor.
+  - Consider: What happens when we want to intercept `obj.Prop += a`? We can't choose a single method to replace usage of `Prop`.
+  - Conclusion: We should probably favor intercepting a property with an extension property. We are likely blocked here until we ship extensions.
+
+### Intercepting constructors and other member kinds
+
+For all the member kinds we want to support, we will need to decide how usages of each member kind are denoted syntactically. Where possible, we should try to be consistent with [UnsafeAccessorAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-8.0), which also works by referencing members in attribute arguments.
```
RikkiGibson, line 43:
```suggestion
For all the member kinds we want to support, we will need to decide how usages of each member kind are denoted syntactically. In terms of signature matching requirements, we should try to be consistent with [UnsafeAccessorAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute?view=net-8.0), which also works by referencing members in attribute arguments.
```

RikkiGibson:
@jaredpar I consider this ready to merge, please give a review and/or assign reviewers.

I'd like to send a follow up PR which essentially migrates the speclet from the Roslyn repo (linked in this document), into this same document, and renames it to match the other proposal docs (e.g. renaming it to interceptors.md, including the speclet content first, then as a subsequent section include these open questions.)

jaredpar (approved):

PR #7850, opened 2024-01-16
Update `Metadata` section in params-collections.md

AlekseyTs:


proposals/params-collections.md
```diff
@@ -498,7 +498,21 @@ the construction of parameter collections when APIs are invoked in their expande
 
 ### Metadata
 
-In metadata `params` parameters are marked with `System.ParamArrayAttribute` as `params` arrays are marked today.
+In metadata we could mark non-array `params` parameters with `System.ParamArrayAttribute`, as `params` arrays are marked today.
+However, it looks like we will be much safer to use a different attribute for non-array `params` parameters.
+For example, the current VB compiler will not be able to consume them decorated with `ParamArrayAttribute` neither in normal, nor in expanded form. Therefore, an addition of 'params' modifier is likely to be break VB consumers, and very likely consumers from other languages or tools.
```
jjonescz, line 503:
```suggestion
For example, the current VB compiler will not be able to consume them decorated with `ParamArrayAttribute` neither in normal, nor in expanded form. Therefore, an addition of 'params' modifier is likely to break VB consumers, and very likely consumers from other languages or tools.
```

333fred (approved):
We'll need to make sure we bring this API design to the BCL's API review process so we can add it to the BCL for .NET 9+. I don't expect them to have any feedback on the shape, but I've been surprised before.

PR #7865, opened 2024-01-23
Update ref struct interfaces based on LDM discussions

jaredpar:


proposals/ref-struct-interfaces.md
```diff
@@ -147,61 +151,113 @@ The new rules apply to the `heap` constraint:
...
+
 - Preventing default interface methods from applying to `ref struct`
-- API in `System.Reflection.Metadata` for encoding the `gpAcceptByRefLike` value.
+- API in `System.Reflection.Metadata` for encoding the `gpAcceptByRefLike` value
 - Support for generic parameters being a `ref struct`
 
-Most of this support is likely already in place. The general `ref struct` as generic parameter support is already implemented as described [here](https://github.com/dotnet/runtime/blob/main/docs/design/features/byreflike-generics.md). It's possible the DIM implementation already account for `ref struct`. But each of these items needs to be tracked down.
+Most of this support is likely already in place. The general `ref struct` as generic parameter support is already implemented as described [here][byref-like-generics]. It's possible the DIM implementation already account for `ref struct`. But each of these items needs to be tracked down.
+
+## API versioning
```
jaredpar, line 243: @stephentoub section I promised about how I think the anti-constraint impacts API versioning.

proposals/ref-struct-interfaces.md
```diff
@@ -147,61 +151,113 @@ The new rules apply to the `heap` constraint:
...
+
+The `allows ref struct` anti-constraint can be safely applied to a large number of generic definitions that do not have implementations. That means most delegates, interfaces and `abstract` methods can safely apply `allows ref struct` to their parameters. These are just API definitions without implementations and hence expanding the set of allowed types is only going to result in errors if they're used as type arguments where `ref struct` are not allowed.
```
stephentoub, line 245: So effectively there'd be zero downside to our marking the `T` in `Func<T>`, `IEnumerable<T>`, `IList<T>`, etc. as `allow ref struct`, yes?

proposals/ref-struct-interfaces.md
```diff
@@ -147,61 +151,113 @@ The new rules apply to the `heap` constraint:
...
+
+API owners can rely on a simple rule of "if it compiles, it's safe". The compiler will error on any unsafe uses of `allows ref struct`, just as it does for other `ref struct` uses.
+
+At the same time though there are versioning considerations API authors should consider. Essentially API owners should avoid adding `allows ref struct` to type parameters where the owning type / member may change in the future to be incompatible with `allows ref struct`. For example:
+
+- An `abstract` method which may later change to a `virtual` method
+- An `abstract` type which may later add implementations
+- An `interface` which may later add a default interface method
```
stephentoub, line 253: This last one is interesting. That means we're going to want to be thoughtful about which interfaces this applies to, e.g. we probably want it on `IEnumerable<T>` but probably not on `IList<T>`.

proposals/ref-struct-interfaces.md
```diff
@@ -147,61 +151,113 @@ The new rules apply to the `heap` constraint:
...
+## API versioning
+
+The `allows ref struct` anti-constraint can be safely applied to a large number of generic definitions that do not have implementations. That means most delegates, interfaces and `abstract` methods can safely apply `allows ref struct` to their parameters. These are just API definitions without implementations and hence expanding the set of allowed types is only going to result in errors if they're used as type arguments where `ref struct` are not allowed.
+
+API owners can rely on a simple rule of "if it compiles, it's safe". The compiler will error on any unsafe uses of `allows ref struct`, just as it does for other `ref struct` uses.
+
+At the same time though there are versioning considerations API authors should consider. Essentially API owners should avoid adding `allows ref struct` to type parameters where the owning type / member may change in the future to be incompatible with `allows ref struct`. For example:
+
+- An `abstract` method which may later change to a `virtual` method
+- An `abstract` type which may later add implementations
```
stephentoub, line 252: That basically means you can't use it with a public abstract type?

proposals/ref-struct-interfaces.md
jaredpar: Correct. The only real downside is you effectively give up on having DIM in those interfaces in the future but that seems like an unlikely future path anyways.

proposals/ref-struct-interfaces.md
jaredpar: You can but you need to be very certain in the future you won't want to use `T` in a way that violates `ref struct` rules. Practically though I think that means .NET team at least would avoid doing so.

proposals/ref-struct-interfaces.md
jaredpar: This is similar to the existing challenges posed by DIM. Specifically [cases][1] where adding members but only re-compiling _some_ of the DLLs will lead to a runtime exception. This is effectively another instance of that. Adding a DIM will cause a runtime exception if it's ever invoked on a `ref struct` that _inheritted_ that DIM. If the code containing the `ref struct` were re-compiled the compiler would issue an error forcing an implementation that avoided the exception. 

[1]: https://github.com/dotnet/csharplang/blob/main/meetings/2018/LDM-2018-10-17.md#diamond-inheritance

proposals/ref-struct-interfaces.md
KennethHoff: Isn't there a plan to add DIMs to IList literally this upcoming release? https://github.com/dotnet/runtime/pull/95830

stephentoub: yes, hence my comment https://github.com/dotnet/csharplang/pull/7865#discussion_r1464080189

RikkiGibson: I'm not sure I grasp why adding `IList<T> where allow T : ref struct` makes it a breaking change to add a DIM to `IList<T>`. It seems like adding a DIM will be a breaking change for ref structs implementing the interface, regardless of whether the type parameter has this constraint.

```cs
interface I<T>
{
    void M()
    {
        Console.Write(1);
    }
}

class C : I<int> { } // ok
ref struct RS : I<int> { } // error: must implement I<int>.M()
```

jaredpar: That's a good point. There are essentially two facets to how DIM can create breaking changes for `ref struct`. 

1. The problem where a `ref struct` implements the interface and it leads to runtime behavior because `ref stuct` can't inherit the DIM. This doesn't require any use of `allows ref struct`
2. When a DIM uses `T` in a way that is incompatible with `allows ref struct`. This is similar to the future proofing concerns around `abstract` methods. 

Overall (2) feels more like a theoretical concern vs. a practical one. A DIM that uses `T` in `allows ref struct` would be hard pressed to create a lifetime issue because all such `T` are going to be _safe-to-return_. The `this` is _safe-to-return_ hence all values sourced from it are going to be too. 

The only actual concern I can think of is for (2) you fashion some API that takes and returns `ref` values of type `T` **and** there is a `ref struct` that implements the interface and sources `T` from a field. 

```csharp
interface I1<T> where T : allows ref struct
{
  ref T M(ref T p);
}

ref struct S : I1<Span<char> {
  Span<char> span;
  public Span<char> M(ref Span<char> span) { ... } 
}
```

If you toyed around with the lifetimes there a bit I'm sure you could find a way  to make it difficult to implement on the `ref struct`. Feels very corner case though.

RikkiGibson: > It seems like adding a DIM will be a breaking change for ref structs implementing the interface

BTW, I wanted to kind of.. re-emphasize this. DIM, as far as I know, was supposed to be a feature which permitted library authors to evolve interfaces without imposing breaking changes on their consumers. It feels like this feature actually takes that away. Library authors can no longer rely on it. We should be very deliberate in deciding whether that is acceptable, or, otherwise, we should consider devising some way to declare a DIM such that it can be called on a type parameter constrained to ref struct, and library authors who are concerned about compatibility can use that.

proposals/ref-struct-interfaces.md
hamarb123: It's a shame that DIMs box on usage of `this`, I usually avoid them for that reason alone, other than with static DIMs obviously which would presumably(?) be fine with `ref struct`s.

proposals/ref-struct-interfaces.md
```diff
@@ -30,7 +31,7 @@ File f = ...;
 IDisposable d = f;
 ```
 
-The ability to implement interfaces is only useful when combined with the ability for `ref struct` to participate in generic arguments (as [laid out later](#ref-struct-generic)).
+The ability to implement interfaces is only useful when combined with the ability for `ref struct` to participate in generic arguments (as [laid out later][ref-struct-generics]).
 
 To allow for interfaces to cover the full expressiveness of a `ref struct`, the language will allow `[UnscopedRef]` to appear on interface methods and properties. When a `ref struct` member implements an interface member with a `[UnscopedRef]` attribute, that `ref struct` member must also be decorated with `[UnscopedRef]`. The attribute is ignored when a `class` or non-ref `struct` implements the interface.
```
hamarb123, line 36: I'm confused what is happening with `[UnscopedRef]` here, why would non-ref `struct`s ignore it? Does it have the same meaning as normal? Is there any reason why a ref struct implementing member would *have* to have `[UnscopedRef]`?

For example, if I write:
```csharp
interface I1
{
    int FieldAccessor { get; set; }
    [UnscopedRef] ref int RefFieldAccessor { get; } //I assume this is supposed to mean the current meaning - which in this case is that the result may be a memory location on the struct passed by `this` (if a struct obviously)
}

class C1 : I1
{
    int field;
    int FieldAccessor { get => field; set => field = value; }
    ref int RefFieldAccessor => ref field; //ignore [UnscopedRef] as expected (lifetime is always long enough)
}

ref struct R1 : I1
{
    int field;
    int FieldAccessor { get => field; set => field = value; }
    [UnscopedRef] ref int RefFieldAccessor => ref field; //require [UnscopedRef] as expected since it returns a ref to a field, allowed since definition has [UnscopedRef]
}

ref struct R2 : I1
{
    C1 instance;
    int FieldAccessor { get => instance.FieldAccessor; set => instance.FieldAccessor = value; }
    /*[UnscopedRef] - should be optional*/ ref int RefFieldAccessor => ref instance.RefFieldAccessor; //I see no reason for this to require [UnscopedRef] on implementation, since it just means the lifetime is longer than was required by [UnscopedRef]
}

ref struct R2 : I1
{
    ref int fieldRef;
    int FieldAccessor { get => fieldRef; set => fieldRef = value; }
    /*[UnscopedRef]*/ ref int RefFieldAccessor => ref fieldRef; //I see no reason for this to require [UnscopedRef] on implementation, since it just means the lifetime is longer than was required by [UnscopedRef]
}

struct S1 : I1
{
    int field;
    int FieldAccessor { get => field; set => field = value; }
    //why would I not also be able to define this?
    [UnscopedRef] ref int RefFieldAccessor => ref field; //require [UnscopedRef] as expected since it returns a ref to a field, allowed since definition has [UnscopedRef]
}
```

i.e., what I'm suggesting is that to me it would make sense for the rule to be this: interface members can be marked as `[UnscopedRef]`, and interface member implementations in `struct`s or `ref struct`s can only be marked as `[UnscopedRef]` if their definition was also marked `[UnscopedRef]` (and never in `class`es obviously). (simplified to only deal with `this` parameter)

jaredpar: The `[UnscopdRef]` annotation is necessary for consumption. Consider as an example: 

```csharp
interface I1 {
    [UnscopedRef] ref P1 { get; } 
    ref P2 { get; } 
}

T M<T>(T p) where T : struct, allows ref struct, I1 {
    // Illegal
    return p.P1;

    // Legal
    return p.P2;
}
```

Just as `[UnscopedRef]` is necessary to distinguish between these two cases when interacting directly with a `ref struct`, it's needed when through an interface that abstracts over a `ref struct`

proposals/ref-struct-interfaces.md
jaredpar: >  It feels like this feature actually takes that away

How does it take it away though? I've already pointed out that there are existing cases where adding a DIM member leads to runtime exceptions. This is one of the hesitations runtime has in adding DIM to existing API. So this is already very much an existing problem. All this proposal does is add one more case of it. 

>  we should consider devising some way to declare a DIM such that it can be called on a type parameter constrained to ref struct, and library authors who are concerned about compatibility can use that.

The issue is not a language one, it is a runtime one. The act of invoking a DIM forces a box. This was called out during the implementation of DIM as a potential future limitation if `ref struct` were allowed to implement interfaces. The decision was made that it was a reasonable trade off. If in the future the runtime wants to go back on that then we could evaluate the intersection of DIM and `ref struct`. 

I think that is unlikely though. As mentioned there are existing limitations of DIM and those are more prevalent than the `ref struct` one. Given that we haven't felt the need to revisit those limitations I think it's unlikely runtime will revisit this one.

proposals/ref-struct-interfaces.md
jaredpar: I don't think there are any issues with static DIMs

proposals/ref-struct-interfaces.md
RikkiGibson: > I've already pointed out that there are existing cases where adding a DIM member leads to runtime exceptions.

I wasn't able to find the description of these existing cases, could you please point me to it?

jaredpar: https://github.com/dotnet/csharplang/blob/main/meetings/2018/LDM-2018-10-17.md#diamond-inheritance

proposals/ref-struct-interfaces.md
```diff
@@ -35,6 +35,34 @@ The ability to implement interfaces is only useful when combined with the abilit
...
+```csharp
+interface I1
+{
+    [UnscopedRef]
+    ref int Prop { get; }
+}
+
+struct S1 : I1
+{
+    // Error: missing [UnscopedRef]
```
hamarb123, line 47: Please read the whole comment if you're not convinced, I explain my logic from multiple angles, including one based on already existing scoped rules in C# (linked).

I don't see why this wouldn't be allowed, consider that I can always safely work around this arbitrary limitation in the case of properties:
```csharp
struct S1 : I1
{
    ref int Prop { get; }
    [UnscopedRef] ref int I1.Prop => ref Prop;
}
```
By using an explicit interface implementation.

To me, this indicates that the implementation should not require `[UnscopedRef]` to be on the member, even if it's on the interface declaration, since a non-`[UnscopedRef]` implementation is always a valid `[UnscopedRef]` implementation.

This is because, `[UnscopedRef]` on a struct member (not the parameter, the member itself) simply indicates that the return value may include a ref which overlaps with `ref this` (or via `RefStruct&` with applicable annotations), it does not necessitate that it does. That's also why you cannot have an `[UnscopedRef]` member in a class, since it can never return a ref which overlaps with `ref this` (unless it's passed in separately obviously).

In fact, if you pretend that we pass this as an actual parameter to structs, a normal method `ref int X() => ...;` is effectively `ref int X(scoped ref ThisType this) => ...;`, whereas an unscoped-ref method `[UnscopedRef] ref int Y() => ...;` is effectively `ref int Y(ref ThisType this) => ...;`. If you look at the [delegate conversion rules](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements#parameter-scope-variance) around `scoped`, you'll notice that you are allowed to add `scoped` to a `ref` parameter in an override. Therefore, if I'm expected to implement `ref int Y(ref ThisType this) => ...;` (`[UnscopedRef]`), then I can implement it with `ref int Y(scoped ref ThisType this) => ...;` instead if I want.

The first step of my logic is explained [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements#provide-unscoped) (that is, how it impacts the implicit `ref this` parameter). And the second is explained on the delegate conversion & overriding rules linked above.

> When applying `[UnscopedRef]` to an instance method of a `struct` it has the impact of modifying the implicit `this` parameter. This means `this` acts as an unannotated `ref` of the same type.

I have come up with all permutations of parameters to show that you can always implement a `[UnscopedRef]` member with a non-`UnscopedRef` member, no matter what parameters a method has, and as shown earlier it also works for properties: [link](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgBgAJDSA6AEWwEMBzAOwgGcM2MHyYBhCABMYAQW7tcATz7Y+Abiq0GzAEoBXboIC2MMREMAHPDCgBlawDchMNVQ0BmOgKi6wGOjYpKAG8qOjC6WAAzOmwDOmlSOgBeAD4ImGiAVW4+dkiTADldXFxtDIAeWIwUgAoASnVKcLoAbWy+SHMYCTLIgF106Kr44mS0qPjSRtDwieGAIUDmmpnmsLm49jRPTu7BmLjgbdiDvzBtjoguiQPTugk0VbWIXT9hmG22nN2ejIGXt5xSLbGzmdjcSoGNKcC4/fxgiFVNIAC0eTTW6wy8PBkOqMVhVz2E1BOKRdAAVtsAdjEVC6ABrT7tH69f6vGm4tK4J51J4hdEYmDJe4ZdjFDCNDFhaJJEWRMW4CVPZr04VSeXiyVS3Bq0Wa5XhQgAdn27TyhWKpQqSPqWrCAF9XALWszCb9+vsFsQnitnc0Nn4tjs3Z7DsduHdzsHrrdhg8DWFqe8md83ay6EmgSCEZy6DDo3sSbS8aiE/si7nsASY8Sc2TKRn2RWyYyXanrunqc26dznXUxvtFjUJkGJkdblGTg9G34PjO6MC89tUZ7thMG9TW9TcA0nc1u3jREtwr6paHA9WiVjhuOTsMo5cY3e4vG/eFM7OU4/up32cNFwe0KXjcgF0KWb6YtEoFVgWNy1qSdIbk2dZ0q2Xzfu6bJ+KBvbNLyzr8meQqyuqCpKhBC66hqip2msqokXqNFljqDHUeRZ7GnKZG0Y6lBPOhLJ/ByZKiN6zqnlKAZ0EGGHnnQt4RvewGxi+aJnh+dBzgJaZCRpAEoXi+ayaB4FnvBxZpDBsnmbmSHYQZaRoa6Ha6chCF4rh4T9qkdBHsOWKjliCl0JOEbTtSc7Uou+YrnMa5YnZDJUuyO7TM6+i5Pk5YOXQLDHmEEkYlJMlwlJwVKbBKl+K+6l/nEWnOT+rmAn4+nuUBlUmWpklYtByk2fWyX2e1SVthhv7DRZdCeWE+HNIRUrEVx+oUTKy1MRR9Hrex2pUdxZacRMABUNSAUddQ0Dxe7hNpLkegNdIsGJyxlsV/XXmGVUhcpz7Vd1GIaQ17ZNR6enZiNRlwl1r29TlVmlbDI2JaBTnA5h844TyA4THl/nRIF0TBaF9xDZppPRcu8VDFTFKk1uKW7pQvFAA=).

But you would definitely not want the implementation of a non-`[UnscopedRef]` interface member to have `[UnscopedRef]`, since that would break the contract (as is the case currently).

jaredpar: > I have come up with all permutations of parameters to show that you can always implement ...

The permutations example isn't persuasive to me. Had cases in the past where did permutation arguments but found later that we missed permutations / interactions with other features. Lifetime interactions are very complex. An argument on the underlying lifetime rules would be more persuasive. 

> Please read the whole comment if you're not convinced ... 

One point you haven't addressed is what is the value in removing this requirement? The value in keeping is that it's a simple rule, no need to prove the absence of it is safe, etc ... What is the value in going through that exercise?

hamarb123: > The permutations example isn't persuasive to me. Had cases in the past where did permutation arguments but found later that we missed permutations / interactions with other features. Lifetime interactions are very complex. An argument on the underlying lifetime rules would be more persuasive.

Which I also presented (repeated more clearly below).

It would effectively be a change to existing rules to disallow it. The existing rules explain how to interpret `[UnscopedRef]` members on structs: 

From https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements#provide-unscoped

> When applying `[UnscopedRef]` to an instance method of a `struct` it has the impact of modifying the implicit `this` parameter. This means `this` acts as an unannotated `ref` of the same type.

By default (without `[UnscopedRef]`), `this` is passed as `scoped ref this`, but with `[UnscopedRef]` on the method, it's passed as `ref this`.

From https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements#parameter-scope-variance

> The `scoped` modifier and `[UnscopedRef]` attribute (see below) on parameters also impacts our object overriding, interface implementation and `delegate` conversion rules. The signature for an override, interface implementation or `delegate` conversion can:
- Add `scoped` to a `ref` or `in` parameter
- ...

Note these rules apply to interface implementation, which is what we're discussing here. It says we can add `scoped` to a `ref` parameter, meaning that we can implement `ref int M(ref this)` (`[UnscopedRef] ref int M()`) with `ref int M(scoped ref this)` (`ref int M()`). To change this would effectively mean changing the existing rules on lifetimes.

Furthermore, to prove its safety more directly, since `ref int M()` is effecively `ref int M(scoped ref this)`, and `[UnscopedRef] ref int M()` is effecively `ref int M(ref this)` - the `scoped` in `ref int M(scoped ref this)` means that `ref this` will not ever escape; this means that it is compatible with `ref int M(ref this)` where `ref this` is allowed to escape. i.e., `[UnscopedRef] ref int M()` requires that `ref this` is considered to be potentially stored in any `ref` after the method exits, whereas `ref int M()` guarantees that it won't be - if something won't ever happen (`ref this` escaping - no `[UnscopedRef]`), it is safe to treat it as a possibility (`[UnscopedRef]`), it just means that there are other safe operations which wouldn't be allowed.

> One point you haven't addressed is what is the value in removing this requirement? The value in keeping is that it's a simple rule, no need to prove the absence of it is safe, etc ... What is the value in going through that exercise?

Consider that I have 2 implementations for something, which are exposed through structs. One has the fields directly on itself, whereas the other one has some of the fields on helper (class) fields. If I want to expose an interface for retrieving one of these fields by reference, then I need to expose it as `[UnscopedRef]`, but only the implementation which has the fields on the struct directly needs to actually use `[UnscopedRef]` on its public member, the other one can just do it as a normal ref.

To mark the other one as `[UnscopedRef]` would potentially (in my case, actually) unnecessarily reduce the possible ways it could be used. Having to define an explicit implementation is something I can do (and will if I have to), but it also should be unnecessary since it's safe as I showed above with the current rules.

This is not an academic example, this is something I have in (non-exposed) code I actually have already written (just without the safety of being able to use `[UnscopedRef]` on the interface member definition currently). I would like to gain that safety for obvious reasons - no need to be unsafe when there's not supposed to actually be anything unsafe about the code.

e.g., what I do today:
```csharp
interface I1
{
    ref int PotentialField { get; }
}

struct S1 : I1
{
    int field;
    [UnscopedRef] public ref int PotentialField => ref field;
    unsafe ref int I1.PotentialField => ref PotentialField; //ignore warning
}

struct S2 : I1
{
    SomeClass field;
    public int PotentialField => ref field.fieldOfTypeInt; //can implement it here
}
```

Ideally I could just write:
```csharp
interface I1
{
    [UnscopedRef] ref int PotentialField { get; }
}

struct S1 : I1
{
    int field;
    [UnscopedRef] public ref int PotentialField => ref field;
}

struct S2 : I1
{
    SomeClass field;
    public int PotentialField => ref field.fieldOfTypeInt; //can implement it here
}
```

hamarb123: > Had cases in the past where did permutation arguments but found later that we missed permutations / interactions with other features

Out of curiosity - what was an example of this :)

jaredpar: Thought about it some more this morning and decided I agreed that it's safe. I added a rationale at the bottom of the spec. Essentially, though the approach I took is make `this` explicit parameter and verify that _effectively adding / removing `scoped` in the ways that `[UnscopedRef]` work here is already legal for parameters so this is already established.

jaredpar: > Out of curiosity - what was an example of this :)

There were a number of them. Probably the one I most remember, and was missed in the initial ref reassignment implementation in 7.3, is the following: 

```csharp
Span<int> stackSpan = stackalloc int[42];
Span<int> heapSpan = new int[42];

ref Span<int> refStackSpan = ref stackSpan;
ref Span<int> refHeapSpan = ref heapSpan;

// Must be an error
refStackSpan = ref refHeapSpan;

// If it's allowed then you can do this which writes a stack value to 
// heapSpan ... oops
refStackSpan = new stackalloc int[13];
```

This is a case where we went through the permutations and decided it was legal. While we succeeded in enumerating the permutations, we failed in understanding their implications. We errored in saying that it's always safe to make the lifetime smaller as _seems_ to occur here. In reality it's only safe to make the lifetime smaller in some cases but not others. 

The [annotations][1] part of the spec does a cleaner job of expressing this type of problem as lifetimes are modeled with variance. In most cases lifetime is covariant (can be made more specific / smaller). When visualizing `ref` as a type where lifetime is a parameter (as it is on other types) it becomes clear that lifetime there must be invariant. Otherwise it's easy to violate lifetimes. A few other cases like this were found when we modeled in this fashion. 

[1]: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#annotations

proposals/ref-struct-interfaces.md
RikkiGibson: Thanks, I think I'm satisfied that we're not entering new territory with this interaction between ref structs and DIMs, it just means my dreams were never reality 😄

PR #7906, opened 2024-02-03
Add proposal for overload resolution priority

333fred:
This is my replacement proposal for what I originally wrote up in https://github.com/dotnet/csharplang/pull/7707.

### DIFF
--- a/proposals/overload-resolution-priority.md
+++ b/proposals/overload-resolution-priority.md
@@ -26,28 +26,17 @@ ambiguity, so that they can evolve their API surface areas and steer users towar
 We define a new concept, ***overload_resolution_priority***, which is used during the process of resolving a method group. ***overload_resolution_priority*** is a 32-bit integer
 value. All methods have an ***overload_resolution_priority*** of 0 by default, and this can be changed by applying
 [`OverloadResolutionPriorityAttribute`](#systemruntimecompilerservicesoverloadresolutionpriorityattribute) to a method. We update section 
-[§12.8.9.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#12892-method-invocations) of the C# specification as
+[§12.6.4.1](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#12641-general) of the C# specification as
 follows (change in **bold**):
 
-> For a method invocation, the *primary_expression* of the *invocation_expression* shall be a method group. The method group identifies the one method to invoke or the set of overloaded methods from which to choose a specific method to invoke. In the latter case, determination of the specific method to invoke is based on the context provided by the types of the arguments in the *argument_list*.
->
-> The binding-time processing of a method invocation of the form `M(A)`, where `M` is a method group (possibly including a *type_argument_list*), and `A` is an optional *argument_list*, consists of the following steps:
+> Once the candidate function members and the argument list have been identified, the selection of the best function member is the same in all cases:
 > 
-> - The set of candidate methods for the method invocation is constructed. For each method `F` associated with the method group `M`:
->   - If `F` is non-generic, `F` is a candidate when:
->     - `M` has no type argument list, and
->     - `F` is applicable with respect to `A` ([§12.6.4.2](expressions.md#12642-applicable-function-member)).
->   - If `F` is generic and `M` has no type argument list, `F` is a candidate when:
->     - Type inference ([§12.6.3](expressions.md#1263-type-inference)) succeeds, inferring a list of type arguments for the call, and
->     - Once the inferred type arguments are substituted for the corresponding method type parameters, all constructed types in the parameter list of `F` satisfy their constraints ([§8.4.5](types.md#845-satisfying-constraints)), and the parameter list of `F` is applicable with respect to `A` ([§12.6.4.2](expressions.md#12642-applicable-function-member))
->   - If `F` is generic and `M` includes a type argument list, `F` is a candidate when:
->     - `F` has the same number of method type parameters as were supplied in the type argument list, and
->     - Once the type arguments are substituted for the corresponding method type parameters, all constructed types in the parameter list of `F` satisfy their constraints ([§8.4.5](types.md#845-satisfying-constraints)), and the parameter list of `F` is applicable with respect to `A` ([§12.6.4.2](expressions.md#12642-applicable-function-member)).
-> - The set of candidate methods is reduced to contain only methods from the most derived types: For each method `C.F` in the set, where `C` is the type in which the method `F` is declared, all methods declared in a base type of `C` are removed from the set. Furthermore, if `C` is a class type other than `object`, all methods declared in an interface type are removed from the set.  
->   > *Note*: This latter rule only has an effect when the method group was the result of a member lookup on a type parameter having an effective base class other than `object` and a non-empty effective interface set. *end note*
-> - If the resulting set of candidate methods is empty, then further processing along the following steps are abandoned, and instead an attempt is made to process the invocation as an extension method invocation ([§12.8.9.3](expressions.md#12893-extension-method-invocations)). If this fails, then no applicable methods exist, and a binding-time error occurs.
-> - **The resulting set of candidate methods is grouped by ***overload_resolution_priority***. All candidates not in the highest (by standard integer comparison) ***overload_resolution_priority*** group are removed from the set.**
-> - The best method of the set of candidate methods is identified using the overload resolution rules of [§12.6.4](expressions.md#1264-overload-resolution). If a single best method cannot be identified, the method invocation is ambiguous, and a binding-time error occurs. When performing overload resolution, the parameters of a generic method are considered after substituting the type arguments (supplied or inferred) for the corresponding method type parameters.
+> - First, the set of candidate function members is reduced to those function members that are applicable with respect to the given argument list ([§12.6.4.2](expressions.md#12642-applicable-function-member)). If this reduced set is empty, a compile-time error occurs.
+> - **Then, the reduced set of candidate members is grouped by declaring type. Within each group:**
+>     - **Candidate function members are ordered by ***overload_resolution_priority***.
+>     - **All members that have a lower ***overload_resolution_priority*** than the highest found within its declaring type group are removed.**
+> - **The reduced groups are then recombined into the final set of applicable candidate function members.**
+> - Then, the best function member from the set of applicable candidate function members is located. If the set contains only one function member, then that function member is the best function member. Otherwise, the best function member is the one function member that is better than all other function members with respect to the given argument list, provided that each function member is compared to all other function members using the rules in [§12.6.4.3](expressions.md#12643-better-function-member). If there is not exactly one function member that is better than all other function members, then the function member invocation is ambiguous and a binding-time error occurs.
 
 As an example, this feature would cause the following code snippet to print "Span", rather than "Array":
 
@@ -98,7 +87,7 @@ We introduce the following attribute to the BCL:
 ```cs
 namespace System.Runtime.CompilerServices;
 
-[AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+[AttributeUsage(AttributeTargets.Method | AttributeTargets.Constructor | AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
 public sealed class OverloadResolutionPriorityAttribute(int priority)
 {
     public int Priority => priority;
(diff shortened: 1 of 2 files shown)


yaakov-h:
The file name looks wrong to me, and as such is annoyingly not triggering GitHub’s markdown renderer.

TahirAhmadov:
> Should the attribute be inherited? If not, what is the priority of the overriding member?
> If the attribute is specified on a virtual member, should an override of that member be required to repeat the attribute?

With `Obsolete`, you have to specify the `Obsolete` on overriding members to avoid the warning; I think it should be a similar approach here:
- If overriding member doesn't have the attribute, it defaults to 0, and shows a warning;
- If overriding member has the attribute, then the specified value is used.

> Should the attribute support constructors?

Yes.

Re. member types - I second @KalleOlaviNiemitalo 's concerns - the attribute should be applicable to methods, constructors, and indexer properties, but not to accessors (`get`, `set`, `add`, `remove`).

TonyValenti:
I'm not a fan of it being completely open ended either the numbers. I think there should be some predefined values/enums that help. Ie:
OverloadResolutionPriority.Default
OverloadResolutionPriority.PreferSpans
OverloadResolutionPriority.BinaryCompatibilityOnly

proposals/proposals
```diff
@@ -0,0 +1,214 @@
...
+[AttributeUsage(AttributeTargets.Method AllowMultiple = false, Inherited = false)]
+public sealed class OverloadResolutionPriorityAttribute(int priority)
+{
+    public int Priority => priority;
+}
+```
+
+**Open question**: Should the attribute be inherited? If not, what is the priority of the overriding member?  
+**Open question**: If the attribute is specified on a virtual member, should an override of that member be required to repeat the attribute?  
+**Open question**: Should the attribute support constructors?
```
hamarb123, line 110: My opinions are:
- For inheritance, it should just default to 0. It shouldn't warn if one isn't specified, but it would be fine to have a suggestion. Rationale: The specific value is only relevant within a specific class, so it doesn't make much sense to me to require, or even warn, on missing value, since the chosen values are only relevant within a specific class.
- For constructors: it would be good if it supported constructors, and indexers also. It should potentially likewise be disallowed on property getters & setters, event `add` and `remove` also.

proposals/proposals
```diff
@@ -0,0 +1,214 @@
...
+
+The effect of this change is that, like pruning for most-derived types, we add a final pruning for overload resolution priority. Because this pruning occurs at the very end of the overload resolution
+process, it does mean that a base type cannot make its members higher-priority than any derived type. This is intentional, and prevents an arms-race from occuring where a base type may try to always
+be better than a derived type. For example:
+
+```cs
+using System.Runtime.CompilerServices;
+
+var d = new Derived();
+d.M([1, 2, 3]); // Prints "Derived", because members from Base are not considered due to finding an applicable member in Derived
```
hamarb123, line 78: I understand why this is done, but it is a shame that there isn't a way to allow this to occur.

For example, we could support the following:
```csharp

var d = new Derived();
d.M([1, 2, 3]); // Prints "Base", because member in Derived is treated as if it were defined in Base, therefore the ROS overload wins

class Base
{
    [OverloadResolutionPriority(1)]
    public void M(ReadOnlySpan<int> s) => Console.WriteLine("Base");
}

class Derived : Base
{
    [OverloadResolutionPriority(0, TreatAsBase = typeof(Base))]
    public void M(int[] a) => Console.WriteLine("Derived");
}
```

I would find this option useful (I've run into this issue a few times), but I understand that it might be considered unnecessarily complicated.

proposals/proposals
```diff
@@ -0,0 +1,214 @@
...
+Negative numbers are allowed to be used, and can be used to mark a specific overload as worse than all other default overloads.
+
+### `System.Runtime.CompilerServices.OverloadResolutionPriorityAttribute`
+
+We introduce the following attribute to the BCL:
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+[AttributeUsage(AttributeTargets.Method AllowMultiple = false, Inherited = false)]
```
hamarb123, line 101: minor typo :)
```diff
-[AttributeUsage(AttributeTargets.Method AllowMultiple = false, Inherited = false)]
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
```

hamarb123 (approved):
Just added my thoughts on a few aspects. Mostly looks good to me :)

proposals/proposals
333fred: > since all derived overloads are considered first

This is actually not how overload resolution works.

hamarb123: I was referring to:

> The set of candidate methods is reduced to contain only methods from the most derived types

Have I misinterpreted it? If so, what does it mean, so I can understand how it works?

I interpreted it as: once you have a list of methods you could call, only consider the definitions in the most derived type that defines at least one of them (i.e., don't consider a virtual override in this list, since it's not a 'definition').

333fred: > I interpreted it as: once you have a list of methods you could call, only consider the definitions in the most derived type that defines at least one of them (i.e., don't consider a virtual override in this list, since it's not a 'definition').

You're almost there. The key thing is that overrides are _not_ definitions in the most derived type. You can see that in [this](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAIwCwAUBQG4TIAEGdAvHTAKYDudAIm8lFTYYAFAEoA3BVgI6UZnWKTyGAHQBZYVAl0A9DroAFfjAQBnOgCIANgHsYAcwsUK+AMx0IwUwmQQAxjJEdABCEKZsdADeFHSxdG4eXj7+gQAsdBrSshIUAL7O5AlBvPyCjCAhYRHR5HHx7jaCyPwYEfjpmSbZMXE1dXX4xACcwhbSFjm1cflTsT2xRR3Ctg50VqLzUZsDw6MrjpN1M7lAA==) example:

```cs
using System;

var d = new Derived();
int i = 1;
d.M(i); // Prints "long"

public abstract class Base {
    public abstract void M(int i);
}

public class Derived : Base {
    public override void M(int i)
    {
        Console.WriteLine("int");
    }
    
    public void M(long l)
    {
        Console.WriteLine("long");
    }
}
```

proposals/proposals
```diff
@@ -0,0 +1,214 @@
...
+    public int Priority => priority;
+}
+```
+
+**Open question**: Should the attribute be inherited? If not, what is the priority of the overriding member?  
+**Open question**: If the attribute is specified on a virtual member, should an override of that member be required to repeat the attribute?  
+**Open question**: Should the attribute support constructors?
+
+All methods in C# have a default ***overload_resolution_priority*** of 0, unless they are attributed with `OverloadResolutionPriorityAttribute`. If they are
+attributed with that attribute, then their ***overload_resolution_priority*** is the integer value provided to the first argument of the attribute.
```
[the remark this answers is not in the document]
333fred: I expect we will update the VB and F# compilers. Other languages would need to adjust as well, but that's the same for any such feature.

proposals/overload-resolution-priority.md
```diff
@@ -0,0 +1,214 @@
...
+>   > *Note*: This latter rule only has an effect when the method group was the result of a member lookup on a type parameter having an effective base class other than `object` and a non-empty effective interface set. *end note*
+> - If the resulting set of candidate methods is empty, then further processing along the following steps are abandoned, and instead an attempt is made to process the invocation as an extension method invocation ([§12.8.9.3](expressions.md#12893-extension-method-invocations)). If this fails, then no applicable methods exist, and a binding-time error occurs.
+> - **The resulting set of candidate methods is grouped by ***overload_resolution_priority***. All candidates not in the highest (by standard integer comparison) ***overload_resolution_priority*** group are removed from the set.**
```
AlekseyTs, line 49: First, I think that a better place for the change is actually in Overload Resolution part of the language spec. In [12.6.4.1 General](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#12641-general). Perhaps a new bullet point right after:
>- First, the set of candidate function members is reduced to those function members that are applicable with respect to the given argument list ([§12.6.4.2](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#12642-applicable-function-member)). If this reduced set is empty, a compile-time error occurs.



Second, I think the proposed wording of the bullet point doesn't accurately reflect the intent of filtering per declaring type: "adjust the relative priority of overloads within a single type". 


It should probably say some thing like: "Group candidates per declaring type. Determine the highest overload_resolution_priority in each group. Candidates with lower overload_resolution_priority from each group are removed from the set of candidates."

proposals/overload-resolution-priority.md
```diff
@@ -0,0 +1,214 @@
...
+Negative numbers are allowed to be used, and can be used to mark a specific overload as worse than all other default overloads.
+
+### `System.Runtime.CompilerServices.OverloadResolutionPriorityAttribute`
+
+We introduce the following attribute to the BCL:
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
```
AlekseyTs, line 101: We probably want the attribute to be applicable to properties as well.

AlekseyTs:
I understand that this is a C# repo. However, in order to achieve the motivating goal, I think we need VB to pay attention to the attributes as well. Therefore, it would be good to create a VB specific proposal as well.

hez2010:
Just to note that allowing overload priority could open a can of worms and shouldn't be rushed out to the public. I would instead recommend making `OverloadResolutionPriorityAttribute` an internal type for now if we really want this feature, so that only BCL can make use of this feature. If there're 3rd library authors wanting to use this feature, they can define their own `OverloadResolutionPriorityAttribute` type to opt-in.

TonyValenti:
I've had mixed feelings about that as well. I have kind of thought that this should be tagged as a "requires preview features" feature. 

I also really think that, in order to be truly useful, we need the c# compiler to allow methods that vary only by return type. I have plenty of methods I want to upgrade the return type on.

colejohnson66:
Could `OverloadResolutionPriorityAttribute` itself be marked with the new `ExperimentalAttribute`? This would require users to acknowledge awareness of said “can of worms”.

jaredpar:
> Just to note that allowing overload priority could open a can of worms and shouldn't be rushed out to the public

This statement without specific examples is not actionable. 

>  I would instead recommend making OverloadResolutionPriorityAttribute an internal type for now if we really want this feature, so that only BCL can make use of this feature

The BCL is not the only team that has found a strong need for this.

jaredpar:
> Could OverloadResolutionPriorityAttribute itself be marked with the new ExperimentalAttribute? This would require users to acknowledge awareness of said “can of worms”.

At the moment the can of worms is empty

jaredpar:
> I imagine that the compiler will follow the common pattern of not caring where the attribute comes from, only the fully qualified name of the attribute class

Agree this is the most likely outcome but the spec should be clear about this. @333fred.

TahirAhmadov:
I've also been wondering what people are referring to by "can of worms". It's strictly an opt-in feature; BCL maintainers generally do a great job at controlling what changes are approved. I don't expect anything "bad" to come out of it, except for potentially some invalid usage by 3-rd party libraries, but that can be said about almost any feature.

333fred:
> Agree this is the most likely outcome but the spec should be clear about this.

No previous specification I'm aware of has been clear about this, I'm not sure what's different here?

333fred:
@AlekseyTs I've addressed your feedback on this.

AlekseyTs (approved):
LGTM (commit 4)

hez2010:
Does this also apply to extension methods?

Assuming we already have:
```cs
class C1
{
    public void M(int[] a) => Console.WriteLine("Array");
}
```

Now we have a
```cs
static class Ext
{
    [OverloadResolutionPriority(1)]
    public static void M(this C1 c, ReadOnlySpan<int> s) => Console.WriteLine("Span");
}
```
IIRC `c.M([42])` would be resolved to `Span`, right?

However, what if we have

```cs
class C1
{
    [OverloadResolutionPriority(1)]
    public void M(Span<int> s) => Console.WriteLine("Span");
    [OverloadResolutionPriority(2)]
    public void M(ReadOnlySpan<int> s) => Console.WriteLine("ReadOnlySpan");
    public void M(int[] a) => Console.WriteLine("Array");
}
```

and now a developer who doesn't own the type `Ext` wants to cut in between the `Span` and `ReadOnlySpan` so that the priority  should be greater than 1 but less than 2, what should the code be like? Should we change the priority to a floating-point number?

We have extension methods, and, in the future, we will also have extension types, there'll be many conflicts due to this if we introduce explicit overload priority. 

This is one of what I said by the can of worms.

hez2010:
> Do you have any concerns not addressed by the proposal?

Nope. I'm okay with it as long as the overload priority only affects the resolution of overloads which are defined in the type itself.

TonyValenti:
I think they should affect Extension methods as well. This way I can upgrade those too.

333fred:
> I think they should affect Extension methods as well. This way I can upgrade those too.

They do, but as worded currently, only within the same type. For example:

```cs
static class Ext1
{
    [OverloadResolutionPriority(1)]
    public static void M(this C1 c, Span<int> s) => Console.WriteLine("Span");
    [OverloadResolutionPriority(0)]
    public static void M(this C1 c, ReadOnlySpan<int> s) => Console.WriteLine("ReadOnlySpan");
}

static class Ext2
{
    [OverloadResolutionPriority(0)]
    public static void M(this C1 c, ReadOnlySpan<int> s) => Console.WriteLine("ReadOnlySpan");
}
```

Within `Ext1`, the span overload would be preferred over the readonlyspan overload. However, between `Ext1` and `Ext2`, overload resolution won't attempt to find any priorities, so resolution would likely end up picking `Ext2.M` as it is a better function member (because Span is convertible to a ReadOnlySpan).

333fred:
I'm going to go ahead and merge this so we can put the open questions on the LDM agenda.

PR #7928, opened 2024-02-10
Add description of extension interface implementation

agocke:
There are a lot of unknowns and difficult semantic constraints with bringing extension interface implementation to C#. This documents many of the challenges and outlines some of the most desirable semantics. When a particular semantic is suggested, example code is provided to show why that semantic is valuable.

### DIFF
--- a/proposals/extension-interfaces.md
+++ b/proposals/extension-interfaces.md
@@ -1,0 +1,192 @@
+# Extension Interface Implementation
+
+This covers open semantic questions and unresolved problems in extension interface implementation.
+
+## What
+
+The basic concept for extension interfaces is to allow additional interfaces to be implemented on existing types. For instance,
+```C#
+extension EquatableUri for Uri : IEquatable<Uri> {
+    public bool Equals(Uri other) => this == other;
+}
+```
+This extension would “implement” the interface IEquatable<T> for Uri. The C# compiler would, in an appropriate context, “attach” this implementation to a given instance of System.Uri to satisfy an implementation requirement.
+This is a very high-level explanation. The exact semantics and requirements of the feature will be detailed below.
+
+## Why
+The basic idea and inspiration traces back to Wadler’s paper [1] on a feature called “type classes.” The basic goal of the feature is to allow the description of arbitrary sets of functions (interfaces) and arbitrary implementations of those sets (interface implementations). C# already allows some of those features, but the current design has severe limitations in two areas:
+1.	Existing types cannot be made to conform to new abstractions without changing the original type
+2.	Interface implementation is tied to type definitions, meaning that different implementations cannot be provided for different substituted generic types, nor can implementations only be valid for certain generic substitutions and not others.
+
+These limitations commonly cause problems for developers, but here are two representative examples:
+1.	A developer wants to build a new abstract framework, like a new logging system. There is no universally accepted implementation of logging for arbitrary types, so they want to provide an interface that users implement to customize logging for their types. However, there’s no way to provide reasonable default implementations for built-in types in the core framework. This creates friction both in logging instances of those types directly, and in implementing logging in their own types that contain instances of built-in types.
+2.	A developer would like to build a new serialization framework. Serialization of certain types, like collections, is not valid unless the generic type parameters to the collection are also serializable. In other words, if serialization is defined using an interface `ISerialize`, `List<T>` should only be considered as implementing `ISerialize` when `T` implements `ISerialize`. Because users cannot implement interfaces on individual type substitutions, and instead must implement an interface using a type definition, it is impossible to encode the correct serialization contract in .NET.
+
+## Usage
+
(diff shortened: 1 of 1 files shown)


proposals/extension-interfaces.md
```diff
@@ -0,0 +1,194 @@
...
+This case is almost identical to _Interface parameter_, but with a generic type parameter as the parameter type instead.
+```
+string M1(string s, int i) => M2(s) + M2(i);
+string M2<T>(T p) where T : IPrettyPrint { ... }
+```
+Unlike in the Interface parameter example, no conversion is performed. However, in the above example, both `string` and `int` should be considered valid substitutions for the type parameter `T`, even though they would not have originally satisfied the `IPrettyPrint` constraint. Once again, the interface members on `IPrettyPrint` should be invocable on `p` and they should use the implementation provided in the extension implementations.
+The main differences in this example and Interface parameter are around boxing. While a value type would have been copied and boxed with an `IPrettyPrint` parameter type, a generic substitution would normally not box the parameter during the invocation of `M2` or during a call to `PrettyPrint` if the input argument directly implemented `IPrettyPrint`. The same should be true of an extension implementation, and side-effects of `M2` and `PrettyPrint` should be propagated back to the `i` parameter in the call `M2(i)`.
```
FaustVX, line 63: > The same should be true of an extension implementation, and side-effects of `M2` and `PrettyPrint` should be propagated back to the `i` parameter in the call `M2(i)`.

I don't understand this.
In today's C#, a value type is copied in a method call, even in a generic one.
What do you mean by: should be propagated back to the `i` parameter.

proposals/extension-interfaces.md
```diff
@@ -0,0 +1,194 @@
...
+```C#
+extension EquatableUri for Uri : IEquatable<Uri> {
+    public bool Equals(Uri other) => this == other;
+}
+```
+This extension would “implement” the interface IEquatable<T> for Uri. The C# compiler would, in an appropriate context, “attach” this implementation to a given instance of System.Uri to satisfy an implementation requirement.
+This is a very high-level explanation. The exact semantics and requirements of the feature will be detailed below.
+
+## Why
+The basic idea and inspiration traces back to Wadler’s paper [1] on a feature called “type classes.” The basic goal of the feature is to allow the description of arbitrary sets of functions (interfaces) and arbitrary implementations of those sets (interface implementations). C# already allows some of those features, but the current design has severe limitations in two areas:
```
jjonescz, line 17: Should there be a link to the paper `[1]`?

proposals/extension-interfaces.md
```diff
@@ -0,0 +1,194 @@
...
+2.	A developer would like to build a new serialization framework. Serialization of certain types, like collections, is not valid unless the generic type parameters to the collection are also serializable. In other words, if serialization is defined using an interface `ISerialize`, `List<T>` should only be considered as implementing `ISerialize` when `T` implements `ISerialize`. Because users cannot implement interfaces on individual type substitutions, and instead must implement an interface using a type definition, it is impossible to encode the correct serialization contract in .NET.
+
+## Usage
+
+Above we defined the general idea of extension interfaces. We will now elaborate on the exact semantics. We will do so using individual motivating examples and describe the reasoning behind preferring certain semantics.
+
+### Interface Parameter
+
+First, let’s consider one of the simplest possible examples: extending an existing primitive type with a new interface, and passing an instance of this type to a parameter of the interface type.
+```
```
jjonescz, line 32: (applies to other code snippets below)

````suggestion
```C#
````

File: proposals/extension-interfaces.md (hunk 1 above)
Comment (agocke) [reply proposals/extension-interfaces.md:63]: On a constrained instance invocation on a generic type parameter, the receiver is a `ref` variable, so the interface invocation is side-effect-preserving.

File: proposals/extension-interfaces.md (hunk 2 above)
Comment (agocke) [reply proposals/extension-interfaces.md:17]: Yup, endnotes got lost somehow. It's https://dl.acm.org/doi/pdf/10.1145/75277.75283

File: proposals/extension-interfaces.md (hunk 1 above)
Comment (FaustVX) [reply proposals/extension-interfaces.md:63]: > a constrained instance invocation on a generic type parameter

Can you give an example please, I'm still not sure to fully understand this line :)

Comment (agocke) [reply proposals/extension-interfaces.md:63]: Yup, so if you have a call like

```C#
interface IFace {
    void Mutate();
}
struct S : IFace {
  public int F;
  public void Mutate() {
    F++;
  }
}
void M<T>(T t) where T : IFace {
  Console.WriteLine(t.F);
  t.Mutate();
  Console.WriteLine(t.F);
}
```

You can see that the call to the interface Method `Mutate` does not copy the struct `t`. Instead, the function mutates the struct variable directly. If you look underneath, you'll see the reason why is that the `this` parameter in the `S.Mutate` method is actually `ref S this`, that is it's passing the struct by-ref.

PR #7977, opened 2024-02-28
Add syntax grammar for `allows` constraint to ref-struct-interfaces.md

AlekseyTs:


CyrusNajmabadi (approved):

proposals/ref-struct-interfaces.md
```diff
@@ -146,6 +146,58 @@ Detailed Notes:
...
+constructor_constraint
+    : 'new' '(' ')'
+    ;
+
+allows_type_parameter_constraints_clause
+    : 'allows' allows_type_parameter_constraints
+
+allows_type_parameter_constraints
+    : allows_type_parameter_constraint
+    | allows_type_parameter_constraints ',' allows_type_parameter_constraint
```
jjonescz, line 192: Why allow multiple if we have just single option now?

cston (approved):

proposals/ref-struct-interfaces.md
AlekseyTs: > Why allow multiple if we have just single option now?

Future proofing, we might want to allow more things in the future.

PR #7989, opened 2024-03-07
Update `Lock` object speclet

jjonescz:
- suppress warning in object equality (`_lock != null`)

### DIFF
--- a/proposals/lock-object.md
+++ b/proposals/lock-object.md
@@ -67,9 +67,6 @@ Note that the shape might not be fully checked (e.g., there will be no errors no
 but the feature might not work as expected (e.g., there will be no warnings when converting `Lock` to a derived type,
 since the feature assumes there are no derived types).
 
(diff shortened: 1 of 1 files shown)


proposals/lock-object.md
```diff
@@ -67,6 +67,9 @@ Note that the shape might not be fully checked (e.g., there will be no errors no
 but the feature might not work as expected (e.g., there will be no warnings when converting `Lock` to a derived type,
 since the feature assumes there are no derived types).
 
+Unlike the equivalent `using` statement, the `lock` statement on the new `Lock` type is allowed in `async` methods
+(but `await`s inside the `lock`'s body are still disallowed).
```
jjonescz, line 71: Clarified with Jared offline this is going to be a separate feature for `using`+`ref struct` in general, so it doesn't have to be mentioned here. Will remove.

PR #7990, opened 2024-03-07
refine speclet disclaimer language

BillWagner:
Add details about the lifetime of these documents on Learn as it pertains to the standardization process.

Contributes to dotnet/docs#39814


proposals/speclet-disclaimer.md
```diff
@@ -1,2 +1,6 @@
...
+> This article is a feature specification. The specification serves as the design document for the feature. It includes proposed specification changes, along with information needed during the design and development of the feature. These articles are published until the proposed spec changes are finalized and incorporated in the current ECMA specification.
+>
+> There may be some discrepancies between the feature specification and the completed implementation. Those differences are captured in the pertinent [language design meeting (LDM) notes](https://github.com/dotnet/csharplang/tree/main/meetings). Links to pertinent meetings are included at the bottom of the spec.
```
333fred, line 4: > Links to pertinent meetings are included at the bottom of the spec.

Are they? I certainly don't go updating the individual feature specifications for each meeting.

KathleenDollard:
Love it! 

I think this is pretty clear that this is our current best effort at describing the feature.

proposals/speclet-disclaimer.md
BillWagner: They seem to be. I'll remove that text if we're not comfortable saying we do it. Readers will certainly find any links we do have.

PR #8000, opened 2024-03-12
Update ref-struct-interfaces.md

AlekseyTs:


jaredpar (approved):

proposals/ref-struct-interfaces.md
```diff
@@ -253,17 +253,44 @@ T M3<T>(Span<T> span)
 }
 ```
 
-These parameters will be encoded in metadata as described in the [byref-like generics doc][byref-like-generics]. Specifically by using the `gpAcceptByRefLike(0x0020)` attribute value.
-
+The anti-constraint is not "inherited" from a type parameter type constraint.
+For example, `S` in the code bellow cannot be substituted with a ref struct:
```
cston, line 257: >bellow [](http://example.com/codeflow?start=29&length=6)

Typo?

cston (approved):

jjonescz (approved):

PR #8012, opened 2024-03-15
Adjust extension member lookup to prefer more specific members

jcouv:

### DIFF
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -1,12 +1,12 @@
 # Extension types
 
-TODO3 No duplicate base extensions (to avoid ambiguities)  
-TODO3 issue with variance of extended type if we erase to a ref struct with a ref field.  
+TODO(inheritance) No duplicate base extensions (to avoid ambiguities)  
+TODO(inheritance) issue with variance of extended type if we erase to a ref struct with a ref field.  
 
-TODO2 need to spec why extension properties are not found during lookup for attribute properties, or explicitly disallow them  
-TODO2 adjust scoping rules so that type parameters are in scope within the 'for'  
+TODO(instance) need to spec why extension properties are not found during lookup for attribute properties, or explicitly disallow them  
+TODO(static) adjust scoping rules so that type parameters are in scope within the 'for'  
 TODO2 check Method type inference: 7.5.2.9 Lower-bound interfaces  
-TODO2 extensions are disallowed within interfaces with variant type parameters  
+TODO(static) extensions are disallowed within interfaces with variant type parameters  
 TODO2 We should likely allow constructors and `required` properties  
 TODO attributes and attribute targets  
 
@@ -218,7 +218,7 @@ An example with multiple base explicit extension:
 explicit extension DiamondExtension for NarrowerUnderlyingType : BaseExtension1, Interface1, BaseExtension2, Interface2 { }`
 ```
 
-TODO should we have a naming convention like `Extension` suffixes? (`DataObjectExtension`)  
+TODO(static) should we have a naming convention like `Extension` suffixes? (`DataObjectExtension`)  
 
 ## Extension type
 
@@ -232,13 +232,13 @@ There is an identity conversion between an extension and its underlying type,
 and between an extension and its base extensions.
 
 An extension type satisfies the constraints satisfied by its underlying type (see section on constraints). 
-In phase C, some additional constraints can be satisfied (additional implemented interfaces).  
+In the Interface Phase, some additional constraints can be satisfied (additional implemented interfaces).  
 
 ### Underlying type
 
 The *extension_underlying_type* type may not be `dynamic`, a pointer, 
 a ref struct type, a ref type or an extension.  
-The underlying type may not include an *interface_type_list* (this is part of Phase C).  
+The underlying type may not include an *interface_type_list* (this is part of the Interface Phase).  
 The extension type must be static if its underlying type is static.  
 
 When a partial extension declaration includes an underlying type specification,
@@ -250,9 +250,9 @@ an underlying type specification.
 It is a compile-time error if the underlying type differs amongst all the parts
 of an extension declaration.  
 
-TODO2 we need rules to only allow an underlying type that is compatible with the base extensions.  
+TODO2(inheritance) we need rules to only allow an underlying type that is compatible with the base extensions.  
 
-TODO should the underlying type be inferred when none was specified but
+TODO(inheritance) should the underlying type be inferred when none was specified but
   base extensions are specified?
 
 ### Modifiers
@@ -328,7 +328,7 @@ other derived extensions, even when the members are inherited from the same base
 
 Note: the rules still disallow access to protected members of the underlying type through the extension type.
 
-TODO
+TODO(static)
 
 Let `B` be a base class that declares a protected instance member `M`, 
 and let `D` be a class that derives from `B`. Within the *class_body* of `D`, 
@@ -371,7 +371,7 @@ The existing [rules for signatures](https://github.com/dotnet/csharpstandard/blo
 Two signatures differing by an extension vs. its underlying type, or an extension vs. 
 one of its base extensions are considered to be the *same signature*.
 
-TODO2 this needs to be refined to allow overload on different underlying types.
+TODO2(inheritance) this needs to be refined to allow overload on different underlying types.
 ```
 explicit extension ObjectExtension : object;
 explicit extension StringExtension : string, ObjectExtension;
@@ -413,7 +413,7 @@ Otherwise, existing [rules for fields](https://github.com/dotnet/csharpstandard/
 
 #### Methods
 
-TODO allow `this` (of type current extension).  
+TODO(instance) allow `this` (of type current extension).  
 Parameters with the `this` modifier are disallowed.
 Otherwise, existing [rules for methods](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#146-methods) apply.
 In particular, a static method does not operate on a specific instance, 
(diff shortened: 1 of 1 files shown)


proposals/extensions.md
```diff
@@ -882,11 +883,9 @@ and when successful this process yields an extension type we can use
 
 An extension type `X` is compatible with given type `U` if:
 - `X` is non-generic and its underlying type is `U`, a base type of `U` or an implemented interface of `U`
-- a possible type substitution on the type parameters of `X` yields underlying type `U`, 
+- we collect all the possible type substitutions on the type parameters of `X` yields underlying type `U`,
```
AlekseyTs, line 886: >we collect all the possible type substitutions on the type parameters of `X` yields underlying type `U`, [](http://example.com/codeflow?start=2&length=104)

The wording doesn't quite work in the list of "An extension type `X` is compatible with given type `U` if"  conditions. This item describes some form of action, not a condition.

proposals/extensions.md
```diff
@@ -693,7 +693,8 @@ TODO4 need to merge extension members and extension methods
     - Merge the results
     - Next, members that are hidden by other members are removed from the set.  
       (note: "base types" means "base extensions and underlying type" for extension types)
-    - Finally, having removed hidden members:
+    - Next, extension members are removed if any more specific extension member is applicable.
```
AlekseyTs, line 696: >Next, extension members are removed if any more specific extension member is applicable. [](http://example.com/codeflow?start=6&length=88)

This is very vague. What does it mean to be more specific should be clearly specified. Perhaps in a separate section. I think our goal at this point should be to clearly describe what the implementation is doing, so that the rules could be reviewed/discussed/approved. Like less specific methods are not removed by more specific methods, etc.

proposals/extensions.md
```diff
@@ -948,24 +959,21 @@ We process as follows:
     - Perform member lookup for `I` in each compatible substituted extension type `X` 
       (note this takes into account whether the member is invoked).
     - Merge the results
-TODO4 spec how we deal with duplicate or near duplicate entries (for example, field E1.Member and method E2.Member)
   - Next, members that are hidden by other members are removed from the set.  
     (note: "base types" means "base extensions and underlying type" for extension types)
+  - Next, extension members are removed if they are less specific than another extension member.
```
AlekseyTs, line 964: >- Next, extension members are removed if they are less specific than another extension member.   [](http://example.com/codeflow?start=2&length=96)

This is very vague

proposals/extensions.md
```diff
@@ -995,6 +1003,15 @@ extension member lookup will find the `int` property and stop there.
...
 For context see [extension method invocation rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11783-extension-method-invocations).
 
+### 12.6.4 Overload resolution
+
+We update the [overload resolution section](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#12641-general) as follows:
+
+Each of these contexts defines the set of candidate function members and the list of arguments in its own unique way. 
+For instance, the set of candidates for a method invocation does not include methods marked override, 
+methods in a base class are not candidates if any method in a derived class is applicable,
+\***and extension methods in a less specific extension type are not candidates if any extension method in a more specific extension type is applicable**.
```
AlekseyTs, line 1013: This is very vague

proposals/extensions.md
```diff
@@ -1125,36 +1142,40 @@ of a class, struct, interface, /***or extension** type `T`, and `A` is an *argum
...
-  of multiple extension methods, and even though the scenario remains an error for other reasons.
-  > A method group has a natural type if all candidate methods in the method group have a common signature.
-    (If the method group may include extension methods, the candidates include the containing type and all extension method scopes.)
 
 ### Pattern-based invocations and member access
 
-TODO Need to scan through all pattern-based rules for known members to consider whether to include extension type members.
+TODO(static) Need to scan through all pattern-based rules for known members to consider whether to include extension type members.
   If extension methods were already included, then we should certainly include extension type methods.
   Otherwise, we should consider it (for example `Current` property in `foreach`).
```
jjonescz, line 1167: Looks like this example is unnecessary since there's now a whole list of examples.

```suggestion
  Otherwise, we should consider it:
```

proposals/extensions.md
```diff
@@ -693,7 +693,11 @@ TODO4 need to merge extension members and extension methods
     - Merge the results
     - Next, members that are hidden by other members are removed from the set.  
       (note: "base types" means "base extensions and underlying type" for extension types)
-    - Finally, having removed hidden members:
+    - Next, extension members are removed if they are "hidden" by more specific extension members. 
+      For every member `X.M` in the set, where `X` is the type in which the member `M` is declared, the following rules are applied:
+      - If `M` is a method, then all non-method members declared in a less specific type than `X` are removed from the set.
```
AlekseyTs, line 698: >less specific type than `X` [](http://example.com/codeflow?start=70&length=27)

> less specific type than `X`

Do we specify what does it mean to be less specific?

proposals/extensions.md
```diff
@@ -995,6 +1015,25 @@ extension member lookup will find the `int` property and stop there.
 
 For context see [extension method invocation rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11783-extension-method-invocations).
 
+### More specific extension member
```
AlekseyTs, line 1018: >More specific extension member [](http://example.com/codeflow?start=4&length=30)

Should this section also define a concept of "less specific type than `X`"?

jcouv: I reframed this section to define "less specific" to avoid confusion

proposals/extensions.md
```diff
@@ -693,7 +693,11 @@ TODO4 need to merge extension members and extension methods
     - Merge the results
     - Next, members that are hidden by other members are removed from the set.  
       (note: "base types" means "base extensions and underlying type" for extension types)
-    - Finally, having removed hidden members:
+    - Next, extension members are removed if they are "hidden" by more specific extension members. 
+      For every member `X.M` in the set, where `X` is the type in which the member `M` is declared, the following rules are applied:
```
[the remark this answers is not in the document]
jcouv: I think there is a deeper problem :-/ 
These rules require a specific iteration order. Otherwise, in a "method hiding a field hiding a method" situation, we could end up removing the last method, which is wrong.
I'll take another stab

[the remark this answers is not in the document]
jcouv: Note, I think this spec problem is already present in the [current member lookup rules](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#125-member-lookup)...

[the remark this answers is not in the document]
AlekseyTs: > These rules require a specific iteration order. Otherwise, in a "method hiding a field hiding a method" situation, we could end up removing the last method, which is wrong.

I do not think the suggested wording has this problem. We can discuss offline.

>Note, I think this spec problem is already present in the current member lookup rules...

I don't think this helps us in any way. We need to specify clear rules so that they can be reviewed, clearly understood by others, and then approved or rejected at design meeting.

proposals/extensions.md
AlekseyTs: This thread was marked as resolved, but I do not see it addressed. The question was about "specific **type**" vs. "specific member".

AlekseyTs:
Done with review pass (commit 10)

jcouv:
@AlekseyTs I addressed the remaining feedback. Please take another look. Thanks

proposals/extensions.md
```diff
@@ -1125,36 +1168,60 @@ of a class, struct, interface, /***or extension** type `T`, and `A` is an *argum
...
+
+The current [simple assignment rules](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#12212-simple-assignment) state:
 
-## Implementation details
+> When a property or indexer declared in a struct_type is the target of an assignment,
+> the instance expression associated with the property or indexer access shall be classified as a variable.
+> If the instance expression is classified as a value, a binding-time error occurs.
 
-TODO3 revise this to use a regular struct
+We're expecting those rules to be updated to check whether the receiver is a reference or value type instead.
```
AlekseyTs, line 1210: I think the proposed new rules should be explicitly stated, completely. Also, it sounds like the proposed change is going to "break" some existing code. If that is the case, consider adding an example.

proposals/extensions.md
```diff
@@ -948,24 +927,27 @@ We process as follows:
     - Perform member lookup for `I` in each compatible substituted extension type `X` 
       (note this takes into account whether the member is invoked).
     - Merge the results
-TODO4 spec how we deal with duplicate or near duplicate entries (for example, field E1.Member and method E2.Member)
   - Next, members that are hidden by other members are removed from the set.  
     (note: "base types" means "base extensions and underlying type" for extension types)
+  - Next, less specific extension members are removed if they are "hidden" by more specific extension members.  
+    For every member `X.M` in the set, where `X` is the extension type in which the member `M` is declared, the following rules are applied:
```
AlekseyTs, line 933: >For every member `X.M` in the set, where `X` is the extension type in which the member `M` is declared [](http://example.com/codeflow?start=4&length=102)

Should the following be mentioned here?

> if no member in the set has a containing type `Y` that is more specific than `X` then the following rules are applied

proposals/extensions.md
```diff
@@ -995,6 +977,67 @@ extension member lookup will find the `int` property and stop there.
...
+
+If `X` extends `C` and `Y` extends `D`, `X` is considered less specific than `Y`
+when:
+- `C` is a base type of `D`, or
+- `C` is an interface implemented by `D`.
+
+### Less specific extension member
+
+TL;DR: As part of extension member lookup, extension invocations and overload resolution,
+we consider that "less specific" extension members are "hidden" by "more specific" extension members.
```
AlekseyTs, line 990: It looks like the rules are specified in terms of specificity of extension type instead. It also looks like the rule specified in the paragraph below is not used anywhere. Perhaps that paragraph should be removed, and the rest of the current section (the examples, etc.) should be merged into the previous section

AlekseyTs:
Done with review pass (commit 12)

proposals/extensions.md
jcouv: Indeed. Thanks

AlekseyTs (approved):
LGTM (commit 13)

PR #8027, opened 2024-04-01
Adjust dynamic binding rules for a situation of a single applicable candidate.

AlekseyTs:
Meant to mitigate breaking changes like https://github.com/dotnet/roslyn/issues/72750.

### DIFF
--- a/proposals/params-collections.md
+++ b/proposals/params-collections.md
@@ -341,8 +341,12 @@ If only a single candidate passes the test, the invocation of the candidate is s
 - the candidate is either not generic, or its type arguments are explicitly specified;
 - there is no ambiguity between normal and expanded forms of the candidate that cannot be resolved at compile time. 
 
-In this case, if the single candidate is not a local function and returns a value (doesn't have type `void`, doesn't return a `ref`) and
-there is an implicit conversion from result type to `dynamic`, then result of the invocation is the returned value converted to `dynamic`.
+In this case, if the single candidate is not a local function and all the following conditions are met,
+then result of the invocation is the returned value converted to `dynamic`.
+- the candidate returns a value (doesn't have type `void`, doesn't return a `ref`);
+- there is an implicit conversion from result type to `dynamic`;
+- the receiver and argument list would be supported for dynamically bound invocation.
+
 For example:
 ```csharp
 unsafe public class C
@@ -359,12 +363,19 @@ unsafe public class C
         static int test4(int x) => x;
 
         var f = Test5(d); // int
+
+        var g = Test6(name: "name", d); // int : named before regular argument
+
+        string name = "name";
+        var h = Test(&name, d); // int : pointer argument
     }
     
     static void Test1(int x) {}
     static void* Test2(int x) => null;
     static int Test3(int x) => x;
     static ref int Test5(int x) => ...;
+    static int Test6(string name, object value) => ...;
+    static int Test7(string* name, object value) => ...;
 }
 ```
 
--- a/proposals/params-collections.md
+++ b/proposals/params-collections.md
@@ -341,8 +341,8 @@ If only a single candidate passes the test, the invocation of the candidate is s
 - the candidate is either not generic, or its type arguments are explicitly specified;
 - there is no ambiguity between normal and expanded forms of the candidate that cannot be resolved at compile time. 
 
-In this case, if the single candidate is not a local function and returns a value (doesn't have type `void`) and there is an
-implicit conversion from result type to `dynamic`, then result of the invocation is the returned value converted to `dynamic`.
+In this case, if the single candidate is not a local function and returns a value (doesn't have type `void`, doesn't return a `ref`) and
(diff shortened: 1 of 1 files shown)


proposals/params-collections.md
```diff
@@ -341,6 +341,44 @@ If only a single candidate passes the test, the invocation of the candidate is s
 - the candidate is either not generic, or its type arguments are explicitly specified;
 - there is no ambiguity between normal and expanded forms of the candidate that cannot be resolved at compile time. 
 
+In this case, if the single candidate is not a local function and all the following conditions are met,
```
jjonescz, line 344: For consistency/clarity the "if the single candidate is not a local function" condition could be also one of the bullet points below

proposals/params-collections.md
```diff
@@ -341,6 +341,44 @@ If only a single candidate passes the test, the invocation of the candidate is s
...
+        var b = Test2(d); // void*, there is no implicit conversion to dynamic
+        var c = Test3(d); // dynamic
+        c.InvokeOnDynamic(); // No error
+
+        var e = test4(d); // int
+        static int test4(int x) => x;
+
+        var f = Test5(d); // int
+
+        var g = Test6(name: "name", d); // int : named before regular argument
```
jjonescz, line 367:
```suggestion
        var g = Test6(name: "name", d); // error: named before regular argument
```

proposals/params-collections.md
```diff
@@ -341,6 +341,44 @@ If only a single candidate passes the test, the invocation of the candidate is s
...
+
+        string name = "name";
+        var h = Test(&name, d); // int : pointer argument
```
jjonescz, line 370:
```suggestion
        var h = Test7(&name, d); // error: pointer argument
```

proposals/params-collections.md
AlekseyTs: This isn't an error for static binding

proposals/params-collections.md
AlekseyTs: This isn't an error for static binding

proposals/params-collections.md
```diff
@@ -341,6 +341,45 @@ If only a single candidate passes the test, the invocation of the candidate is s
...
+        dynamic d = 1;
+        var a = Test1(d); // error CS0815: Cannot assign void to an implicitly-typed variable
+        var b = Test2(d); // void*, there is no implicit conversion to dynamic
+        var c = Test3(d); // dynamic
+        c.InvokeOnDynamic(); // No error
+
+        var e = test4(d); // int
+        static int test4(int x) => x;
+
+        var f = Test5(d); // int
```
jaredpar, line 366: Shouldn't this be an error given it's a `ref` return?

AlekseyTs: To clarify, consider the following example:
```
public class C
{
    public static void Main()
    {
        dynamic d = 1;
        var a = Test1(d);
        System.Console.WriteLine(a);        
    }

    static int _test1 = 0;    
    static ref int Test1(int x) => ref _test1;
}
```

When compiled in C# 12, 'a' is dynamic and the code produces the following output:
```
System.InvalidCastException: The result type 'System.Int32&' of the dynamic binding produced by binder 'Microsoft.CSharp.RuntimeBinder.CSharpInvokeMemberBinder' is not compatible with the result type 'System.Object' expected by the call site.
   at System.Dynamic.DynamicMetaObjectBinder.Bind(Object[] args, ReadOnlyCollection`1 parameters, LabelTarget returnLabel)
   at System.Runtime.CompilerServices.CallSiteBinder.BindCore[T](CallSite`1 site, Object[] args)
   at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)
   at C.Main()
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

When compiled in preview (prior to this design change), 'a' is int and the code produces the following output:
```
0
```

With the proposed change in design, there will be no change in behavior by comparison to the preview behavior.

jaredpar (approved):

proposals/params-collections.md
```diff
@@ -341,6 +341,45 @@ If only a single candidate passes the test, the invocation of the candidate is s
...
+
+        var g = Test6(name: "name", d); // int : named before regular argument
+
+        string name = "name";
+        var h = Test(&name, d); // int : pointer argument
```
jjonescz, line 371:
```suggestion
        var h = Test7(&name, d); // int : pointer argument
```

333fred:
Discussed in LDM on May 1st: https://github.com/dotnet/csharplang/blob/main/meetings/2024/LDM-2024-05-01.md#adjust-binding-rules-in-the-presence-of-a-single-candidate

PR #8055, opened 2024-04-15
Take "non-enumerable collection types" out of C# 12

jcouv:
I'd incorrectly merged the whole change into C# 12: the change from iteration-type to element-type is fine, but the part about inferring the element-type from the `Create` method won't be handled as a C# 12 bugfix so should be taken out. It'll be re-introduced into a later speclet. 
FYI @cston @AlekseyTs @RikkiGibson


cston (approved):

jcouv:
@cston Can I let you revert this change into a new speclet for C# 13 collection-related changes?

PR #8056, opened 2024-04-16
Update ref-in-async speclet

jjonescz:


cston (approved):

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -155,6 +162,9 @@ class C
   > or the variable needs to be hoisted as part of an async ([§15.15][async-funcs]) or an iterator ([§15.14][iterators]) method**.
   > - [...]
 
+  - Currently, we have an existing C# 12 warning for address-of in async methods and a proposed C# 13 error for address-of in iterators.
```
AlekseyTs, line 165: >a proposed C# 13 error for address-of in iterators [](http://example.com/codeflow?start=85&length=50)

What does "a proposed C# 13 error for address-of in iterators" mean? Is it a new error reported by new compiler, or it is an error reported for language version C# 13 and later? Consider clarifying.

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -47,7 +47,11 @@ and C# 13 will lift these restrictions as described below.
 [§13.3.1 Blocks > General][blocks-general]:
 
 > ~~It is a compile-time error for an iterator block to contain an unsafe context ([§23.2][unsafe-contexts]).~~
-> An iterator block always defines a safe context, even when its declaration is nested in an unsafe context.
+> An iterator block always defines a safe context, even when its declaration is nested in an unsafe context
+> **unless the iterator method itself is marked with the `unsafe` modifier ([§23.2][unsafe-contexts])**.
```
AlekseyTs, line 51: I think we need to clarify what effect, if any, does this have on the signature of the declaration.
For example, is the signature of the method in an `unsafe` context in the following example?
```
unsafe class C
{
    System.Collections.Generic.IEnumerable<int> M() 
    {
        yield return 1;
    }
}
```

Also, consider the following scenario:
```
unsafe class C
{
    System.Collections.Generic.IEnumerable<int> M() 
    {
         unsafe
         {
              { // Does this define a safe context?
                    yield return 1;
              }
         }
    }
}
```

According to the current definition
> A block that contains one or more yield statements is called an iterator block.

Given that, the innermost block is an iterator block Does it define a safe context or not?

I think we also need clarification for properties. For example, is body of the getter in a safe or `unsafe` context in this example?
```
class C
{
     unsafe System.Collections.Generic.IEnumerable<int> P
    {
        get
        {
            yield return 1;
        }
        set
        {}
    }
}
```

Is context safe or unsafe in the setter for this scenario?
```
unsafe class C
{
    System.Collections.Generic.IEnumerable<int> P
    {
        get
        {
            yield return 1;
        }
        set
        {}
    }
}
```

proposals/ref-unsafe-in-iterators-async.md
jjonescz: It doesn't really make a difference if it were reported for earlier langversions since address-of in iterators could not be used anyway (because unsafe code in general could not be used inside iterators). But will clarify, thanks.

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -43,11 +43,98 @@ lock (this)
...
-> ~~It is a compile-time error for an iterator block to contain an unsafe context ([§23.2][unsafe-contexts]).~~
-> An iterator block always defines a safe context, even when its declaration is nested in an unsafe context.
+> A *block* that contains one or more `yield` statements ([§13.15][yield-statement]) is called an iterator block,
+> **even if those `yield` statements are contained only indirectly in nested blocks (excluding nested lambdas and local functions).**
+>
+> [...]
+>
+> ~~It is a compile-time error for an iterator block to contain an unsafe context ([§23.2][unsafe-contexts]).
+> An iterator block always defines a safe context, even when its declaration is nested in an unsafe context.~~
+> **The iterator block used to implement an iterator ([§15.14][iterators])
```
AlekseyTs, line 57: > **The iterator block used to implement an iterator ([§15.14][iterators]) [](http://example.com/codeflow?start=1&length=73)

Wouldn't it be more intuitive to treat the entire iterator declaration as defining a safe context, since presence of `unsafe` modifier makes a difference? Was this option seriously considered?

jjonescz: > Wouldn't it be more intuitive to treat the entire iterator declaration as defining a safe context

That might be a more straightforward alternative, thanks. The way I thought about this was that an iterator is defined by the block (method body or property accessor) which contains `yield`, so that's where the safe context should begin (and hence the signature and the `set` accessor are not in that context).

jjonescz: Will it be possible to implement that though? Do we know whether a method is an iterator when binding its signature?

AlekseyTs: >  Also I'm not sure if partial declarations without bodies can "see" the corresponding partial implementation body to determine whether it's an iterator.

I think partial declarations are never iterators. They lack an iterator block

AlekseyTs (approved):
LGTM (commit 9)

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -43,11 +43,112 @@ lock (this)
...
+            };
+        }
+        async void local()
+        { // safe context
+          // note: in Roslyn, this is an unsafe context in LangVersion 12 and lower
+            await Task.Yield();
+        }
+        local();
+    }
+    partial IEnumerable<int> M5(); // unsafe context, no iterator body
```
cston, line 138: >unsafe [](http://example.com/codeflow?start=38&length=6)

Does that mean an explicit `unsafe` will be allowed on the definition part only for iterators with language version 13?
```csharp
partial IEnumerable<int> M6();
partial unsafe IEnumerable<int> M6() { yield break; }
```

cston (approved):

proposals/ref-unsafe-in-iterators-async.md
jjonescz: Yes, exactly.

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -43,11 +43,128 @@ lock (this)
...
+partial class C1
+{
+    public partial IEnumerable<int> M5(); // safe context (inherits from parent)
+}
+class C2
+{ // safe context
+    [/* unsafe context */ A]
+    unsafe IEnumerable<int> M(
+        /* unsafe context */ int*[] x)
+    { // unsafe context (the iterator declaration is unsafe)
```
AlekseyTs, line 154: >// unsafe context (the iterator declaration is unsafe) [](http://example.com/codeflow?start=6&length=54)

This contradicts the following statement from above: "If an iterator declaration is marked with the `unsafe` modifier, the signature is in an unsafe scope
  but the iterator block used to implement that iterator still defines a safe scope."

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -43,11 +43,128 @@ lock (this)
...
+        yield return 1; // error: `yield return` in unsafe context
+    }
+    unsafe IEnumerable<int> this[
+        /* unsafe context */ int*[] x]
+    { // unsafe context (the iterator declaration is unsafe)
+        get
+        { // unsafe context
```
AlekseyTs, line 161: >// unsafe context [](http://example.com/codeflow?start=10&length=17)

This one too

AlekseyTs (approved):
LGTM (commit 12)

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -43,11 +43,128 @@ lock (this)
...
-> An iterator block always defines a safe context, even when its declaration is nested in an unsafe context.
+> A *block* that contains one or more `yield` statements ([§13.15][yield-statement]) is called an iterator block,
+> **even if those `yield` statements are contained only indirectly in nested blocks (excluding nested lambdas and local functions).**
+>
+> [...]
+>
+> ~~It is a compile-time error for an iterator block to contain an unsafe context ([§23.2][unsafe-contexts]).
+> An iterator block always defines a safe context, even when its declaration is nested in an unsafe context.~~
+> **The iterator block used to implement an iterator ([§15.14][iterators])
+> always defines a safe context, even when its declaration is nested in an unsafe context.**
```
cston, line 58: >even when its declaration [](http://example.com/codeflow?start=33&length=25)

Consider using "even when **the iterator** declaration" to clarify "its" here.

proposals/ref-unsafe-in-iterators-async.md
cston: Shouldn't the mismatch (of safe/unsafe contexts for the partial method signatures) result in an error?

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -43,11 +43,128 @@ lock (this)
...
+
+From this spec it also follows:
+- If an iterator declaration is marked with the `unsafe` modifier, the signature is in an unsafe scope
+  but the iterator block used to implement that iterator still defines a safe scope.
+- The `set` accessor of an iterator property or indexer (i.e., its `get` accessor is implemented via an iterator block)
+  "inherits" its safe/unsafe scope from the declaration.
+- This does not affect partial declarations without implementation as they are only signatures and cannot have an iterator body.
+- In C# 12 it is an error to have an iterator method marked with the `unsafe` modifier, but that is allowed in C# 13.
```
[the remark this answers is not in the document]
jjonescz: Well, you're right, the C# 12/13 thing does not follow just from the spec but also from this:

https://github.com/dotnet/csharplang/blob/f0732b08451be96ef3d4f9f4e82377011e416f7f/proposals/ref-unsafe-in-iterators-async.md?plain=1#L43-L46

I'm not sure, but I thought the spec is usually not conditional on langversion, the whole spec is for one specific langversion. Anyway I will clarify this, thanks.

cston (approved):

Issue #6420, opened 2022-08-30
[Proposal]: Partial properties


RikkiGibson:
# Partial properties

* [x] Proposed
* [ ] Prototype: [Not Started](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
* [ ] Implementation: [Not Started](https://github.com/dotnet/roslyn/BRANCH_NAME)
* [ ] Specification: [Not Started](pr/1)

## Summary
[summary]: #summary

Allow the `partial` modifier on properties to separate declaration and implementation parts, similar to partial methods.

```cs
// UserCode.cs
public partial class ViewModel : INotifyPropertyChanged
{
    public event PropertyChangedEventHandler PropertyChanged;

    [NotifyPropertyChanged]
    public partial string UserName { get; set; }
}

// Generated.cs
public partial class ViewModel
{
    private string __generated_userName;

    public partial string UserName
    {
        get => __generated_userName;
        set
        {
            if (value != __generated_userName)
            {
                __generated_userName = value;
                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(UserName)));
            }
        }
    }
}
```

## Motivation
[motivation]: #motivation

When we did [extended partial methods](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/extending-partial-methods.md), we indicated we would like to consider adding support for other kinds of partial members in the future. The community has [shown enthusiasm](https://github.com/dotnet/csharplang/discussions/3412) for partial properties in particular.

.NET has a number of scenarios where a property implementation is some kind of boilerplate. One of the most prominent cases is `INotifyPropertyChanged`, as seen above. Another is [dependency properties](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/properties/dependency-properties-overview?view=netdesktop-6.0). There are currently production source generators designed to handle these scenarios. These currently work by having the user write a field and having the generator add the corresponding property.

```cs
// UserCode.cs
public partial class ViewModel : INotifyPropertyChanged
{
    public event PropertyChangedEventHandler PropertyChanged;

    [NotifyPropertyChanged]
    private string _userName;
}

// Generated.cs
public partial class ViewModel
{
    public string UserName
    {
        get => /* ... */;
        set
        {
            // ...
        }
    }
}
```

[...]

PR #8065, opened 2024-04-19
Create partial-properties.md

RikkiGibson:
@jcouv

The goal is to migrate the proposal details in #6420 nearer to its "final" location.

### DIFF
--- a/proposals/partial-properties.md
+++ b/proposals/partial-properties.md
@@ -17,7 +17,7 @@ property_declaration
 ### Defining and implementing declarations
 When a property declaration includes a *partial* modifier, that property is said to be a *partial property*. Partial properties may only be declared as members of partial types.
 
-A *partial property* declaration is said to be a *defining declaration* when its accessors all have semicolon bodies, and it lacks the `extern` modifier. Otherwise, it is an *implementing declaration*. A *partial property* cannot be an auto-property.
+A *partial property* declaration is said to be a *defining declaration* when its accessors all have semicolon bodies, and it lacks the `extern` modifier. Otherwise, it is an *implementing declaration*.
 
 ```cs
 partial class C
@@ -29,6 +29,10 @@ partial class C
     public partial string Prop { get => field; set => field = value; }
 }
 ```
+
+Because we have reserved the syntactic form with semicolon accessor bodies for the *defining declaration*, a partial property cannot be *automatically implemented*. We therefore adjust [Automatically implemented properties (§15.7.4)](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/classes.md#1574-automatically-implemented-properties) as follows:
+
+> An automatically implemented property (or auto-property for short), is a non-abstract, non-extern, **non-partial,** non-ref-valued property with semicolon-only accessor bodies.
 
 **Remarks**. It is useful for the compiler to be able to look at a single declaration in isolation and know whether it is a defining or an implementing declaration. Therefore we don't want to permit auto-properties by including two identical `partial` property declarations, for example. We don't think that the use cases for this feature involve implementing the partial property with an auto-property, but in cases where a trivial implementation is desired, we think the `field` keyword makes things simple enough.
 
--- a/proposals/partial-properties.md
+++ b/proposals/partial-properties.md
@@ -87,25 +87,52 @@ partial class C3
 }
 ```
 
-## Drawbacks
-[drawbacks]: #drawbacks
+### Indexers
 
-As always, this feature adds to the language concept count and must be weighed accordingly.
+Per [LDM meeting on 2nd November 2022](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-11-02.md#partial-properties), indexers will be supported with this feature.
 
-The fact that adding the `partial` modifier can change an auto-property declaration to a defining partial declaration may be confusing.
+Indexer parameters must match in the same way that partial method parameters must match, and merging of parameter attributes across partial indexer declarations occurs in the same way that it does for partial methods.
 
-## Alternatives
-[alternatives]: #alternatives
 
-We could consider more flexible designs which permit different accessor definitions or implementations to be spread across different declarations.
+```cs
+partial class C
+{
+    public partial int this[int x] { get; set; }
+    public partial int this[int x]
+    {
+        get => this._store[x];
+        set => this._store[x] = value;
+    }
+}
 
-We could consider introducing some special way to denote that a partial property implementation is an auto-property, separate from the `field` keyword.
+// attribute merging
+partial class C
+{
+    public partial int this[[Attr1] int x]
+    {
+        [Attr2] get;
+        set;
+    }
 
-We could also consider doing nothing, which means that source generators and perhaps our tooling will need to establish conventions for working around the limitations of the field-based approach.
+    public partial int this[[Attr3] int x]
+    {
+        get => this._store[x];
+        [Attr4] set => this._store[x] = value;
+    }
 
-## Unresolved questions
-[unresolved]: #unresolved-questions
+    // results in a merged member emitted to metadata:
+    public int this[[Attr1, Attr3] int x]
+    {
+        [Attr2] get => this._store[x];
+        [Attr4] set => this._store[x] = value;
+    }
+}
+```
 
-Should we permit partial indexers as part of this feature? It would increase orthogonality to allow this, but the word "indexer" is mentioned zero times since 2020 in the community discussion for this feature.
+## Open Issues
 
-Similarly, should we permit other kinds of partial members like fields, events, constructors, operators, etc? The same is mentioned in [extending-partial-methods.md](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/extending-partial-methods.md#partial-on-all-members).
+### Other member kinds
+
+A community member opened a discussion to request support for [partial events](https://github.com/dotnet/csharplang/discussions/8064). In the [LDM meeting on 2nd November 2022](https://github.com/dotnet/csharplang/blob/main/meetings/2022/LDM-2022-11-02.md#partial-properties), we decided to punt on support for events, in part because nobody at the time had requested it. We may want to revisit this question, since this request has now come in, and it has been over a year since we last discussed it.
+
+We could also go even further in permitting partial declarations of constructors, operators, fields, and so on, but it's unclear if the design burden of these is justified, just because we are already doing partial properties.
--- a/proposals/partial-properties.md
+++ b/proposals/partial-properties.md
@@ -86,3 +86,26 @@ partial class C3
     public partial string Prop { get => field; set => field = value; }
 }
 ```
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+As always, this feature adds to the language concept count and must be weighed accordingly.
+
+The fact that adding the `partial` modifier can change an auto-property declaration to a defining partial declaration may be confusing.
+
+## Alternatives
+[alternatives]: #alternatives
+
+We could consider more flexible designs which permit different accessor definitions or implementations to be spread across different declarations.
+
+We could consider introducing some special way to denote that a partial property implementation is an auto-property, separate from the `field` keyword.
+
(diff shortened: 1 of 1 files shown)


proposals/partial-properties.md
```diff
@@ -0,0 +1,88 @@
...
+    public C() // warning CS8618: Non-nullable property 'Prop' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.
+    {
+    }
+}
+```
+
+### Matching signatures
+The property declarations must have the same type and ref kind. The property declarations must have the same set of accessors.
+
+The property declarations and their accessor declarations must have the same modifiers, though the modifiers may appear in a different order. This does not apply to the `extern` modifier, which may only appear on an *implementing declaration*.
```
jaredpar, line 60: The `required` modifier is a new consideration as it can't apply to `partial` methods. Does that pose any particular challenges here or does it just fall out?

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
...
+https://github.com/dotnet/csharplang/issues/6420
+
+### Grammar
+
+The *property_declaration* grammar [(§14.7.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1471-general) is updated as follows:
+
+```diff
+property_declaration
+-    : attributes? property_modifier* type member_name property_body
++    : attributes? property_modifier* 'partial'? type member_name property_body
```
jcouv, line 11: 1. We'll probably want to confirm whether `partial` should still be in locked position for new features (that's something we considered relaxing for existing features)
2. The structure for `method_header` is a bit different (`partial` is not inlined in the declaration, but rather inside the `method_modifiers` production). Should we spec it the same way for partial properties?
```antlr
method_declaration
    : attributes? method_modifiers return_type method_header method_body
    | attributes? ref_method_modifiers ref_kind ref_return_type method_header
      ref_method_body
    ;

method_modifiers
    : method_modifier* 'partial'?
    ;
```

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
+# Partial properties
+https://github.com/dotnet/csharplang/issues/6420
+
+### Grammar
+
+The *property_declaration* grammar [(§14.7.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1471-general) is updated as follows:
+
+```diff
+property_declaration
```
jcouv, line 9: Have we discussed whether the change should extend to indexers too (that feels like it should go along with properties), and possibly events as well?

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
...
+    public partial string Prop { get => field; set => field = value; }
+
+    public C() // warning CS8618: Non-nullable property 'Prop' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.
+    {
+    }
+}
+```
+
+### Matching signatures
+The property declarations must have the same type and ref kind. The property declarations must have the same set of accessors.
```
jcouv, line 62: "must have the same type"

I think this should be adjusted. For example, the following should be allowed:
```
public partial string { get => ""; }
public partial string? { get; }
```

We should follow what partial methods do.

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
...
+partial class C3
+{
+    public partial string Prop { get; }
+
+    // Error: implementation of 'Prop' cannot have a 'set' accessor because the definition does not have a 'set' accessor.
+    public partial string Prop { get => field; set => field = value; }
+}
+```
+
+### Indexers
```
jcouv, line 94: Let's flesh out the indexers section too. For instance, the grammar.

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
...
+```diff
+property_declaration
+-    : attributes? property_modifier* type member_name property_body
++    : attributes? property_modifier* 'partial'? type member_name property_body
+    ;  
+```
+
+**Remarks**: This is similar to how *method_header* [(§14.6.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1461-general) and *class_declaration* [(§14.2.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1421-general) are specified. (Note that [Issue #946](https://github.com/dotnet/csharplang/issues/946) proposes to relax the ordering requirement, and would probably apply to all declarations which allow the `partial` modifier.)
+
+### Defining and implementing declarations
```
jcouv, line 17: We need to add rules for `property_initializer`

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
...
+    public C() // warning CS8618: Non-nullable property 'Prop' must contain a non-null value when exiting constructor. Consider declaring the property as nullable.
+    {
+    }
+}
+```
+
+### Matching signatures
+The property declarations must have the same type and ref kind. The property declarations must have the same set of accessors.
+
+The property declarations and their accessor declarations must have the same modifiers, though the modifiers may appear in a different order. This does not apply to the `extern` modifier, which may only appear on an *implementing declaration*.
```
jcouv, line 64: FWIW, it looks like partial methods allow some variation in `scoped` (see usage of `CheckValidScopedOverride` in `PartialMethodChecks`)
Some variation in tuple names is also allowed.

jcouv (commented):
Done with review pass (iteration 4)

proposals/partial-properties.md
```diff
@@ -0,0 +1,142 @@
...
+The *property_declaration* grammar [(§14.7.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1471-general) is updated as follows:
+
+```diff
+property_declaration
+-    : attributes? property_modifier* type member_name property_body
++    : attributes? property_modifier* 'partial'? type member_name property_body
+    ;  
+```
+
+**Remarks**: This is similar to how *method_header* [(§14.6.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1461-general) and *class_declaration* [(§14.2.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1421-general) are specified. (Note that [Issue #946](https://github.com/dotnet/csharplang/issues/946) proposes to relax the ordering requirement, and would probably apply to all declarations which allow the `partial` modifier.)
```
jcouv, line 15: FWIW, it looks like some of those links are out-of-date. But I think it's unavoidable as the standard evolves... :-/

proposals/partial-properties.md
RikkiGibson: My preference here would be to use similar requirements as the "partial method 'strict' comparison", except to report any difference under this comparison as an error. Essentially, let's tolerate nullable oblivious differences, but almost nothing else, as much as is practical.

I'll try and look for a straightforward way to specify this.

jaredpar: Keep in mind that one of our goals of NRT is that user code and generated code can adopt NRT independently of each other. A `partial` member is squarely in the generated code scenario. The decision we make here needs to facilitate users and generators adopting NRT at different times. 

Given that I think it makes the most sense to allow NRT differences between the `partial` definition and declaration.

RikkiGibson: I think it would be reasonable to:
1. give no diagnostics on nullable-oblivious difference. This addresses the needs of source generators.
2. give a *warning* diagnostic on any nullable-enabled difference (annotated vs. not-annotated).
3. The declaration part "wins" re: what types are actually used in the signature of the property for use sites and emit, just as with partial method declarations.

The reason I phrased (2) in this particular way is, I don't want to have to specify the conditions under which a nullability difference is considered "safe" across partial property declarations, which would include considerations around the set of accessors (is the property readable? writable?) and so on.

RikkiGibson: If you don't want to know about nrt, then you need to disable nullable, and you will never get a warning resulting from (2).

jaredpar: Consider the case where a generator doesn't know about NRT. Now the developer adds NRT annotations to a partial property. If the warnings are in the generated file then only the generator can fix them. That goes against the "adopt independently" goal.

jcouv: I would stick to the same rules as partial methods (which I think are the rules of overrides, see `SourceMemberContainerTypeSymbol.CheckValidNullableMethodOverride` used in `PartialMethodChecks`):
```
    // no warning
    public partial string? M();
    public partial string M() { return ""; }

    // warning
    public partial string M();
    public partial string? M() { return ""; }
```
With the exception of null-oblivious, no variance would be allowed for a property with both getter and setter, but some variance would be allowed for read-only or write-only properties.

RikkiGibson:
```cs
    // no warning
    public partial string? M();
    public partial string M() { return ""; }
```
Try this in a console app. It is a warning.

RikkiGibson: What I am intending to propose at this point, is something which falls entirely along the precedent that we have for partial methods, in terms of which scenarios receive warnings or errors by default, and where those diagnostics are reported. This is a precedent which we have shipped and through the following years have found that it works for all the usage scenarios that we care about. There is nothing about the nature of partial properties which fundamentally changes the question of which signature differences to warn/error about or not, or where to report the warnings.

The only difference I am putting forward on the properties side is that I want to simplify the design in comparison to partial methods, by not having both "strict" versus "legacy" mode of comparison, and essentially only having one mode equivalent to "strict" for methods. This is because the strict comparison for partial methods was only added later and in a warning wave for compat purposes. There is no such compat concern for partial properties.

jcouv: I'd tried in sharplab (no warning) but in VS I see the warning (CS8826 `WRN_PartialMethodTypeDifference`). Sharplab seems to not have warning waves turned on...
Sounds good. Thanks for the clarification :-)

proposals/partial-properties.md
RikkiGibson: I would like to address (1) in a subsequent PR, and to do so for both properties and methods in C# 13. I recall soft approval from LDM for this, and at the time I make the change I'll also follow up by email to confirm that approval.

I would prefer to not make changes for (2) at this time, as I think the spec is conveying what it is intending to with adequate clarity, and that we'll address the concern before the spec is committed to its final state for C# 13.

proposals/partial-properties.md
RikkiGibson: Yes, but we can probably reduce the breaks by preferring references to `standard-` instead of `draft-` branches in the future. I'll go ahead and update the references now.

proposals/partial-properties.md
jcouv: Were you able to dig up our rules for `scoped` differences?

jcouv (commented):
Done with review (iteration 5)

proposals/partial-properties.md
RikkiGibson: Scoped is just a modifier, I am expecting the normal rules for modifier differences to apply.

RikkiGibson: This can be observed for partial methods here

```cs
partial class C
{
    public partial Span<int> M(Span<int> x);
    public partial Span<int> M(scoped Span<int> x) => default; // The 'scoped' modifier of parameter 'x' doesn't match partial method declaration.CS8988
    
    public partial Span<int> M2(scoped Span<int> x);
    public partial Span<int> M2(Span<int> x) => default; // The 'scoped' modifier of parameter 'x' doesn't match partial method declaration.CS8988
}
```

jcouv: It looks like we allow some variations specifically for `scoped` on partial method, so the normal rule of "must have the same modifiers" doesn't seem right

RikkiGibson: I am not aware of any permitted variations on `scoped` modifiers on partial methods. The method in the implementation which checks this is called `CheckValidScopedOverride`, which may be misleading, but it is also passing argument `allowVariance: false`. In other words, scoped-ness of parameters across partial methods must be exactly the same, otherwise an error (not a warning) occurs.

jcouv: Ah, thanks for catching that. Cool :-)

jcouv (approved):
LGTM Thanks (iteration 5)

PR #8099, opened 2024-05-07
Update README.md to point to recently added notes

alrz:


meetings/2024/README.md
```diff
@@ -58,32 +58,41 @@ All schedule items must have a public issue or checked-in proposal that can be l
 
 - Meeting on meeting (Mads)
 
+
+### Mon Feb 26, 2024 (unresolved topics)
+
+- Breaking change with collection expressions and UI collection types (https://github.com/dotnet/roslyn/issues/72098).  Cyrus + Chuck
+- Dictionary-expressions + Collection-expressions future: WG to go over our plans, and the initial design space (Cyrus)
```
cston, line 65: The first item has been resolved and can be removed.

@CyrusNajmabadi, do we still need the second item? If not, the entire block can be removed.

333fred: This is effectively unchanged from the baseline. We can update the missing topics in a separate change.

333fred (approved):

333fred:
Thanks @alrz!

PR #8113, opened 2024-05-10
Use inline differences for specification updates in semi-auto-properties.md

cston:


proposals/semi-auto-properties.md
```diff
@@ -12,61 +12,31 @@ In these cases by now you always have to create an instance field and write the
...
+> **For auto-properties, any semicolon-only `get` accessor is implemented to read from, and any semicolon-only `set` accessor to write to its backing field.**
+> If the auto-property has ~~no set accessor~~ **only a semicolon-only get accessor**, the backing field is considered `readonly` ([§14.5.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/classes.md#1453-readonly-fields)).
+> Just like a `readonly` field, a getter-only auto-property **(without a set accessor or an init accessor)** can also be assigned to in the body of a constructor 
+> of the enclosing class. Such an assignment assigns directly to the ~~readonly~~ backing field of the property.
+> 
+> **The backing field can be referenced directly using the `field` keyword within all accessors and within the property expression body.**
+> **Because the field is unnamed, it cannot be used in a `nameof` expression.**
 
 ````diff
 +The following example:
```
RikkiGibson, line 42: If you don't mind, I think we should also remove the surrounding diff block from here, and the + at the start of every line, and just indicate this as being a wholly new section that should be added to the standard.

RikkiGibson (approved):

RikkiGibson:
I still find the "Automatically implemented properties" section difficult to read but I think that goes back to 2 things:
- The standard itself is not inserting enough line breaks.
- The changes to the standard here are fairly extensive.

I am wondering if this feature, in the end, should be spec'd such that "auto-properties" no longer exists as a concept, and "auto-accessors" exists instead, when an accessor has a semicolon body and the containing property meets certain conditions.

CyrusNajmabadi:
> I am wondering if this feature, in the end, should be spec'd such that "auto-properties" no longer exists as a concept, and "auto-accessors" exists instead

i like that idea :)

proposals/semi-auto-properties.md
cston: Removed `diff`.

cston:
@CyrusNajmabadi, please review, thanks.

proposals/semi-auto-properties.md
```diff
@@ -12,114 +12,85 @@ In these cases by now you always have to create an instance field and write the
 
 ## Specification changes
 
-The following changes are to be made to [§14.7.4](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/classes.md#1474-automatically-implemented-properties):
+[§14.7.1](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/classes.md#1471-general) *Properties - General*
+
+> A *property_initializer* may only be given for ~~an automatically implemented property, and~~ **a property that has a backing field that will be emitted and the property either does not have a setter, or its setter is auto-implemented. The *property_initializer*** causes the initialization of the underlying field of such properties with the value given by the *expression*.
```
CyrusNajmabadi, line 17: i honestly ca'nt made heads or tails of this sentence.  there are too many ands, ors, and nots.  And, based on my simple reading of this, this would disallow allow:

```c#
strnig X { get => field; set => field = value.Trim() } = "foo";
```

From my read, this clause doesn't apply "does not have a setter".  nor does "its setter is auto-implemented" apply.  

But the above should work.

Can we break this up into clearer rules?

proposals/semi-auto-properties.md
```diff
@@ -12,114 +12,85 @@ In these cases by now you always have to create an instance field and write the
...
+
+[§14.7.4](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/classes.md#1474-automatically-implemented-properties) *Automatically implemented properties*
+
+> An automatically implemented property (or ***auto-property*** for short), is a non-abstract non-extern
+> property with ~~semicolon-only accessor bodies. Auto-properties must have a get accessor and can optionally~~
+> ~~have a set accessor.~~ **either or both of:**
+> 1. **an accessor with a semicolon-only body**
+> 2. **usage of the `field` contextual keyword within the accessors or**
+>    **expression body of the property. The `field` identifier is only considered the `field` keyword when there is**
+>    **no existing symbol named `field` in scope at that location.**
```
CyrusNajmabadi, line 27: i thought we were gonig the breaking change route with `field`.  are we going back on that?

proposals/semi-auto-properties.md
```diff
@@ -12,114 +12,85 @@ In these cases by now you always have to create an instance field and write the
...
+> 
+> When a property is specified as an ~~automatically implemented property~~ **auto-property**, a hidden **unnamed** backing field is automatically
+> available for the property ~~, and the accessors are implemented to read from and write to that backing field~~.
```
CyrusNajmabadi, line 30: i would leave out. ` and the accessors are implemented to read from and write to that backing field~~`.  it redudnant with the two sentences after.

proposals/semi-auto-properties.md
```diff
@@ -12,114 +12,85 @@ In these cases by now you always have to create an instance field and write the
...
+> **For auto-properties, any semicolon-only `get` accessor is implemented to read from, and any semicolon-only**
+> **`set` accessor to write to its backing field.**
+> 
+> **The backing field can be referenced directly using the `field` keyword**
+> **within all accessors and within the property expression body. Because the field is unnamed, it cannot be used in a**
+> **`nameof` expression.**
+> 
+> If the auto-property has ~~no set accessor~~ **only a semicolon-only get accessor**, the backing field is considered `readonly` ([§14.5.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/classes.md#1453-readonly-fields)).
+> Just like a `readonly` field, a getter-only auto property **(without a set accessor or an init accessor)** can also be assigned to in the body of a constructor
+> of the enclosing class. Such an assignment assigns directly to the ~~readonly~~ backing field of the property.
```
CyrusNajmabadi, line 40: i'd prefer to say that `get;` is shorthand for `get => field;` and if there is no setter then the backing field is readonly if the getter doesn't mutate `field`.  It is not readonly if it does.

proposals/semi-auto-properties.md
cston: My intent for the PR was to simply re-format the existing statements that are in `diff` blocks as *inline diffs* - that is, using ~~strike-through~~ for deletions and **bold* for additions.

We can certainly modify the existing statements, but I'd prefer to do that in a follow-up change.

cston:
@CyrusNajmabadi, let's make changes to the actual text (rather than the formatting) in a separate PR.

CyrusNajmabadi:
i guess i'm confused at what i'm seeing:

image

It looks like the content is changing.

cston:
> It looks like the content is changing.

In general, the content should not have changed, although I fixed or removed broken links, and removed a duplicate sentence.

CyrusNajmabadi:
I'm confused then.  The screenshot i took seems to show the content changing.

cston:
> The screenshot i took seems to show the content changing.

It looks like the screenshot is displaying the `diff` block before the change. Viewing the PR diffs side-by-side might make the PR differences clearer.

proposals/semi-auto-properties.md
CyrusNajmabadi: gotcha.

CyrusNajmabadi (approved):

PR #8121, opened 2024-05-14
Add extension codegen details

jcouv:

### DIFF
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -1130,17 +1130,6 @@ Instance methods/accessors in the extension type are emitted as static
 methods with an additional parameter for the instance.  
 The instance parameter is by-value for reference types and `ref` otherwise.
 
-## Type references
-
-TODO4 this section is not finalized
-
-Extensions appearing in signatures are emitted as the extension's underlying type
-marked with a modopt of the extension type.
-
-```
-void M(Extension r) // emitted as `void M(modopt(Extension) UnderlyingType r)`
-```
-
 ### Extension type members
 
 #### Fields
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -1219,34 +1219,68 @@ implicit extension E for int
 
 # Implementation details
 
-TODO3(static) revise this to use a regular struct
+Extensions are emitted as structs with an extension marker method and an instance field for non-static extensions.  
+The type is marked with Obsolete and CompilerFeatureRequired attributes.  TODO we'll need to relax that
 
-struct Extension
-{
-  UnderlyingType underlyingValue;
-}
-// Avoid copying via Unsafe.As 
+## Marker method
 
-Extensions are implemented as ref structs with an extension marker method.  
-The type is marked with Obsolete and CompilerFeatureRequired attributes.  
 The extension marker method encodes the underlying type and base extensions as parameters in that order.  
-The marker method is called `<ImplicitExtension>$` for implicit extensions and 
+It is private and static, and is called `<ImplicitExtension>$` for implicit extensions and 
 `<ExplicitExtension>$` for explicit extensions.  
+
 For example: `implicit extension R for UnderlyingType : BaseExtension1, BaseExtension2` yields
 `private static void <ImplicitExtension>$(UnderlyingType, BaseExtension1, BaseExtension2)`.  
 
-If the extension has any instance member, then we emit a ref field (of underlying type)
-into the ref struct and a constructor.  
-TODO2(instance) The wrapping can be done with a static unspeakable factory method  
+## Instance field
 
-Values of extension types are left as values of the underlying value type, until an extension
-member is accessed. When an extension member is accessed, an extension instance is created
-with a reference to the underlying value and the member is accessed on that instance.
+If the extension not static, then we emit a private instance field of the underlying type
+into the struct.  
+We use the default layout for structs (sequential layout, with pack and size 0)
+ensuring that the single instance field is placed at offset zero.
 
+Note: Although we could not find an explicit statement to that effect,
+this behavior falls out from ECMA 335 (II.10.1.2), which states for sequential layout:
+> The CLI shall lay out the fields in sequential order, based on the order of the fields in the logical metadata table (§II.22.15).
+
+> [Rationale: ... sequential layout is intended to instruct the CLI to match layout rules commonly followed by languages like C and C++
+> on an individual platform, where this is possible while still guaranteeing verifiable layout. ...]
+
+and from the C99 standard section 6.7.2.1 bullet point 13 (gated):
+
+> Within a structure object, the non-bit-field members and the units in which bit-fields 
+> reside have addresses that increase in the order in which they are declared. 
+> A pointer to a structure object, suitably converted, points to its initial member 
+> (or if that member is a bit-field, then to the unit in which it resides), 
+> and vice versa. There may be unnamed padding within a structure object, but not at its beginning.
+
+For example `implicit extension E for UnderlyingType` yields
+```csharp
+struct E
+{
+	private UnderlyingType <UnderlyingInstance>$;
+	private static void <ImplicitExtension>$(UnderlyingType) { }
+}
 ```
-Extension r = default(UnderlyingType); // emitted as a local of type `UnderlyingType`
-r.ExtensionMember(); // emitted as `new Extension(ref r).ExtensionMember();`
-```
+
+## Instance invocations
+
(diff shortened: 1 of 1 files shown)


proposals/extensions.md
```diff
@@ -1219,34 +1219,68 @@ implicit extension E for int
...
-into the ref struct and a constructor.  
-TODO2(instance) The wrapping can be done with a static unspeakable factory method  
+## Instance field
 
-Values of extension types are left as values of the underlying value type, until an extension
-member is accessed. When an extension member is accessed, an extension instance is created
-with a reference to the underlying value and the member is accessed on that instance.
+If the extension not static, then we emit a private instance field of the underlying type
+into the struct.  
+We use the default layout for structs (sequential layout, with pack and size 0)
```
jcouv, line 1238: Tagging @davidwrighton @jkotas to confirm that we can rely on this layout (first field is guaranteed to be at offset zero).

proposals/extensions.md
```diff
@@ -1219,34 +1219,68 @@ implicit extension E for int
...
+ensuring that the single instance field is placed at offset zero.
 
+Note: Although we could not find an explicit statement to that effect,
+this behavior falls out from ECMA 335 (II.10.1.2), which states for sequential layout:
```
jkotas, line 1242: The sequential/explicit/auto type layout attributes describe unmanaged layout of the type (ie what you get when you marshal that type using built-in runtime marshaller. If the type is blittable, the managed layout of the type is going to match that unmanaged. If the type is not blittable, the runtime is free to use any layout it wants for the type.

For example, consider the following struct. It is non-blittable structs. The comments have offsets assigned to the fields by the current CoreCLR type loader. You can see that it is anything but sequential:
```
struct MyStruct
{
    public int a; // offset 16
    public object b; // offset 0
    public int c; // offset 20
    public object d; // offset 8
}
```

proposals/extensions.md
```diff
@@ -1219,34 +1219,68 @@ implicit extension E for int
...
+}
 ```
-Extension r = default(UnderlyingType); // emitted as a local of type `UnderlyingType`
-r.ExtensionMember(); // emitted as `new Extension(ref r).ExtensionMember();`
-```
+
+## Instance invocations
+
+The design and layout of the instance field will allow us to re-interpret an instance of `UnderlyingType`
+as an instance of `E` with `public static ref TTo Unsafe.As<TFrom,TTo>(ref TFrom source)`.
```
jkotas, line 1268: Unsafe casts like this have a good chance of confusing the JIT type-based optimizations. I am not able to tell for sure from what is written here whether it can turn into a real problem or not. It is very subtle.

proposals/extensions.md
jkotas: (If you have a non-blittable sequential struct with a single field, I believe that all current runtimes are going to place that field at offset 0 in the managed view. It do not think it is written down anywhere. )

proposals/extensions.md
jkotas: The simple rule is: The unsafe casts are fine if they are just a performance optimization ie they are replacing an equivalent slower safe cast.

https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafe.as has more details.

proposals/extensions.md
```diff
@@ -1219,34 +1219,68 @@ implicit extension E for int
...
+If the receiver of an extension member invocation on a value of the underlying type is `r`,
+we will replace it with `Unsafe.As<UnderlyingType, E>(ref r)`.
+If `r` does not refer to a location, a temporary variable will be created and initialized, and its reference
+will be used.
+
+The re-interpreted receiver will be only computed once where possible. For example, in a compound assignment
+`r.P += value;`.
+
+TODO we may wrap this method in a compiler-generated helper to increase verifiability of the generated code.  
+TODO there is also a verifiability issue with taking a reference to `this` (which is readonly)
```
AlekseyTs, line 1279: There is a helper that allows to drop the `readonly` property of a `ref` (Unsafe.AsRef<T>), but in some scenarios I think we should capture `readonly` targets by value

proposals/extensions.md
huoyaoyuan: Some context: https://github.com/dotnet/runtime/pull/99778#issuecomment-2007289974

huoyaoyuan: Honestly, this is a reason why I expect extension to get more runtime support. The cast with underlying type can be canonicalized in this case.

proposals/extensions.md
```diff
@@ -1219,34 +1219,68 @@ implicit extension E for int
...
-r.ExtensionMember(); // emitted as `new Extension(ref r).ExtensionMember();`
-```
+
+## Instance invocations
+
+The design and layout of the instance field will allow us to re-interpret an instance of `UnderlyingType`
+as an instance of `E` with `public static ref TTo Unsafe.As<TFrom,TTo>(ref TFrom source)`.  
+
+If the receiver of an extension member invocation on a value of the underlying type is `r`,
+we will replace it with `Unsafe.As<UnderlyingType, E>(ref r)`.
```
agocke, line 1271: This is not going to work in trimming or Native AOT. If the IL never contains a reference to the private field of the lowered struct, then it will trim it away. This will break the layout and the runtimes will crash on invocation.

agocke: Actually, even with the marker we can't do anything for trimming/AOT in .NET 8. If someone compiles with extensions for netstandard2.0, and someone attempts to trim code with this pattern, they will hit code corruption and crash.

proposals/extensions.md
hamarb123: It might be good to just guarantee this specific case (1 field in a sequential struct) in the augments.

proposals/extensions.md
hamarb123: It would be great if there was a way to specify whether a reference is wanted or not for a struct extension's member. My understanding is that a number of VectorXxx members are extensions / static methods / operators because they can be passed by value instead of reference. And I've seen similar for enums too, along with other primitives. But there are also many scenarios where it's great to pass by reference instead (like calling an mutating interface member in a constrained way, or any other mutating method, or for very large structs). Don't really see how this could work with the current design though, unless we lowered these special (by value) instance methods/members as static methods with extra first parameter of extension type and some name prefix/suffix to indicate to the compiler that they're instance methods, but by value instead.

proposals/extensions.md
hamarb123: It should be able to work if it generates something like this in private implementation details, right?
```csharp
static class Helper<T1, T2>
{
    [DynamicDependency(.NonPublicFields, typeof(T2))] static ref T2 AsExtension(ref T1 ut) => ref Unsafe.As<T1, T2>(ref r);
    [DynamicDependency(.NonPublicFields, typeof(T1))] static ref T2 AsUnderlying(ref T1 ut) => ref Unsafe.As<T1, T2>(ref r);
}
```

proposals/extensions.md
```diff
@@ -1117,75 +1117,19 @@ implicit extension E for int
...
-
-Extensions appearing in signatures are emitted as the extension's underlying type
-marked with a modopt of the extension type.
-
-```
-void M(Extension r) // emitted as `void M(modopt(Extension) UnderlyingType r)`
-```
+Instance methods/accessors in the extension type are emitted as static
+methods with an additional parameter for the instance.  
+The instance parameter is by-value for reference types and `ref` otherwise.
```
Sergio0694, line 1132: Small question on "`ref` otherwise": what about invoking extension members on value types that are passed as `readonly` values, or `ref readonly` references? Will it be possible to mark instance extension members as `readonly` like with members of `struct` types, and have Roslyn omit defensive copies in the same way?

jcouv: That's an open issue. At the moment, the spec disallows `readonly` ("Member methods may not use the readonly modifier."). But I think we'll want to allow it at some point.

jcouv:
@AlekseyTs Please take another look. Thanks

proposals/extensions.md
```diff
@@ -1117,75 +1148,19 @@ implicit extension E for int
...
-
-TODO4 this section is not finalized
-
-Extensions appearing in signatures are emitted as the extension's underlying type
-marked with a modopt of the extension type.
-
-```
-void M(Extension r) // emitted as `void M(modopt(Extension) UnderlyingType r)`
-```
+Instance methods/accessors in the extension type are emitted as static
```
AlekseyTs, line 1161: I suggest to remove any details about instance to static translation for now. I think it will be better to add them separately later once the details are more o less crisp. Deleting information about the approach that we decided to abandon, is fine, of course

PR #8126, opened 2024-05-14
Remove mentions of extension inheritance and interfaces

jcouv:


KennethHoff:
This is worrying; what's the premise? Why's this being removed? 

Is it de-emphasizing implementing interfaces on behalf of another type for the first version, or is this a "we might not ever get this" (.. To a greater extent than the default "we might not ever get this" that all new features inherently bring)?

jcouv:
@KennethHoff I'm scoping down the spec to what we're targeting for first release.

KennethHoff:
Ah, that's what I was assuming, but I was worried, so I had to ask 😅

jcouv:
@AlekseyTs for review. Thanks

proposals/extensions.md
```diff
@@ -166,18 +150,12 @@ extension_modifier
     | 'partial'
     | 'unsafe'
     | 'static'
-    | 'protected'
```
AlekseyTs, line 169: It looks like the modifiers on a type are modified here. Did you really mean to do that?

AlekseyTs:

proposals/extensions.md
jcouv: That was incorrect. Restored. Thanks

AlekseyTs (approved):
LGTM (commit 3)

PR #8131, opened 2024-05-15
Spec partial properties behavior with doc comments

RikkiGibson:
Test plan: https://github.com/dotnet/roslyn/issues/73090

@jcouv @333fred

### DIFF
--- a/proposals/partial-properties.md
+++ b/proposals/partial-properties.md
@@ -127,6 +127,54 @@ partial class C4
 }
 ```
 
+### Documentation comments
+
+We want the behavior of doc comments on partial properties to be consistent with what we shipped for partial methods. That behavior is detailed in https://github.com/dotnet/csharplang/issues/5193.
+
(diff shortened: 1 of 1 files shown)


proposals/partial-properties.md
```diff
@@ -127,6 +127,54 @@ partial class C4
 }
 ```
 
+### Documentation comments
+
+We want the behavior of doc comments on partial properties to be consistent with what we shipped for partial methods. That behavior is detailed in https://github.com/dotnet/csharplang/issues/5193.
+
+It is permitted to include doc comments on either the partial or implementation part of a partial property. (Note that doc comments are not supported on property accessors.)
```
RikkiGibson, line 134:
```suggestion
It is permitted to include doc comments on either the definition or implementation part of a partial property. (Note that doc comments are not supported on property accessors.)
```

jcouv (approved):
LGTM Thanks (iteration 2)

333fred (approved):

Issue #946, opened 2017-09-28
Champion: relax ordering constraints around `ref` and `partial` modifiers on type declarations


agocke:
Currently, `partial` must appear directly before `struct`, `class`, or another type declaration keyword. If the type is a `ref` struct, `ref` must appear immediately before `partial` or `struct`. It seems likely that various other keywords could be used to disambiguate these contextual modifiers and allow us to relax the constraints on where `partial` and `ref` can appear in the modifier list.

[ ] Proposal added
[ ] Discussed in LDM
[ ] Decision in LDM
[ ] Finalized (done, rejected, inactive)
[ ] Spec'ed

alrz:
>Currently, partial must appear directly before struct, class

No idea why was that required from the beginning. Allowing `partial` among other modifiers didn't seem to be ambigious at any time. Just like `async` for methods.

CyrusNajmabadi:
Note: i've looked at https://github.com/dotnet/csharplang/blob/master/meetings/2017/LDM-2017-12-04.md, but i can't find where in it it refers to this issue.  Can you link/copy the appropriate section?

Thanks!

Joe4evr:
Can restrictions around `partial` just be relaxed entirely? I'd like it if `partial abstract class` worked just the same as `abstract partial class`.

Joe4evr:
> See https://github.com/dotnet/csharplang/blob/master/meetings/2017/LDM-2017-12-04.md

So that link does not talk about `partial` *anywhere* (as @CyrusNajmabadi noted), so I did a quick search for the paper trail myself:
* https://github.com/dotnet/csharplang/blob/main/meetings/2017/LDM-2017-09-27.md (which was before this issue)
* https://github.com/dotnet/csharplang/blob/main/meetings/2018/LDM-2018-04-04.md (where it was approved)
* https://github.com/dotnet/csharplang/blob/main/meetings/2019/LDM-2019-09-16.md (where it was initially slated for 9.0)
* https://github.com/dotnet/csharplang/blob/main/meetings/2020/LDM-2020-09-09.md (where it was punted to later)

alrz:
> It seems likely that various other keywords could be used to disambiguate these contextual modifiers


I think this is now far simpler with https://github.com/dotnet/csharplang/discussions/7033

Make `partial` a definite keyword in the context of a modifier-list and relax ordering.

PR #8134, opened 2024-05-17
Propose relaxed ordering for partial and ref modifiers

RikkiGibson:
dotnet/roslyn#73090
#946

Spec'ing this related small feature as the partial properties speclet said we would do.

### DIFF
--- a/proposals/partial-properties.md
+++ b/proposals/partial-properties.md
@@ -3,16 +3,28 @@ https://github.com/dotnet/csharplang/issues/6420
 
 ### Grammar
 
-The *property_declaration* grammar [(§14.7.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1471-general) is updated as follows:
+The partial modifier on properties is treated just like any other modifier and can be included in any position in a modifier list.
+
+The property grammar [(§15.7.1)](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/classes.md#1571-general) is updated as follows:
 
 ```diff
-property_declaration
--    : attributes? property_modifier* type member_name property_body
-+    : attributes? property_modifier* 'partial'? type member_name property_body
-    ;  
+ property_modifier
+     : 'new'
+     | 'public'
+     | 'protected'
+     | 'internal'
+     | 'private'
+     | 'static'
+     | 'virtual'
+     | 'sealed'
+     | 'override'
+     | 'abstract'
+     | 'extern'
++    | 'partial'
+     ;
 ```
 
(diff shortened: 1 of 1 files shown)


proposals/partial-properties.md
```diff
@@ -3,16 +3,28 @@ https://github.com/dotnet/csharplang/issues/6420
...
 
-The *property_declaration* grammar [(§14.7.1)](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1471-general) is updated as follows:
+The partial modifier on properties is treated just like any other modifier and can be included in any position in a modifier list.
+
+The property grammar [(§15.7.1)](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/classes.md#1571-general) is updated as follows:
 
 ```diff
-property_declaration
--    : attributes? property_modifier* type member_name property_body
-+    : attributes? property_modifier* 'partial'? type member_name property_body
```
RikkiGibson, line 11: reading diffs of diffs can be challenging even in simple cases like this, so consider just reading the old version then the new version to understand what's going on.

333fred (approved):
Love it.

RikkiGibson:
@jcouv suggested this aspect of the feature could probably wait till after merge to main. I will probably leave this PR open until it's time to do that work (which shouldn't be too long).

RikkiGibson:
We still want to do this but will be punting impl work on it out to .NET 10.

jcouv:
> We still want to do this but will be punting impl work on it out to .NET 10.

Should this go into a separate proposal then?

RikkiGibson:
Yeah I'll push such a change to this PR when I get the chance.

proposals/relaxed-partial-ref-ordering.md
```diff
@@ -0,0 +1,265 @@
...
+[motivation]: #motivation
+
+Most modifiers in the language can be written in any order. It feels arbitrary that `partial` and `ref` use a more stringent rule, and need to come at the end of the modifier list. This is especially cumbersome in the case of `ref partial` on structs, where even the opposite order `partial ref` is not permitted.
+
+In order to avoid getting in the user's way needlessly, we should allow `partial` and `ref` modifiers to appear in any position in a modifier list, just like the other modifiers.
+
+## Detailed design
+[design]: #detailed-design
+
+Note that some of the below grammars are from C# 7 and are missing more recent modifiers such as `required` and `file.`
```
jjonescz:
```suggestion
Note that some of the below grammars are from C# 7 and are missing more recent modifiers such as `required` and `file`.  
```

333fred (approved):

PR #8141, opened 2024-05-21
Update `is` pattern section of "first-class Span"

jjonescz:
I think `is` pattern cannot be supported. See [ref struct interface tests](https://github.com/dotnet/roslyn/blob/a2f37e7051d904aabb3648ddaaba49f2a14eced5/src/Compilers/CSharp/Test/Emit3/RefStructInterfacesTests.cs#L23461-L23522) for example. I assume it's not the intention to change the runtime or codegen to support this as part of the feature.

### DIFF
--- a/proposals/first-class-span-types.md
+++ b/proposals/first-class-span-types.md
@@ -62,37 +62,46 @@ as if the `T` was declared as `out T` in some scenarios. We do not, however, plu
 variance-convertible in [§18.2.3.3](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/interfaces.md#18233-variance-conversion). If in the future, we change the runtime
 to more deeply understand the variance here, we can take the minor breaking change to fully recognize it in the language.
 
-Practically, this will also mean that in pattern matching for generic scenarios, we'd have behavior as follows:
-
-```cs
-using System;
-
-M<object[]>(["0"]); // Does not print
-M<ReadOnlySpan<string>>(["1"]); // Does not print
-M<Span<object>>(["2"]); // Does not print
-M<ReadOnlySpan<object>>(["3"]); // Prints
-
-void M<T>(T t) where T : allows ref struct
-{
-    if (t is ReadOnlySpan<object> r) Console.WriteLine(r[0]);
-}
-```
-
-In array variance scenarios, this pattern would return true for all reference type arrays:
-
-```cs
-using System;
-
-M<object[]>(["0"]); // Prints
-M<string[]>(["1"]); // Prints
-
-void M<T>(T t)
-{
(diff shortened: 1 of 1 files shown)


proposals/first-class-span-types.md
```diff
@@ -62,37 +62,46 @@ as if the `T` was declared as `out T` in some scenarios. We do not, however, plu
...
+```cs
+class C<T> where T : allows ref struct
 {
-    if (t is ReadOnlySpan<object> r) Console.WriteLine(r[0]);
+    void M1(T t) { if (t is T x) { } } // ok
+    void M2(R r) { if (r is R x) { } } // ok
+    void M3(T t) { if (t is R x) { } } // error
+    void M4(R r) { if (r is T x) { } } // error
 }
+ref struct R { }
```
jaredpar, line 79: @AaronRobinsonMSFT wanted to double check with you that this is not intended to work at the IL level with your changes.

proposals/first-class-span-types.md
```diff
@@ -62,37 +62,46 @@ as if the `T` was declared as `out T` in some scenarios. We do not, however, plu
 variance-convertible in [§18.2.3.3](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/interfaces.md#18233-variance-conversion). If in the future, we change the runtime
 to more deeply understand the variance here, we can take the minor breaking change to fully recognize it in the language.
 
-Practically, this will also mean that in pattern matching for generic scenarios, we'd have behavior as follows:
+There is also an open question below about participation in delegate signature matching.
```
333fred, line 65: This line can actually be removed.

proposals/first-class-span-types.md
```diff
@@ -62,37 +62,46 @@ as if the `T` was declared as `out T` in some scenarios. We do not, however, plu
...
 
-```cs
-using System;
+#### Patterns
 
-M<object[]>(["0"]); // Does not print
-M<ReadOnlySpan<string>>(["1"]); // Does not print
-M<Span<object>>(["2"]); // Does not print
-M<ReadOnlySpan<object>>(["3"]); // Prints
+Note that `ref struct`s are not supported in patterns, except for trivial identity patterns:
```
333fred, line 69: I don't think is quite correct. For example, this works:

```cs
Span<int> ints = [1, 2, 3];

if (ints is [1, 2, 3])
{
    Console.WriteLine("Here");
}
```

I think this needs to be more specific: When used as a `type` in any pattern, only identity conversions are allowed.

333fred (approved):

proposals/first-class-span-types.md
AaronRobinsonMSFT: Correct.

PR #8149, opened 2024-05-23
Add betterness rule for first-class Span

jjonescz:


cston (approved):

proposals/first-class-span-types.md
```diff
@@ -109,6 +109,55 @@ The compiler expects to use the following helpers or equivalents to implement th
...
+> Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a *better conversion* than `C₂` if one of the following holds:
+>
+> - `E` is a *collection expression* and one of the following holds:
+>   - `T₁` is `System.ReadOnlySpan<E₁>`, and `T₂` is `System.Span<E₂>`, and an implicit conversion exists from `E₁` to `E₂`.
+>   - `T₁` is `System.ReadOnlySpan<E₁>` or `System.Span<E₁>`, and `T₂` is an *array_or_array_interface* with *element type* `E₂`, and an implicit conversion exists from `E₁` to `E₂`.
+>   - `T₁` is not a *span_type*, and `T₂` is not a *span_type*, and an implicit conversion exists from `T₁` to `T₂`.
+> - `E` is not a *collection expression* and one of the following holds:
+>   - `E` exactly matches `T₁` and `E` does not exactly match `T₂`
+>   - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂`
+>   - **the rules above do not hold and `C₁` is an implicit span conversion and `C₂` is not an implicit span conversion**
```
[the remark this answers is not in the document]
cston: Let's capture this breaking change example in this spec so we don't lose it.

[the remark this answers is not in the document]
jjonescz: Captured, thanks. But I think we should try to update some rules to avoid this break. It seems like a common scenario. Also if the BCL tried to mitigate this break, I think they would need to add a special overload like `ArrayExtensions.Contains<T>(T[], T)` which is exactly what we want to prevent with this feature.

cston (approved):

proposals/first-class-span-types.md
```diff
@@ -121,6 +121,55 @@ The compiler expects to use the following helpers or equivalents to implement th
...
+
+> Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a *better conversion* than `C₂` if one of the following holds:
+>
+> - `E` is a *collection expression* and one of the following holds:
+>   - `T₁` is `System.ReadOnlySpan<E₁>`, and `T₂` is `System.Span<E₂>`, and an implicit conversion exists from `E₁` to `E₂`.
+>   - `T₁` is `System.ReadOnlySpan<E₁>` or `System.Span<E₁>`, and `T₂` is an *array_or_array_interface* with *element type* `E₂`, and an implicit conversion exists from `E₁` to `E₂`.
+>   - `T₁` is not a *span_type*, and `T₂` is not a *span_type*, and an implicit conversion exists from `T₁` to `T₂`.
+> - `E` is not a *collection expression* and one of the following holds:
+>   - `E` exactly matches `T₁` and `E` does not exactly match `T₂`
+>   - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂`
```
cston, line 137: >- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` [](http://example.com/codeflow?start=4&length=104)

In the call to `M(new string[0])`, will we prefer `M(object[])` over `M(ReadOnlySpan<object>)`? If so, is that intended?

jjonescz: According to the spec here we would prefer `object[]` since it's implicitly convertible to `ROS<object>` (and not the other way around) hence it's a better conversion target.

The intention was to prefer implicit span conversion, so this goes against that intention. I will update the rule. Thanks.

I'm not sure if in general it makes sense for `ROS<object>` to be the better target for `string[]` than `object[]`. I guess one potential advantage of ROS is that it's readonly, unlike `object[]` which the method could potentially assign a non-string object into?

proposals/first-class-span-types.md
```diff
@@ -121,6 +121,58 @@ The compiler expects to use the following helpers or equivalents to implement th
...
+
+> Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a *better conversion* than `C₂` if one of the following holds:
+>
+> - `E` is a *collection expression* and one of the following holds:
+>   - `T₁` is `System.ReadOnlySpan<E₁>`, and `T₂` is `System.Span<E₂>`, and an implicit conversion exists from `E₁` to `E₂`.
+>   - `T₁` is `System.ReadOnlySpan<E₁>` or `System.Span<E₁>`, and `T₂` is an *array_or_array_interface* with *element type* `E₂`, and an implicit conversion exists from `E₁` to `E₂`.
+>   - `T₁` is not a *span_type*, and `T₂` is not a *span_type*, and an implicit conversion exists from `T₁` to `T₂`.
+> - `E` is not a *collection expression* and one of the following holds:
+>   - `E` exactly matches `T₁` and `E` does not exactly match `T₂`
+>   - **`E` exactly matches both or neither of `T₁` and `T₂`,
```
cston, line 137: >both or [](http://example.com/codeflow?start=28&length=7)

Can "both or" be dropped? (If `E` exactly matches both, then won't `C1` and `C2` both be something other than implicit span conversions?)

jjonescz: That should work. Thanks.

cston (approved):

cston (approved):

PR #8194, opened 2024-06-08
Add extension WG meeting notes on type erasure codegen

jcouv:

### DIFF
--- a/meetings/working-groups/roles/extension-wg-2024-06-07.md
+++ b/meetings/working-groups/roles/extension-wg-2024-06-07.md
@@ -6,7 +6,7 @@
 We need to encode the extension type (for extension-aware compiler) and the underlying type (erasure and for non-extension-aware tools).
 
 We'll need to encode tuple names, dynamic bits, nullability, native integer for both types. 
-Ideally, this should be done in a way that leave any tools that understand current attribute/constructor.
+Ideally, this should be done in a way that doesn't break tools that understand the current attribute/constructor encoding.
 
 Options for encoding extension type:
 - generic attribute (pinging David to ask limitations/bugs on older frameworks)
@@ -33,6 +33,7 @@ override void M2(E e)
 
 Con: binary break to change API from extension to underlying type and back
 Con: problem for override at language level (might be solvable)
+Con: need to new solution encode the second set of tuples names/dynamic/nullability
 
 ### Generic attribute
 ```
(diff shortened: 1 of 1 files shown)


meetings/working-groups/roles/extension-wg-2024-06-07.md
```diff
@@ -0,0 +1,146 @@
...
+```
+virtual void M2(C c)
+// other assembly
+override void M2(E e)
+```
+
+Con: binary break to change API from extension to underlying type and back
+Con: problem for override at language level (might be solvable)
+
+### Generic attribute
```
hez2010, line 37: Another con would be open generics support in generic attributes is not present yet, so a generic extension cannot use it. I'm adding the support in the runtime, so that generic extensions can use it in the future: https://github.com/dotnet/runtime/pull/103213

meetings/working-groups/roles/extension-wg-2024-06-07.md
```diff
@@ -0,0 +1,146 @@
+
+# Extensions WG 2024-06-07
+
+## Type erasure encoding
+
+We need to encode the extension type (for extension-aware compiler) and the underlying type (erasure and for non-extension-aware tools).
+
+We'll need to encode tuple names, dynamic bits, nullability, native integer for both types. 
+Ideally, this should be done in a way that leave any tools that understand current attribute/constructor.
```
jaredpar, line 9: This sentence seems incomplete. Did you means something like 

> leaves any tools that understand current attribute/constructor functioing?

meetings/working-groups/roles/extension-wg-2024-06-07.md
```diff
@@ -0,0 +1,146 @@
...
+
+// codegen
+void M([Extension<E>] Underlying e) // System.Extension`1 (new)
+```
+Attribute doesn't exist downlevel (we should synthesize)
+Generic attribute not supported before .NET 7 (prove there's enough support, or block off)
+
+### Attribute with typeof
+
+Limitation with typeof in attribute referencing type parameter (typeof encoded as fully-qualified name, context-free, but no such syntax exists for type parameters)
```
jaredpar, line 50: Won't this also lose the attributes that we use to encode nullability, dynamic, etc ... Basically I'm thinking of this as the following:

```csharp
// user writes
extension E<T> for List<T> { ... }
E<dynamic> M() => ...;

// compiler emits
[typeof(E<object>)]
```

Believe that essentially this approach also loses information encoded in attributes.

PR #8209, opened 2024-06-12
Extract open issues and organize Expressions section

jcouv:

### DIFF
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -342,12 +342,16 @@ Will need to check with the WG as I don't recall the reasoning for this.
 
 The current proposal is to use an attribute with a string representing the type with extensions un-erased.
 For example: `void M(E)` would be emitted as `void M([Attribute("E")] UnderlyingType)`.
+
 The serialization format would be the same as the one used for `typeof` in attributes, but with support for type parameters (using `T!` and `T!!` syntax).  
 Note: the serialization format does not support function pointers at the moment. Tracked by https://github.com/dotnet/roslyn/issues/48765
 
 The attribute would also encode the tuple names, dynamic, native integer and nullability information for the type with extensions un-erased.
 For example: `void M(E<dynamic>)` with `C<(dynamic a, dynamic b)>` as the underlying type for `E<dynamic>` would be emitted as
 `void M([Attribute("E<object>"", TupleNames = "..."}] [... existing attributes for dynamic and tuple names ... ] C<ValueTuple<object, object>>)`.  
+
+Should we instead extend support for type parameters in `typeof` in attributes (including runtime/reflection),
+and use a `Type` parameter in the attribute constructor?
 
 ## Summary
 [summary]: #summary
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -1,7 +1,5 @@
 # Extension types
 
-TODO(static) extensions are disallowed within interfaces with variant type parameters  
-
 ## Open issue: merging extension methods and extension members
 
 We need to revise the preference of extension types over extension methods (should mix and disambiguate duplicates if needed instead)
@@ -270,11 +268,11 @@ Need to spec why extension properties are not found during lookup for attribute 
 From WG discussion, we'd allow attributes on extensions types, using the AttributeTargets.Class target,
 since extensions are emitted as static classes.
 
-## Open issue: usings
+## Open issue: extension member lookup in usings
 
-TODO
-static usings
-cycles
+We don't resolve implicit extension members in usings, due to cycles.  
+This needs to be investigated further to disentangle whether all cycles are implementation-specific
+or some are inherent to the language feature.
 
 ## Open issue: need to specify scoping rules for `for UnderlyingType`
 
@@ -323,7 +321,7 @@ void M<T>(T t1, T t2) { }
 
 Should we have a naming convention like `Extension` suffixes? (`DataObjectExtension`)  
 
-### Open issue: need to specify pattern-based invocations and member access
+## Open issue: need to specify pattern-based invocations and member access
 
 We need to scan through all pattern-based rules for known members to consider whether to include extension type members.
   If extension methods were already included, then we should certainly include extension type methods.
@@ -335,6 +333,21 @@ We need to scan through all pattern-based rules for known members to consider wh
   - `GetPinnableReference` in fixed statements
   - `GetAwaiter` in `await` expressions
   - `Dispose` in `using` statements
+
+## Open issue: need to disallow extensions within interfaces with variant type parameters  
+
+Will need to check with the WG as I don't recall the reasoning for this.
+
+## Open issue: need to specify the type erasure design
+
+The current proposal is to use an attribute with a string representing the type with extensions un-erased.
+For example: `void M(E)` would be emitted as `void M([Attribute("E")] UnderlyingType)`.
+The serialization format would be the same as the one used for `typeof` in attributes, but with support for type parameters (using `T!` and `T!!` syntax).  
+Note: the serialization format does not support function pointers at the moment. Tracked by https://github.com/dotnet/roslyn/issues/48765
+
+The attribute would also encode the tuple names, dynamic, native integer and nullability information for the type with extensions un-erased.
+For example: `void M(E<dynamic>)` with `C<(dynamic a, dynamic b)>` as the underlying type for `E<dynamic>` would be emitted as
+`void M([Attribute("E<object>"", TupleNames = "..."}] [... existing attributes for dynamic and tuple names ... ] C<ValueTuple<object, object>>)`.  
 
 ## Summary
 [summary]: #summary
@@ -889,7 +902,7 @@ implicit extension E for C
 ```
 
 #### Extension indexer access
-TODO3(instance) write this section, including preference for more specific extension indexers
+TODO write this section, including preference for more specific extension indexers
 
 In an element access of one of the forms
 
@@ -1237,11 +1250,8 @@ TODO3
 
 ### Method group conversions
 
-TODO4 spec this section
-
-https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/conversions.md#108-method-group-conversions
-TODO A single method is selected corresponding to a method invocation, 
-but with some tweaks related to normal form and optional parameters.  
+No changes to [method group conversion rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/conversions.md#108-method-group-conversions).  
+The extension behavior falls out of method invocation rules.
 
 ### Simple assignment
 
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -1,15 +1,10 @@
 # Extension types
 
-TODO(instance) need to spec why extension properties are not found during lookup for attribute properties, or explicitly disallow them  
-TODO(static) adjust scoping rules so that type parameters are in scope within the 'for'  
-TODO2 check Method type inference: 7.5.2.9 Lower-bound interfaces  
 TODO(static) extensions are disallowed within interfaces with variant type parameters  
-TODO2 We should likely allow constructors and `required` properties  
-TODO attributes and attribute targets  
 
 ## Open issue: merging extension methods and extension members
 
-TODO(instance) revise preference of extension types over extension methods (should mix and disambiguate duplicates if needed instead)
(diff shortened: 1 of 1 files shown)


proposals/extensions.md
```diff
@@ -601,41 +760,61 @@ The preceding rules mean:
 - and that extension methods declared directly in a namespace take precedence
   over extension methods imported into that same namespace with a using namespace directive.
 
-TODO(static) clarify behavior for extension on `object` or `dynamic` used as `dynamic.M()`?
```
jcouv, line 604: :memo: This is converted to a test with a PROTOTYPE comment about diagnostic quality in [PR](https://github.com/dotnet/roslyn/pull/73994)

MadsTorgersen (approved):
This is a good restructuring! Makes it a lot easier to get an overview on the current state of design, and which are the most pressing issues.

proposals/extensions.md
```diff
@@ -116,6 +109,254 @@ public implicit extension E for S
...
+  - `Add` in collection initializers
+  - `GetPinnableReference` in fixed statements
+  - `GetAwaiter` in `await` expressions
+  - `Dispose` in `using` statements
+
+## Open issue: need to disallow extensions within interfaces with variant type parameters  
+
+Will need to check with the WG as I don't recall the reasoning for this.
+
+## Open issue: need to specify the type erasure design
```
jjonescz, line 341: What's the difference between this and "Open issue: need to specify type erasure" above?

proposals/extensions.md
```diff
@@ -632,41 +855,62 @@ The preceding rules mean:
...
-TL;DR: If no candidate is applicable, then we attempt extension indexer access instead.
+TL;DR: For non-extension types, we'll fall back to an implicit extension member lookup. For extension types, we include indexers from the underlying type.
 
 We modify the [element access rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#128111-general) as follows:
 
-/[...]
 An *element_access* is dynamically bound if \[...]
 If the *primary_no_array_creation_expression* of an *element_access* is a value of an *array_type*, the *element_access* is an array access. 
-Otherwise, the *primary_no_array_creation_expression* shall be a variable or value of a class, struct, or interface type  TODOTODO
+Otherwise, the *primary_no_array_creation_expression* shall be a variable or value of a class, struct, interface, /***or extension** type
```
jjonescz, line 866: Did you mean to write `\*` instead of `/*` ? But I don't know why would that be needed, this seems to work fine:

```suggestion
Otherwise, the *primary_no_array_creation_expression* shall be a variable or value of a class, struct, interface, **or extension** type
```

Applies to other similar changes throughout the doc - I see usages of `/*` and `/[` that were probably meant to be `\*` and `\[` (although the escaping might be also unnecessary).

jcouv: Tried again and it looks like the escape sequence is no longer needed. Removed :-)

proposals/extensions.md
jcouv: Merged now. Thanks

proposals/extensions.md
```diff
@@ -500,7 +497,7 @@ We modify the [accessibility constraints](https://github.com/dotnet/csharpstanda
 
 The following accessibility constraints exist:
 - [...]
-- \***The underlying type of an extension type shall be at least as accessible as the extension type itself.**
+- ***The underlying type of an extension type shall be at least as accessible as the extension type itself.**
```
jjonescz, line 500: Nit: it seems excessive to use three `*` - the first is rendered as a literal `*` and the remaining two make the text bold:

image

jcouv: This is intended. If found that the bold is not always easy to see, so the `*` helps spot additions even when the bold isn't.
It looks like GitHub may have adjusted their fonts and the bold seems easier to see than before, but I think it's better to keep for now. It'll be easy to remove if we decide this extra formatting isn't necessary.

jcouv:
@AlekseyTs for review. Thanks

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+
+static class Extensions
+{
+    public static void M(this C c) { }
+}
+```
+
+## Open issue: removing zero-arity-matches-any from var scenario
+
+By the current rules, there are some unfortunate interactions between member access and natural function type.
```
AlekseyTs, line 134: >there are some unfortunate interactions between member access and natural function type [](http://example.com/codeflow?start=22&length=87)

This is not specific. What interactions? Why might one consider them unfortunate?

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+> Note: When the result of such a member lookup is a method group and K is zero,
```
AlekseyTs, line 135: It is not obvious what is the context for the quote (where it is taken from) and how it is relevant to the issue.

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+> the method group can contain methods having type parameters.
+> This allows such methods to be considered for type argument inferencing. end note
+
+I would like to brainstorm tweaks to member access and natural function type to make this work better.
```
AlekseyTs, line 139: Better is a subjective term, consider describing the behavior that you think is better.

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+
+```csharp
+var x = C.Member; // error: member lookup finds C.Member (method group) and lacks type arguments to apply to that match
```
AlekseyTs, line 142: It might help including what does the error say exactly

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+
+## Open issue: what kind of type is an extension type?
+
+Many sections of the spec need to consider the kind of type. Consider a few examples:  
+- the spec for a conditional element access `P?[A]B`
+considers whether `P` is a nullable value type. So it will need to handle the case
+where `P` is an instance of an extension type on a nullable value type,
+- the spec for an object creation considers whether the type is
+a value_type, a type_parameter, a class_type or a struct_type,
+- the spec for satisfying constraints also consider what kind of type were dealing with.
```
AlekseyTs, line 195: Typo? "what kind of type **were** dealing with"

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+
+It may be possible to address all those cases without changing each such section of the spec,
+but rather by adding general rules ("an extension on a class type is considered a class type" or some such).
+
+## Open issue: need to specify type erasure
+
+It should be done in a way that allows switching between the extension type and the underlying type without binary break.  
+It should allow for using type parameters as type parameters: `void M<T>(E<T> e)`.
```
AlekseyTs, line 203: Did you mean to say: "It should allow for using type parameters as type **arguments**" ?

proposals/extensions.md
```diff
@@ -116,6 +109,251 @@ public implicit extension E for S
...
+## Open issue: need to specify type erasure
+
+It should be done in a way that allows switching between the extension type and the underlying type without binary break.  
+It should allow for using type parameters as type parameters: `void M<T>(E<T> e)`.  
+It should allow for using extension types as type arguments without violating type parameter constraints: `C<E>` with `class C<T> where T : struct, I { }`.  
+
+The current proposal is to use an attribute with a string representing the type with extensions un-erased.
+For example: `void M(E)` would be emitted as `void M([Attribute("E")] UnderlyingType)`.
+
+The serialization format would be the same as the one used for `typeof` in attributes, but with support for type parameters (using `T!` and `T!!` syntax).
```
AlekseyTs, line 209: I thought the proposal was to follow the way specified in ECMA-335 (i.e. to refer by ordinal, rather than by name):
>II.9 Generics
>Within a generic type definition, its generic parameters are referred to by their index. Generic
parameter zero is referred to as !0, generic parameter one as !1, and so on. Similarly, within the body
of a generic method definition, its generic parameters are referred to by their index; generic parameter
zero is referred to as !!0, generic parameter one as !!1, and so on.

proposals/extensions.md
```diff
@@ -473,45 +669,73 @@ We modify the [member access rules](https://github.com/dotnet/csharpstandard/blo
...
-2. indexer access scenarios where the set of *applicable* candidate indexers is empty.
-3. TODO(static) there may be more scenarios (operator resolution, delegate conversion, natural function types)
 
-That is covered below.
-
-TODO3(static) Is the "where `T` is not a type parameter" portion still relevant?
+Note: We allow static lookups on type parameters only for members that are static virtual members. 
+Since extensions members are not virtual, we don't allow static extension lookups on type parameters either.
+```
+void M<T>(T t) where T
```
AlekseyTs, line 695: The constraint is not finished. Did you mean "where T : U"?

proposals/extensions.md
```diff
@@ -715,6 +956,30 @@ The search proceeds as follows:
...
+#### This access
+
+We modify the [this access rules](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12813-this-access) as follows:
+
+A this_access has one of the following meanings:
+- [...]
+- ***When this is used in a primary_expression within an instance method or instance accessor of an extension with a class underlying type,
+  it is classified as a value. The type of the value is the instance type of the extension within which the usage occurs, 
+  and the value is a reference to the object for which the method or accessor was invoked.**
+- ***When this is used in a primary_expression within an instance method or instance accessor of an extension with a struct underlying type,
```
AlekseyTs, line 968: Need to specify the meaning for a scenario when the underlying type is a type parameter not known to be a reference type or a value type.

proposals/extensions.md
```diff
@@ -715,6 +956,30 @@ The search proceeds as follows:
 - If no extension indexer is found to be suitable for the element access
   in any enclosing scope, a compile-time error occurs.
 
+#### This access
+
+We modify the [this access rules](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12813-this-access) as follows:
+
+A this_access has one of the following meanings:
+- [...]
+- ***When this is used in a primary_expression within an instance method or instance accessor of an extension with a class underlying type,
```
AlekseyTs, line 965: > with a class underlying type

A class underlying type, or an underlying type known to be a reference type? For example, an interface type is not a class and is not a struct, and is not a type parameter.

AlekseyTs:
Done with review pass (commit 10)

proposals/extensions.md
jcouv: See the example below (line 142). A member that cannot work in the end gets in the way. We've taken steps in overload resolution and more recently in natural type to reduce such behaviors (discard bad candidates earlier).
I'm not sure we should do something here, maybe the current behavior is desirable fallout of current rules...

jcouv: I'm going to remove this open issue. I don't have a good proposal for how to make this scenario work and I'm no longer convinced it should.

Here's an updated summary of the question for the record:

## Open issue: removing zero-arity-matches-any from var scenario

By the current rules, a generic method which cannot possibly work in a function type scenario can hide
a member that would work:

```csharp
var x = C.Member; // error: member lookup finds C.Member (method group) and lacks type arguments to apply to that match

class C
{
    public static void Member<T>() { }
}

implicit extension E for C
{
    public static int Member = 42;
}
```

Is this a scenario we think should work?

For context, from the [member access rules](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#12871-general),
we perform a member lookup which includes generic methods even when no type arguments are present:
> [...] if a member lookup of I in E with K type parameters produces a match, then E.I is evaluated and classified as follows
> Note: When the result of such a member lookup is a method group and K is zero,
> the method group can contain methods having type parameters.
> This allows such methods to be considered for type argument inferencing. end note

See details from the [member lookup rules](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1251-general):

> When K is zero, methods having type parameters are not removed, since the type inference process might be able to infer the type arguments.

See details on [natural type of method groups](https://github.com/dotnet/csharplang/blob/main/proposals/method-group-natural-type-improvements.md)

proposals/extensions.md
jcouv: Yes. I meant to have this as an open issue but forgot. Added.

jcouv:
@AlekseyTs Please take another look. Thanks

AlekseyTs (approved):
LGTM (commit 11)

PR #8217, opened 2024-06-14
Add notes for extensions WG 2024-06-14

jcouv:

### DIFF
--- a/meetings/working-groups/roles/extension-wg-2024-06-14.md
+++ b/meetings/working-groups/roles/extension-wg-2024-06-14.md
@@ -1,0 +1,50 @@
+# Extensions WG 2024-06-14
+
+Agenda
+- type erasure
+
+TL;DR: we'll pursue the attribute with typeof-like string.
+
+## Type erasure
+
+We did a quick recap of various options we'd considered so far, and explored using an attribute
+with a string serialization to represent the erased types.
+
+1. modopts
+2. generic attribute
+3. attribute+typeof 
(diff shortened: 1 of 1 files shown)


meetings/working-groups/roles/extension-wg-2024-06-14.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. modopts
+2. generic attribute
+3. attribute+typeof 
+4. parallel method
+5. typeref/blobs
+
+### Attribute with typeof-like string
+
+The main problem is how to represent type parameters, since they are currently disallowed in typeof inside attributes.
```
jjonescz, line 21: Should we allow type parameters in typeof inside attributes and use that instead of inventing another way?

jcouv: Thinking about your question more, I think the implication is if we could extend `typeof` first, then the attribute constructor should take a `Type` instead of a `string`, if the runtime and other tools could recognize it. I'll capture an open question.

hez2010: BTW the prototype of using type parameters in generic attributes is here: https://github.com/dotnet/runtime/pull/103213
It's not a non-trivial work for runtime as the coreclr already supports looking up type parameters for an attribute, it's just that the reflection library didn't pass the generic context to the runtime while looking up attributes before.

PR #8220, opened 2024-06-17
Remove warning for `yield return` in `lock`

jjonescz:
See https://github.com/dotnet/roslyn/issues/72443#issuecomment-2147915772.

### DIFF
--- a/proposals/ref-unsafe-in-iterators-async.md
+++ b/proposals/ref-unsafe-in-iterators-async.md
@@ -287,10 +287,10 @@ class C
   }
   ```
 
-- `yield return` inside `lock` could be an error (like `await` inside `lock` is) or a warning-wave warning but that would be a breaking change: https://github.com/dotnet/roslyn/issues/72443.
+- `yield return` inside `lock` could be an error (like `await` inside `lock` is) or a warning-wave warning,
+  but that would be a breaking change: https://github.com/dotnet/roslyn/issues/72443.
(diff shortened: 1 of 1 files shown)


AlekseyTs (approved):

proposals/ref-unsafe-in-iterators-async.md
```diff
@@ -308,6 +287,10 @@ class C
   }
   ```
 
+- There is a related issue to make `yield return` inside a `lock` a warning-wave warning or an error: https://github.com/dotnet/roslyn/issues/72443.
+  Note that [the new `Lock`-object-based `lock`][lock-object] reports compile-time errors for `yield return`s in its body,
+  because such `lock` statement is equivalent to a `using` on a `ref struct` which disallows `yield return`s in its body.
+
 - `yield` inside `lock` could be an error (like `await` inside `lock` is) but that would be a breaking change.
```
cston, line 294: Are the two bullets referring to the same alternative? If so, consider combining. Perhaps:
```suggestion
- `yield return` inside `lock` could be an error (like `await` inside `lock` is) or a warning-wave warning but that would be a breaking change: https://github.com/dotnet/roslyn/issues/72443.
  Note that [the new `Lock`-object-based `lock`][lock-object] reports compile-time errors for `yield return`s in its body,
  because such `lock` statement is equivalent to a `using` on a `ref struct` which disallows `yield return`s in its body.

```

cston (approved):

PR #8231, opened 2024-06-20
Open questions for field-backed properties

cston:


KennethHoff:
Aren't all auto properties "field backed"? That feels like a bad name for what is essentially a subset.

I prefer semi-auto while realizing that there's a bias there; it's always been referred to as semi-auto, so I'm naturally more inclined to call it that.

proposals/semi-auto-properties.md
```diff
@@ -156,6 +156,64 @@ public class Point
...
+```csharp
+class MyClass
+{
+    private const int field = -1;
+
+    public object Property { get; } = field; // bind to const (ok) or backing field (error)?
+}
+```
+
+In the example above, binding to the backing field should result in an error: "initializer cannot reference non-static field".
```
jaredpar, line 179: Is there an example where field binding to the backing property would have desirable behavior?

cston: I haven't found an example where binding to the backing field could be used in the property initializer.

jaredpar: If so then why do we recommend making it a keyword? Seems like a breaking change that isn't balanced with a positive scenario.

RikkiGibson: Also, if we did make `field` a keyword in property initializer, I would expect to actually be able to read the `field` in the initializer if the property is static. I feel like it gets wonky.

HaloFour: Got a weirder one, which isn't on the list (and I don't know that I've seen it mentioned elsewhere).  What about properties with expression body definitions?

```cs
public int X => Increment(ref field);

private int Increment(ref int field) => field++;
```

RikkiGibson: Yeah those are property accessors and we expect field to be a keyword in them.

cston: It seems like the backing field of a `static` property might be usable in the property initializer.
```csharp
class MyClass
{
    static object Property { get; } = Initialize(ref field);
    
    static object Initialize(ref object obj)
    {
        return obj = new Random().Next();
    }
}
```

RikkiGibson: Yeah that's wonky and not useful IMO (you're already assigning the field, why do you need to be able to otherwise use it?). So put me in for wanting 'field' to just not be a keyword in the initializer, even though it might feel a little confusing when an accessor is using 'field' keyword in the same property.

hez2010:
> Aren't all auto properties "field backed"?

You can definitely create a property which is not backed by a field:

```cs
public int CurrentNumber => Random.Shared.Next();
```

KennethHoff:
@hez2010 that's not an auto property though, is it? I might be wrong, but I thought an auto property was properties where you don't reference any storage area (basically just `{ get; set/init; }`)

cston:
@dotnet-policy-service rerun

proposals/semi-auto-properties.md
jnm2: > Got a weirder one, which isn't on the list (and I don't know that I've seen it mentioned elsewhere). What about properties with expression body definitions?

@HaloFour This was part of the intended design from the beginning, supporting lazy initialization such as that example and `=> field ??=`. https://github.com/dotnet/csharplang/pull/8069 fleshes out the proposal with examples like these.

PR #8241, opened 2024-06-25
Update first-class Span speclet with LDM decisions

jjonescz:


proposals/first-class-span-types.md
```diff
@@ -61,11 +64,21 @@ The explicit span conversions are the following:
...
 User-defined conversions are not considered when converting between
 - any single-dimensional `array_type` and `System.Span<T>`/`System.ReadOnlySpan<T>`,
 - any combination of `System.Span<T>`/`System.ReadOnlySpan<T>`,
 - `string` and `System.ReadOnlySpan<char>`.
 
+The implicit span conversions are exempted from the rule
+that it is not possible to define a user-defined operator between types for which a non-user-defined conversion exists
+([§10.5.2 Permitted user-defined conversions][permitted-udcs]).
+This is needed so BCL can keep defining the existing Span conversion operators even when they switch to C# 13
+(to avoid binary breaking changes and also because we use these operators in codegen of the new standard span conversion).
```
cston, line 78: Perhaps:
```suggestion
(to avoid binary breaking changes and also because these operators are used in codegen of the new standard span conversion).
```

proposals/first-class-span-types.md
```diff
@@ -490,6 +476,14 @@ We see a couple of ways of solving this concern:
    On the other hand, that could affect partners abilities to take them up.
 2. We couple type conversions to a Compilation or look at other ways of providing the well-known type to it. This will take a bit of investigation.
 
+#### Answer
```
cston, line 479: Do we need to keep the open question (and answer) or is the update to the spec from that decision (copied below) sufficient?

> Any Span/ReadOnlySpan types are considered applicable for the conversion if they match by their fully-qualified name.

jjonescz: I wasn't sure how this is usually done, but your suggestion makes sense to me, thanks.

cston (approved):

proposals/first-class-span-types.md
```diff
@@ -50,6 +50,9 @@ An implicit span conversion permits `array_types`, `System.Span<T>`, `System.Rea
 
 ------
 
+Any Span/ReadOnlySpan types are considered applicable for the conversion if they match by their fully-qualified name.
+If they do not match the types that are selected by the compiler's well-known type resolution logic, a compile-time error is reported.
```
333fred, line 54: This is not what we decided? We won't check that, and we can't really specify it either, since "the compiler's well-known type resolution logic" isn't a thing in the spec.

jjonescz: Ah, okay, I thought that was part of the option we discussed, but I might have misunderstood.

jjonescz: Then the question is how will we emit - do we need to find the operators on the Span types manually? It would be much easier to just use the well-known member machinery; and hence report an error if the Span type doesn't match the well-known type. Maybe that doesn't have to be specified, it can be an implementation detail like well-known types are.

333fred (approved):

PR #8242, opened 2024-06-25
Provide details about representation of instance members in metadata in extensions.md

AlekseyTs:


AlekseyTs:
@jcouv Please review

SunnieShine:
Hello. Question here.

Due to such adjustment here, can we allow `ref struct`s to be used also? After all declarations will be changed into `static` declarations, and there's no backing field inside the extension.

Thanks!

AlekseyTs:
> Due to such adjustment here, can we allow `ref struct`s to be used also? After all declarations will be changed into `static` declarations, and there's no backing field inside the extension.
> 

At the moment there are no plans to change that.

proposals/extensions.md
```diff
@@ -1158,6 +1158,205 @@ It allows roundtripping of extension symbols through metadata (full and referenc
...
+	}
+```
+
+Note, the extra parameter for event accessors are not CLS compliant, therefore, tools and other compilers
+likely won't be able to consume them as regular static events. For example, VB compiler is not able to and
+the event accessors cannot be used directly in VB as well.
+
+Since at the moment we are not supporting overriding or interface implementation by extension types, presence of the
+`modopt(System.Runtime.CompilerServices.ExtensionAttribute)` is sufficient to avoid a signature conflict with a user
+defined static member because, because given the restriction, there is no way for any `modopt` to get its way into
```
jjonescz, line 1272:
```suggestion
defined static member because, given the restriction, there is no way for any `modopt` to get its way into
```

jjonescz (approved):

proposals/extensions.md
```diff
@@ -1158,6 +1158,205 @@ It allows roundtripping of extension symbols through metadata (full and referenc
 For example: `implicit extension R for UnderlyingType` yields
 `public static void <ImplicitExtension>$(UnderlyingType)`.  
 
+Instance method/property/indexer declarations in source are represented as static declarations in metadata:
```
jcouv, line 1161: Let's add open issues at the top for the follow-up questions (how to expose the new APIs to other compilers, capturing receiver in type parameter scenario, ...).

jcouv (approved):
LGTM Thanks (iteration 3)

PR #8253, opened 2024-06-28
Object initializer updates

jaredpar:
This updates the ref safety rules to better explain how object initializers work.

### DIFF
--- a/proposals/csharp-11.0/low-level-struct-improvements.md
+++ b/proposals/csharp-11.0/low-level-struct-improvements.md
@@ -1,5 +1,4 @@
-Low Level Struct Improvements
-====
+# Low Level Struct Improvements
 
 [!INCLUDE[Specletdisclaimer](../speclet-disclaimer.md)]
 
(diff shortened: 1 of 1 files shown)


proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -411,6 +410,30 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
...
+var x = new S(ref heapSpan)
+{
+    Field = stackSpan;
+}
+
+// Can be modeled as 
+var x = new S(ref heapSpan, stackSpan)
+```
+
+This modeling is important because it demonstrates that our [MAMM][rules-method-arguments-must-match] need to account specially for member initializers. Consider that this particular case needs to be illegal as it allows for a value with a narrower *safe-context* to be assigned to a higher one.
```
cston, line 435: Should the link be `(#rules-method-arguments-must-match)` instead?

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -411,6 +410,30 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
 }
 ```
 
+#### Rules for object initializers
+
+The *safe-context* of an object initializer expression is narrowest of:
+
+1. The *safe-context* of the constructor call.
+2. The *safe-context* arguments to member initializer indexers that can escape to the receiver (can escape to *caller-context*)
```
cston, line 418: Should this be *ref-safe-context* of `ref` arguments to indexers (specifically, an `in` argument)?

jaredpar: This is a case of practically speaking no. The `in` part of an indexer can't escape into `this` because it's return only. Hmm ... guess you could if it was `[UnscopedRef]`. Not sure if we have good testing for that. The 99% case is the value is what matters. 

Spec wise can say both thogugh.

cston (approved):

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -411,6 +410,30 @@ Span<int> ComplexScopedRefExample(scoped ref Span<int> span)
...
+```c#
+Span<int> heapSpan = default;
+Span<int> stackSpan = stackalloc int[42];
+var x = new S(ref heapSpan)
+{
+    Field = stackSpan;
+}
+
+// Can be modeled as 
+var x = new S(ref heapSpan, stackSpan)
```
jjonescz, line 432: Cannot it be modeled simply as this?

```cs
var x = new S(ref heapSpan);
x.Field = stackSpan;
```

jaredpar: Doing it that way would mean that the *safe-context* of `x` is *calling method* because only `heapSpan` is used in the variable declaration. The motivation for putting all of the other values in the constructor is to solve issues like this.

PR #8257, opened 2024-06-30
Start proposal for more constant types

BillWagner:
First draft at a proposal to enable more types to be declared as `const` types.

### DIFF
--- a/proposals/more-const-types.md
+++ b/proposals/more-const-types.md
@@ -1,0 +1,79 @@
+# Unmanaged constant types
+
+* [x] Proposed
+* [ ] Prototype: [Complete](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
+* [ ] Implementation: [In Progress](https://github.com/dotnet/roslyn/BRANCH_NAME)
+* [ ] Specification: [Not Started](pr/1)
+
+## Summary
+[summary]: #summary
+
+Expand the allowable types for compile-time constants (`const` declarations) to any value type that is fully blittable.
+
+The current specification [§12.23](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1223-constant-expressions) limits constant expressions to an enumerated set of types, or default reference expressions (`default`, or `null`).
+
+This proposal expands the types allowed. The allowed types should include any unmanaged type ([§8.8](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/types.md#88-unmanaged-types))
+
+It could be expanded to include any `readonly struct` type.
+
+We should consider if a `ref struct` type, such as `ReadOnlySpan<T>` could be a compile-time constant.
+
+## Motivation
+[motivation]: #motivation
+
+Only constant expressions can be used in pattern matching expressions. This expands the types that can be used in patterns. For example, it's common to use GUIDs as identifiers. This feature enables matching on GUID values. Currently, the only way to do that is to convert a GUID to a ReadOnlySpan<byte> and use a list pattern on the list of bytes.
+
+Similarly, any other type that holds multiple values, like record struct types, could be used to match patterns.  For example, a `Point` type could be matched agains a `const` for the origin.
+
+Other uses for constant expressions include:
+
+- default argument values.
+- Attribute parameters
+
+## Detailed design
+[design]: #detailed-design
+
+This is the bulk of the proposal. Explain the design in enough detail for somebody familiar with the language to understand, and for somebody familiar with the compiler to implement,  and include examples of how the feature is used. This section can start out light before the prototyping phase but should get into specifics and corner-cases as the feature is iteratively designed and implemented.
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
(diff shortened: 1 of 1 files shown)


HaloFour:
> Attribute parameters

This proposal includes changing the runtime to understand and support for more types in attribute parameters?  Would that also include the pseudoconstant `decimal`?

KathleenDollard:
I love it!

proposals/more-const-types.md
```diff
@@ -0,0 +1,79 @@
...
+
+* [x] Proposed
+* [ ] Prototype: [Complete](https://github.com/PROTOTYPE_OWNER/roslyn/BRANCH_NAME)
+* [ ] Implementation: [In Progress](https://github.com/dotnet/roslyn/BRANCH_NAME)
+* [ ] Specification: [Not Started](pr/1)
+
+## Summary
+[summary]: #summary
+
+Expand the allowable types for compile-time constants (`const` declarations) to any value type that is fully blittable.
```
tannergooding, line 11: C# notably doesn't have a concept of `blittable`, only `unmanaged`. `blittable` has a specific meaning, as it pertains to `interop`, and types like `bool` aren't considered `blittable` unless the user opts-in via `DisableRuntimeMarshalling`

proposals/more-const-types.md
```diff
@@ -0,0 +1,79 @@
+# Unmanaged constant types
```
tannergooding, line 1: I have a similar proposal, from 2017, that goes over a different approach to supporting this functionality here: https://github.com/dotnet/csharplang/discussions/688

proposals/more-const-types.md
```diff
@@ -0,0 +1,78 @@
...
+    // V2:
+    
+    public struct Point
+    {
+        public int X;
+        public int Y;
+        public object SomeReference;
+    }
+    ```
+
```
mattwar, line 75: How do we reconcile this with the runtime concept of field constant?  

Currently C# constants match the runtime's version of field constants.  For example, a field can be declared that is a constant which is different than a static field with a value. The constant value is encoded in the metadata which the compiler uses to emit as load instructions at the use site instead of referencing the field at runtime, because no field exists for it at runtime.  The kinds of values that can be field constants is constrained to those that can be encoded in metadata. I'm not sure what those are, but it probably does not include arbitrary types even if they are blittable.

Of course, C# could deviate from the runtime and allow other types to be constant but would have to change to use static fields for these instead of field constants.

333fred: C# already diverges from the runtime's concept of constants. Consider `decimal`:

```cs
const decimal d = 1.0M;
```

Results in:

```
    // Fields
    .field public static initonly valuetype [System.Runtime]System.Decimal d
    .custom instance void [System.Runtime]System.Runtime.CompilerServices.DecimalConstantAttribute::.ctor(uint8, uint8, uint32, uint32, uint32) = (
        01 00 01 00 00 00 00 00 00 00 00 00 0a 00 00 00
        00 00
    )
```

proposals/more-const-types.md
```diff
@@ -0,0 +1,78 @@
...
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still undecided?
+
+- **What types should be allowed?**:  Preventing mutation for types whose members are all unmanaged types (or `struct` types whose members are all unmanaged types) is a manageable problem: a `const` instance can't be accessed as a writable `ref`. However, if any member is a reference type, that problem is harder to solve.
+- **Are there implications for the runtime?** Depending on how constants are stored in memory, is there an impact on the runtime?
+- **What are the implications of allowing constant ref structs? Are they even useful?**: Are the rules surrounding ref safety consistent with where constants are used?
+- **How will down-level scenarios work?** Can these constants be accessed by code using previous compilers? What about using Reflection to access fields?
```
jkotas, line 56: Yes, we should make sure that reflection handles these new constant types. Our users expect reflection to return C# constants as properly typed values (https://github.com/dotnet/runtime/issues/104270#issuecomment-2207589673 has some discussion about issues we have in this area currently.)

proposals/more-const-types.md
```diff
@@ -0,0 +1,78 @@
...
+This proposal expands the types allowed. The allowed types should include any unmanaged type ([§8.8](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/types.md#88-unmanaged-types))
+
+It could be expanded to include any `readonly struct` type.
+
+We should consider if a `ref struct` type, such as `ReadOnlySpan<T>` could be a compile-time constant.
+
+## Motivation
+[motivation]: #motivation
+
+Only constant expressions can be used in pattern matching expressions. This expands the types that can be used in patterns. For example, it's common to use GUIDs as identifiers. This feature enables matching on GUID values. Currently, the only way to do that is to convert a GUID to a ReadOnlySpan<byte> and use a list pattern on the list of bytes.
```
jaredpar, line 24: Expanding the notion of `const` in the language is a decent sized feature as it ends up touching on a number of areas. If the main motivation is to expand the set of types usable in patterns should we just approach that more directly?

PR #8264, opened 2024-07-03
Rename semi-auto-properties.md to reflect new feature name

jnm2:
This PR consists of a file rename `proposals/semi-auto-properties.md` to `proposals/field-keyword.md` to align with the current name of the feature as indicated by LDM.

It also prevents two links from breaking inside this repo. I updated the links to point to the version of the spec that was being referenced at the time.


jnm2:
Wait, a meeting notes link is still referencing main, I'll fix

CyrusNajmabadi (approved):

jcouv:
Could you also do the rename in https://github.com/dotnet/roslyn/blob/main/docs/Language%20Feature%20Status.md and fix links from test plan?

PR #8283, opened 2024-07-15
`field` as keyword in primary expression only

cston:


proposals/field-keyword.md
```diff
@@ -481,24 +500,7 @@ class MyClass
 
 In the example above, binding to the backing field should result in an error: "initializer cannot reference non-static field".
 
-### `field` and `value` in event accessor
```
333fred, line 484: I would generally prefer if we kept old questions and simply marked them answered, so that when you go from the notes to the current version of the speclet, you can see the question those notes talked about.

333fred (approved):

PR #8287, opened 2024-07-16
Update first-class Span open question about extension receiver break

jjonescz:
Based on LDM decision and discussion.

### DIFF
--- a/proposals/first-class-span-types.md
+++ b/proposals/first-class-span-types.md
@@ -104,7 +104,11 @@ static class E
 }
 ```
 
-There's [an open question](#delegate-extension-receiver-break) whether this break should be avoided or not.
+As possible future work, we could consider removing this condition that span conversion is not considered for extension receiver in method group conversions
+and instead implement changes so the scenario like the one above would end up successfully calling the `Span` overload instead:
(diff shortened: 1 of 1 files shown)


proposals/first-class-span-types.md
```diff
@@ -445,25 +449,6 @@ However, that would mean users could get different behavior after updating the t
 On the other hand, the new Span APIs would still be ambiguous from VB unless we implement the betterness rule there as well.
 So it might be easier if API authors solve this themselves via the `OverloadResolutionPriorityAttribute`.
 
-### Delegate extension receiver break
```
333fred, line 448: Rather than removing this section, please mark it answered and put the answer as a subheading below, like the other answered questions above.

jjonescz: Previously it was suggested to remove the resolved question if its answer is somewhere else in the speclet: https://github.com/dotnet/csharplang/pull/8241#discussion_r1653603180. I find that reasonable

333fred: Let's please keep the question, as that will make it make more sense when people go from the notes to the current version of the speclet.

333fred (approved):

cston (approved):

PR #8295, opened 2024-07-16
Fix lists in an LDM note

xparadoxical:
The note used `a.` `b.`, which aren't recognized, and used `-` incorrectly (with >= 4 space indent), so the lists didn't display correctly


333fred (approved):

333fred:
Thanks @xparadoxical!

PR #8296, opened 2024-07-17
Add open questions for LDM

333fred:


proposals/overload-resolution-priority.md
```diff
@@ -258,6 +258,32 @@ Our options are:
...
+* Destructors - again, currently blocked.
+* Static constructors - again, currently blocked.
+* Local functions - These are not currently blocked, because they _do_ undergo overload resolution, you just can't overload them. This is simlar to how we don't
+  error when the attribute is applied to a member of a type that is not overloaded. Should this behavior be confirmed?
+
+### Langversion Behavior
+
+The implementation currently only issues langversion errors when `OverloadResolutionPriorityAttribute` is applied, _not_ when it actually influences anything. This
+decision was made because the there are APIs that the BCL will add (both now and over time) that will start using this attribute; if the user manually sets their
+language version back to C# 12 or prior, they may see these members and, depending our langversion behavior, either:
```
jjonescz, line 277: I assume the attribute won't be applicable in VB, but will be consumable by VB (e.g., via BCL). That would be similar to the proposed C# <12 behavior and hence would be consistent. Perhaps it's worth mentioning.

proposals/overload-resolution-priority.md
```diff
@@ -258,6 +258,32 @@ Our options are:
...
+
+* If we ignore the attribute in C# <13, run into an ambiguity error because the api is truly ambiguous without the attribute, or;
+* If we error when the attribute affected the outcome, run into an error that the api is unconsumable. This will be especially bad because `Debug.Assert(bool)`
+  is being de-prioritized in .NET 9, or;
+* If we silently change resolution, encounter potentially different behavior between machines if one machine understands the attribute and another doesn't.
```
jjonescz, line 282: I guess we should have a disclaimer in the attribute docs that this can happen and that it should be used only on overloads which have equivalent behavior.

333fred:
@jcouv @cston for review

proposals/overload-resolution-priority.md
```diff
@@ -258,6 +267,40 @@ Our options are:
...
+  error when the attribute is applied to a member of a type that is not overloaded. Should this behavior be confirmed?
+
+#### Answer
+
+All of the locations listed above are blocked.
+
+### Langversion Behavior (Answered)
+
+The implementation currently only issues langversion errors when `OverloadResolutionPriorityAttribute` is applied, _not_ when it actually influences anything. This
+decision was made because the there are APIs that the BCL will add (both now and over time) that will start using this attribute; if the user manually sets their
```
cston, line 289: Typo: "because the there"

proposals/overload-resolution-priority.md
```diff
@@ -258,6 +267,40 @@ Our options are:
...
+language version back to C# 12 or prior, they may see these members and, depending our langversion behavior, either:
+
+* If we ignore the attribute in C# <13, run into an ambiguity error because the api is truly ambiguous without the attribute, or;
+* If we error when the attribute affected the outcome, run into an error that the api is unconsumable. This will be especially bad because `Debug.Assert(bool)`
```
cston, line 293: Consider using "API" (uppercase) to match other uses in the document.

proposals/overload-resolution-priority.md
```diff
@@ -258,6 +267,40 @@ Our options are:
...
+  is being de-prioritized in .NET 9, or;
+* If we silently change resolution, encounter potentially different behavior between machines if one machine understands the attribute and another doesn't.
```
cston, line 295: > If we silently change resolution, encounter potentially different behavior between machines if one machine understands the attribute and another doesn't.

Perhaps "between versions of the compiler if one version understands ..."

333fred:
@jcouv @cston for review.

proposals/overload-resolution-priority.md
```diff
@@ -258,6 +267,40 @@ Our options are:
...
+
+All of the locations listed above are blocked.
+
+### Langversion Behavior (Answered)
+
+The implementation currently only issues langversion errors when `OverloadResolutionPriorityAttribute` is applied, _not_ when it actually influences anything. This
+decision was made because there are APIs that the BCL will add (both now and over time) that will start using this attribute; if the user manually sets their
+language version back to C# 12 or prior, they may see these members and, depending our langversion behavior, either:
+
+* If we ignore the attribute in C# <13, run into an ambiguity error because the api is truly ambiguous without the attribute, or;
```
cston, line 292: >api [](http://example.com/codeflow?start=80&length=3)

"API" (upper-case)?

cston (approved):

proposals/overload-resolution-priority.md
333fred:
```suggestion
* If we ignore the attribute in C# <13, run into an ambiguity error because the API is truly ambiguous without the attribute, or;
```

PR #8311, opened 2024-07-24
Fix notes

jnm2:
@333fred

### DIFF
--- a/meetings/2024/LDM-2024-07-22.md
+++ b/meetings/2024/LDM-2024-07-22.md
@@ -21,7 +21,7 @@ We started today by looking at the latest proposed changes for extensions. There
 * How will the format work for local functions?
 * We don't think that function pointers are necessary to support right now; extensions already have limited interactions with pointer types in general. If we can come up with specific scenarios, we could look
   at them at that point.
-* We talked briefly about simply upgrading the `Type` format to allow encoding, however that runs into the issue that the extension types would not satisfy type constraints. Any solution that involves concrete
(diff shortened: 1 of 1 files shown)


meetings/2024/LDM-2024-07-17.md
```diff
@@ -45,7 +45,7 @@ can break collection expressions because we don't have a betterness tiebreaker a
 a different preferred span type than the rest of the language.
 
 Discussion in LDM showed a general reluctance to go with the narrow proposal, and we much preferred the version that applied better conversion from element recursively to solve this problem. That version
-gets us closer to unification with `params` collections and is more complete. While we do think this will be a problem in .NET 9, we're hopeful that we can use `OverloadResolutionPriorityAttribute` to solve
+gets us closer to unification with `params` collections and is more complete. While we do think this more complete version will ship later than .NET 9, we're hopeful that we can use `OverloadResolutionPriorityAttribute` to solve
```
333fred, line 48: The original wording is intentional here. `this` refers to overloading problem, not the solution.

jnm2: Okay, thanks. I'm still struggling to follow what it's saying. Why do we think that overloading will be a problem in .NET 9?

333fred: That's the motivating scenario we're as to why we've talked about this twice in as many weeks. Concat is a problem in NET 9.

jnm2: Sorry, I did get that, I just don't see it stated anywhere that better conversion from element is assumed not to be available in .NET 9, which would explain why Concat would then still be a problem in .NET 9. I can leave it though.

333fred (approved):

PR #8321, opened 2024-07-29
Update first-class Span speclet to consider only structs

jjonescz:


proposals/first-class-span-types.md
```diff
@@ -50,7 +50,7 @@ An implicit span conversion permits `array_types`, `System.Span<T>`, `System.Rea
 
 ------
 
-Any Span/ReadOnlySpan types are considered applicable for the conversion if they match by their fully-qualified name.
+Any Span/ReadOnlySpan types are considered applicable for the conversion if they are `struct`s and they match by their fully-qualified name.
```
cston, line 53: Should we require `ref struct` rather than `struct`?

jjonescz: I don't think the compiler needs it but we could.

333fred: I think we should. We care very specifically about the semantics of the type here, and the `ref`-ness is part of that.

333fred (approved):

PR #8338, opened 2024-08-06
Collection expressions: better conversion from element

cston:
Update [*better conversion from expression*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11644-better-conversion-from-expression) to prefer the target with the better conversion from *each element*.

See [LDM-2024-07-24](https://github.com/dotnet/csharplang/blob/main/meetings/2024/LDM-2024-07-24.md#better-conversion-from-collection-expression-with-readonlyspant-overloads)

### DIFF
--- a/proposals/collection-expressions-updates.md
+++ b/proposals/collection-expressions-updates.md
@@ -1,0 +1,36 @@
+# Collection expressions (updates)
+
+## Summary
+[summary]: #summary
+
+Updates to [*collection expressions*](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md) for C#13.
+
+## Overload resolution
+[overload-resolution]: #overload-resolution
+
+[*Better conversion from expression*](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11644-better-conversion-from-expression) is updated to prefer the target with the better conversion from **each element**.
+
+> Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a *better conversion* than `C₂` if one of the following holds:
+>
+> * `E` is a *collection expression*, and one of the following holds:
+>   * **`T₁` and `T₂` have the same *collection type definition*, and `T₁` has *element type* `S₁`, and `T₂` has *element type* `S₂`, and both of the following hold:**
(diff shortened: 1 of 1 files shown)


proposals/collection-expressions-updates.md
```diff
@@ -0,0 +1,36 @@
...
+>   * **`T₁` and `T₂` have the same *collection type definition*, and `T₁` has *element type* `S₁`, and `T₂` has *element type* `S₂`, and both of the following hold:**
+>     * **For each element `Eᵢ` in `E`:**
+>       * **If `Eᵢ` is an expression element, the conversion from `Eᵢ` to `S₂` is not better than the conversion from `Eᵢ` to `S₁`**
+>       * **If `Eᵢ` is a spread element with iteration type `Sᵢ`, the conversion from `Sᵢ` to `S₂` is not better than the conversion from `Sᵢ` to `S₁`**
+>     * **For at least one element `Eᵢ` in `E`:**
+>       * **If `Eᵢ` is an expression element, the conversion from `Eᵢ` to `S₁` is better than the conversion from `Eᵢ` to `S₂`**
+>       * **If `Eᵢ` is a spread element with iteration type `Sᵢ`, the conversion from `Sᵢ` to `S₁` is better than the conversion from `Sᵢ` to `S₂`**
+>
+>   * `T₁` is `System.ReadOnlySpan<E₁>`, and `T₂` is `System.Span<E₂>`, and an ~~implicit~~ **identity** conversion exists from `E₁` to `E₂`
+>   * `T₁` is `System.ReadOnlySpan<E₁>` or `System.Span<E₁>`, and `T₂` is an *array_or_array_interface* with *element type* `E₂`, and an ~~implicit~~ **identity** conversion exists from `E₁` to `E₂`
```
333fred, line 25: Mentioned this offline as well, but I'm not a fan of this change. Between `ReadOnlySpan<string>` and `Span<object>`, it feels like `ReadOnlySpan<string>` is clearly the better choice for a collection expression. Particularly since between `ReadOnlySpan<object>` and `ReadOnlySpan<string>`, the rule we're adding _would_ make that unambiguous.

Think the next step here is to collect various examples of what we want to enable, vs what harm we think will come from leaving this overly broad. For example, the `ReadOnlySpan<byte>` vs `Span<int>` question.

cston:
Replaced by https://github.com/dotnet/csharplang/blob/main/proposals/csharp-13.0/collection-expressions-better-conversion.md.

PR #8342, opened 2024-08-07
Polish and clarify prior field WG notes

jnm2:
@CyrusNajmabadi

### DIFF
--- a/meetings/working-groups/field-keyword/FK-2024-06-26.md
+++ b/meetings/working-groups/field-keyword/FK-2024-06-26.md
@@ -1,10 +1,10 @@
-# `field` keyword working group nodes for 06/26/2024
+# `field` keyword working group notes for June 26, 2024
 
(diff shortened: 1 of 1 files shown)


meetings/working-groups/field-keyword/FK-2024-06-26.md
```diff
@@ -235,7 +235,7 @@ string Prop { get => field ?? ""; } // or
 string Prop { get => field ??= ComputeInitialValue(); }
 ```
 
-With both properties, the initial analysis is done with the backing field being `string?`.  The analysis proves that the get-accessor always exist with the backing field in a non-null state.  As such, it's "safe" to have a nullable backing field.  And because it is safe, we don't need a constructor initializer.
+With both properties, the initial analysis is done with the backing field being `string?`.  The analysis proves that the get accessor never returns a maybe-null expression.  As such, it's "safe" to have a nullable backing field.  And because it is safe, we don't need a constructor initializer.
```
jnm2, line 238: In this one especially, it's clear that @CyrusNajmabadi meant to write what the PR is updating to. The first property doesn't exit with the backing field in a non-null state.

CyrusNajmabadi (approved):

cston (approved):

PR #8358, opened 2024-08-13
Update span breaking changes

jjonescz:


333fred (approved):

proposals/first-class-span-types.md
```diff
@@ -303,6 +303,24 @@ However, the upper-bound span inference would only apply if the source type were
...
 As any proposal that changes conversions of existing scenarios, this proposal does introduce some new breaking changes. Here's a few examples:
 
+#### Ambiguities
+
+The following examples previously failed type inference for the Span overload,
+but now type inference from array to Span succeeds, hence these are ambiguous.
+
+```cs
+var x = new long[] { 1 };
+Assert.Equal([2], x); // previously Assert.Equal<T>(T[], T[]), now ambiguous with Assert.Equal<T>(ReadOnlySpan<T>, Span<T>)
```
cston, line 313: Should this be `Assert.Equal<T>(ReadOnlySpan<T>, ReadOnlySpan<T>)`?

jjonescz: No, the error message I got in BCL really said `Assert.Equal<T>(ReadOnlySpan<T>, Span<T>)`. There are [all combinations](https://github.com/xunit/assert.xunit/blob/1b39b380a3d3ab50db1434ef33c216f1606391ae/SpanAsserts.cs#L143-L192) of Assert.Equal overloads (with ROS/Span). And I think for the collection expression, ROS is chosen probably due to some collection expression rules, and for array, Span is better because it is convertible to ROS?

cston (approved):

PR #8390, opened 2024-08-24
Incorporate the accepted field nullability proposal

jnm2:
Replaces the stale Nullability section by incorporating <https://github.com/dotnet/csharplang/pull/8346> which was accepted (https://github.com/dotnet/csharplang/blob/main/meetings/2024/LDM-2024-08-21.md#field-keyword-nullability).

### DIFF
--- a/meetings/2024/LDM-2024-08-21.md
+++ b/meetings/2024/LDM-2024-08-21.md
@@ -46,4 +46,4 @@ they might also still cause some confusion. We'll consider these details again i
 
 #### Conclusion
 
-General proposal is adopted. Specific behavior still need more review.
+General proposal is adopted. Specific behavior still needs more review.
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -20,6 +20,11 @@ In these cases by now you always have to create an instance field and write the 
 
 - **Full accessor**: This is an accessor that has a body. The implementation is not provided by the compiler, though the backing storage may still be (as in the example `set => field = value;`).
 
+- **Field-backed property**: This is either a property using the `field` keyword within an accessor body, or an auto property.
+
+- **Backing field**: This is the variable denoted by the `field` keyword in a property's accessors, which is also implicitly read or written in automatically implemented accessors (`get;`, `set;`, or `init;`).
+
+
 ## Detailed design
 
 For properties with an `init` accessor, everything that applies below to `set` would apply instead to the `init` accessor.
@@ -251,22 +256,9 @@ public class C
 
 The following nullability rules will apply not just to properties that use the `field` keyword, but also to existing auto properties.
 
-#### Terms
-
-A property is a *field-backed property* if it meets any of the following conditions:
-1. It contains any automatically implemented accessors (`get;`/`set;`/`init;`).
-2. It uses the `field` keyword within any accessor bodies.
-
-Some examples of *field-backed properties* include:
-```cs
-public string Prop1 { get; set; }
-public string Prop2 { get => field; set; }
-public string? Prop3 { get => field; }
-```
-
-The variable denoted by the `field` keyword in a property's accessors is the *backing field* of that property.
-
 #### Nullability of the *backing field*
+
+See [Glossary](#glossary) for definitions of new terms.
 
 The *backing field* has the same type as the property. However, its nullable annotation may differ from the property. To determine this nullable annotation, we introduce the concept of *null-resilience*. *Null-resilience* intuitively means that the property's `get` accessor preserves null-safety even when the field contains the `default` value for its type.
 
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -287,7 +287,7 @@ The nullability of the backing field is determined as follows:
 
 Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
 
-We update the following spec language from [An alternative approach (TODO rename that section?)](nullable-constructor-analysis.md#an-alternative-approach) to accomplish this:
+We update the following spec language from the previous [proposed approach](nullable-constructor-analysis.md#proposed-approach) to accomplish this:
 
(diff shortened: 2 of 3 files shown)


meetings/2024/LDM-2024-08-21.md
```diff
@@ -46,4 +46,4 @@ they might also still cause some confusion. We'll consider these details again i
 
 #### Conclusion
 
-General proposal is adopted. Specific behavior still need more review.
+General proposal is adopted. Specific behavior still needs more review.
```
jnm2, line 49: Fixed this text before copying it into the answered questions section.

proposals/field-keyword.md
```diff
@@ -230,74 +230,110 @@ public struct S
...
-```cs
-class C
-{
-    public C() { P = ""; }
+- Patterns that use the `field` keyword should feel as though they've always been part of the language. Avoid making the user jump through hoops to enable Nullable Reference Types in code that is perfectly idiomatic for the `field` keyword feature.
 
-    public string P { get => field; set => field = value; }
-}
-```
+One of the key scenarios is lazily initialized properties:
```
jnm2, line 241: I rewrote this goals/motivation section to include some of the rationale that came out of LDM, such as the thought experiment of designing the NRTs feature after the `field` keyword had already been in use.

Suggestions/nits welcome!

proposals/field-keyword.md
```diff
@@ -230,74 +230,110 @@ public struct S
...
-            field = value; // Stop following the parent's value
-    }
-}
+public string Prop1 { get; set; }
+public string Prop2 { get => field; set; }
+public string? Prop3 { get => field; }
 ```
 
-`var` was designed to declare nullability so that subsequent assignments to the variable could be nullable, due to established patterns in C#. It's expected that the same rationale would apply to property backing fields.
+The variable denoted by the `field` keyword in a property's accessors is the *backing field* of that property.
```
jnm2, line 267: Should I move this up into the Glossary section at the top of the file?

proposals/field-keyword.md
```diff
@@ -230,74 +230,110 @@ public struct S
...
+
+#### Nullability of the *backing field*
+
+The *backing field* has the same type as the property. However, its nullable annotation may differ from the property. To determine this nullable annotation, we introduce the concept of *null-resilience*. *Null-resilience* intuitively means that the property's `get` accessor preserves null-safety even when the field contains the `default` value for its type.
+
+A *field-backed property* is determined to be *null-resilient* or not by performing a special nullable analysis of its `get` accessor.
+- For the purposes of this analysis, `field` is temporarily assumed to have *annotated* nullability, e.g. `string?`. This causes `field` to have *maybe-null* or *maybe-default* initial state in the `get` accessor, depending on its type.
```
jnm2, line 274: @RikkiGibson For an `int` property, does "annotated nullability" mean `int?`- that is, to literally add a syntactic `?`- or is it defined to mean "maybe-default"?

I'm curious in case we would want to explicitly say anything about value types remaining unaffected by this. Similar question lower down; would we want an explicit rule that bails out of "The nullability of the backing field is determined as follows:" like we do for if `[field: MaybeNull]` is specified?

proposals/field-keyword.md
```diff
@@ -230,74 +230,110 @@ public struct S
...
+
+#### Constructor analysis
+
+Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
 
-To land in this sweet spot implicitly, without having to write an attribute each time, nullability analysis will combine an inherent nullability of the field with the behavior of `[field: NotNull]`. This allows maybe-null assignments without warning, which is desirable as shown above, while simultaneously allowing a scenario like `=> field.Trim();` without requiring an intervention to silence a warning that `field` could be null. Making sure `field` has been assigned is already covered by the warning that ensures non-nullable properties are assigned by the end of each constructor.
+We update the following spec language from [An alternative approach (TODO rename that section?)](nullable-constructor-analysis.md#an-alternative-approach) to accomplish this:
```
jnm2, line 290: @RikkiGibson The TODO is moving

proposals/field-keyword.md
RikkiGibson: Nullable annotation is not a thing for value types. In that case, a question mark is a `System.Nullable` (reified) type rather than merely a type annotation.

*maybe-default* is only really used for type parameters. In that case, the meanings of the flow states are:
- maybe-default: the value may be null, even if `T` is non-nullable reference type.
- maybe-null: the value may be null, only if `T` itself permits null (e.g. it is ok to return this value when the return type is `T`, but it is not ok to dereference it.)
- not-null: same as usual.

proposals/field-keyword.md
RikkiGibson: can I push to your branch to fix?

jnm2:
@RikkiGibson @333fred I did not move the [Alternatives](https://github.com/dotnet/csharplang/blob/5f86d734337b5113518df206bd64a1ad42a7371e/proposals/field-keyword-nullability.md#alternatives) section. How do we normally handle that?

proposals/field-keyword.md
jnm2: Works for me!

proposals/field-keyword.md
jnm2: Does that mean that when considering a value-typed property, it goes without saying that you stop after the first sentence and consider everything else irrelevant?

> The *backing field* has the same type as the property. However, its nullable annotation may differ from the property. To determine this nullable annotation, we introduce the concept of *null-resilience*.

proposals/field-keyword.md
RikkiGibson: Yeah feel free to do this and eliminate anything redundant.

proposals/field-keyword.md
RikkiGibson: Yes exactly.

proposals/field-keyword.md
jnm2: Done.

jnm2:
Done!

RikkiGibson (approved):

PR #8393, opened 2024-08-29
Update the params proposal in line with the new betterness rules

333fred:
The main change here is that we now require an identity conversion between element types to tiebreak among collection types.

### DIFF
--- a/proposals/params-collections.md
+++ b/proposals/params-collections.md
@@ -162,12 +162,12 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂,
 - If for at least one parameter `Mᵥ` uses the ***better parameter-passing choice*** ([§12.6.4.4](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#12644-better-parameter-passing-mode)) than the corresponding parameter in `Mₓ` and none of the parameters in `Mₓ` use the better parameter-passing choice than `Mᵥ`, `Mᵥ` is better than `Mₓ`.
 - **Otherwise, if both methods have params collections and are applicable only in their expanded forms then
    `Mᵢ` is better than `Mₑ` if the same set of arguments corresponds to the collection elements for both methods, and one of the following holds
-   (this corresponds to https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md#overload-resolution):**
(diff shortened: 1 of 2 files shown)


cston (approved):

proposals/params-collections.md
```diff
@@ -261,6 +261,48 @@ It doesn't feel reasonable to "compare" collections that are built from differen
...
+    
+    public static void M3(params ReadOnlySpan<int> a) {
+    }
+    
+    public static void M3(params ReadOnlySpan<int?> a) {
+    }
+}
+```
+
+Given that we prioritize element type over all else, this seems reasonable; there's nothing to tell the langauge whether the user would prefer `int?`
```
jnm2, line 303:
```suggestion
Given that we prioritize element type over all else, this seems reasonable; there's nothing to tell the language whether the user would prefer `int?`
```

jaredpar (approved):

cston:
> - `T₁` or `T₂` is not a *span type*, and `T₁` is implicitly convertible to `T₂`, and `T₂` is not implicitly convertible to `T₁`, or

Does this mean at least one of `T₁` and `T₂` is not a *span type*, or neither `T₁` nor `T₂` are *span types*? If the latter, consider using wording similar to the params-collections.md change:
> both `T₁` and `T₂` are not *span types*

---
Refers to: proposals/collection-expressions-better-conversion.md:38 in 6e17646. [](commit_id = 6e176467e36249a213c6eccf1c1f8eacfc47ef7f, deletion_comment = False)

cston:
> If there is an identity conversion from `E₁` to `E₂`, then the element conversions are as good as each other. Otherwise, the element conversions to `E₁` are better than the element conversions to `E₂` if:

> better than the element conversions

Is this a definition? (We use the same term below within the definition of ***better collection conversion from expression***.) If so, consider using ***bold*** here.

---
Refers to: proposals/collection-expressions-better-conversion.md:31 in 6e17646. [](commit_id = 6e176467e36249a213c6eccf1c1f8eacfc47ef7f, deletion_comment = False)

cston:
> - `E₁` does not have an identity conversion to `E₂`, and the element conversions to `E₁` are better than the element conversions to `E₂`, or

Is it possible for `T₁` to be better than `T₂` using the first rule, and `T₂` better than `T₁` using the second?


---
Refers to: proposals/collection-expressions-better-conversion.md:39 in 6e17646. [](commit_id = 6e176467e36249a213c6eccf1c1f8eacfc47ef7f, deletion_comment = False)

333fred:
> Is it possible for T₁ to be better than T₂ using the first rule, and T₂ better than T₁ using the second?

Yes, in theory. It would have to be a particularly convoluted case, but it's possible.

cston (approved):

PR #8398, opened 2024-09-02
Field keyword/updates

jnm2:
Fixes a code sample reported by a community member, removes an extra blank line from another code sample, and documents decisions from Monday's LDM.

### DIFF
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -305,12 +305,12 @@ class C
 
     public C()
     {
-        F = "a"; // ok
+        Prop = "a"; // ok
     }
 
     public static void Main()
     {
-        new C().F.ToString(); // NRE at runtime
+        new C().Prop.ToString(); // NRE at runtime
     }
 }
 ```
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -631,16 +631,6 @@ Should the proposed nullability of `field` be accepted? See the [Nullability](#n
 
 General proposal is adopted. Specific behavior still needs more review.
 
-## Open LDM questions
-
-### Feature name
-
-Some options for the name of the feature:
-1. semi-auto properties
-1. field access for auto properties [LDM-2023-07-17](https://github.com/dotnet/csharplang/blob/main/meetings/2023/LDM-2023-07-17.md#compiler-check-in)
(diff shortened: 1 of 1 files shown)


proposals/field-keyword.md
```diff
@@ -687,6 +681,10 @@ partial class C
 
 **Recommendation**: Permit an initializer on either part of a partial property when the implementation part uses `field`. Report an error if both parts have an initializer.
 
+#### Answer
+
+Allow `field` in implementing declaration and allow initializer in defining declaration (which compels implementation to be field-backed).
```
jnm2, line 686: @cston Should I move this back to the Open Questions section? I'm not sure whether we confirmed that initializers would be allowed on either part (not just the defining part) or that it would be an error if both parts have an initializer (even when set to the same constant, etc).

proposals/field-keyword.md
```diff
@@ -666,6 +656,10 @@ class B
 
 In the example above, binding to the backing field should result in an error: "initializer cannot reference non-static field".
 
+#### Answer
+
+`field` should not be a keyword in a property initializer. It will continue to bind (or fail to bind) as an identifier.
```
jaredpar, line 661:
```suggestion
`field` will not be a keyword in a property initializer. It will continue to bind (or fail to bind) as an identifier.
```

proposals/field-keyword.md
jaredpar: Think this section should move to having a more full example that helps demonstrate why we chose to allow this behavior.

cston: > I'm not sure whether we confirmed that initializers would be allowed on either part

From [LDM-2024-08-26](https://github.com/dotnet/csharplang/blob/main/meetings/2024/LDM-2024-08-26.md#conclusion-1), an initializer is allowed on either part, but not both.

proposals/field-keyword.md
jnm2: Thanks! Copying in from Fred's notes now.

proposals/field-keyword.md
jnm2: Updating using the wording from the official notes!

cston (approved):

333fred (approved):

PR #8413, opened 2024-09-06
Allow yield inside of try / catch and catch

jaredpar:
This proposal expands the use of `yield` to

- `try` portion of a `try / catch` block
- `catch` portion of a `try / catch` block with some restrictions around `finally`

### DIFF
--- a/proposals/yield-in-try-catch.md
+++ b/proposals/yield-in-try-catch.md
@@ -1,0 +1,388 @@
+# Permit yield in a try with a catch clause
+
+## Summary
+
+This proposal will allow `yield` statements to be written inside `try` and `catch` blocks.
+
+## Motivation
+
+The inability to use `yield` inside a `try` and `catch` block is a long standing pain point for customers. The restriction prevents the use of `yield` in a number of appealing scenarios. For example:
+
+```csharp
+IEnumerable<int> M1(IEnumerable<e> col)
+{
+    try
+    {
+        foreach (var item in col)
+        {
+            var otherItem = Translate(item);
+            yield return otherItem;
+        }
+    }
+    catch (Exception ex)
+    {
+        throw new MoreSpecificException(ex);
+    }
+}
+
+IEnumerable<int> M2(IEnumerable<e> col)
+{
+    foreach (var item in col)
+    {
+        try
+        {
+            var otherItem = Translate(item);
+            yield return otherItem;
+        }
+        catch (Exception ex)
+        {
+            Log(ex);
+        }
+    }
+}
+```
+
+These restrictions exist in large part because it presented technical challenges for the native compiler and it wasn't a high enough priority item. The state machine support in Roslyn already supports the type of transforms necessary for this feature.
+
+This proposal will allow for the common cases of `yield` within `try` and `catch` without the awkward workarounds that are necessary today to move the `yield` outside the `try`.
+
+## Detailed Design
+
+### yield in try and catch
+
+The `yield` statement will be allowed inside `try` and `catch` blocks. The behavior will be the same as a `yield` statement today:
+
+- `yield return` will cause the method to suspend and return the value to the caller via `Current`.
+- `yield break` will cause the iterator to return `false` from `MoveNext`.
+
+For example:
+
+```csharp
+foreach (var e in Iterator())
+{
+    Console.WriteLine(e);
+}
+
+IEnumerable<int> Iterator()
+{
+    try
+    {
+        yield return 1;
+        throw new Exception("");
+    }
+    catch (Exception ex)
+    {
+        yield return 2;
+    }
+}
+```
+
+This code will output:
+
+```cmd
+1
+2
+```
+
+The `catch` block will go through the same rewriting as an `await` inisde of a `catch` block. That will be observable when the `throw;` statement is used to rethrow an exception as it will reset the stack trace vs. perserving it (just as it is for an `async` method).
+
+The `yield` statement will not be allowed inside a `catch` when there is an associated or nested `finally` block. That would allow `yield` to be executed in the `Dispose` method which is not supported ([more details][catch-finally]).
+
+Detailed notes:
+
+- The `yield` statement will be allowed in a `try` block.
+- The `yield` statement will be allowed in a `catch` block provided that:
+  - The `try` block does not contain a `finally` block.
+  - The `try` block does not have a nested `finally` block.
+
+### Dispose and finally
+
+A lesser known detail of iterators is that `finally` blocks can be executed as part of the `IDisposable.Dispose` implementation. The `Dispose` method has the same state machine implementation as the generated `MoveNext` except it only has the parts necessary for executing the `finally` blocks. That allows `Dispose` to _resume_ the method from the last suspend and execute the `finally` that were _active_ at the last suspend point.
+
+For example consider this code sample:
+
+```csharp
+var e = Iterator().GetEnumerator();
+e.MoveNext();
+e.Dispose();
+
+static T M<T>(T t) => t;
+
+static IEnumerable<int> Iterator()
+{
+    try
+    {
+        try
+        {
+            yield return M(1);
+            Console.WriteLine("After yield");
+        }
+        finally
+        {
+            Console.WriteLine("Finally Inner");
+        }
+    }
+    finally
+    {
+        Console.WriteLine("Finally Outer");
+    }
+}
+```
+
+This program will output:
+
+```cmd
+Finally Inner
+Finally Outer
+```
+
+The code generation for the `Dispose` method is meant to mirror the original `finally` structure as closely as possible. This includes execution of the code in the face of an exception during `Dispose`. This is achieved by refactoring the contents of the `finally` block into a method on the iterator and then having both `MoveNext` and `Dispose` generate the same `try / finally` structure and call into the methods.
+
+For example this is the `Dispose` method for the above iterator:
+
+```csharp
+[DebuggerHidden]
+void IDisposable.Dispose()
+{
+    int num = <>1__state;
+    if ((uint)(num - -4) > 1u && num != 1)
+    {
+        return;
+    }
+    try
+    {
+        if (num != -4 && num != 1)
+        {
+            return;
+        }
+        try
+        {
+        }
+        finally
+        {
+            <>m__Finally2();
+        }
+    }
+    finally
+    {
+        <>m__Finally1();
+    }
+}
+```
+
+This behavior is important to understand when considering the code generation for `try / catch` blocks.
+
+### Code generation yield inside try with catch
+
+The code generation for iterators that have `yield` inside `try` blocks must preserve the same exception semantics for the `catch` in both `MoveNext` and `Dispose`. To achieve this the compiler will take a similar approach to what it does for `finally` blocks.
+
+For every `catch` block where the `try` has a nested `try / finally` with `yield`:
+
+1. The contents of the `catch` block will be generated into a parameterless method on the iterator with a `void` return.
+2. The contents of the `when` clause will be generated into a parameterless method on the iterator with a `bool` return.
+3. The `catch` block will be replaced with a call to the generated method in `MoveNext`.
+4. The `when` clause will be replaced with a call to the generated method in `MoveNext`.
+5. The `Dispose` method will mirror `catch` and `finally` blocks in the same way it mirrors `finally` blocks today and dispatch to the appropriate method
+
+The exception object will be lifted into a field just as any other local would be and accesses to it in the generated `catch` and `when` methods will use the field. In the case the language allowed `when` clauses the pattern variables would be lifted into fields as well.
+
+For example consider this code sample:
+
+```csharp
+static IEnumerable<int> Iterator()
+{
+    try
+    {
+        try
+        {
+            yield return M(1);
+        }
+        finally
+        {
+            Console.WriteLine("Finally Inner");
+        }
+    }
+    catch (InvalidOperationException ex1) when (ex.Message.Contains("hello"))
+    {
+        Console.WriteLine("Catch1");
+    }
+    catch (Exception ex2) 
+    {
+        Console.WriteLine("Catch2");
+    }
+    finally
+    {
+        Console.WriteLine("Finally Outer");
+    }
+}
+static T M<T>(T t) => t;
+```
+
+Will generate the following `Dispose` method:
+
+```csharp
+[DebuggerHidden]
+void IDisposable.Dispose()
+{
+    int num = <>1__state;
+    if ((uint)(num - -4) > 1u && num != 1)
+    {
+        return;
+    }
+    try
+    {
+        if (num != -4 && num != 1)
+        {
+            return;
+        }
+        try
+        {
+        }
+        finally
+        {
+            <>m__Finally2();
+        }
+    }
+    catch (InvalidOperationException ex) when (<>1__ex1 = ex, <>m__When1())
+    {
+        <>m__Catch1();
+    }
+    catch (Exception ex)
+    {
+        <>1__ex2 = ex;
+        <>m__Catch1();
+    }
+    finally
+    {
+        <>m__Finally1();
+    }
+}
+```
+
+The `<>1__ex1 = ex` in the `when` clause is not legal but the IL generated for the `when` will conceptually have this behavior.
+
+### Code generation yield inside try / catch in async iterators
+
+The code generation for `try / catch` blocks in async iterators will be largely the same as traditional iterators. The difference is that the return type of generated `catch` methods will be `ValueTask` instead of `void`.
+
+### Code generation yield inside catch
+
+The restrictions on the feature mean that `yield` inside a `catch` cannot be observed from a `finally` block. That means the code generation does not need to consider the impact on `Dispose` and can focus soley on `MoveNext`.
+
+Given that the code generation for `yield` inside `catch` will have the same structure as `await` inside of `catch`. Essentially the user written contents of the `catch` block will be moved outside the `catch`. The `catch` block will be replaced with saving the `Exception` object into the state machine and updating of the state variable to reflect execution is logically inside the catch block.
+
+For example consider this code sample:
+
+```csharp
+IEnumerable<int> M()
+{
+    try
+    {
+        M();
+    }
+    catch (Exception ex)
+    {
+        Console.WriteLine("Catch1");
+        yield return 1;
+        Console.WriteLine("Catch2");
+    }
+    Console.WriteLine("Done");
+}
+```
+
+This would be generated as effectively:
+
+```csharp
+bool MoveNext()
+{
+    switch (<>1__state)
+    {
+    case 0:
+        <>1__state = -1;
+        try
+        {
+            M();
+        }
+        catch (Exception ex)
+        {
+            <>3__ex = ex;
+            <>1__state = 1;
+        }
+
+        int num2 = <>1__state;
+        if (num2 != 1)
+        {
+            goto case 2;
+        }
+
+        Console.WriteLine("Catch1");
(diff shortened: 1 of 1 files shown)


timcassell:
Regarding `yield-inside-catch-with-nested-finally`, could that be resolved by generating 2 separate code paths, one for `MoveNext` and another for `Dispose`?

jaredpar:
> Regarding yield-inside-catch-with-nested-finally, could that be resolved by generating 2 separate code paths, one for MoveNext and another for Dispose?

I think the big problem is agreeing on what behavior `yield` should have when it's encountered in the `Dispose` path? Is it ignored? Do we throw an exception? etc ... My suspicion is that there won't be broad agreement on what is to be done there. 

Basically I think the problem is more along agreeing what behavior is correct vs. implementing that behavior.

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,388 @@
...
+This behavior is important to understand when considering the code generation for `try / catch` blocks.
+
+### Code generation yield inside try with catch
+
+The code generation for iterators that have `yield` inside `try` blocks must preserve the same exception semantics for the `catch` in both `MoveNext` and `Dispose`. To achieve this the compiler will take a similar approach to what it does for `finally` blocks.
+
+For every `catch` block where the `try` has a nested `try / finally` with `yield`:
+
+1. The contents of the `catch` block will be generated into a parameterless method on the iterator with a `void` return.
+2. The contents of the `when` clause will be generated into a parameterless method on the iterator with a `bool` return.
```
jcouv, line 182: Ah, I see that we don't need parameters for this extracted method after all. Makes sense :-)

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,388 @@
...
+        <>m__Finally1();
+    }
+}
+```
+
+The `<>1__ex1 = ex` in the `when` clause is not legal but the IL generated for the `when` will conceptually have this behavior.
+
+### Code generation yield inside try / catch in async iterators
+
+The code generation for `try / catch` blocks in async iterators will be largely the same as traditional iterators. The difference is that the return type of generated `catch` methods will be `ValueTask` instead of `void`.
```
jcouv, line 266: > The code generation for `try / catch` blocks in async iterators will be largely the same as traditional iterators.

I didn't follow this. Async iterators are more like async state machines than iterator state machines. 
Instead of generating a MoveNext and a Dispose method with similar structure, we use a single method (MoveNext) for both normal and disposal execution. A flag indicates if we're in disposal and we use it to skip over non-disposal code as appropriate.

Here are some notes on [disposal in async iterators.](https://github.com/jcouv/roslyn/blob/d0a56dda8423ad4ddd1f100c66ab56c0f9a343c3/async-iterator.md#design) Happy to discuss offline.

proposals/yield-in-try-catch.md
jaredpar: Yeah I realized it too once I started digging through a bunch of our generated code.

proposals/yield-in-try-catch.md
jaredpar: Hmm. Maybe i was looking at the wrong code gen here. Will take a look at this.

poizan42:
May I suggest that instead of resetting the stack trace, `throw;` is lowered to `System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(ex).Throw()` so the original stack trace is preserved like with a normal `throw;`?

(It should be `.Capture(ex).Throw()` and not `.Throw(ex)` as the latter is a static method and cannot be polyfilled if people want to use the new C# version with runtimes earlier than .net core 2.0)

proposals/yield-in-try-catch.md
```diff
@@ -42,13 +42,13 @@ IEnumerable<int> M2(IEnumerable<e> col)
 }
 ```
 
-These restrictions exist in large part because it presented technical challenges for the native compiler and it wasn't a high enough priority item. The state machine support in Roslyn already supports the type of transforms necessary for this feature.
+These restrictions exist in large part because it presented technical challenges for the native compiler and it wasn't a high enough priority item. That is no longer a blocker as the state machine in Rsolyn supports the types of transforms necessary to support this feature. There are still several semantic challenges to work through and this proposal has been written to address those.
```
jjonescz, line 45:
```suggestion
These restrictions exist in large part because it presented technical challenges for the native compiler and it wasn't a high enough priority item. That is no longer a blocker as the state machine in Roslyn supports the types of transforms necessary to support this feature. There are still several semantic challenges to work through and this proposal has been written to address those.
```

proposals/yield-in-try-catch.md
```diff
@@ -172,100 +170,130 @@ void IDisposable.Dispose()
...
 
-Will generate the following `Dispose` method:
+This program will output:
+
+```cmd
+inner finally
+outer finally
+```
+
+The `"after catch"` is not printed beacuse only the `finally` structure is mirrored in the `Dispose` method. The `catch`, like all other statements between `finally` is not included. statements in between the `catch` and `finally` are not executed in `Dispose`. This may seem odd at first glance but is leaning into the specified behavior for iterator `Dispose`.
```
jjonescz, line 222:
```suggestion
The `"after catch"` is not printed beacuse only the `finally` structure is mirrored in the `Dispose` method. The `catch`, like all other statements between `finally` is not included. Statements in between the `catch` and `finally` are not executed in `Dispose`. This may seem odd at first glance but is leaning into the specified behavior for iterator `Dispose`.
```

proposals/yield-in-try-catch.md
```diff
@@ -172,100 +170,130 @@ void IDisposable.Dispose()
...
+
+### Dispose and finally in async iterators
+
+The `DiposeAsync` behavior for async iterators mirrors that of traditional iterators in that it executes the `finally` blocks at the the point the state machine is suspended. However it does this by setting the `state` variable to a value that represents disposing and then calls `MoveNextAsync`. The implementation of `MoveNextAsync` will then execute only the `finally` blocks for the suspend point. The `DisposeAsync` method does not have a mirror copy of the `finally` blocks.
```
jjonescz, line 226: Curious, is there some reason why the codegen of DisposeAsync is different from Dispose?

proposals/yield-in-try-catch.md
```diff
@@ -172,100 +170,130 @@ void IDisposable.Dispose()
...
+
+### Code generation of yield inside try / catch in async iterators
+
+The code generation of async iterators will change such that `catch` blocks do not observably execute during `DisposeAsync`. To achieve this all `catch` blocks visible from a `yield` will rethrow exceptions if the state machine is in a disposing state. For example consider the following code::
```
jjonescz, line 230:
```suggestion
The code generation of async iterators will change such that `catch` blocks do not observably execute during `DisposeAsync`. To achieve this all `catch` blocks visible from a `yield` will rethrow exceptions if the state machine is in a disposing state. For example consider the following code:
```

proposals/yield-in-try-catch.md
```diff
@@ -172,100 +170,130 @@ void IDisposable.Dispose()
...
+inner finally
+outer finally
+```
+
+To achieve this the `catch` will be effectively rewritten as follows:
+
+```csharp
+catch (Exception ex)
+{
+    if (<>1__state == /* dispatching state */)
```
jjonescz, line 278:
```suggestion
    if (<>1__state == /* disposing state */)
```

proposals/yield-in-try-catch.md
```diff
@@ -172,100 +170,130 @@ void IDisposable.Dispose()
...
+The generator will also need to modify any `when` clauses on `catch` blocks to ensure they don't execute during `DisposeAsync`. To do this a prefix will be generated that returns `false` when the state machine is in a disposing state. For example:
 
-### Code generation yield inside try / catch in async iterators
+```csharp
+// User code
+catch (Exception ex) when (SomeMethod(ex))
 
-The code generation for `try / catch` blocks in async iterators will be largely the same as traditional iterators. The difference is that the return type of generated `catch` methods will be `ValueTask` instead of `void`.
+// Generated code
+catch (Exception ex) when (<>1__state == /* dispatching state */ ? false : SomeMethod(ex))
```
jjonescz, line 293:
```suggestion
catch (Exception ex) when (<>1__state == /* disposing state */ ? false : SomeMethod(ex))
```

proposals/yield-in-try-catch.md
```diff
@@ -172,100 +170,130 @@ void IDisposable.Dispose()
...
 }
-static T M<T>(T t) => t;
 ```
 
-Will generate the following `Dispose` method:
+This program will output:
+
+```cmd
+inner finally
+outer finally
```
jjonescz, line 219: Consider adding the exception to the output since it seems it will be thrown from the Dispose invocation, right?

proposals/yield-in-try-catch.md
```diff
@@ -334,52 +362,135 @@ bool MoveNext()
...
+1. Throw an exception when `yield` is encountered in `Dispose`.
+2. Silently ignore the value and continue executing the `finally` block without suspend.
+
+Neither of these seem like desirable outcomes and as such `yield` will not be allowed in `finally` blocks.
+
+## Open Issues
+
+### Preserve catch in Dispose paths
+
+The `Dispose` method could include mirroring both `catch` and `finally` blocks. This would allow the `catch` block to be executed in `Dispose` when `finally` blocks threw an exceptoin. The approach for this would be to do the followng.
```
jjonescz, line 398:
```suggestion
The `Dispose` method could include mirroring both `catch` and `finally` blocks. This would allow the `catch` block to be executed in `Dispose` when `finally` blocks threw an exception. The approach for this would be to do the following.
```

proposals/yield-in-try-catch.md
```diff
@@ -334,52 +362,135 @@ bool MoveNext()
...
+
+For every `catch` block where the `try` has a nested `try / finally` with `yield`:
+
+1. The contents of the `catch` block will be generated into a parameterless method on the iterator with a `void` return.
+2. The contents of the `when` clause will be generated into a parameterless method on the iterator with a `bool` return.
+3. The `catch` block will be replaced with a call to the generated method in `MoveNext`.
+4. The `when` clause will be replaced with a call to the generated method in `MoveNext`.
+5. The `Dispose` method will mirror `catch` and `finally` blocks in the same way it mirrors `finally` blocks today and dispatch to the appropriate method
```
jjonescz, line 406: Could we also do codegen similar to async iterators, i.e., share MoveNext and Dispose implementations?

proposals/yield-in-try-catch.md
```diff
@@ -334,52 +362,135 @@ bool MoveNext()
...
+        <>m__Finally1();
+    }
+}
+```
 
-### yield inside finally
+The `<>1__ex1 = ex` in the `when` clause is not legal but the IL generated for the `when` will conceptually have this behavior.
 
-The `yield` statement inside a `finally` creates the same type of code generation issues as [catch with finally][catch-finally]. For that reason it was excluded from this proposal.
+This would add a bit of complexity the feature and it only produces observable differences when a `finally` block throws an exception on the `Dispose` path. That is likely a rare case. Further it potentially increases the complexity for developers. The `Dispose` method at first glance is likely unintuitive in that it only mirrors `finally` blocks and no other statements but that is also a very simple rule to learn. Preserving `catch` and `finally` and discussing how the code flows between them is potentially more complex.
```
jjonescz, line 485:
```suggestion
This would add a bit of complexity to the feature and it only produces observable differences when a `finally` block throws an exception on the `Dispose` path. That is likely a rare case. Further it potentially increases the complexity for developers. The `Dispose` method at first glance is likely unintuitive in that it only mirrors `finally` blocks and no other statements but that is also a very simple rule to learn. Preserving `catch` and `finally` and discussing how the code flows between them is potentially more complex.
```

proposals/yield-in-try-catch.md
jaredpar: Yse good point.

proposals/yield-in-try-catch.md
jaredpar: That is definitely an option, essentially changing iterators to favor the async iterators code gen approach. I _think_ that would be fine as I can't see any technical reason why iterators have to be done that way but we'd want to vet it out a bit

proposals/yield-in-try-catch.md
jaredpar: I suspect it's mostly an accident of history. The `async` state machine didn't have to deal with `Dispose` scenarios initially hence never went through the necessary rewriting process. At the point we added async iterators it was easier to just tweak the `state` field to represent disposing and using the existing state machine. 

@jcouv can confirm or give a better explanation ;)

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+        finally
+        {
+            Console.WriteLine("inner finally");
+            if (b)
+            {
+                throw new Exception();
+            }
+        }
+    }
+    catch
```
jcouv, line 256: What if the catch has a `when` clause with side-effects? Would the side-effects be executed?

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+inner finally
+outer finally
+Unhandled exception. System.Exception: Exception of type 'System.Exception' was thrown
+```
+
+The `"after catch"` is not printed beacuse only the `finally` structure is mirrored in the `Dispose` method. The `catch`, like all other statements between `finally` is not included. Statements in between the `catch` and `finally` are not executed in `Dispose`. This may seem odd at first glance but is leaning into the specified behavior for iterator `Dispose`.
+
+### Dispose and finally in async iterators
+
+The `DiposeAsync` behavior for async iterators mirrors that of traditional iterators in that it executes the `finally` blocks at the the point the state machine is suspended. However it does this by setting the `state` variable to a value that represents disposing and then calls `MoveNextAsync`. The implementation of `MoveNextAsync` will then execute only the `finally` blocks for the suspend point. The `DisposeAsync` method does not have a mirror copy of the `finally` blocks.
```
jcouv, line 227: >it does this by setting the `state` variable [](http://example.com/codeflow?start=183&length=44)

It's not the "state" variable that indicates disposal mode, but a dedicated variable. It's called "disposeMode" in the speclet: [https://github.com/dotnet/roslyn/blob/main/docs/features/async-streams.md#detailed-design-for-async-iterator-methods](https://github.com/dotnet/roslyn/blob/main/docs/features/async-streams.md#detailed-design-for-async-iterator-methods)

proposals/yield-in-try-catch.md
jcouv: Yes, it was simpler and more elegant/lightweight to use a single state machine with a "disposal" flag for async-iterators. I was aware of the codegen for iterators but didn't see a benefit to the added overhead (second state machine with bunch of logic extracted into helper methods).

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+// User code
+catch (Exception ex) when (SomeMethod(ex))
+
+// Generated code
+catch (Exception ex) when (<>1__state == /* disposing state */ ? false : SomeMethod(ex))
+```
+
+### Code generation of yield inside catch
+
+The restrictions on the feature mean that `yield` inside a `catch` cannot be observed from a `finally` block. That means the code generation does not need to consider the impact on `Dispose` and can focus soley on `MoveNext`.
```
jcouv, line 299: typo: soley -> solely
```suggestion
The restrictions on the feature mean that `yield` inside a `catch` cannot be observed from a `finally` block. That means the code generation does not need to consider the impact on `Dispose` and can focus solely on `MoveNext`.
```

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+Detailed notes:
+
+- The `yield` statement will be allowed in a `try / catch` block
+- The `yield` statement will be allowed in a `catch` block
+
+### Dispose and finally in iterators
+
+A lesser known detail of iterators is that `finally` blocks can be executed as part of the `IDisposable.Dispose` implementation. The `Dispose` method has the same state machine structure as the generated `MoveNext` except it only has the parts necessary for executing the `finally` blocks. That allows `Dispose` to _resume_ the method from the last suspend and execute the `finally` that were _active_ at the last suspend point.
+
+The `Dispose` method _only_ executes the `finally` blocks and ignores all code inbetween them. For example consider this code sample:
```
jcouv, line 98: Typo: inbetween doesn't appear to be a word
```suggestion
The `Dispose` method _only_ executes the `finally` blocks and ignores all code between them. For example consider this code sample:
```

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+
+This program will output:
+
+```cmd
+inner finally
+outer finally
+Unhandled exception. System.Exception: Exception of type 'System.Exception' was thrown
+```
+
+The `"after catch"` is not printed beacuse only the `finally` structure is mirrored in the `Dispose` method. The `catch`, like all other statements between `finally` is not included. Statements in between the `catch` and `finally` are not executed in `Dispose`. This may seem odd at first glance but is leaning into the specified behavior for iterator `Dispose`.
```
Rageking8, line 223:
```suggestion
The `"after catch"` is not printed because only the `finally` structure is mirrored in the `Dispose` method. The `catch`, like all other statements between `finally` is not included. Statements in between the `catch` and `finally` are not executed in `Dispose`. This may seem odd at first glance but is leaning into the specified behavior for iterator `Dispose`.
```

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+
+### Dispose and finally in async iterators
+
+The `DiposeAsync` behavior for async iterators mirrors that of traditional iterators in that it executes the `finally` blocks at the the point the state machine is suspended. However it does this by setting the `disposeMode` variable and then calls `MoveNextAsync`. The implementation of `MoveNextAsync` will then execute only the `finally` blocks for the suspend point. The `DisposeAsync` method does not have a mirror copy of the `finally` blocks.
```
Rageking8, line 227:
```suggestion
The `DisposeAsync` behavior for async iterators mirrors that of traditional iterators in that it executes the `finally` blocks at the point the state machine is suspended. However it does this by setting the `disposeMode` variable and then calls `MoveNextAsync`. The implementation of `MoveNextAsync` will then execute only the `finally` blocks for the suspend point. The `DisposeAsync` method does not have a mirror copy of the `finally` blocks.
```

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+        try
+        {
+            yield return M(1);
+        }
+        finally
+        {
+            Console.WriteLine("Finally Inner");
+        }
+    }
+    catch (InvalidOperationException ex1) when (ex.Message.Contains("hello"))
```
Rageking8, line 427:
```suggestion
    catch (InvalidOperationException ex1) when (ex1.Message.Contains("hello"))
```

Rageking8 (changes requested):
Bunch of small typos.

proposals/yield-in-try-catch.md
```diff
@@ -0,0 +1,500 @@
...
+  - The `try` block does not contain a `finally` block.
+  - The `try` block does not have a nested `finally` block.
+
+On the whole it seems simpler to not support `catch` in `Dispose` paths.
+
+## Related Issues
+
+Related Items:
+
+- https://github.com/dotnet/csharplang/discussions/765
```
hawkerm, line 500: Could be good to link to other duplicate items:
- https://github.com/dotnet/csharplang/discussions/6819
- https://github.com/dotnet/csharplang/discussions/8994
- https://github.com/dotnet/docs/issues/42112

PR #8466, opened 2024-09-30
Field-backed properties: open question for interface properties

cston:


RikkiGibson:
What would the alternative be? Consider it an ordinary interface property where one of the accessors has a default implementation?

cston:
> What would the alternative be? Consider it an ordinary interface property where one of the accessors has a default implementation?

Yes, or disallow a combination of manually- and auto-implemented accessors.

RikkiGibson:
> Yes, or disallow a combination of manually- and auto-implemented accessors.

so if we did that, then `static string Prop { get; set => ... }` would also be disallowed?

jnm2 (approved):

PR #8485, opened 2024-10-04
Update proposal for simple lambda parameters with modifiers

jnm2:
https://github.com/dotnet/csharplang/issues/338

### DIFF
--- a/meetings/2024/LDM-2024-02-21.md
+++ b/meetings/2024/LDM-2024-02-21.md
@@ -17,7 +17,7 @@
 ### Declaration of ref/out parameters in lambdas without typename
 
 https://github.com/dotnet/csharplang/issues/338  
-https://github.com/dotnet/csharplang/blob/main/proposals/ref-out-lambda-params.md
+https://github.com/dotnet/csharplang/blob/main/proposals/simple-lambda-parameters-with-modifiers.md
 
 We started today by looking at an Any Time proposal specification to address a common cliff: when using a delegate type with a modifier of any kind, users are required to spell out the parameter types
 of that lambda. This is a change that the LDM has long mentioned in passing, but has never actually gone through and specified. In addition to the main proposal, we also discussed few potential
--- a/proposals/ref-out-lambda-params.md
+++ b/proposals/ref-out-lambda-params.md
@@ -1,6 +1,6 @@
 # Declaration of lambda parameters with modifiers without type name
 
-## Summary  
+## Summary
 
 Allow lambda parameter declarations with modifiers (`in` / `ref` / `out` / etc.) to be declared without requiring their type names.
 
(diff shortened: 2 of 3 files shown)


proposals/implicit-lambda-parameters-with-modifiers.md
```diff
@@ -30,25 +30,32 @@ implicit_anonymous_function_signature
...
 
-implicit_anonymous_function_parameter_ex
-    : anonymous_function_parameter_modifier? identifier
+explicit_anonymous_function_parameter
+    : 'scoped'? anonymous_function_parameter_modifier? type identifier
+    ;
+
+anonymous_function_parameter_modifier
+    : 'in'
+    | 'ref' 'readonly'?
```
jnm2, line 47: I'm not able to find where `readonly` started being valid syntax within a lambda parameter declaration. <https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/ref-readonly-parameters.md> says "No changes in grammar are necessary," but searching the repo for `'readonly'` as well as for `anonymous_function_parameter_modifier` failed to show where the the grammar had been updated.

CyrusNajmabadi (approved):

proposals/implicit-lambda-parameters-with-modifiers.md
BillWagner: Well, `ref readonly` was added in C# 7: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/variables#971-general

That may be what this means

proposals/implicit-lambda-parameters-with-modifiers.md
```diff
@@ -1,10 +1,10 @@
-# Declaration of lambda parameters with modifiers without type name
+# Implicit lambda parameters with modifiers
```
333fred, line 1: Can't say I like this rename. Perhaps `implicitly-typed`? The parameter isn't implicit.

proposals/implicit-lambda-parameters-with-modifiers.md
```diff
@@ -103,3 +109,11 @@ We would also update the semantic specification to say:
 + is converted to a compatible delegate type or expression tree type, that type 
 + provides the parameter types (§10.7).
 ```
+
+#### Answer
+
+Neither attributes nor default parameter values will be supported without a fully-typed lambda.
+
+## Design meetings
```
333fred, line 117: Let's please not include these links. The champion issue collates all the design meetings, and it is much more difficult for me to keep links in the repo up to date than it is to keep the champion issues up to date.

proposals/implicit-lambda-parameters-with-modifiers.md
jnm2: The spec calls them "implicit parameters" versus "explicit parameters"- but I don't disagree. What if I drop the commit so that I call them "simple parameters" instead?

proposals/implicit-lambda-parameters-with-modifiers.md
jnm2: So perhaps the C# 12 "ref readonly parameters" proposal overlooked adding `'readonly'` to `anonymous_function_parameter_modifier`?

proposals/implicit-lambda-parameters-with-modifiers.md
333fred: That works for me.

333fred (approved):

PR #8525, opened 2024-10-18
Create YAEP (Yet Another Extensions Proposal)

DustinCampbell:
This change adds my 'Extensions: An Evolution of Extension Methods' proposal.

### DIFF
--- a/proposals/extensions-an-evolution-of-extension-methods.md
+++ b/proposals/extensions-an-evolution-of-extension-methods.md
@@ -1,0 +1,591 @@
+# Extensions: An Evolution of Extension Methods
+
+* [x] Proposed
+* [ ] Prototype
+* [ ] Implementation
+* [ ] Specification
+
+## Summary
+[summary]: #summary
+
+This proposal presents a design for "extension everything" as an evolution of classic extension methods and allows for
+future expansion.
+
+### Goals
+- Describe a new syntax for extension members that scales to member kinds beyond just instance methods.
+- Provide support for all functionality available to classic extension methods.
+- Guarantee binary and source compatibility for classic extension method scenarios in the new syntax.
+- Design support for the highest value extension member kinds that have the best chance of being implemented in the
+  C# 14 timeframe.
+- Lay the design groundwork for further extension member kinds, if or when the scenarios are compelling enough
+  to pursue.
+
+### Non-Goals
+- Do not consider syntax for a possible future “roles” or "explicit extensions" feature. This proposal assumes that
+  “extensions” and “roles” have no dependency on one another. However, it allows for future synergy through additional
+  language features.
+- Do not consider syntax to allow type parameters on member kinds that do not support them today. Although, this could
+  be supported by future work.
+
+## Motivation
+[motivation]: #motivation
+
+[Extension methods](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods)
+are wildly popular! They first appeared in C# 3 as a smallish feature intended to support the syntactic rewrites of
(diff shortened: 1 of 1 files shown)


proposals/extensions-an-evolution-of-extension-methods.md
```diff
@@ -0,0 +1,591 @@
...
+{
+    public static string string.Create<TState>(int length, TState state, SpanAction<char, TState> action)
+    {
+    }
+}
+```
+
+Like an instance extension method's receiver parameter, it is necessary to state the target type of the static
+extension method before the method name[^8]. This allows the declaration syntax to align with the calling syntax. In
+order to allow an  A downside is that the type must be restated even if an optional _for-clause_ is defined, but this
```
jnm2, line 255: Missing words?

proposals/extensions-an-evolution-of-extension-methods.md
```diff
@@ -0,0 +1,591 @@
...
+
+    public static implicit operator int(string x) => ...;
+}
+```
+
+Instance indexers are straightforward and would likely be useful. Events are possible, but similar to extension
+properties, they would require `add` and `remove` accessors to avoid creating state that wouldn't flow with the
+receiver.
+
+Note that operators overloads and user-defined conversions are declared using the same syntax as always,. It is not
```
jnm2, line 444:
```suggestion
Note that operators overloads and user-defined conversions are declared using the same syntax as always. It is not
```

proposals/extensions-an-evolution-of-extension-methods.md
```diff
@@ -0,0 +1,591 @@
...
+## Design meetings
+
+<!-- Link to design notes that affect this proposal, and describe in one sentence for each what changes they led to. -->
+
+[^1]: For example, the query expression, `from x in Enumerable.Range(1, 10) select x * x` is rewritten syntactically at
+compile-time as `Enumerable.Range(1, 10).Select(x => x * x)`. The `Enumerable.Select(...)` extension method ensure that
+this rewrite compiles.
+[^2]: Consider the [`ILogger`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger)
+interface, which only has three interface members that need to be implemented. A much larger API surface is available
+for an `ILogger` implementation by the 29 (as of this writing) extension methods defined by [`LoggerExtesnions`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions)
```
jnm2, line 576:
```suggestion
for an `ILogger` implementation by the 29 (as of this writing) extension methods defined by [`LoggerExtensions`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions)
```

proposals/extensions-an-evolution-of-extension-methods.md
DustinCampbell: Thanks!

proposals/extensions-an-evolution-of-extension-methods.md
```diff
@@ -0,0 +1,591 @@
...
+    | 'internal'
+    | unsafe_modifier   // unsafe code support
+    ;
+    
+for_clause
+    : 'for' type
+    ;
+
+extension_container_body
+    : '{' extension_member_declaration* class_member_declaration* '}'
```
RikkiGibson, line 97: It looks like this grammar supports "0 or more extension member decls, then 0 or more class member decls"? Is it intended that all the `class_member_declaration`s appear after the `extension_member_declaration`s?

DustinCampbell: Not intended, no. Just an artifact of being quick and dirty. Will fix, thanks!

PR #8531, opened 2024-10-23
use nameof in caller-argument-expression

WeihanLi:
use `nameof` in `CallerArgumentExpression` for better maintainability

from comment: https://github.com/dotnet/aspnetcore/pull/58514#pullrequestreview-2386629058

Since the `nameof` is available since C# 11, not sure if it's the right way to go, maybe an additional comment for using the `nameof`?


333fred:
We should leave this specification as is. As you mentioned, nameof parameters were not available when this feature was created, so the spec cannot use them in examples or risk causing confusion.

WeihanLi:
Get it, thanks very much

PR #8536, opened 2024-10-24
Fix cyclic assignment ref safety docs

jjonescz:
Related to https://github.com/dotnet/roslyn/issues/75592.

### DIFF
--- a/proposals/csharp-11.0/low-level-struct-improvements.md
+++ b/proposals/csharp-11.0/low-level-struct-improvements.md
@@ -604,8 +604,12 @@ The [rationale](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-
 <a name="rules-unscoped"></a>
 
 To fix this the language will provide the opposite of the `scoped` lifetime annotation by supporting an `UnscopedRefAttribute`. This can be applied to any `ref` and it will change the *ref-safe-context* to be one level wider than its default. For example:
-- if applied to a `struct` instance method it will become *return only* where previously it was *function-member*.
-- if applied to a `ref` parameter it will become *caller-context* where previously it was *return only*
+
+| UnscopedRef applied to | Original *ref-safe-context* | New *ref-safe-context* |
+| --- | --- | --- |
+| instance member | function-member | return-only |
+| `in` / `ref` parameter | return-only | caller-context |
+| `out` parameter | function-member | return-only |
 
 When applying `[UnscopedRef]` to an instance method of a `struct` it has the impact of modifying the implicit `this` parameter. This means `this` acts as an unannotated `ref` of the same type. 
 
@@ -634,7 +638,7 @@ ref int SneakyOut([UnscopedRef] out int i)
 }
 ```
 
-For the purposes of ref safe cotnext rules, such an `[UnscopedRef] out` is considered simply a `ref`. Similar to how `in` is considered `ref` for lifetime purposes. 
+For the purposes of ref safe context rules, such an `[UnscopedRef] out` is considered simply a `ref`. Similar to how `in` is considered `ref` for lifetime purposes. 
 
 The `[UnscopedRef]` annotation will be disallowed on `init` members and constructors inside `struct`. Those members are already special with respect to `ref` semantics as they view `readonly` members as mutable. This means taking `ref` to those members appears as a simple `ref`, not `ref readonly`. This is allowed within the boundary of constructors and `init`. Allowing `[UnscopedRef]` would permit such a `ref` to incorrectly escape outside the constructor and permit mutation after `readonly` semantics had taken place.
 
@@ -1810,7 +1814,8 @@ void Usage()
 
 To make these APIs usable the compiler ensures that the `ref` lifetime for a `ref` parameter is smaller than lifetime of any references in the associated parameter value. This is the rationale for having *ref-safe-context* for `ref` to `ref struct` be *return-only* and `out` be *caller-context*. That prevents cyclic assignment because of the difference in lifetimes.
 
-Note that `[UnscopedRef]` promotes the *ref-safe-context* of any `ref` to `ref struct` values to *caller-context*:
+Note that `[UnscopedRef]` [promotes](#rules-unscoped) the *ref-safe-context* of any `ref`/`out` to `ref struct` values to *caller-context*
+and hence it allows for cyclic assignment and forces a viral use of `[UnscopedRef]` for a `ref struct`:
 
 ```c#
 S F()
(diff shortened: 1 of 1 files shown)


proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -1810,47 +1810,27 @@ void Usage()
 
 To make these APIs usable the compiler ensures that the `ref` lifetime for a `ref` parameter is smaller than lifetime of any references in the associated parameter value. This is the rationale for having *ref-safe-context* for `ref` to `ref struct` be *return-only* and `out` be *caller-context*. That prevents cyclic assignment because of the difference in lifetimes.
 
-It is also why `[UnscopedRef]` only promotes the *ref-safe-context* of any `ref` to `ref struct` values to *return-only* and not *caller-context*. Consider that using *caller-context* allows for cyclic assignment and would force a viral use of `[UnscopedRef]` for a `ref struct`:
```
jjonescz, line 1813: > It is also why `[UnscopedRef]` only promotes the *ref-safe-context* of any `ref` to `ref struct` values to *return-only* and not *caller-context*.

This was incorrect, I believe. Replaced with an `[UnscopedRef]` self-assignment example.

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -1810,47 +1810,27 @@ void Usage()
...
+    S local = new();
+    // Error: self assignment possible inside `S.M`.
+    S.M(ref local);
+    return local;
 }
-```
-
-This is correctly illegal in that case because the compiler has to consider the pathological case that `S.Data` could cyclic assign via `this`. That forces methods all methods that call `S.Data` to further mark their `ref` parameters as `[UnscopedRef]`. This is viral until the method which creates the value as a local. This is why *return-only* exists as a *safe-context*. It does complicate the spec / implementation but it serves to make the feature significantly more usable.
 
-Note: this cyclic assignment problem does continue to exist for `[UnscopedRef] out` to `ref struct` because that causes the *safe-context* and *ref-safe-context* to be equivalent.
```
jjonescz, line 1833: > this cyclic assignment problem does continue to exist for `[UnscopedRef] out` to `ref struct` because that causes the *safe-context* and *ref-safe-context* to be equivalent. 

This example didn't make sense to me - it doesn't compile simply because the `out` parameter is unassigned. Removed.

proposals/csharp-11.0/low-level-struct-improvements.md
jaredpar: Pretty sure this was correct. The core problem once you allow the `ref` and the `ref struct` of a parameter to have the same context is it effectively mandates that the corresponding argument be scoped. It means there is really no way to call that method on a non-scoped value. 

This is particularly problematic when you apply this to `this` (which is another parameter) because it means every instance method call is potentially a cyclic assignment. 

```csharp

void Test()
{
  S local = new();

  // Error: local could now contain a ref to the stack
  local.M1();

  // Error: local could now contain a ref to the stack
  S.M2(ref local);
}

ref struct S { 
  [UnscopedRef]
  void M1() { }

  static void M2([UnscopedRef] ref S p) { }
}
```

jaredpar: Think we should change this section to have a list that describes what effect `[UnscopedRef]` has on all the different ref locations:

| ref Location | Original Context | New Context |
| --- | --- | --- |
| instance member | function-member | return-only |
| in / ref | return-only | caller-context |
| out | function-member | return-only |

Essentially a list to that effect.

proposals/csharp-11.0/low-level-struct-improvements.md
jaredpar: I would like us to keep at least the note about `out` because it's likely unintuitive at a glance. This would make it clear that it was deliberate.

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -1810,47 +1810,27 @@ void Usage()
...
-void M2([UnscopedRef] out RS p)
-{
-    // The lifetimes of LHS and RHS are equivalent here and hence this is legal
-    p.refField = ref p.Field;
+    static void M([UnscopedRef] ref S s)
+    {
+        // Allowed: s has both safe-context and ref-safe-context of caller-context
+        s.refField = ref s.field;
+    }
 }
```
jaredpar, line 1834: Think that we should have a sample where `[UnscopedRef]` to assign a `ref` field. Basically demonstrate that this behavior we chose provides value when the type is not a `ref struct`. Essentially: 

```csharp
ref struct RS
{
   ref int refField;

   public void Assign([UnscopedRef] int data)
   {
      refField = ref data;
   }
}
```

RikkiGibson (commented):
LGTM modulo Jared's comments.

proposals/csharp-11.0/low-level-struct-improvements.md
jjonescz: > change this section to have a list that describes what effect `[UnscopedRef]` has

This section is about the silly cyclic assignment, it feels like a better place for such table is in the `#rules-unscoped` section above.

jaredpar: That works for me. Just want it somewhere in the doc.

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -1810,47 +1814,56 @@ void Usage()
...
 ```
 
-This is correctly illegal in that case because the compiler has to consider the pathological case that `S.Data` could cyclic assign via `this`. That forces methods all methods that call `S.Data` to further mark their `ref` parameters as `[UnscopedRef]`. This is viral until the method which creates the value as a local. This is why *return-only* exists as a *safe-context*. It does complicate the spec / implementation but it serves to make the feature significantly more usable.
+Similarly `[UnscopedRef] out` allows a cyclic assignment because the parameter has both *safe-context* and *ref-safe-context* of *return-only*.
 
-Note: this cyclic assignment problem does continue to exist for `[UnscopedRef] out` to `ref struct` because that causes the *safe-context* and *ref-safe-context* to be equivalent. 
+Promoting `[UnscopedRef] ref` to *caller-context* is useful when the type is *not* a ref struct
```
jaredpar, line 1844:
```suggestion
Promoting `[UnscopedRef] ref` to *caller-context* is useful when the type is *not* a `ref struct`
```

jaredpar (approved):

RikkiGibson (approved):

PR #8577, opened 2024-11-11
Add 'extending extensions' proposal

DustinCampbell:


meetings/working-groups/extensions/extending-extensions-a-guide-to-relaxation.md
```diff
@@ -0,0 +1,316 @@
...
+    public static bool IsKind(this SyntaxTrivia trivia, SyntaxKind kind) => ...;
+    public static bool IsKind([NotNullWhen(true)] this SyntaxNode? node, SyntaxKind kind) => ...;
+    public static bool IsKind(this SyntaxNodeOrToken nodeOrToken, SyntaxKind kind) => ...;
+    public static bool ContainsDirective(this SyntaxNode node, SyntaxKind kind) => ...;
+}
+```
+
+If these were written using the base design, they would need to be declared across four extension declarations:
+
+```c#
```
HaloFour, line 169: If the braces are optional this doesn't feel too bad?

```cs
public static class CSharpExtensions
{
    extension(SyntaxToken) public bool IsKind(SyntaxKind kind) => ...;
    extension(SyntaxTrivia) public bool IsKind(SyntaxKind kind) => ...;

    extension(SyntaxNode?)
        [param: NotNullWhen(true)]
        public bool IsKind(SyntaxKind kind) => ...;

    extension(SyntaxNodeOrToken) public bool IsKind(SyntaxKind kind) => ...;
    extension(SyntaxNode) public bool ContainsDirective(SyntaxKind kind) => ...;
}
```

DustinCampbell: > If the braces are optional this doesn't feel too bad?
> 
> ```cs
> public static class CSharpExtensions
> {
>     extension(SyntaxToken) public bool IsKind(SyntaxKind kind) => ...;
>     extension(SyntaxTrivia) public bool IsKind(SyntaxKind kind) => ...;
>     extension([NotNullWhen(true)] SyntaxNode?) public bool IsKind(SyntaxKind kind) => ...;
>     extension(SyntaxNodeOrToken) public bool IsKind(SyntaxKind kind) => ...;
>     extension(SyntaxNode) public bool ContainsDirective(SyntaxKind kind) => ...;
> }
> ```

This is definitely a possibility! It's a bit verbose for my personal taste though, and I'm not sure how I feel about it being ahead of all modifiers and the return type. I aimed for a syntax that corresponds with what an extension member looks like when it's called in its reduced form. The way I was thinking about it was to take the `this` parameter from a classic extension method and shift it to the left. Essentially, it embraces the idea that the "underlying type" is really just a parameter.

image

DustinCampbell: > tangentially related: since this is literally a sort of modifier on members (either inline or as a block), does this practically exclude interfaces (in which case youd have use another type kind) or they're going to appear together with nested extensions?

I have no idea. :smile: I'm expecting that a lot of design will have happened and moved things around before we get to extension interfaces.

meetings/working-groups/extensions/extending-extensions-a-guide-to-relaxation.md
```diff
@@ -0,0 +1,316 @@
...
+While this leads to a clear design, that assertion will not be true for all. Parameter names provide important value
+to a member’s signature and the code within.
+
+- Parameter names are an important form of self-documentation when the type system isn’t expressive enough to clearly
+  state the author's intent. For example, it is helpful when a parameter of type `Person` has the parameter name
+  `teacher`. This is a clear indicator to the caller what sort of `Person` to pass and helps make the code within the
+  member more readable.
+- Parameter names can synergize semantically with other parameter names. For example, the parameter name `source` might
+  be linked semantically to another parameter named `destination`.
+- Parameter names can be used to disambiguate overloads by applying a named argument.
```
jnm2, line 53: Though, this would not apply when used *as* an extension member, and would not apply at all unless it was decided that these things could be called as ordinary members.

DustinCampbell: That's true! Here, I'm trying to list reasons why parameter names are generally important. Using a `this` parameter name as a named argument for disambiguation of classic extension method is definitely super-rare, but it is certainly possible.

On a side note, I think it's an open question to allow any new disambiguation form to work with classic extension methods or the classic static-method-call to work with new extension methods. However, I would advocate for making both cases work if only because consumers shouldn't have to know how the extension method was implemented. To the consumer, it should just be an extension method.

PR #8574, opened 2024-11-11
Add open question about preferring ROS over Span

jjonescz:

PR #8580, opened 2024-11-12
Note a possible ambiguity do to the new proposed betterness rule

jjonescz:
Follow up on https://github.com/dotnet/csharplang/pull/8574.

### DIFF
--- a/proposals/first-class-span-types.md
+++ b/proposals/first-class-span-types.md
@@ -582,6 +582,19 @@ The full change to the spec of "better conversion from expression" would be:
 >     and `T₁` is a better conversion target than `T₂`
 > - `E` is a method group, `T₁` is compatible with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
 
+Note that adding a betterness rule can lead to more ambiguities when two betterness rules have an effect in opposite directions on a given set of overloads:
+
+```cs
+var x = new int[0];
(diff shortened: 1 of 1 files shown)


proposals/first-class-span-types.md
```diff
@@ -582,6 +582,19 @@ The full change to the spec of "better conversion from expression" would be:
...
+var x = new int[0];
+C.M(x, x); // 2 previously, ambiguous now
+
+static class C
+{
+    public static void M(IEnumerable<int> a, ReadOnlySpan<int> b) => Console.Write(1);
+    public static void M(Span<int> a, Span<int> b) => Console.Write(2);
+}
+```
+
```
cston, line 597: Will the new rule also be a breaking change for the following?
```csharp
var x = new int[0];
C.M1(x); // Span<int> previously, ReadOnlySpan<int> now

static class C
{
    public static void M1(ReadOnlySpan<int> a) => Console.Write(1);
    public static void M1(Span<int> a) => Console.Write(2);
}
```

jjonescz: Yes, but that seems clear directly from the rule. Also it's one of the "ok" breaking changes (there are many like that as part of the feature where a different overload is chosen but we expect the overloads to be semantically equivalent) as opposed to a "bad" breaking change (e.g., build/runtime failing where it was previously succeeding).

cston (approved):

PR #8695, opened 2024-11-28
Add syntax for extensions

jcouv:


proposals/extensions.md
```diff
@@ -1,30 +1,58 @@
...
+    ;
+
+receiver_parameter // add
+    : attributes? parameter_modifiers? type identifier?
+    ;
+```
 
-## Declaration
+Extension containers shall only be declared in non-generic, non-nested static classes.  
+It is an error for a type to be named `extension`.
```
jjonescz, line 50: > It is an error for a type to be named `extension`.

Is it because it would be ambiguous with an instance constructor declaration like this?

```cs
class extension
{
   extension() { }
}
```

But only static classes can contain `extension`s, and static classes cannot have instance constructors.

jcouv: This is not because of possibly ambiguity. It will simplify parsing and improve error recovery. We've been moving towards reserving names for new contextual keywords we introduce (`scoped`, `record`, `file`, `required`). 
We've made all-lowercase-alpha names a warning wave warning as a preparation step. Now (with `field`) we've introduced a process for introducing such breaks.

jaredpar: Can we note that this error is tied to language version?

PR #8766, opened 2024-12-04
Consider silly cyclic assignment in scoped variance

jjonescz:
This brings back the scope variance exceptions removed from the speclet in https://github.com/dotnet/csharplang/pull/8777 and updates them.
Corresponding implementation bugfix: https://github.com/dotnet/roslyn/pull/76261


proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -469,8 +469,11 @@ The `scoped` modifier and `[UnscopedRef]` attribute (see [below](#rules-unscoped
 Any other difference with respect to `scoped` or `[UnscopedRef]` is considered a mismatch.
 
 The compiler will report a diagnostic for _unsafe scoped mismatches_ across overrides, interface implementations, and delegate conversions when:
-- The method returns a `ref struct` or returns a `ref` or `ref readonly`, or the method has a `ref` or `out` parameter of `ref struct` type, and
-- The method has at least one additional `ref`, `in`, or `out` parameter, or a parameter of `ref struct` type.
+- The method has a `ref` or `out` parameter of `ref struct` type with a mismatch of adding `[UnscopedRef]` (not removing `scoped`).
+  (In this case, a [silly cyclic assignment](#cyclic-assignment) is possible, hence no other parameters are necessary.)
```
jaredpar, line 473: I don't think it's so much that it allows silly cyclic assignment as it's just a basic type system violation. Consider that it allows you to do things like:

```csharp
interface I1 {
  ref int M(); 
}
struct S {
  int field;
  [UnscopedRef] ref int M() => ref field;
}
```

This is a similar case where introducing `[UnscopedRef]` in the implementing type leads to lifetime violation.

jjonescz: Yes, but that example is already covered by the existing rules and [is an error today](https://sharplab.io/#v2:EYLgtghglgdgPgAQAwAIEEYB0ARKEDmMA9gM4AuUAxiZgMJEAmApgIIwQA2AniVCQNwBYAFCwyTAE4AzCJSYoAkuhQBvEShQSmUlGJQBZABQBKfihEBfEeQkBXSmRQBlFCEXK1wjXqlQmHBiEvFABtAFUYEkoiAAcmBgAlbQBdNABmTW1dGEcjYxQAXgA+TJ1ff0DLIA). Those rules are

>- Or both of these are true:
>   - The method returns a `ref struct` or returns a `ref` or `ref readonly`, or the method has a `ref` or `out` parameter of `ref struct` type.
>   - The method has at least one additional `ref`, `in`, or `out` parameter, or a parameter of `ref struct` type.

This new bullet point

> - The method has a `ref` or `out` parameter of `ref struct` type with a mismatch of adding `[UnscopedRef]` (not removing `scoped`).

is only necessary because of the silly cyclic assignment.

jjonescz: Discussed offline and it's not clear why the paragraph starting with

> The compiler will report a diagnostic for _unsafe scoped mismatches_ across overrides, interface implementations, and delegate conversions when:

is in the speclet in the first place.

In other words, it seems that adding `[UnscopedRef]` or removing `scoped` should always be a mismatch error without exceptions. cc @cston

We should also investigate this sentence in the light of ref structs being allowed to implement interfaces:

> The rules above ignore `this` parameters because `ref struct` instance methods cannot be used for overrides, interface implementations, or delegate conversions.

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -478,7 +478,19 @@ The `scoped` modifier and `[UnscopedRef]` attribute (see [below](#rules-unscoped
...
+- Or both of these are true:
+  - The method returns a `ref struct` or returns a `ref` or `ref readonly`, or the method has a `ref` or `out` parameter of `ref struct` type.
+  - The method has at least one additional `ref`, `in`, or `out` parameter, or a parameter of `ref struct` type.
+
+The diagnostic is not reported in other cases because
+- the methods with such signatures cannot capture the refs passed in, so any scoped mismatch is not dangerous,
+- these include very common and simple scenarios (e.g., plain old `out` parameters which are used in `TryParse` method signatures)
+  and reporting scoped mismatches just because they are used across LangVersion 11 (and hence the `out` parameter is differently scoped) would be confusing.
```
jaredpar, line 491:
```suggestion
The diagnostic is not reported in other cases because:
- The methods with such signatures cannot capture the refs passed in, so any scoped mismatch is not dangerous,
- These include very common and simple scenarios (e.g., plain old `out` parameters which are used in `TryParse` method signatures)
  and reporting scoped mismatches just because they are used across LangVersion 11 (and hence the `out` parameter is differently scoped) would be confusing.
```

Nit: formatting consistency

jaredpar (approved):

proposals/csharp-11.0/low-level-struct-improvements.md
```diff
@@ -478,7 +478,19 @@ The `scoped` modifier and `[UnscopedRef]` attribute (see [below](#rules-unscoped
...
-The mismatch is reported as an _error_ if the mismatched signatures are both using C#11 ref safe context rules; otherwise, the diagnostic is a _warning_.
+The compiler will report a diagnostic for _unsafe scoped mismatches_ across overrides, interface implementations, and delegate conversions when:
+- The method has a `ref` or `out` parameter of `ref struct` type with a mismatch of adding `[UnscopedRef]` (not removing `scoped`).
+  (In this case, a [silly cyclic assignment](#cyclic-assignment) is possible, hence no other parameters are necessary.)
+- Or both of these are true:
+  - The method returns a `ref struct` or returns a `ref` or `ref readonly`, or the method has a `ref` or `out` parameter of `ref struct` type.
+  - The method has at least one additional `ref`, `in`, or `out` parameter, or a parameter of `ref struct` type.
+
+The diagnostic is not reported in other cases because:
+- The methods with such signatures cannot capture the refs passed in, so any scoped mismatch is not dangerous,
```
jjonescz, line 489:
```suggestion
- The methods with such signatures cannot capture the refs passed in, so any scoped mismatch is not dangerous.
```

cston (approved):

Issue #8098, opened 2024-05-06
[Proposal]: Edit speclets to distinguish features already delivered from potential future enhancements


BillWagner:
We've been publishing the [feature speclets](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-10.0/enhanced-line-directives) on Learn so that customers can easily access the detailed information about new features.

There have been a few comments from readers that are confused because the speclets include specification language for features that haven't been implemented yet.  For example:

- [Low level struct improvements](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements) The section on `scoped` has been implemented. Much of the remaining sections could be implemented later
- [Covariant return types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns#implicit-interface-implementations) has a section on explicit interface implementation that hasn't been implemented.

There are likely others.

One fix would be to make the following edits:

1. Separate out sections that haven't been implemented into new proposals in the [Proposals](https://github.com/dotnet/csharplang/tree/main/proposals). 
2. Link to the new "proposed additions" speclet from the existing specs to provide a roadmap for future work. That link would be an absolute GitHub link, so that readers on Learn would be redirected to GitHub. That would be a signal that it's a future enhancement.

That would benefit Learn readers, and provide a clear distinction for designers and the compiler team on what behavior has been implemented, and what behavior is still in a proposed state.

LDM should weigh in on this before editing the docs.

333fred:
Another part to consider is the sections of the specs that call out invalid scenarios: I've had several questions on some of the DIMs examples of what _doesn't_ work from people just copy/pasting examples.

KathleenDollard:
@333fred We might want a docs feature on this. I'd love to see some different visual appearance when it is an example of what not to do, either because it won't work or it's just a terrible idea.

That is probably a docs infrastructure issue. @BillWagner should we open one?

jcouv:
Found another one: the inline-arrays spec suggests that inline array types can be initialized from a collection expression, but that's not supported.
https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/inline-arrays.md#collection-literals

FYI @AlekseyTs @cston

jnm2:
From the perspective of someone who has consumed the C# spec proposals as a library and tooling author, this spec documentation is well worth solving and delivering along with each release of C#.

[...]

PR #8769, opened 2024-12-04
Edit pass.

BillWagner:
Make sure code is correct, as commented for C# 13. If an example is invalid, add a clear comment why.

I found one or two other editorial nits as well.

Starting with C# 13, and I'll open other PRs for older versions.

Contributes to #8098

### DIFF
--- a/proposals/csharp-13.0/overload-resolution-priority.md
+++ b/proposals/csharp-13.0/overload-resolution-priority.md
@@ -121,7 +121,7 @@ It is an error to apply `OverloadResolutionPriorityAttribute` to the following l
 * Conversion operators
 * Lambdas
 * Local functions
-* Destructors
+* Finalizers
 * Static constructors
 
 Attributes encountered on these locations in metadata are ignored by C#.
--- a/proposals/csharp-13.0/ref-struct-interfaces.md
+++ b/proposals/csharp-13.0/ref-struct-interfaces.md
@@ -48,24 +48,25 @@ interface I1
 struct S1
 {
     [UnscopedRef]
-    ref int P1 { get; }
-    ref int P2 { get; }
+    internal ref int P1 { get {...} }
+
+    internal ref int P2 { get {...} }
 }
 
-int M<T>(T t, S1 s)
-    where T : allows ref struct, I1
+ref int M<T>(T t, S1 s)
+    where T : I1, allows ref struct
 {
     // Error: may return ref to t
-    return t.P1;
+    return ref t.P1;
 
     // Error: may return ref to t
-    return s.P1;
+    return ref s.P1;
 
     // Okay
-    return t.P2;
+    return ref t.P2;
 
     // Okay
-    return s.p2;
+    return ref s.P2;
 }
 ```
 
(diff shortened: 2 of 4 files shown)


proposals/csharp-13.0/overload-resolution-priority.md
```diff
@@ -121,7 +121,7 @@ It is an error to apply `OverloadResolutionPriorityAttribute` to the following l
 * Conversion operators
 * Lambdas
 * Local functions
-* Destructors
+* Finalizers
```
BillWagner, line 124: I made this change to conform to updated spec language. "Destructor" was replaced everywhere with "finalizer".

proposals/csharp-13.0/ref-struct-interfaces.md
```diff
@@ -48,24 +48,25 @@ interface I1
 struct S1
 {
     [UnscopedRef]
-    ref int P1 { get; }
-    ref int P2 { get; }
+    internal ref int P1 { get {...} }
+
+    internal ref int P2 { get {...} }
```
BillWagner, line 53: Added boilerplate because `ref` auto properties aren't legal.

proposals/csharp-13.0/ref-struct-interfaces.md
```diff
@@ -115,13 +116,14 @@ interface I1
 {
     void M()
     {
-        // Error: both of these box if I1 is implemented by a ref struct
+        // Danger: both of these box if I1 is implemented by a ref struct
```
BillWagner, line 119: This doesn't produce a compiler error, the only error is on line 125.

proposals/csharp-13.0/ref-struct-interfaces.md
```diff
@@ -233,33 +235,25 @@ Examples of these rules in action:
...
-    return default;
-    return p;
-}
-
-T M3<T>(Span<T> span)
-    where T : allows ref struct
-{
-    return default;
+    return ref p;
 }
```
BillWagner, line 257: This is the sample that should get the most thorough review. Much of it didn't compile as written because `T` can't be allowed to be a `ref struct` if it will be used as a `Span<T>`. That meant much of `M2` needed edits.

333fred (approved):

proposals/csharp-13.0/ref-unsafe-in-iterators-async.md
```diff
@@ -260,7 +260,7 @@ class C
         await Task.Yield();
         using (new R()) { } // allowed under this proposal
         foreach (var x in new C()) { } // allowed under this proposal
-        foreach (ref int x in new int[0]) { } // allowed under this proposal
+        foreach (ref int x in new int[0]) { } // allowed under this proposal, but not implemented in C# 13
```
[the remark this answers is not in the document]
jjonescz: This example could be changed to

```cs
using System.Threading.Tasks;

ref struct R
{
    public ref int Current => throw null;
    public bool MoveNext() => false;
    public void Dispose() { }
}
class C
{
    public R GetEnumerator() => new R();
    async void M()
    {
        await Task.Yield();
        using (new R()) { } // allowed under this proposal
        foreach (var x in new C()) { } // allowed under this proposal
        foreach (ref int x in new C()) { } // allowed under this proposal
        lock (new System.Threading.Lock()) { } // allowed under this proposal
        await Task.Yield();
    }
}
```

PR #8880, opened 2024-12-13
Add section on metadata for extensions

jcouv:

### DIFF
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -331,11 +331,13 @@ is so that IL verification could run and pass (thus validating the completeness 
 The extension marker method encodes the receiver parameter.  
 - It is public and static, and is called `<Extension>$`.  
 - If the receiver parameter doesn't specify a name, then the parameter name is unspeakable.  
+
 Note: This allows roundtripping of extension container symbols through metadata (full and reference assemblies).  
 
 The method bodies for method/property/indexer declarations in an extension container in source are emitted 
 as static implementation methods in the top-level static class.  
-- The name of an implementation method prepends `<Extension>` to the name of the original method. For example: `set_Property` => `<Extension>set_Property`.
+- The name of an implementation method prepends `<Extension>` to the name of the original method.  
+  For example: `set_Property` => `<Extension>set_Property`.
 - An implementation method has type parameters derived from the extension container prepended to the type parameters of the original method.  
 - An implementation method has the same accessibility as the original method.  
 - The implementation method corresponding to a static method has the same parameters and return type as the original method. 
@@ -406,7 +408,7 @@ static class IEnumerableExtensions
 }
 ```
 
-Whenever extension members are used in source, we will emit those as reference to implementation methods.
+Whenever extension members are used in source, we will emit those as reference to implementation methods.  
 For example: an invocation of `enumerableOfInt.Method()` would be emitted as a static call 
 to `IEnumerableExtensions.Method<int>(enumerableOfInt)`.  
 
@@ -417,6 +419,8 @@ to `IEnumerableExtensions.Method<int>(enumerableOfInt)`.
 - How do we associate signatures and implementations? Attribute+number? Naming convention? Using the prepended parameter name?
 - Should we flag some methods with special name?
 - Where should attributes on the receiver parameter go?
+- Should we emit implementation methods with speakable names instead, as a disambiguation strategy and also to allow
+  usage from other languages?
 
 ### Add support for more member kinds
 
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -315,7 +315,10 @@ These requirements need more refinement as implementation progresses, and may ne
 
 ### Metadata for declarations
 
-Each extension container is emitted as a nested public static class with a marker method.  
+Each extension container is emitted as a nested private static class with a marker method and skeleton members.  
+Each skeleton member is accompanied by a top-level static implementation method with a modified signature.  
+
+#### Skeletons
 
 Its name is unspeakable and determined based on the lexical order in the program. 
 The name is not guaranteed to remain stable across re-compilation. 
@@ -324,6 +327,10 @@ Below we use `<Extension>ee_` followed by an index. For example: `<Extension>ee_
 Method/property/indexer declarations in an extension container in source are represented as skeleton declarations in metadata.  
 The signatures of the original methods are maintained, but their bodies are replaced with `throw null`.  
 Those should not be referenced in IL.  
+
+The nested types declarations in an extension container in source are represented as skeleton nested types in metadata.  
+The name, accessibility and type parameters of the original nested type are maintained, but the members are emitted as
+skeleton declarations in metadata.  
 
 Note: This is similar to ref assemblies. The reason for using `throw null` bodies (as opposed to no bodies) 
 is so that IL verification could run and pass (thus validating the completeness of the metadata).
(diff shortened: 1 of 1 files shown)


proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
 
 These requirements need more refinement as implementation progresses, and may need to be compromised in corner cases in order to allow for a reasonable implementation approach.
 
+### Metadata for declarations
+
+Each extension container is emitted as a nested public static class with a marker method.
```
AlekseyTs, line 318: Does it have to be public? Is IL going to reference this type?

AlekseyTs: Especially given the fact that the names aren't stable

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+
+    // Implementation for Method
+    public static void <Extension>Method<T>(IEnumerable<T> source) { ... }
+
+    // Implementation for Property
+    internal static int <Extension>get_Property<T>(IEnumerable<T> source) { ... }
+    internal static void <Extension>set_Property<T>(IEnumerable<T> source, int value) { ... }
+
+    // Implementation for NestedType
+    public class NestedType<V>
```
AlekseyTs, line 401: Is the name of the type going to be mangled?

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+    {
+        public (T, U) NestedMethod<T, U>(IEnumerable<T> source, V v) { ... }
```
AlekseyTs, line 403: It looks like `T` references an undefined type parameter.

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+{
+    public static class <Extension>ee_1<T>
+    {
+        public static <Extension>$(IEnumerable<T> source) => throw null;
+        public void Method() => throw null;
+        public static int Property { get => throw null; set => throw null; }
+
+        public class NestedType<V>
+        {
+            public (T, U) NestedMethod<U>(V v) => throw null;
```
AlekseyTs, line 383: If we are going to support non-private nested types, I think we can drop all members from them in extension container.

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+```
+is emitted as
+```
+static class IEnumerableExtensions
+{
+    public static class <Extension>ee_1<T>
+    {
+        public static <Extension>$(IEnumerable<T> source) => throw null;
+        public void Method() => throw null;
+        public static int Property { get => throw null; set => throw null; }
```
jjonescz, line 379: Do we need these stub definitions? It feels like the marker methods specifying the extended types would be enough (they wouldn't even need to be in nested types, they could contain the container type parameters themselves). We could reference the marker method (e.g., by name) from each implementation method by some attribute / modreq.

We could even skip that for instance extension members, they already specify the receiver (we would just use a different unspeakable name prefix to recognize they are instance extension members). That would need some other way of specifying where the container type parameters end though (like some attribute or even encoding that in the name). And we would have to emit the receiver attributes to each implementation method, but that's what I would expect anyway.

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+
+        public class NestedType<V>
```
AlekseyTs, line 381: Is this type supposed to be an extension member? Is there a real scenario for an extension type?

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+        {
+            public (T, U) NestedMethod<U>(V v) => throw null;
+        }
+    }
+
+    public static class <Extension>ee_2
+    {
+        public static <Extension>$(IAsyncEnumerable<int> values) => throw null;
+        public Task<int> SumAsync() => throw null;
```
AlekseyTs, line 390: It looks like there is no method with this name in the original source

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+    internal static void <Extension>set_Property<T>(IEnumerable<T> source, int value) { ... }
+
+    // Implementation for NestedType
+    public class NestedType<V>
+    {
+        public (T, U) NestedMethod<T, U>(IEnumerable<T> source, V v) { ... }
+    }
+
+    // Implementation for SumAsync
+    public static int SumAsync(IAsyncEnumerable<int> values) { ... }
```
AlekseyTs, line 407: Is the name going to be mangled?

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
 
 These requirements need more refinement as implementation progresses, and may need to be compromised in corner cases in order to allow for a reasonable implementation approach.
```
jjonescz, line 314: One of the bullet points above says the generated artifacts should be hidden via modreqs but the new specification below doesn't seem to do that.

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+
+Whenever extension members are used in source, we will emit those as reference to implementation methods.  
+For example: an invocation of `enumerableOfInt.Method()` would be emitted as a static call 
+to `IEnumerableExtensions.Method<int>(enumerableOfInt)`.  
+
 ## Open issues
 
+### Metadata
+
+- How do we associate signatures and implementations? Attribute+number? Naming convention? Using the prepended parameter name?
```
jjonescz, line 419: Name and signature need to be unique for instance extension members so that should be enough to match them, right? I guess that's not true for static extension members.

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
...
+Note: This allows roundtripping of extension container symbols through metadata (full and reference assemblies).  
+
+The method bodies for method/property/indexer declarations in an extension container in source are emitted 
+as static implementation methods in the top-level static class.  
+- The name of an implementation method prepends `<Extension>` to the name of the original method.  
+  For example: `set_Property` => `<Extension>set_Property`.
+- An implementation method has type parameters derived from the extension container prepended to the type parameters of the original method.  
+- An implementation method has the same accessibility as the original method.  
+- The implementation method corresponding to a static method has the same parameters and return type as the original method. 
+- The implementation method corresponding to an instance method adds a prepended parameter to the signature of the original method.
```
AlekseyTs, line 344: How do we avoid a possible name and signature collision with static extension method after this parameter is added to the signature?

AlekseyTs:

proposals/extensions.md
```diff
@@ -313,8 +313,115 @@ However, beyond implementing the language semantics it must satisfy certain requ
 
 These requirements need more refinement as implementation progresses, and may need to be compromised in corner cases in order to allow for a reasonable implementation approach.
 
+### Metadata for declarations
+
+Each extension container is emitted as a nested public static class with a marker method.  
+
+Its name is unspeakable and determined based on the lexical order in the program. 
+The name is not guaranteed to remain stable across re-compilation.
```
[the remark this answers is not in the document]
jcouv: Yes. It is determined by the lexical order, thus deterministic if no compilation inputs change.

proposals/extensions.md
jcouv: The name was munged to avoid collisions. If we choose to keep speakable names instead, we'll likely have to stick a modopt or a rule to avoid collisions.

proposals/extensions.md
jcouv: I didn't follow. `T` is defined on the method.
Alternatively, we could move `T` to the implementation nested type and then we probably don't need skeleton members for the nested type. We can discuss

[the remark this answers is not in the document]
jcouv: I'd missed `<Extension>` on the implementation nested type. Added

proposals/extensions.md
jcouv: I'm not sure either. We may not need any additional mechanism to associate skeletons and implementations. Given a skeleton, we can always find the corresponding implementation by repeating the transformation process. But that may be heavy-handed

proposals/extensions.md
jcouv: Sorry, I'd misunderstood. The problem was with `static void M(IEnumerable<T>)` and `void M()`. From offline discussion, this is now solved by using a different prefix for static methods (`<StaticExtension>`).

proposals/extensions.md
```diff
@@ -38,7 +38,6 @@ extension_member_declaration // add
     | property_declaration
     | indexer_declaration
     | operator_declaration
-    | type_declaration
```
jcouv, line 41: :memo: removed extension types for now. We have an encoding solution available so no need to front-load extra design that doesn't seem to have unique challenges.

proposals/extensions.md
jcouv: Good question. Not sure where all we should put modreqs. I'll add an open question

proposals/extensions.md
jcouv: You're right, there is a lot of redundant information between the skeleton members and the implementation members.
However, if we only emitted skeletons for the extension container type, with an extension marker, but none of the skeleton members, it would be harder to reconstitute the right symbols from metadata. 
For example, how would we reconstitute a property and where would we get the attribute for that property from?

PR #8964, opened 2025-01-06
Create extension-member-disambiguation.md

MadsTorgersen:
Draft discussion document for adding an extension member disambiguation syntax.


bartdesmet:
Random thought: extension methods today are brought in with `using <namespace>;` so maybe that keyword could be used as an infix for `(e using MyExtension).Prop`.

jcouv:
I'd like to add one more candidate to the hat: static method invocation syntax (same as for classic extension methods) With a simple naming convention (get_...,  set_..., op_..., ...).

PR #9028, opened 2025-01-08
Add proposal for `fieldof()`

RikkiGibson:
See also discussion #8704.


HaloFour:
`fieldof` feels like it has something to do with reflection to me.

Also, In Foof We Trust.

PR #9038, opened 2025-01-09
Field-backed properties: language version diagnostics

cston:
Document the breaking change diagnostics that may be reported when re-compiling with the language version that includes field-backed properties.

### DIFF
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -102,7 +102,7 @@ Since `field` is a keyword and not an identifier, it can only be "shadowed" by a
 
 If a variable named `field` is declared in a property accessor, an error is reported.
 
-If an expression `field` is used as a *primary expression* and would bind to an existing symbol with language version 13 or earlier, a warning is reported.
+In language version 14 or later, a warning is reported if a *primary expression* `field` refers to the backing field, but would have referred to a different symbol in an earlier language version.
 
 ### Field-targeted attributes
 
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -102,7 +102,7 @@ Since `field` is a keyword and not an identifier, it can only be "shadowed" by a
 
 If a variable named `field` is declared in a property accessor, an error is reported.
 
-If an expression `field` would bind to a symbol with language version 13 or earlier, a warning may be reported.
+If an expression `field` would bind to an existing symbol with language version 13 or earlier, a warning is reported.
 
 ### Field-targeted attributes
 
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -102,7 +102,7 @@ Since `field` is a keyword and not an identifier, it can only be "shadowed" by a
 
 If a variable named `field` is declared in a property accessor, an error is reported.
 
-If an expression `field` would bind to an existing symbol with language version 13 or earlier, a warning is reported.
+If an expression `field` is used as a *primary expression* and would bind to an existing symbol with language version 13 or earlier, a warning is reported.
 
 ### Field-targeted attributes
 
--- a/proposals/field-keyword.md
+++ b/proposals/field-keyword.md
@@ -96,9 +96,13 @@ Set-only properties may also use `field`:
 
 ### Breaking changes
 
-The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change, proposed as part of a larger [Breaking Changes](https://github.com/dotnet/csharplang/issues/7964) feature.
+The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change.
 
-Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 13 by adding the initial `@`.
+Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 14 by adding the initial `@`.
+
+If a variable named `field` is declared in a property accessor, an error is reported.
+
+If an expression `field` would bind to a symbol with language version 13 or earlier, a warning may be reported.
 
 ### Field-targeted attributes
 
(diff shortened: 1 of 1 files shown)


proposals/field-keyword.md
```diff
@@ -96,9 +96,13 @@ Set-only properties may also use `field`:
...
 
-The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change, proposed as part of a larger [Breaking Changes](https://github.com/dotnet/csharplang/issues/7964) feature.
+The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change.
 
-Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 13 by adding the initial `@`.
+Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 14 by adding the initial `@`.
+
+If a variable named `field` is declared in a property accessor, an error is reported.
+
+If an expression `field` would bind to a symbol with language version 13 or earlier, a warning may be reported.
```
KathleenDollard, line 105: Why "may" here? Can we make it "will" and better describe the situation if necessary.

Would it be more clear to say something like

```suggestion
If a symbol named `field` is accessible in the body of a setter or getter that uses the `field` keyword, a warning will be reported. "Accessible" here means that a symbol is available that would be used for binding, if it did not conflict with the `field` keyword. 
```

cston: Thanks. Changed to "is" rather than "may be".

In terms of accessibility, we should state that it's accessible *at the reference to* `field`. For instance, the following is valid, even though the `field` local is accessible within (parts of) the accessor body.
```csharp
object Property
{
    get
    {
        if (field == null)
        {
            int @field = 0;
            return @field;
        }
        return field;
    }
}
```

proposals/field-keyword.md
```diff
@@ -96,9 +96,13 @@ Set-only properties may also use `field`:
...
 
-The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change, proposed as part of a larger [Breaking Changes](https://github.com/dotnet/csharplang/issues/7964) feature.
+The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change.
 
-Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 13 by adding the initial `@`.
+Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 14 by adding the initial `@`.
+
+If a variable named `field` is declared in a property accessor, an error is reported.
+
+If an expression `field` would bind to an existing symbol with language version 13 or earlier, a warning is reported.
```
jaredpar, line 105: Can we clarify that this must be a _primary expression_? Essentially, `field` will warn but `this.field` willl not.

jaredpar (approved):

jnm2 (commented):
My hesitation with "warnings after upgrade" instead of "warnings before upgrade" is the small chance of this scenario:

- There's something called `field` in scope. Let's say a protected field in a base class you're implementing from a library that you can't edit.

- You write a _new_ property using `field` as a keyword. This code is good code.

- The brand-new, correct code gets hit with a warning that `field` would have bound to the inherited field named `field` in C# 13, despite this new property never having compiled with C# 13.

There are variations on this scenario as well. Switching to langversion 14, and bulk applying "use `field`," could land you with this same warning despite the new occurrences of `field` being correct, and never having existed under C# 13.

So what does a user do?

- Rename the class field, but this might be a bitter pill if `field` was a genuinely good name for the class field. This might be impossible too, if inheriting from a third-party class.

- Or, put a `#pragma warning disable` on each prop. That's a lot of noise especially for a view model class.

- Or, put a `#pragma warning disable` once at the top of the file because of how noisy the warning is: it's basically all false positives. **This is weird**, because now in the _one_ place where confusion is likely (where there's a class field called `field`), the user has disabled the warning.

  - Or, disable the warning project-wide. 

- Or, quit using the `field` keyword feature in scope of the class field.

In short: Anytime this warning matters (in scenarios where there's a true conflict that breaks 13->14), I expect the occurrences of our breaking changes warning to be absolutely dominated by false positives, and for it to be tuned out one way or another. At the same time I agree we need something.

Regardless of the above, making it an error to declare something called `field` inside a property accessor seems like a solid move.

Does this apply to lambda parameters and local function parameters named `field`, or only locals?

jaredpar:
The cases that would require a permanent warning are essentially when you both: 

1. Have an accessible API named `field` which the developer cannot or will not rename 
2. Have a property where you desire the _new_ behavior of the `field` keyword

Our belief is the intersection of these two items will be low. It also falls in line with our stance that this is a breaking change and potentially requires changes to adopt. Yes there are a few cases that can be outlined where you _cannot_ rename the member but those feel sufficiently small. The vast majority of use cases we've seen don't have this intersection of problems.

Warnings on upgrade is a problematic area that carry a lot of risk for the product. In this case we feel like the scenarios where customers would be required to use a permanent `#pragma warning disable` are small enough that we'd rather take the certainty around warn on upgrade vs. the risk of warn on upgrade.

cston:
> Regardless of the above, making it an error to declare something called `field` inside a property accessor seems like a solid move.
> 
> Does this apply to lambda parameters and local function parameters named `field`, or only locals?

The error applies to parameters named `field` from lambdas and local functions nested in the accessor, in addition to locals declared in the accessor.

proposals/field-keyword.md
```diff
@@ -96,9 +96,13 @@ Set-only properties may also use `field`:
...
+The existence of the `field` contextual keyword within property accessor bodies is a potentially breaking change.
 
-Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 13 by adding the initial `@`.
+Since `field` is a keyword and not an identifier, it can only be "shadowed" by an identifier using the normal keyword-escaping route: `@field`. All identifiers named `field` declared within property accessor bodies can safeguard against breaks when upgrading from C# versions prior to 14 by adding the initial `@`.
+
+If a variable named `field` is declared in a property accessor, an error is reported.
+
+If an expression `field` is used as a *primary expression* and would bind to an existing symbol with language version 13 or earlier, a warning is reported.
```
RikkiGibson, line 105:
```suggestion
In language version 14 or later, a warning is reported if a *primary expression* `field` refers to the backing field, but would have referred to a different symbol in an earlier language version.
```

I was confused by the language here at first so made an attempt to revise. Up to you whether to take it, but I managed to un-confuse myself.

RikkiGibson (approved):

KathleenDollard:
@jnm2 

Is the wording of the warning, the fact that it refers to compiling in C# 13, which may be nonsensical to the user (if they are on LTS, C# 13 is irrelevant to them, for example) or do you think we are underestimating how often people will encounter the warning?

I am concerned about the wording, but have not yet come up with an alternative.

jnm2:
@KathleenDollard I was only worried about whether there was a purpose to the warning. It seemed that anyone who meets it will very likely need to turn it off. But Jared's rationale explains to me why that's a desirable tradeoff. I took him to be saying that it's not too annoying because it will happen sufficiently rarely. And in those rare cases, it's not purposeless even though the user will have to turn off the warning; it's important to have _something_ to alert people that they need to look closely at the situation.

@jaredpar Did I get what you were saying?

PR #9042, opened 2025-01-10
Edit pass for C# 8 speclets

### LINKED ISSUE #8098: [Proposal]: Edit speclets to distinguish features already delivered from potential future enhancements

BillWagner:
Closes #8098

Notes on this PR for reviewers:

1. I only edited the speclets that haven't been incorporated into the current draft V8 specification. (Thankfully, that means I didn't edit the NRT spec in this folder).
1. I did try to mark areas in the default interface members specification that weren't implemented in C# 8. I intended to mark all those as "not implemented in **C# 8**" because some involved early questions on static virtual (or abstract) methods or operators. I think it got most of these correct. However, that's the one area that could use the most careful review.

In particular, none of the examples that specify *which* base interface method to invoke from an override compiled. I believe that wasn't implemented, even though it was specified with decisions to implement a given syntax.

### DIFF
--- a/proposals/csharp-8.0/default-interface-methods.md
+++ b/proposals/csharp-8.0/default-interface-methods.md
@@ -207,7 +207,7 @@ Because interfaces may now contain executable code, it is useful to abstract com
 
 > ***Open issue***: should we permit interface methods to be `protected` or `internal` or other access? If so, what are the semantics? Are they `virtual` by default? If so, is there a way to make them non-virtual?
 
-> ***Open issue***: If we support static methods, should we support (static) operators? **Decision: NO**
+> ***Closed issue***: If we support static methods, should we support (static) operators? **Decision: YES**
 
 ### Base interface invocations
 
@@ -748,17 +748,11 @@ but `C` is not recompiled. What happens when the program is run? An invocation o
 
 ### Permit `partial` in interface? (closed)
 
-This was not implemented in C# 8.
-
-<details>
-
 Given that interfaces may be used in ways analogous to the way abstract classes are used, it may be useful to declare them `partial`. This would be particularly useful in the face of generators.
 
 > ***Proposal:*** Remove the language restriction that interfaces and members of interfaces may not be declared `partial`.
 
 ***Decision***: Yes. See <https://github.com/dotnet/csharplang/blob/master/meetings/2018/LDM-2018-10-17.md#permit-partial-in-interface>.
-
-</details>
 
 ### `Main` in an interface? (closed)
 
@@ -788,7 +782,7 @@ There are a few ways to observe whether an override declaration introduces a new
 ```csharp
 interface IA
 {
-void M(int x) { }
+    void M(int x) { }
 }
 interface IB : IA
 {
--- a/proposals/csharp-8.0/default-interface-methods.md
+++ b/proposals/csharp-8.0/default-interface-methods.md
@@ -171,7 +171,7 @@ abstract class D : IA, IB, IC // ok
 {
     public abstract void M();
 }
-public class E : ID { }// Error. No most specific implementation for 'IA.M'
+public class E : ID { } // Error. No most specific implementation for 'IA.M'
 ```
 
 The most specific implementation rule ensures that a conflict (i.e. an ambiguity arising from diamond inheritance) is resolved explicitly by the programmer at the point where the conflict arises.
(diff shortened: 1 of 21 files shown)


proposals/csharp-8.0/default-interface-methods.md
333fred, line 1: @AlekseyTs would you be able to give the edits here a review?

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -169,13 +165,13 @@ interface IC : IA
...
 }
-interface ID : IB, IC { } // error: no most specific implementation for 'IA.M'
+interface ID : IB, IC { } // compiles, but error when a class implements 'ID'
 abstract class C : IB, IC { } // error: no most specific implementation for 'IA.M'
 abstract class D : IA, IB, IC // ok
 {
     public abstract void M();
 }
-
+public class E : ID { }// Error. No most specific implementation for 'IA.M'
```
333fred, line 174:
```suggestion
public class E : ID { } // Error. No most specific implementation for 'IA.M'
```

333fred (commented):
Generally LGTM, but I'd like someone with better understanding of DIM to look at that one.

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -211,10 +207,14 @@ Because interfaces may now contain executable code, it is useful to abstract com
 
 > ***Open issue***: should we permit interface methods to be `protected` or `internal` or other access? If so, what are the semantics? Are they `virtual` by default? If so, is there a way to make them non-virtual?
 
-> ***Open issue***: If we support static methods, should we support (static) operators?
+> ***Open issue***: If we support static methods, should we support (static) operators? **Decision: NO**
```
AlekseyTs, line 210: This feels inaccurate. I think we supported operators too.

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -741,12 +748,18 @@ but `C` is not recompiled. What happens when the program is run? An invocation o
 
 ### Permit `partial` in interface? (closed)
 
+This was not implemented in C# 8.
```
AlekseyTs, line 751: I am not sure about this. I see a lot of DIM tests using partial interfaces. If partial interfaces were not supported prior to DIM support, it looks like they become available at the same time.

proposals/csharp-8.0/default-interface-methods.md
```diff
@@ -775,19 +788,19 @@ There are a few ways to observe whether an override declaration introduces a new
 ```csharp
 interface IA
 {
-    void M(int x) { }
+void M(int x) { }
```
AlekseyTs, line 791: Is the offset change intentional?

BillWagner: It was not. Reverted.

proposals/csharp-8.0/default-interface-methods.md
BillWagner: Thanks @AlekseyTs   

I see what I had done wrong:  I created an interface that declared partial methods with arguments and different return types. Those weren't supported in C# 8. 

Fixed.

proposals/csharp-8.0/default-interface-methods.md
BillWagner: I made the same mistake here, by not *implementing* the operator. So, the compiler interpreted it as a bad declaration for a static virtual member, which isn't supported in C# 8.

BillWagner:
Thanks @AlekseyTs 

Updates from today address all your comments.

PR #9087, opened 2025-01-23
Format meeting note links for collection expressions

jnm2:
Improving this, and providing a handy index now that my memory is fading of which dates corresponded to which topics.

image


CyrusNajmabadi (approved):

333fred (changes requested):
As discussed offline, we'll move these to the champion issues for consistency with the standard process.

jnm2:
Thanks @333fred! Followup to make the proposal prettier: https://github.com/dotnet/csharplang/pull/9088

PR #9091, opened 2025-01-24
Extensions: disallow partial members

jcouv:


proposals/extensions.md
```diff
@@ -126,6 +126,8 @@ public static class Enumerable
 }
 ```
 
+It is an error to specify the `partial` modifier on a member of an extension declaration.
```
RikkiGibson, line 129: It would be good to add some explanation for why we are restricting this.

HaloFour: I can see some interesting source generator use cases arising from partial members in extensions.

jcouv: We're trimming down on complexity. We'll add if we can identify some useful scenarios. Please drop a note if you know of some. For example, do some folks already use partial extension methods?

PR #9126, opened 2025-02-08
Extensions: proposal to use type inference during lookup

jcouv:

### DIFF
--- a/meetings/working-groups/extensions/extensions-lookup.md
+++ b/meetings/working-groups/extensions/extensions-lookup.md
@@ -59,7 +59,7 @@ public interface I<out T> { }
 ```
 
 My proposal is that the implementation continue to diverge from the spec: instead of doing 2-phase lookup 
-(as described in the section above, where we figure find compatible substituted extension containers, then find the candidate members in those)
+(as described in the section above, where we find compatible substituted extension containers, then find the candidate members in those)
 we could do a 1-phase lookup. We would only do this in invocation scenarios (ie. when arguments are provided).
 
 For such invocation scenarios:
--- a/meetings/working-groups/extensions/extensions-lookup.md
+++ b/meetings/working-groups/extensions/extensions-lookup.md
@@ -1,0 +1,98 @@
+There are two parts to this proposal:
+1. adjust how we find compatible substituted extension containers
+2. align with current implementation of extension methods
+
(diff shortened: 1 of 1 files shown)


meetings/working-groups/extensions/extensions-lookup.md
```diff
@@ -0,0 +1,99 @@
...
+   extension<T>(I<T> i)
+   {
+      public void M2(out T t) { t = default; }
+   }
+}
+public interface I<out T> { }
+```
+
+My proposal is that the implementation continue to diverge from the spec: instead of doing 2-phase lookup 
+(as described in the section above, where we figure find compatible substituted extension containers, then find the candidate members in those)
```
jjonescz, line 62:
```suggestion
(as described in the section above, where we find compatible substituted extension containers, then find the candidate members in those)
```

meetings/working-groups/extensions/extensions-lookup.md
```diff
@@ -0,0 +1,99 @@
...
+   {
+      public void M2(out T t) { t = default; }
+   }
+}
+public interface I<out T> { }
+```
+
+My proposal is that the implementation continue to diverge from the spec: instead of doing 2-phase lookup 
+(as described in the section above, where we find compatible substituted extension containers, then find the candidate members in those)
+we could do a 1-phase lookup. We would only do this in invocation scenarios (ie. when arguments are provided).
```
jaredpar, line 63: At this point should we just update the spec with the compiler behavior? If the behavior is so ingrained we can't change it with classic extension methods and will continue it for extension members, instead of it being a spec deviation maybe we change the spec to reflect the implementation.

jjonescz: (otherwise we should document it in https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/Deviations%20from%20Standard.md ?)

jcouv: We'll discuss this topic in LDM tomorrow.
FWIW, I have an issue tracking the discrepancy between spec and compiler here: https://github.com/dotnet/csharpstandard/issues/1203 (we'll convert to roslyn issue if we decide to adjust implementation or document deviation instead of updating spec).

PR #9132, opened 2025-02-11
Add missing "C#" to code block to turn on syntax highlighting

damieng:


proposals/csharp-13.0/params-collections.md
```diff
@@ -512,7 +512,7 @@ However, it looks like we will be much safer to use a different attribute for no
 For example, the current VB compiler will not be able to consume them decorated with `ParamArrayAttribute` neither in normal, nor in expanded form. Therefore, an addition of 'params' modifier is likely to break VB consumers, and very likely consumers from other languages or tools.
 
 Given that, non-array `params` parameters are marked with a new `System.Runtime.CompilerServices.ParamCollectionAttribute`.
-```
+``` C#
```
333fred, line 515: Nit: the document uses `cs` everywhere else.

```suggestion
``` cs
```

333fred: Or, actually, the first one that was visible is actually the oddity, and C# is the commonly-used one... never mind.

333fred (approved):

damieng:
Yeah I had cs then saw C# and thought about changing them all...

PR #9150, opened 2025-02-18
Clarify partial event attribute targets

jjonescz:


jjonescz:
@cston and @RikkiGibson for reviews, thanks

PR #9162, opened 2025-02-21
Dictionary expressions: include collection arguments

cston:
Merge specs for collection expression arguments and dictionary expressions, and update.


cston:
Updates and open questions were included in https://github.com/dotnet/csharplang/pull/9216. The only part that remains is merging the specs which was incomplete.

PR #9165, opened 2025-02-22
Revamp proposal for list patterns on enumerables

jnm2:


jnm2:
@alrz I'm interested in your feedback, too, since I heard you had an impl at one point?

proposals/list-patterns-on-enumerables-2.md
```diff
@@ -0,0 +1,91 @@
+# List patterns on enumerables
+
+Champion issue: <https://github.com/dotnet/csharplang/issues/9005>
```
jjonescz, line 3: Since the original proposal still exists, shouldn't we create a new issue instead?

jnm2: I'm not sure, happy to do so. @333fred is the choice of how to handle inactive proposals something that <https://github.com/dotnet/csharplang/blob/main/Design-Process.md> could speak to?

333fred: It could, I suppose, but we definitely don't want to create a new issue here. The point is trying to keep historical context, we don't want to throw out existing context.

jnm2:
@alrz It differs in two ways. 1) It recommends disallowing subpatterns on the slice pattern, even if the enumerable is sliceable. 2) It uses the enumerator directly like the foreach statement does, rather than a helper type.

proposals/list-patterns-on-enumerables-2.md
```diff
@@ -0,0 +1,91 @@
...
+
+Async enumerables are not supported. So far in the language, consumption of async enumerables requires the `await` keyword which highlights the point where execution may be suspended.
+
+No new syntax is involved in this proposal.
+
+### Design rationale
+
+#### No multiple enumeration
+
+Enumerables cannot be assumed to represent a materialized collection. An enumerable may represent an in-memory collection or a generated sequence, but it may also represent a remote query or an iterator method. As such, an enumerable may return different results on each enumeration, and enumeration may have side effects. Multiple enumeration is considered both a performance smell and a correctness issue in cases where the enumerable is not known to be a materialized collection or a trivially generated, guaranteed-stable sequence. The .NET SDK and other popular tools produce warnings for multiple enumeration of the same enumerable.
```
jnm2, line 33: Needs consideration: is the no-multiple-enumeration guarantee even possible when switching on the enumerable?

333fred:
> @alrz It differs in two ways. 1) It recommends disallowing subpatterns on the slice pattern, even if the enumerable is sliceable. 2) It uses the enumerator directly like the foreach statement does, rather than a helper type.

1 feels like it should just be added as an open question. 2 seems like a negative change, we moved to a helper type because the ldm had expressed some passing concern on codegen size. Neither of these feels like it should be a distinct proposal document; if anything, let's keep history intact by proposing updates to the existing document.

proposals/list-patterns-on-enumerables-2.md
jnm2: I don't think so. I also don't think any of these scenarios should be expected to reduce multiple enumeration compared to each other:
```cs
enumerable is [1] or [2, 3]
enumerable is [1] || enumerable is [2, 3]
enumerable.SequenceEqual([1]) || enumerable.SequenceEqual([2, 3])
```
The existing analyzers that warn about multiple enumeration in the last case can be augmented to see it happening in the prior two cases as well.

jnm2: The compiler _may_ be able to guarantee no multiple enumeration though. And then maybe `is [1] or [1, 2]` may not have to enumerate multiple times, and maybe this is worth guaranteeing.

jaredpar:
```csharp
enumerable is [1] || enumerable is [2, 3]
```

I think that must result in multiple enumerations because the `||` is not pattern matching. Basically it's two independent pattern matching expressions each of which has single enumeration of `enumerable`. 

```csharp
enumerable is [1] or [2, 3]
```

This though is a case where we'd want to avoid multiple enumerations.

jaredpar: > The compiler may be able to guarantee no multiple enumeration though. 

Think that would be important for us to support list patterns on `IEnumerable<T>`. Basically within a single pattern expression or `switch` there should be only a single enumeration of the enumerable. Once you break out of the pattern, with say an `||`, then all bets are off.

Issue #9174, opened 2025-02-25
[Proposal]: Readonly setter calls on non-variables


jnm2:
# Readonly setter calls on non-variables

* Specification: PR coming soon, creating issue to generate issue URL for PR
* Discussion: https://github.com/dotnet/csharplang/discussions/2068

## Summary

Permits a readonly setter to be called on _all_ non-variable expressions:

```cs
var c = new C();
// Remove the current CS1612 error, because ArraySegment<T>.this is readonly:
c.ArraySegmentProp[10] = new object();

// The same restriction has *already* been lifted for invocation expressions:
c.ArraySegmentMethod()[10] = new object();

class C
{
    public ArraySegment<object> ArraySegmentProp { get; set; }
    public ArraySegment<object> ArraySegmentMethod() => ArraySegmentProp;
}
```

Currently, the code above gives the error CS1612 "Cannot modify the return value of 'C.ArraySegmentProp' because it is not a variable." This restriction is meaningless when the setter is readonly. The restriction is there to remind you to assign the modified struct value back to the property. But there is no modification to the struct value when the setter is readonly, so there is no reason to assign back to the property.

```cs
// Requested by the current CS1612 error:
var temp = c.ArraySegmentProp;
temp[10] = new object();
c.ArraySegmentProp = temp; // But this line is purposeless; 'temp' cannot have changed.
```

PR #9175, opened 2025-02-25
[Proposal] Readonly setter calls on non-variables

jnm2:
@jaredpar indicated that this is a change that would need to be discussed. Context: https://github.com/dotnet/csharpstandard/issues/1277#issuecomment-2680341325

### DIFF
--- a/proposals/readonly-setter-calls-on-non-variables.md
+++ b/proposals/readonly-setter-calls-on-non-variables.md
@@ -11,17 +11,22 @@ var c = new C();
 // Remove the current CS1612 error, because ArraySegment<T>.this is readonly:
 c.ArraySegmentProp[10] = new object();
 
-// The same restriction has *already* been lifted for invocation expressions:
+// Invocation expressions already omit the CS1612 error when the setter is readonly:
 c.ArraySegmentMethod()[10] = new object();
+
+// In limited cases, ref-returning indexers can be used to work around this:
+c.RefReturningIndexerWorkaround[10] = new object();
 
 class C
 {
     public ArraySegment<object> ArraySegmentProp { get; set; }
     public ArraySegment<object> ArraySegmentMethod() => ArraySegmentProp;
+
+    public Span<object> RefReturningIndexerWorkaround => ArraySegmentProp.AsSpan();
 }
 ```
 
-Currently, the code above gives the error CS1612 "Cannot modify the return value of 'C.ArraySegmentProp' because it is not a variable." This restriction is meaningless when the setter is readonly. The restriction is there to remind you to assign the modified struct value back to the property. But there is no modification to the struct value when the setter is readonly, so there is no reason to assign back to the property.
+Currently, the code above gives the error CS1612 "Cannot modify the return value of 'C.ArraySegmentProp' because it is not a variable." This restriction is unnecessary when the setter is readonly. The restriction is there to remind you to assign the modified struct value back to the property. But there is no supported modification to the struct value when the setter is readonly, so there is no reason to assign back to the property.
 
 ```cs
 // Requested by the current CS1612 error:
@@ -66,7 +71,7 @@ c.SetXyz("key", c.GetXyz("key") + 1);
 c.Xyz["key"]++;
 ```
 
-The only thing that is blocking these use cases is a misappplied error that has not been fully updated with an understanding of readonly structs or readonly members.
+These use cases would no longer be blocked if CS1612 is fully updated with an understanding of readonly structs and readonly members.
 
 ## Detailed design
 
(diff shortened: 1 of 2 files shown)


proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -0,0 +1,127 @@
...
+// ❌ CS1918 Members of property 'C.ArraySegmentProp' of type 'ArraySegment<object>' cannot be assigned with an object
+// initializer because it is of a value type
+_ = new C { ArraySegmentProp = { [42] = new object() } };
+//          ~~~~~~~~~~~~~~~~
+
+class C
+{
+    public ArraySegment<object> ArraySegmentProp { get; set; }
+}
+```
```
timcassell, line 125: Also for ref properties.

```cs
_ = new C
{
  // CS1918
  S = { Prop = "abc" }
};
public class C {
    private S _s;
    public ref S S => ref _s;
}
public struct S
{
    public string Prop {get; set; }
}
```

proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -0,0 +1,127 @@
...
+c.ArraySegmentMethod()[10] = new object();
+
+class C
+{
+    public ArraySegment<object> ArraySegmentProp { get; set; }
+    public ArraySegment<object> ArraySegmentMethod() => ArraySegmentProp;
+}
+```
+
+Currently, the code above gives the error CS1612 "Cannot modify the return value of 'C.ArraySegmentProp' because it is not a variable." This restriction is meaningless when the setter is readonly. The restriction is there to remind you to assign the modified struct value back to the property. But there is no modification to the struct value when the setter is readonly, so there is no reason to assign back to the property.
```
jaredpar, line 24:
```suggestion
Currently, the code above gives the error CS1612 "Cannot modify the return value of 'C.ArraySegmentProp' because it is not a variable." This restriction is unnecessary when the setter is readonly. The restriction is there to remind you to assign the modified struct value back to the property. But there is no supported modification to the struct value when the setter is readonly, so there is no reason to assign back to the property.
```

Suggest softening the language a bit. There are scenarios where mutations could occur, like `unsafe`, but they're not really in the realm of supported for these type of considerations.

proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -0,0 +1,127 @@
...
+c.ArraySegmentProp[10] = new object();
+
+// The same restriction has *already* been lifted for invocation expressions:
+c.ArraySegmentMethod()[10] = new object();
+
+class C
+{
+    public ArraySegment<object> ArraySegmentProp { get; set; }
+    public ArraySegment<object> ArraySegmentMethod() => ArraySegmentProp;
+}
```
jaredpar, line 21: Another example to consider here is a `readonly struct` that has a `ref` returning indexer. That shows another way that we support getting around this limitation. The only reason that a `readonly struct` can't universally just take that approach is that some data storage backing, like if you backed your `readonly struct` with `List<T>` vs. `T[]` is the former doesn't have a great way at getting `ref` to the values.

proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -0,0 +1,127 @@
...
+
+```cs
+// Undesirable:
+c.SetXyz("key", c.GetXyz("key") + 1);
+
+// Desirable:
+c.Xyz["key"]++;
+```
+
+The only thing that is blocking these use cases is a misappplied error that has not been fully updated with an understanding of readonly structs or readonly members.
```
jaredpar, line 69: Soften

proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -0,0 +1,127 @@
...
+## Summary
+
+Permits a readonly setter to be called on _all_ non-variable expressions:
+
+```cs
+var c = new C();
+// Remove the current CS1612 error, because ArraySegment<T>.this is readonly:
+c.ArraySegmentProp[10] = new object();
+
+// The same restriction has *already* been lifted for invocation expressions:
```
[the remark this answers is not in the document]
jnm2: Oh, thanks for clarifying. This is really what I was asking when I asked "is it a compiler bug that the restriction was lifted for invocations but not for all expressions."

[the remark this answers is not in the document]
jaredpar: I honestly wasn't sure until that comment where I thought to debug through the code

jaredpar (approved):

RikkiGibson (approved):

jaredpar:
@jnm2 more changes coming or want us to merge?

jnm2:
Merge please!

jnm2:
Then if someone would not mind editing https://github.com/dotnet/csharplang/issues/9174 to replace the PR link with:

```markdown
[proposals/readonly-setter-calls-on-non-variables.md](https://github.com/dotnet/csharplang/blob/main/proposals/readonly-setter-calls-on-non-variables.md)
```

(I can't edit locked issues, even my own)

jnm2:
Yes, thanks!

jnm2:
Well, I see a third bullet now, but it's not critical.

jnm2:
Come to think of it, the entire summary is outdated compared to the merged proposal now, too. Sorry for the hassle, wish I could edit it.

PR #9180, opened 2025-02-26
Update based on LDM meeting

BillWagner:
We discussed this in the LDM meeting on 2025-02-24. This edit captures feedback from LDM.

I'll wait to merge until I see the notes from Monday, in case I missed anything.

### DIFF
--- a/meetings/working-groups/extensions/Extension-API-docs.md
+++ b/meetings/working-groups/extensions/Extension-API-docs.md
@@ -96,7 +96,7 @@ extension<T>(IEnumerable<T>)
 > }
 > ```
 >
-> It could save screen space to collect members by receiver declaration and remove the duplication. That would only apply on pages where multiple prototypes are shown together. Another negative is that it could conflict with the current lists in the TOC that have all overloads collected together.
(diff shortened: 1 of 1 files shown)


jcouv (approved):

meetings/working-groups/extensions/Extension-API-docs.md
```diff
@@ -84,6 +85,19 @@ extension<T>(IEnumerable<T>)
...
+> ```csharp
+> extension<T>(IEnumerable<T> source)
+> {
+>     public bool IsEmpty { get; }
+>     public int Length { get; }
+>     public IEnumerable<T> Sample(int sampleInterval);
+> }
+> ```
+>
+> It could save screen space to collect members by receiver declaration and remove the duplication. That would only apply on pages where multiple prototypes are shown together. Another negative is that it could conflict with the current lists in the TOC that have all overloads collected together.
```
gewarren, line 99: If you have multiple extension members on a single page, what would the URL of the docs page be? Something like /dotnet/api/namespace.type.extensions?

BillWagner: Adding a note to define this:

```suggestion
> It could save screen space to collect members by receiver declaration and remove the duplication. That would only apply on pages where multiple prototypes are shown together. Another negative is that it could conflict with the current lists in the TOC that have all overloads collected together. We should specify the URL for the docs page for a single `extension` declaration.
```

PR #9184, opened 2025-02-27
Field nullability revisions

RikkiGibson:


proposals/field-keyword.md
```diff
@@ -273,17 +273,18 @@ See [Glossary](#glossary) for definitions of new terms.
...
 - If the get accessor is auto-implemented, the property is not null-resilient.
 
 The nullability of the backing field is determined as follows:
 - If the field has nullability attributes such as `[field: MaybeNull]`, `AllowNull`, `NotNull`, or `DisallowNull`, then the field's nullable annotation is the same as the property's nullable annotation.
     - This is because when the user starts applying nullability attributes to the field, we no longer want to infer anything, we just want the nullability to be *what the user said*.
 - If the containing property has ***oblivious*** or ***annotated*** nullability, then the backing field has the same nullability as the property.
-- If the containing property has *not-annotated* nullability (e.g. `string` or `T`) or has the `[NotNull]` attribute, and the property is ***null-resilient***, then the backing field has ***annotated*** nullability.
-- If the containing property has *not-annotated* nullability (e.g. `string` or `T`) or has the `[NotNull]` attribute, and the property is ***not null-resilient***, then the backing field has ***not-annotated*** nullability.
```
jnm2, line 286: I guess the only difference this change makes is whether this becomes constructor warning requiring initialization instead of a "possible null reference return" warning on `get`?

```cs
[NotNull]
public string? Prop
{
    get;
    set => field = value ?? "";
}
```

jnm2 (approved):
🎉

proposals/field-keyword.md
RikkiGibson: Yeah I think so. And yes I decided that this isn’t a scenario where we need to do inference. Also, in the case of unconstrained T, inferring not nullable field may not be sufficient for things to work satisfactorily here. You’re already expressing things in a “wonky” way, go ahead and call it [AllowNull] string instead, or, attribute the field so that it works how you want.

cston (approved):

proposals/field-keyword.md
```diff
@@ -273,17 +273,18 @@ See [Glossary](#glossary) for definitions of new terms.
...
+- If the containing property has *not-annotated* nullability (e.g. `string` or `T`), and the property is ***null-resilient***, then the backing field has ***annotated*** nullability.
+- If the containing property has *not-annotated* nullability (e.g. `string` or `T`), and the property is ***not null-resilient***, then the backing field has ***not-annotated*** nullability.
```
jaredpar, line 287: Consider an alternative approach where we say that the backing field is _always_ annotated if we get into this section. This doesn't change our ability to decide if the `get` works with an unannotated return type, it just changes how we emit the field. I'm curious what specific things break, from a user and roslyn API perspective, if the field is always annotated?

RikkiGibson: I've been turning this idea over in my mind quite a bit and I think this is actually going to simplify implementation and improve clarity.

> it just changes how we emit the field

This seems fine. If we say a backing field of reference type is always nullable, when `field` keyword is used, it would affect some reflection behaviors, but I would be surprised if it was anything that breaking.

>I'm curious what specific things break, from a user and roslyn API perspective, if the field is always annotated?

Basically it means that for `field` in `string Prop { get => field; }`, Quick Info shows `string? Prop.field`, instead of `string Prop.field`. I actually think this is a good thing. Currently the implementation *always* shows `string Prop.field`, even when we inferred a nullable annotation for it.

The current design was based on the idea that if we just infer the nullable annotation, then constructor analysis will just do the right thing, in terms of knowing whether to enforce initialization of the property. If field is `string?`, then no need for enforcement. If it is `string!`, then need to enforce.

My stretch goal was to make public API show the inferred annotation. But if we do that, I'm concerned about "Schrodinger's null", where assigning maybe-null to the `field` actually changes its type from `string` to `string?`. This seems confusing, as the point of saying its type is `string` is to convey to the user that they are not allowed to put null in it.

If we go with "field is always annotated", then we need to adjust the constructor analysis design so that it uses the "null resilient" concept for these backing fields instead. "Null-resilient" means no need for enforcement. "Not null-resilient" means need to enforce. But that's not a big deal.

RikkiGibson: Also, the current design is to give the field the same nullability as the property, when any nullability attributes are used on the field. If we said the field is "always annotated when reference type", then we would probably want to do that even when attributes are used on the field.

This may break some existing cases where only a precondition or only postcondition attribute is used. Cases where both are specified (e.g. `[DisallowNull, NotNull]` won't change. I do not think such a break is problematic.

proposals/field-keyword.md
```diff
@@ -273,17 +273,20 @@ See [Glossary](#glossary) for definitions of new terms.
...
+- If there is a nullable diagnostic in the pass where the field was *annotated*, which was not present in the pass where the field is *not-annotated*, then the property is not null-resilient. Otherwise, it is null-resilient.
+    - The implementation can optimize by first performing the pass where the field was *annotated*. If this results in no diagnostics at all, then the property is null-resilient and the *not-annotated* pass can be skipped.
 - If the property does not have a get accessor, it is (vacuously) null-resilient.
 - If the get accessor is auto-implemented, the property is not null-resilient.
 
+Null-forgiving (`!`) operators, directives like `#nullable disable` and `#pragma disable warning`, and conventional project-level settings like `<NoWarn>`, are all respected when deciding if a nullable analysis has diagnostics. [DiagnosticSuppressors](https://github.com/dotnet/roslyn/blob/a91d7700db4a8b5da626d272d371477c6975f10e/docs/analyzers/DiagnosticSuppressorDesign.md) are not respected when deciding if a nullable analysis has diagnostics.
```
cston, line 286: >are not respected [](http://example.com/codeflow?start=377&length=17)

Perhaps "are ignored".

proposals/field-keyword.md
```diff
@@ -270,36 +270,48 @@ The following nullability rules will apply not just to properties that use the `
...
+
+The initial flow state of a *backing field* in its associated `get` accessor is determined as follows:
+- If the associated property's nullable annotation is *annotated* or *oblivious*, then the initial flow state is the same as the initial flow state of the associated property.
+- If the associated property's nullable annotation is *not-annotated*, then:
+    - If the property is *null-resilient*, the initial flow state is the *uninitialized* flow state (e.g. maybe-null, or maybe-default for unconstrained `T`).
+    - If the property is not *null-resilient*, the initial flow state is the *initialized* flow state (e.g. not-null, or maybe-null for unconstrained `T`).
 
 #### Constructor analysis
 
 Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
```
cston, line 296: >nullable-constructor-analysis.md [](http://example.com/codeflow?start=110&length=32)

csharp-9.0/nullable-constructor-analysis.md

proposals/field-keyword.md
```diff
@@ -270,36 +270,48 @@ The following nullability rules will apply not just to properties that use the `
...
 
-We update the following spec language from the previous [proposed approach](nullable-constructor-analysis.md#proposed-approach) to accomplish this:
+We update the following spec language from the previous [proposed approach](csharp-9.0/nullable-constructor-analysis.md#proposed-approach) to accomplish this (new language in **bold**):
```
cston, line 298: >csharp-9.0/nullable-constructor-analysis.md [](http://example.com/codeflow?start=76&length=43)

Consider using a fully-qualified link so the link is still correct when this proposal is moved into a distinct folder.

proposals/field-keyword.md
```diff
@@ -518,6 +530,13 @@ One reason we shied away from using nullability attributes here is that the ones
...
+1) "Schrodinger's null". It was felt to be confusing that a backing field displayed in the IDE like `string Prop.field`, could suddenly change to `string? Prop.field`, because the user assigned null to it. Its type being `string` implies that null can't be assigned to it.
+2) Exposing an inferred nullable annotation through the symbol APIs, internally and externally, was found to be more difficult from an engineering perspective, compared to simply setting an initial flow state. There's a risk that accessing the `FieldSymbol.TypeWithAnnotations`, `IFieldSymbol.Type`, or similar APIs, could result in the binding and flow analysis pulling on those same APIs, in simple or mutual recursion, overflowing the stack at runtime and crashing the host process.
+    - We also considered spec'ing and using this inferred annotation anyways, without exposing it in the usual symbol APIs, which seems again to undermine clarity, as a property displaying as `string Prop.field` can again have null written to it, have maybe-null initial state, etc.
```
cston, line 538: >anyways [](http://example.com/codeflow?start=69&length=7)

anyway

proposals/field-keyword.md
RikkiGibson: I tried to pursue the on the implementation side but it just ended up being too much churn at that time. I ended up pushing a change to bring us back around to being "annotation-oriented" which reflects what we shipped.

RikkiGibson:
@jnm2 @CyrusNajmabadi @333fred This PR now reflects the reality of what we shipped except that setter analysis is not implemented yet (dotnet/roslyn#77215).

proposals/csharp-14.0/field-keyword.md
```diff
@@ -274,34 +274,42 @@ See [Glossary](#glossary) for definitions of new terms.
...
+- Two separate nullable analysis passes are performed: one where the `field`'s nullable annotation is *not-annotated*, and one where its nullable annotation is *annotated*. The nullable diagnostics resulting from each analysis are recorded.
+- If there is a nullable diagnostic in the *annotated* pass, which was not present in the *not-annotated* pass, then the property is not null-resilient. Otherwise, it is null-resilient.
+    - The implementation can optimize by first performing the *annotated* pass. If this results in no diagnostics at all, then the property is null-resilient and the *not-annotated* pass can be skipped.
 - If the property does not have a get accessor, it is (vacuously) null-resilient.
 - If the get accessor is auto-implemented, the property is not null-resilient.
 
-The nullability of the backing field is determined as follows:
-- If the field has nullability attributes such as `[field: MaybeNull]`, `AllowNull`, `NotNull`, or `DisallowNull`, then the field's nullable annotation is the same as the property's nullable annotation.
```
jnm2, line 284: I'm glad we didn't go this route.

proposals/csharp-14.0/field-keyword.md
```diff
@@ -520,6 +528,12 @@ One reason we shied away from using nullability attributes here is that the ones
 - If we pursued this approach, we would consider adding a warning when `[field: AllowNull]` is used, suggesting to also add `MaybeNull`. This is because AllowNull by itself doesn't do what users need out of a nullable variable: it assumes the field is initially not-null when we never saw anything write to it yet.
 - We could also consider adjusting the behavior of `[field: MaybeNull]` on the `field` keyword, or even fields in general, to allow nulls to also be written to the variable, as if `AllowNull` were implicitly also present.
 
+#### Infer the initial flow state of the `field`
+
+Instead of defining the behavior in terms of the `field`'s nullable annotation, we could just the initial flow state of the `field` in accessors. This would remove the surprising behavior that the inferred nullable annotation is not exposed in the symbol APIs. Also, it would reflect the reality that it's ok to assign a possible null value to the `field`--the other forms of analysis we have, would succeed in identifying any resulting null safety issues.
```
jnm2:
```suggestion
Instead of defining the behavior in terms of the `field`'s nullable annotation, we could just define it in terms of the initial flow state of the `field` in accessors. This would remove the surprising behavior that the inferred nullable annotation is not exposed in the symbol APIs. Also, it would reflect the reality that it's ok to assign a possible null value to the `field`--the other forms of analysis we have, would succeed in identifying any resulting null safety issues.
```

jnm2 (approved):
Looks great to me, thank you for the updates!

proposals/csharp-14.0/field-keyword.md
```diff
@@ -274,34 +274,42 @@ See [Glossary](#glossary) for definitions of new terms.
...
+
+The nullable annotation of the backing field is determined as follows:
+- If the associated property's nullable annotation is *annotated* or *oblivious*, then the nullable annotation is the same as the nullable annotation of the associated property.
+- If the associated property's nullable annotation is *not-annotated*, then:
+    - If the property is *null-resilient*, the nullable annotation is *annotated*.
+    - If the property is not *null-resilient*, the nullable annotation is *not-annotated*.
+
+> [!NOTE]
+> The inferred nullable annotation of the backing field is not exposed in the Roslyn symbol APIs due to the dependency on binding and flow analysis of methods. The inferred annotation is instead used only in nullable analysis (and surfaced indirectly through APIs such as GetSymbolInfo/GetTypeInfo for expressions.) The nullable annotation of the property is used instead in the symbol APIs. See also subsection of [Nullability alternatives](#infer-the-initial-flow-state-of-the-field).
```
333fred: I don't really know if this note is appropriate for a csharplang spec?

RikkiGibson: I'm happy to move it to any more appropriate location, I just want the fact recorded somewhere that it can be found in the future when investigating behavior of this feature area.

PR #9185, opened 2025-02-28
Add scoping rules for receiver parameters

MadsTorgersen:
This adds scoping rules for receiver parameters. The upshot is that they should behave like primary constructor parameters: They are in scope within contained members, but it is illegal to reference them from a static context.


proposals/extensions.md
```diff
@@ -50,6 +50,47 @@ receiver_parameter // add
...
+        public int M(int i) // *2*
+        {
+            return s.Length + i;
+        }
+        public static string P => s; // Error: Cannot use s from static context
+    }
+}
+```
+
+In the example, `*1*` and `*2*` denote local variable declaration spaces, where `*2*` is nested within `*1*` and thus prohibited from introducing the same local variable names as `*1*`.
```
AlekseyTs, line 90: Either I am misinterpreting what this statement is saying, or it looks like it contradicts the current language rules, which allow shadowing of both parameters and locals within nested methods. For example, the following code is legal:
```
class C2(int p)
{
    
    void M1(long p)
    {}

    static void M2(long p)
    {}

    void M3()
    {
        long p = 0;
    }

    static void M4()
    {
        long p = 0;
    }
    
    void M5(int p1)
    {
        string local1 (string p1) => p1;
        local1("");

        int l1 = 0;
        
        string local2 ()
        {
            string p1 = "";
            string l1 = "";
            return p1 + l1;
        }
        
        local2();
    }
}
```

AlekseyTs: I think shadowing of locals and parameters within local functions and lambdas was allowed in Roslyn time frame. It is quite possible this relaxation didn't make it into the spec.

AlekseyTs: Here is the relevant proposal - [https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/shadowing-in-nested-functions.md](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/shadowing-in-nested-functions.md). With primary constructor parameters we followed the same rules, i.e. they can be shadowed by member names and by parameters and locals declared by members.

MadsTorgersen:
Will make a separate pull request that corresponds to LDM decisions

PR #9192, opened 2025-03-03
Add proposal for ignored directives

jjonescz:
https://github.com/jjonescz/csharplang/blob/sprint-spec-01/proposals/ignored-directives.md

### DIFF
--- a/proposals/ignored-directives.md
+++ b/proposals/ignored-directives.md
@@ -1,0 +1,126 @@
+# Ignored directives
+
+Champion issue: <https://github.com/dotnet/csharplang/issues/3507>
+
+## Summary
+
+Add `#` directives to be used by tooling, but ignored by the language.
+
+```cs
+#!/usr/bin/dotnet run
+#sdk Microsoft.NET.Sdk.Web
+#package Microsoft.AspNetCore.OpenApi 9.0.2
+
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddOpenApi();
+
+var app = builder.Builder();
+
+app.MapGet("/", () => new { Message = "Hello, World!" })
+    .WithName("HelloWorld");
+
+app.Run();
+```
+
+## Motivation
+
+We are adding `dotnet run file.cs` support [in .NET SDK][dotnet-run-file].
+Real-world file-based programs need to reference NuGet packages.
+It would be also useful if it were possible to execute file-based programs directly like `./file.cs` when they have [the shebang directive][shebang] (`#!`).
+
+The language should ignore these directives, but compiler implementations and other tooling can recognize them.
+We already have similar directives in the language:
+- `#region`
+- `#pragma`: details are implementation specific
+- `#error version`: the `version` part is not in the spec, but Roslyn will report its version
+
+## Detailed design
+
+Introduce new *ignored pre-processing directives* ([§6.5][directives]):
+
+```antlr
+PP_Kind
+    : ... // Existing directive kinds
+    | PP_Ignored
+    ;
+
+PP_Ignored
+    : PP_IgnoredToken Input_Character*
+    ;
+
+PP_IgnoredToken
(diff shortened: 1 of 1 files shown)


proposals/ignored-directives.md
```diff
@@ -0,0 +1,126 @@
...
+
+Champion issue: <https://github.com/dotnet/csharplang/issues/3507>
+
+## Summary
+
+Add `#` directives to be used by tooling, but ignored by the language.
+
+```cs
+#!/usr/bin/dotnet run
+#sdk Microsoft.NET.Sdk.Web
```
CyrusNajmabadi, line 11: instead of blessing certain pp directives (like `sdk/package`) I would prefer if we had some other sigil/marker to indicate "should be generally ignored, but can be used by some tool").  That way we don't have to keep adding new pp keywords to support new scenarios.

For example, the above supports sdk/package.  But what if we want a way to add ParseOptions/CompilationOptions, etc.?

It would be nice if we either made a strong argument that we've determined *all* the directives we are likely to need up front.  Or we have an extensible way of doing this.  For example `#$` or `#@` or something like that.

proposals/ignored-directives.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+### Restrictions
+
+Ignored directives must occur before the first token ([§6.4][tokens]) in the compilation unit, just like `#define`/`#undef` directives.
+This improves readability (all package references and other configuration is in one place), tooling performance (no need to scan long files in full),
+and simplifies the design (no need to figure out what would be the meaning of ignored directives inside `#if false`).
+
+Furthermore, the compiler should report a warning if the `#!` directive is not placed at the first line and the first character in the file
+(not even a BOM marker can be in front of it), because otherwise shells won't recognize it.
```
CyrusNajmabadi, line 66: not a complaint.  but we will have to likely update our tools that deal with headers (and determining if a doc is generated) to let them know that this info must be first, and nothing can precede it.

proposals/ignored-directives.md
yaakov-h: > But what if we want a way to add ParseOptions/CompilationOptions, etc.?

particularly with all the discussion around breaking changes (e.g. `field` keyword), having a `#langver` might be neccesary as well at some stage

jaredpar: > instead of blessing certain pp directives (like sdk/package) I would prefer if we had some other sigil/marker to indicate "should be generally ignored, but can be used by some tool"

Pretty sure this suggestion came up the last time we visited this topic (was several years ago). Essentially what if we just said that `#!` was the reserved prefix and everything after it was just ignored. Then the compiler team never has to come back to LDM again, we just add freely after that. 

I worry that idea won't hold up in practice. The `dotnet run app.cs` experience is going to be part of _the_ C# experience, for many people it will be their first entry into the language. Given that I believe that LDM will want to be involved in the choices that we're making for these directives as they'll define, to some degree, the C# experience. 

Do think we should put this idea in the _alternatives_ section though as it has come up before.

jjonescz: > For example, the above supports sdk/package. But what if we want a way to add ParseOptions/CompilationOptions, etc.?

The idea is to add only as few directives as possible, for anything more advanced, project-based programs should be used.

jaredpar (approved):

RokeJulianLockhart:
> https://github.com/dotnet/csharplang/pull/9192#event-16676173617

With this merged, do `.CS` files gain the ability to include a shebang (`#!`) on their first line?

jjonescz:
> With this merged, do `.CS` files gain the ability to include a shebang (`#!`) on their first line?

Not yet, this is just a proposal. After it's approved and implemented, then yes.

Do you have a use-case for `#!` in `.cs` files?

yaakov-h:
if only Roslyn development worked that quickly 😁

PR #9195, opened 2025-03-03
Extensions: use speakable implementation names

jcouv:

### DIFF
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -496,7 +496,7 @@ But that has some limitations, as roslyn only allows named type symbols (so no t
 - Should betterness be adjusted for resolution of static extension methods?
 - How to resolve properties? (answered in broad strokes LDM 2025-03-03, but needs follow-up for betterness)
 - Scoping and shadowing rules for extension parameter and type parameters?
-- How should ORPA apply to new extension methods and properties?
+- How should ORPA apply to new extension methods?
 - How to retcon the classic extension resolution rules? Do we 
   1. update the standard for classic extension methods, and use that to also describe new extension methods,
   2. keep the existing language for classic extension methods, use that to also describe new extension methods, but have a known spec deviation for both,
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -496,7 +496,7 @@ But that has some limitations, as roslyn only allows named type symbols (so no t
 - Should betterness be adjusted for resolution of static extension methods?
 - How to resolve properties? (answered in broad strokes LDM 2025-03-03, but needs follow-up for betterness)
 - Scoping and shadowing rules for extension parameter and type parameters?
-- How should ORPA apply to new extension methods?
+- How should ORPA apply to new extension methods and properties?
 - How to retcon the classic extension resolution rules? Do we 
   1. update the standard for classic extension methods, and use that to also describe new extension methods,
   2. keep the existing language for classic extension methods, use that to also describe new extension methods, but have a known spec deviation for both,
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -178,11 +178,9 @@ public static class NullableExtensions
 }
 ```
 
-### Compatible extension methods
+### Compatibility with classic extension methods
 
-By default, extension members are lowered in such a way that the generated artifacts are not visible at the language level.
-However, if the receiver specification is in the form of a parameter and specifies the `this` modifier, 
-then any extension instance methods in that extension declaration will generate visible classic extension methods.
+Instance extension methods are lowered in such a way that the generated artifacts match those produced by classic extension methods.
 
 Specifically the generated static method has the attributes, modifiers and name of the declared extension method, 
 as well as type parameter list, parameter list and constraints list concatenated from the extension declaration and the method declaration in that order:
@@ -190,7 +188,7 @@ as well as type parameter list, parameter list and constraints list concatenated
 ``` c#
 public static class Enumerable
 {
-    extension<TSource>(this IEnumerable<TSource> source) // Generate compatible extension methods
+    extension<TSource>(IEnumerable<TSource> source) // Generate compatible extension methods
     {
         public IEnumerable<TSource> Where(Func<TSource, bool> predicate) { ... }
         public IEnumerable<TSource> Select<TResult>(Func<TSource, TResult> selector)  { ... }
@@ -203,13 +201,13 @@ Generates:
 ``` c#
 public static class Enumerable
 {
+    [Extension]
     public static IEnumerable<TSource> Where<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate) { ... }
+
+    [Extension]
     public static IEnumerable<TSource> Select<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, TResult> selector)  { ... }
 }
 ```
-
-This does not change anything else about how the declared extension method works. 
-However, adding the `this` modifier may lead to a binary break for consumers because the generated artifact may change.
 
 ### Operators
 
@@ -444,6 +442,7 @@ static class CollectionExtensions
 - How to resolve properties? (answered in broad strokes LDM 2025-03-03, but needs follow-up for betterness)
 - Scoping and shadowing rules for extension parameter and type parameters?
 - How should ORPA apply to new extension methods?
+- How to retcon the classic extension resolution rules?
 
 ### Accessibility
 
--- a/proposals/extensions.md
+++ b/proposals/extensions.md
@@ -199,13 +199,14 @@ public static class Enumerable
 Generates:
 
 ``` c#
+[Extension]
 public static class Enumerable
 {
     [Extension]
-    public static IEnumerable<TSource> Where<TSource>(this IEnumerable<TSource> source, Func<TSource, bool> predicate) { ... }
+    public static IEnumerable<TSource> Where<TSource>(IEnumerable<TSource> source, Func<TSource, bool> predicate) { ... }
 
     [Extension]
-    public static IEnumerable<TSource> Select<TSource, TResult>(this IEnumerable<TSource> source, Func<TSource, TResult> selector)  { ... }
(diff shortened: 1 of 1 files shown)


proposals/extensions.md
```diff
@@ -277,44 +275,37 @@ public static class Enumerable
...
-As a strawman, consider `E @ T` as a disambiguation syntax meaning on a given expression `E` begin member lookup for an immediately enclosing expression in type `T`. For instance:
+Similarly to classic extension methods, the emitted implementation methods can be invoked statically.  
+This allows passing all the type arguments explicitly, and allows the compiler to disambiguate between extension members with the same name and arity.  
 
 ``` c#
 string[] strings = ...;
-var query  = (strings @ Enumerable).Where(s => s.Length > 10);
+
+var query = string.Select(s => s.Length); // extension invocation
+var query2 = string.Select<string, int>(s => s.Length); // ... with explicit full set of type arguments
```
jjonescz, line 287:
```suggestion
var query = strings.Select(s => s.Length); // extension invocation
var query2 = strings.Select<string, int>(s => s.Length); // ... with explicit full set of type arguments
```

proposals/extensions.md
```diff
@@ -353,14 +344,15 @@ Note: we may choose to only emit one extension skeleton type in metadata when du
...
-  For example: `set_Property` => `<Extension>set_Property`.
+- An implementation method has the same name as the original method.  
 - It has type parameters derived from the extension declaration prepended to the type parameters of the original method (including attributes).  
 - It has the same accessibility and attributes as the original method.  
 - If it implements a static method, it has the same parameters and return type. 
 - It if implements an instance method, it has a prepended parameter to the signature of the original method. 
   This parameter's attributes, refness, type, and name are derived from the receiver parameter declared in the relevant extension declaration.
 - The parameters in implementation methods refer to type parameters owned by implementation method, instead of those of an extension declaration.  
+- If the original member is an instance method, the implementation method is marked with an `[Extension]` attribute.
+- If the original member is static, the return type of the implementation method is marked with `modopt` to the parameter type of the containing extension declaration.
```
jjonescz, line 355: What's the purpose of emitting a `modopt`? For metadata roundtrip we can use the `<>E__` skeleton nested types, right?

proposals/extensions.md
```diff
@@ -398,23 +390,27 @@ static class IEnumerableExtensions
...
 
     // Implementation for Property
-    internal static int <StaticExtension>get_Property<T>() { ... }
-    internal static void <StaticExtension>set_Property<T>(int value) { ... }
+    internal static modopt(IEnumerable<T>) int get_Property<T>() { ... }
+    internal static modopt(IEnumerable<T>) void set_Property<T>(int value) { ... }
 
     // Implementation for SumAsync
-    public static int <Extension>SumAsync(IAsyncEnumerable<int> values) { ... }
+    public static int SumAsync(IAsyncEnumerable<int> values) { ... }
```
jjonescz, line 401:
```suggestion
    [Extension]
    public static int SumAsync(IAsyncEnumerable<int> values) { ... }
```

proposals/extensions.md
jaredpar: Using a `modopt` allows us to emit both of these names as speakable. 

```csharp
static class E {
  extension (C1)
    static C1 Create(); 
  extension (C2)
    static C2 Create();
}
```

Using `modopt` on a variation of the extension type it means that we can emit what are otherwise _identical_ signatures and keep speakable names. In the case there are no conflicts other languages can invoke the methods without any additional logic. In the case there are conflicts then other compilers will need to decide how they handle it.

jjonescz: I'd suggest explaining that in the speclet. But later in the speclet we seem to state that the language will not support this?

proposals/extensions.md
```diff
@@ -178,19 +178,17 @@ public static class NullableExtensions
 }
 ```
 
-### Compatible extension methods
+### Compatibility with classic extension methods
 
-By default, extension members are lowered in such a way that the generated artifacts are not visible at the language level.
-However, if the receiver specification is in the form of a parameter and specifies the `this` modifier, 
-then any extension instance methods in that extension declaration will generate visible classic extension methods.
+Instance extension methods are lowered in such a way that the generated artifacts match those produced by classic extension methods.
```
AlekseyTs, line 183: I think the "lowered" term is somewhat vague/undefined, unless it is explained in details later.

proposals/extensions.md
```diff
@@ -277,44 +275,37 @@ public static class Enumerable
 
 When an extension member lookup is attempted, all extension declarations within static classes that are `using`-imported contribute their members as candidates,
 regardless of receiver type. Only as part of resolution are candidates with incompatible receiver types discarded.
-A full generic type inference is attempted between the type of the actual receiver and any type parameters in the declared receiver type.
+A full generic type inference is attempted between the type of the arguments (including the actual receiver) and any type parameters (combining those in the declared receiver type and the extension member).
```
AlekseyTs, line 278:
```suggestion
A full generic type inference is attempted between the type of the arguments (including the actual receiver) and any type parameters (combining those in the extension declaration and in the extension member declaration).
```

proposals/extensions.md
```diff
@@ -277,44 +275,37 @@ public static class Enumerable
...
 
-The inferrability and uniqueness rules mean that the name of the enclosing static type is sufficient to disambiguate 
-between extension members on a given receiver type. 
-As a strawman, consider `E @ T` as a disambiguation syntax meaning on a given expression `E` begin member lookup for an immediately enclosing expression in type `T`. For instance:
+Similarly to classic extension methods, the emitted implementation methods can be invoked statically.  
+This allows passing all the type arguments explicitly, and allows the compiler to disambiguate between extension members with the same name and arity.
```
AlekseyTs, line 281: It would be good to illustrate a scenario where a disambiguation might be necessary, because it won't-help/not-needed for members declared within the same extension type, I think.

proposals/extensions.md
AlekseyTs, line 355: Unless the parameter type is a generic type parameter

proposals/extensions.md
AlekseyTs, line 401:
```suggestion
    [Extension]
    public static int SumAsync(IAsyncEnumerable<int> values) { ... }
```

proposals/extensions.md
```diff
@@ -377,7 +369,7 @@ static class IEnumerableExtensions
         public async Task<int> SumAsync() { ... }
     }
 
-    public void Method() { ... }
+    public void Method2() { ... }
```
AlekseyTs, line 372: It looks like there is no need to rename this method because it doesn't cause a conflict in the given scenario even when speakable names are used. Also, the method should probably be declared `static` since it is in a static class.

proposals/extensions.md
```diff
@@ -424,39 +420,37 @@ static class CollectionExtensions
...
+
 ### Metadata
 
-- Should we emit all implementation methods with speakable names instead, as a disambiguation strategy and also to allow
-  usage from other languages? We could add an attribute to handle compile-time conflicts in factory scenario (`[ExtensionName("CreateList")]`).
-- The metadata format currently doesn't include any modreqs to block other compilers. But the spec does mention we
-  would block those scenarios. Let's either remove this requirement or update the metadata format.  
-- We should follow-up on "factory scenario" where multiple extension declarations have static factory methods 
-  with same parameter types but different return types.
-- Should the extension marker or unspeakable implementation methods be marked with special name?
```
AlekseyTs, line 442: Should the question about special name remain? I didn't see changes related to that in other places and I do not remember LDM making decision on this.

proposals/extensions.md
```diff
@@ -398,23 +390,27 @@ static class IEnumerableExtensions
...
 
 Whenever extension members are used in source, we will emit those as reference to implementation methods.  
 For example: an invocation of `enumerableOfInt.Method()` would be emitted as a static call 
-to `IEnumerableExtensions.<Extension>Method<int>(enumerableOfInt)`.  
+to `IEnumerableExtensions.Method<int>(enumerableOfInt)`.  
 
-Note: multiple extension declarations defining static members with the same signature cannot be represented in metadata.  
-However, differences in return types can be represented, allowing for factory methods extending different types. For example:
+Note: the metadata representation supports static extension methods that differ in receiver type and return type, but
+  this will be disallowed in the language.
```
AlekseyTs, line 412: Might want to double check LDM notes for this. At least I do not remember us making an explicit decision like that.

proposals/extensions.md
```diff
@@ -424,39 +420,37 @@ static class CollectionExtensions
...
 
-- How to resolve invocations in static/type scenario? I assume we follow the same model as for instance methods. This will also work for Color Color scenarios.
-- How to resolve properties?
+- How to resolve instance method invocations now that we have speakable implementation names? We need to avoid ambiguity by understanding that those are the same method.
+- How to resolve static extension methods? (answer: just like instance extension methods, LDM 2025-03-03)
+- Should betterness be adjusted for resolution of static extension methods?
+- How to resolve properties? (answered in broad strokes LDM 2025-03-03, but needs follow-up for betterness)
 - Scoping and shadowing rules for extension parameter and type parameters?
 - How should ORPA apply to new extension methods?
+- How to retcon the classic extension resolution rules?
```
AlekseyTs, line 445: It is not clear to me what is this about. Consider elaborating.

AlekseyTs:

proposals/extensions.md
jcouv: The purpose of rename is to make this more readable (avoid any confusion between the two methods with the same name).

proposals/extensions.md
jcouv: @jaredpar In your example above, I don't think we need modopts, because the return types of the factories are different.

The case where you'd need modopt is when having the same return type. Is that right?
```
static class E {
  extension (C1)
    static Type Create(); 
  extension (C2)
    static Type Create(); // signatures conflict unless we change the signature (for example, using modopts)
}
```

AlekseyTs (approved):
LGTM (commit 8)

PR #9199, opened 2025-03-06
Target typed static lookup/updates

jnm2:


proposals/target-typed-static-member-lookup.md
```diff
@@ -183,6 +185,27 @@ For an invocation expression such as `e(...)` where the invoked expression `e` i
...
+```cs
+du switch
+{
+    .Case1(var arg1, var arg2) => ...
+}
+```
+
+This would continue to be target-typed static member access (since nested types are members of their containing type), which is distinct from target-typed `new` since a definite type is provided to the `new` operator.
+
+TODO: spec the conversion
```
jnm2, line 207: There are other todos in this file. I'm happy merging as-is, since it doesn't seem like there will be an issue in principle coming up with the conversion, and the proposal is being discussed at a higher level currently.

PR #9229, opened 2025-03-17
Describe scoping rules for extension declaration type parameters and receiver parameter

MadsTorgersen:
Based on LDM feedback, these are revised scoping rules for extension declarations.


proposals/extensions.md
```diff
@@ -50,6 +50,48 @@ receiver_parameter // add
...
+> ...
+> 
+> There are several different types of declaration spaces, as described in the following.
+> 
+> ...
+> 
+> - Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through formal parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set accessor for a property or an indexer introduces the name `value` as a formal parameter. 
+**An *extension_declaration* introduces to each member declaration directly contained within it its *type_parameter_list* as type parameters and the *identifier*, if any, of its *receiver_parameter* as a formal parameter.**
+
+There is an existing need to specify (maybe in Section[12.8.4 Simple Names](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#1284-simple-names)) something along the lines of the following:
```
jjonescz, line 66:
```suggestion
There is an existing need to specify (maybe in Section [12.8.4 Simple Names](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#1284-simple-names)) something along the lines of the following:
```

proposals/extensions.md
```diff
@@ -50,6 +50,48 @@ receiver_parameter // add
...
+``` c#
+public static class E
+{
+    extension(string s)
+    {
+        public int M1(int i) // *1*
+        {
+            return s.Length + i;
+        }
+        public int M2(string s) // *2* Error: Cannot reuse name `s`
```
AlekseyTs, line 81: > // *2* Error: Cannot reuse name `s`

Consider adding a similar example with type parameters. Something like:
```
public static class E
{
    extension<T>(T[] s)
    {
        public int M2<T>() // *2* Is this an error or a warning?
```

HaloFour: Same for the following?

```cs
public static class E<T>
{
    extension<T>(T[] s) // *2* Is this an error or a warning?
    {
        // ...
    }
}
```

AlekseyTs: > Same for the following?

Enclosing type cannot be generic

proposals/extensions.md
```diff
@@ -50,6 +50,48 @@ receiver_parameter // add
...
+Section [7.3 Declarations](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/basic-concepts.md#73-declarations) in the C# Standard is updated as follows (additions in **bold non-italic**):
+
+> ...
+> 
+> There are several different types of declaration spaces, as described in the following.
+> 
+> ...
+> 
+> - Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through formal parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set accessor for a property or an indexer introduces the name `value` as a formal parameter. 
+**An *extension_declaration* introduces to each member declaration directly contained within it its *type_parameter_list* as type parameters and the *identifier*, if any, of its *receiver_parameter* as a formal parameter.**
```
AlekseyTs, line 64: Do we expect the following error and, if so, what rules requires it?
```
    extension<T>(long[] p)
    {
        void T(){} // error CS0102: The type 'Extensions.extension<T>(long[])' already contains a definition for 'T'
    }
```

AlekseyTs, line 64: I think this rule can lead to unfriendly typing experience. Let's say I want to declare a method returning ```T``` in ```extension<T>(T[] x)```. Completion list in IDE won't offer ```T```. And if I type ```T``` and then press space or tab, completion is likely to insert something else that starts with "T". That is because ```T``` is not in scope, we are not in context of any member declaration yet.

AlekseyTs: Also, are type parameters of extension in scope inside receiver parameter declaration? Are they in scope inside constraints?

MadsTorgersen: Not an error. From a declaration space point of view it is equivalent to

``` c#
public void T<T>() { }
```

because, per the new rule, the type parameter `T` is added to the local variable declaration space of the member `T`.

proposals/extensions.md
MadsTorgersen: Thanks, fixed.

proposals/extensions.md
MadsTorgersen: Good idea. Added type parameter example.

proposals/extensions.md
MadsTorgersen: I don't think we have to implement it that way! I think the closest analogy is query expressions. From a spec point of view, the variable introduced by e.g. a `from` clause doesn't "exist" but becomes a parameter of several subsequent lambda expressions in a syntactic rewrite. However, the tooling experience treats it as a single variable. I'm imagining we do the same kind of thing here.

AlekseyTs: > I don't think we have to implement it that way!

I do not think this answers the following questions asked earlier on this thread: "Also, are type parameters of extension in scope inside receiver parameter declaration? Are they in scope inside constraints?"

> I think the closest analogy is query expressions.

I am sorry, but I do not get the analogy. 

> I'm imagining we do the same kind of thing here.

I am not sure what is being imagined here.

I suggest we get together for a half an hour an discuss this offline in more details.

AlekseyTs: > Not an error. From a declaration space point of view it is equivalent to ...

Consider adding an example to make this behavior explicit.

MadsTorgersen: I've updated the rules based on feedback. They are no longer expressed as a delta on the language spec, but more directly describe the intended behavior.

AlekseyTs (approved):
LGTM (commit 9)

proposals/extensions.md
```diff
@@ -50,6 +50,38 @@ receiver_parameter // add
 Extension declarations shall only be declared in non-generic, non-nested static classes.  
 It is an error for a type to be named `extension`.  
 
+### Scoping rules
+
+The type parameters and receiver parameter of an extension declaration are in scope within the body of the extension declaration. It is an error to refer to the receiver parameter from within a static member. It is an error for members to declare type parameters or parameters with the same name as a type parameter or receiver parameter of the extension declaration.
```
AlekseyTs, line 55: >It is an error for members to declare type parameters or parameters with the same name as a type parameter or receiver parameter of the extension declaration. [](http://example.com/codeflow?start=209&length=158)

I think it is worth mentioning that the same goes for local variables and local function declared directly by members (not inside lamndas or local functions)

PR #9230, opened 2025-03-17
Document breaking change for proposal discussion Wednesday

jnm2:


proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -77,7 +77,27 @@ These use cases would no longer be blocked if CS1612 is fully updated with an un
...
+
+The [Null-conditional assignment](https://github.com/dotnet/csharplang/blob/main/proposals/null-conditional-assignment.md) proposal enables the following syntax:
+
+```cs
+a?.b = value;
+a?.b.c = value;
+a?[index] = value;
+```
+
+CS1612 errors will be produced for this new syntax in the same way that they are for the corresponding forms when `a` is non-nullable:
```
jaredpar, line 90: Why do they need to be restricted to non-nullable? If it was a nullable the same logic for allowing the `set` applies. It's just writing through to heap state.

proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -145,3 +165,33 @@ class C
...
+
+## Downsides
+
+If this proposal is taken, it becomes a source-breaking change to remove the `readonly` keyword from a struct or setter. Without the `readonly` keyword, the errors would then be relevant and would reappear.
+
+However, in effect, this bridge has already been crossed for invocation expressions. It's already a source-breaking change to remove the `readonly` keyword from a struct or setter, because the struct value may be obtained from an invocation expression where the compiler already omits the CS1612 error for these:
+
+```cs
+// Removing 'readonly' from S1 causes a CS1612 error.
+M().Prop = 1;
```
RikkiGibson, line 177: The following is disallowed, which tripped me up a lot as I was thinking through this space:

```cs
S1? s1 = new S1();
s1.Value.Prop = 1; // error CS1612: Cannot modify the return value of 'S1?.Value' because it is not a variable
```

So, it sounds like a big part of the motivation here is, the current state is super inconsistent and weird.

proposals/readonly-setter-calls-on-non-variables.md
jnm2: I didn't intend this as a restriction, merely "here's the _existing_ syntax that parallels the new syntax mentioned above."

proposals/readonly-setter-calls-on-non-variables.md
jnm2: > The following is disallowed

@RikkiGibson I can see the consistency in this, if we imagine `Nullable<T>.Value` is treated as any other struct-returning property for this example.

jaredpar (approved):

PR #9233, opened 2025-03-18
Update null-conditional-assignment.md

RikkiGibson:


proposals/null-conditional-assignment.md
```diff
@@ -109,6 +109,16 @@ a?.b++; // error
 --a?.b; // error
 ```
 
+- This feature generally doesn't work when the receiver of the conditional access is a value type. This is because it will fall into one of the following two cases:
+```cs
+void Case1(ValueType a)
```
RikkiGibson, line 114: Avoid confusion with `System.ValueType`

```suggestion
void Case1(MyStruct a)
```

proposals/null-conditional-assignment.md
```diff
@@ -109,6 +109,16 @@ a?.b++; // error
...
+    => a?.b = c; // a?.b is not allowed when 'a' is of non-nullable value type
+
+void Case2(ValueType? a)
```
RikkiGibson, line 117:
```suggestion
void Case2(MyStruct? a)
```

RikkiGibson:
@jnm2 for visibility

PR #9295, opened 2025-04-09
Add the speclet disclaimer for published specs

BillWagner:
These are the specifications that are (or will shortly be) published on Learn.

Add the disclaimer that these are the feature specification.


proposals/field-keyword.md
jjonescz, line 1: Should the proposals also be moved to csharp-14 folder or that can happen separately?

BillWagner: That happens separately.

PR #9307, opened 2025-04-12
Work the expansion into the proposal for readonly setter calls on non-variables

jnm2:


proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -17,6 +17,14 @@ c.ArraySegmentMethod()[10] = new object();
 // In limited cases, ref-returning indexers can be used to work around this:
 c.RefReturningIndexerWorkaround[10] = new object();
 
+_ = new C
```
jaredpar, line 20: Think it would help to write out the signatures of `ArraySegment<T>` with the `readonly` modifiers to make it clear how the type is authored and why changes with the spec changes. Saves readers from looking up the type. Also makes spec more future proof in case the type changes shape in the future

jnm2:
@CyrusNajmabadi this is in a similar space to the fix that just went in for removing ToString on span

jnm2:
Wrong thread!

proposals/readonly-setter-calls-on-non-variables.md
jnm2: Updated.

jnm2: Updated!

proposals/readonly-setter-calls-on-non-variables.md
```diff
@@ -17,13 +17,28 @@ c.ArraySegmentMethod()[10] = new object();
...
     public ArraySegment<object> ArraySegmentMethod() => ArraySegmentProp;
 
     public Span<object> RefReturningIndexerWorkaround => ArraySegmentProp.AsSpan();
 }
+
+// Partial declaration of System.ArraySegment<T> for demonstration
+public readonly struct ArraySegment<T>
+{
+    // Implicitly readonly due to the readonly modifier on the struct
+    public T this[int index] { get; set; }
```
jaredpar: Make it clear there is a body here and not some artifact of auto implemented properties

```suggestion
    // Implicitly readonly due to the readonly modifier on the struct
    public T this[int index] { get => ...; set => ...; }
```

PR #9343, opened 2025-04-29
Extensions: capture more open issues

jcouv:


proposals/extensions.md
```diff
@@ -718,6 +718,38 @@ Do we stil need the first part of the rules?
 - Are we supposed to copy doc comments to the implementation methods with speakable names?
 - Should `<param>` element corresponding to receiver parameter be copied from extension container for instance methods? Anything else should be copied from container to implementation methods (`<typeparam>` etc.) ?
 
+### nameof
+
+- Should we disallow extension properties in nameof like we do classic and new extension methods?
```
jaredpar, line 723: With classic extension methods an argument could be made that the developer should just do `nameof` on the extension method definition. For example: `nameof(Enumerable.Select)`. For extension properties there isn't an equivalent as the disambiguation syntax would be `get_M / set_M`. Is that a consideration we should have here?

PR #9372, opened 2025-05-07
Extensions: spec updates and track more open issues

jcouv:
In terms of spec updates:
- remove indexer for now (full support may not make it in .NET 10)
- disallow `[ModuleInitializer]`
- allow `extern`
- add rule for entrypoints
- add proposal for cref
- relax inferrability rule
- mark skeleton types with `specialname`


proposals/extensions.md
```diff
@@ -40,7 +40,6 @@ extension_body // add
 extension_member_declaration // add
     : method_declaration
     | property_declaration
-    | indexer_declaration
```
jaredpar, line 43: Why?

PR #9415, opened 2025-05-29
Expand on the future ref rules a bit

jaredpar:


proposals/expand-ref.md
```diff
@@ -176,31 +194,145 @@ The [method arguments must match](https://github.com/dotnet/csharplang/blob/main
...
+- `__refvalue(e, T)`
+  - When `T` is a `ref struct`: will be treated as accessing a field declared as `ref scoped T` inside `e`.
+  - Will be treated as accessing a field declared as `ref T` inside `e`
 - `__arglist` as a parameter will be implicitly `scoped`
-- `__arglist(...)` as an expression will have a *ref-safe-to-escape* and *safe-to-escape* of *current method*. 
+- `__arglist(...)` as an expression will have a _ref-safe-to-escape_ and _safe-to-escape_ of _current method_.
 
 Conforming runtimes will ensure that `TypedReference`, `RuntimeArgumentHandle` and `ArgIterator` are defined as `ref struct`. Further `TypedReference` must be viewed as having a `ref` field to a `ref struct` for any possible type (it can store any value). That combined with the above rules will ensure references to the stack do not escape beyond their lifetime.
 
+```csharp
```
jaredpar, line 282: @AaronRobinsonMSFT here you go.

proposals/expand-ref.md
```diff
@@ -104,11 +107,19 @@ void M(
 The [method arguments must match](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#rules-method-arguments-must-match) rules will be updated to take `ref scoped` into account. Values passed to such parameters do not need to be considered when calculating the return scopes.
 
 Detailed notes:
+
 - A `ref scoped` parameter is implicitly `scoped ref`
 - An `out scoped` parameter declaration is an error
 
+### ref scoped locals
+
+The language will allow for locals to be declared as `ref scoped`. This will serve to have the lifetime of the value be _current method_. This lifetime is exactly _current method_ unlike paramteers which are just known to be wider than _current method_.
```
AaronRobinsonMSFT, line 116:
```suggestion
The language will allow for locals to be declared as `ref scoped`. This will serve to have the lifetime of the value be _current method_. This lifetime is exactly _current method_ unlike parameters which are just known to be wider than _current method_.
```

proposals/expand-ref.md
```diff
@@ -176,31 +194,145 @@ The [method arguments must match](https://github.com/dotnet/csharplang/blob/main
...
+}
+```
+
+This variance problem only exists for writes because we need to know the _exact_ lifetime in order to validate writes. There is no such issue for reads. Those are legal as long as the lifetime of the read is equal to or narrower than the value being read from. This means that the lifetime of the value referred to by `ref readonly` is variant.
+
+```csharp
+void ReadOnlyExample(ref Span<int> p)
+{
+    Span<int> local = stackalloc int[42];
+    Span<int> local2;
```
AaronRobinsonMSFT, line 234:
```suggestion
```

proposals/expand-ref.md
```diff
@@ -176,31 +194,145 @@ The [method arguments must match](https://github.com/dotnet/csharplang/blob/main
...
+{
+    int x = 42;
+    // This is logically the same as the following which means the lifetime is current method
+    // TypedReference tr = Create(ref x, typeof(int)); 
+    TypedReference tr = __makeref(x);
+
+    // Error: safe-to-escape is current method which is not returnable
+    return tr;
+}
+
```
AaronRobinsonMSFT, line 297: I would like to confirm the following example:

```csharp
struct OS
{
    public IS inner;
}

struct IS
{ }

TypedReference M4(TypedReference os)
{
    // os contains a 'ref OS'
    // Return a 'ref IS' via the TypedReference
}
```

En3Tho:
@jaredpar I wonder if there is some document describing your view of the C# lifetime system with goals, imaginary syntax and so on? I'm really curious how you envision such a big feature/change. (Even if it might never come).

jaredpar:
> I'm really curious how you envision such a big feature/change. (Even if it might never come).

Which change are you referring to here: a world where we have explicit lifetimes?

En3Tho:
@jaredpar Yes, explicit lifetime annotations in C#. How they would work with IDisposable, would they work with reference types, pointers?

What would be design goals: is it to enhance ref usage only or to enable developers with something like automatic disposal on scope exit?

I think it's fun to olay with such features and I'm curious what do you have in mind.

jaredpar:
> Yes, explicit lifetime annotations in C#. How they would work with IDisposable, would they work with reference types, pointers?

I've never sat down and wrote a doc about this because I wanted to focus on how far we could take the existing system. Once we know the limitations of how far you can take `scoped` plus a few _expert_ level attributes it helps define what is left and what level of investment you would use to solve those problems. 

Once I get pas this though, I will likely sit down and write a doc for where we'd go beyond `scoped`.which wolud include explicit lifetime rules.

PR #9455, opened 2025-06-17
Infer type parameters from constraints

333fred:
Adds a championed proposal for inferring type parameters from constraints.

### DIFF
--- a/proposals/type-parameter-inference-from-constraints.md
+++ b/proposals/type-parameter-inference-from-constraints.md
@@ -11,7 +11,7 @@ the bounds of type variables to participate in the inference process. An example
 List<int> l = [1, 2, 3];
 M(l); // Today: TElement cannot be inferred. With this proposal, successful call.
 
-void M<TEnumerable, TElement>(TEnumerable t) where TEnumerable : IEnumerable<T>
+void M<TEnumerable, TElement>(TEnumerable t) where TEnumerable : IEnumerable<TElement>
 {
     Console.WriteLine(string.Join(",", t));
 }
--- a/proposals/type-parameter-inference-from-constraints.md
+++ b/proposals/type-parameter-inference-from-constraints.md
@@ -38,7 +38,8 @@ breaking. Since then, C# has taken larger breaking change steps; most notably fo
 also adding things like target-typing for ternary expressions, adding span conversions as first-class conversions in the language, the `field` keyword, and others. Given this,
 now is an excellent time to re-examine the concern on the breaking change here, and potentially move forward with the proposal.
 
-Credit to [@HellBrick](https://github.com/HellBrick) for the [proposed mechanics](https://github.com/dotnet/roslyn/issues/5023#issuecomment-154728796) of the design, which have been formalized here.
+Credit to [@HellBrick](https://github.com/HellBrick) for the [original proposed mechanics](https://github.com/dotnet/roslyn/issues/5023#issuecomment-154728796) of the design.
+This proposal has been further refined from their original starting point.
 
 ### Use cases this supports:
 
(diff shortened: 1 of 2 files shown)


proposals/type-parameter-inference-from-constraints.md
```diff
@@ -0,0 +1,141 @@
...
+## Summary
+
+Allow type inference to succeed in overload resolution scenarios by promoting generic constraints of inferred type parameters to "fake arguments" during type inference, enabling
+the bounds of type variables to participate in the inference process. An example is:
+
+```cs
+List<int> l = [1, 2, 3];
+M(l); // Today: TElement cannot be inferred. With this proposal, successful call.
+
+void M<TEnumerable, TElement>(TEnumerable t) where TEnumerable : IEnumerable<T>
```
jjonescz: This typo is also in the champion issue description.

```suggestion
void M<TEnumerable, TElement>(TEnumerable t) where TEnumerable : IEnumerable<TElement>
```

proposals/type-parameter-inference-from-constraints.md
```diff
@@ -0,0 +1,141 @@
...
+The following text from [§12.6.3.12 Fixing](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#126312-fixing) is modified:
+
+> **12.6.3.12 Fixing**
+>
+> An *unfixed* type variable `Xᵢ` with a set of bounds is *fixed* as follows:
+>
+> - The set of *candidate types* `Uₑ` starts out as the set of all types in the set of bounds for `Xᵢ`.
+> - Each bound for `Xᵢ` is examined in turn: For each exact bound U of `Xᵢ` all types `Uₑ` that are not identical to `U` are removed from the candidate set. For each lower bound `U` of `Xᵢ` all types `Uₑ` to which there is *not* an implicit conversion from `U` are removed from the candidate set. For each upper-bound U of `Xᵢ` all types `Uₑ` from which there is *not* an implicit conversion to `U` are removed from the candidate set.
+> - If among the remaining candidate types `Uₑ` there is a unique type `V` to which there is an implicit conversion from all the other candidate types, then `Xᵢ` is fixed to `V` **and a lower-bound inference is performed from `V` to each of the types in `Xᵢ`'s constraints, if any**.
+> - Otherwise, type inference fails.
```
agocke, line 93: I haven't thought about the full implications of this algorithm, namely the set of things that will now be inferrable, and what will remain uninferrable.

Ideally I think we'd do as much as possible here. By that I mean, a general statement that if it is theoretically possible to infer the type, we do so. In practice, I'm worried our subtyping + overload resolution rules will make this undecidable.

But it's worth digging in to the literature to figure out precisely what will fall over.

agocke: Btw, starting w/ Algorithm W is not a bad idea.

proposals/type-parameter-inference-from-constraints.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Drawbacks
+
+The primary concern with this proposal is that it introduces potential breaking changes in overload resolution. Code that currently compiles and calls one overload might start
+calling a different overload after this feature is implemented.
```
MadsTorgersen, line 98: This could likely also lead to new ambiguity errors by bringing in new candidates that aren't clearly better or worse than the one called today. Probably worth mentioning.

proposals/type-parameter-inference-from-constraints.md
agocke: I've thought about this for a bit and my new thoughts are: I don't think this change is very complicated, and I'm not sure we have much to worry about.

Basically, we're not doing full type inference, we're just doing type subsitution for particular invocations. It doesn't seem like there's a lot of information that we could be missing with this algorithm.

PR #9463, opened 2025-06-18
Extensions: update nomenclature

jcouv:


proposals/extensions.md
```diff
@@ -399,11 +399,11 @@ These requirements need more refinement as implementation progresses, and may ne
...
 ### Metadata for declarations
 
-Each extension declaration is emitted as a skeleton type with a marker method and skeleton members.  
-Each skeleton member is accompanied by a top-level static implementation method with a modified signature.    
+Each extension declaration is emitted as an extension type with a marker method and extension members.  
+Each extension member is accompanied by a top-level static implementation method with a modified signature.    
 The containing static class for an extension declaration is marked with an `[Extension]` attribute.  
 
-#### Skeletons
+#### Extension types
```
jjonescz, line 406: Consider leaving a note that this used to be called "skeleton" since that term is probably still present in the implementation (and in fact I see it in some places in this proposal as well)

333fred: And certainly, it's present in the notes.

PR #9489, opened 2025-06-26
Extensions: update metadata design to avoid ordinals

jcouv:


proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+Each extension member is emitted with a stub implementation in the right extension grouping type, refers back to an extension marker type by its name via an attribute, and is accompanied by a top-level static **implementation method** with a modified signature.    
 
-#### Extension types
+Here's a schematized overview of metadata encoding:
+```
+[Extension]
+static class EnclosingStaticClass
+{
+    [Extension]
+    public class ExtensionGroupingType1 // has type parameters with minimal constraints sufficient to keep extension member declarations below valid
```
jjonescz: Should be sealed? #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
 
 ### Metadata for declarations
 
-Each extension declaration is emitted as an extension type with a marker method and extension members.  
-Each extension member is accompanied by a top-level static implementation method with a modified signature.    
-The containing static class for an extension declaration is marked with an `[Extension]` attribute.  
+Extension blocks are grouped by their CLR-level signature. Each CLR equivalency group is emitted as an **extension grouping type** with a content-based name.
+Extension blocks within a CLR equivalency group are then sub-grouped by C# equivalency. Each C# equivalency group is emitted as an **extension marker type** with a content-based name, nested in its corresponding extension grouping type.
```
jjonescz, line 424: Consider explaining why we need two levels of grouping as it's not immediately clear.

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- Its name is unspeakable and determined based on the contents of the CLR-level signature. More details below.  
+- Its type parameters have normalized names (`T0`, `T1`, ...) and have no attributes.  
 - Its accessibility is public.
+- It is marked with the `specialname` flag and an `[Extension]` attribute.  
+
+The content-based name of the extension grouping type is based on the CLR-level signature and includes the following:
+- The fully qualified CLR name of the type. This will include namespaces, type constraints, named constraints (like `new`).
+    - Type parameter names will be normalized to `T0`, `T1`, etc ... based on the order they appear in the type declaration. This means an extension block with a type parameter `Dictionary<TKey, TValue>` will result in the fully qualified name being `System.Collections.Generic.Dictionary<T0, T1>`.
+    - The fully qualified name will not include the containing assembly. It is common for types to be moved between assemblies and that should not break the public API.
```
jjonescz: >should not break the public API [](http://example.com/codeflow?start=136&length=31)

What kind of breaks are we talking about? It wouldn't be a binary nor source break anyway, right? #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- Constraints will included and sorted such that reordering them in source code does not change the name. Specifically:
```
jjonescz:
```suggestion
- Constraints will be included and sorted such that reordering them in source code does not change the name. Specifically:
``` #Resolved

File: proposals/extensions.md [hunk 5]
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    - Type parameter constraints will be listed in declaration order. The constraints for the Nth type parameter will occur before the Nth+1 type parameter.
+    - Base type and interface constraints will be sorted by comparing the full names ordinally
+    - Non-type constraints will be sorted by comparing the C# text ordinally
```
jjonescz: Are type and non-type constraints going to be mixed together? That is, changing from `where T : new()` to `where T : @new` (when there is a `class @new` somewhere) won't change the content-based name? #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- Since this does not include attributes, it intentionally ignores C#-isms like tuple names, nullability, etc ...
+
+Note: The name is guaranteed to remain stable across re-compilations, re-orderings and changes of C#-isms (ie. that don't affect the CLR-level signature).
```
jjonescz, line 485: I feel like we should state our goals with the content-based naming. (And it's probably also related to why we need two nested groupings.) I guess making the name remain stable is one of them. But that cannot be the only goal, otherwise their name could be constant. #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
 
 ### Metadata for declarations
```
jjonescz, line 400: Can we explain the motivation for this in this doc? If I understood correctly from https://github.com/dotnet/csharplang/blob/main/meetings/working-groups/extensions/content-based-naming.md#summary, it's for EnC, Public API tracking, and cref. #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+
+    ... ExtensionGroupingType2, etc ...
+
+    ... implementation methods ...
+}
+```
+
+The enclosing static class is emitted with an `[Extension]` attribute.  
+
+Note: This allows roundtripping of extension declaration symbols through metadata (full and reference assemblies).
```
jjonescz: So the extension marker method is preserved in reference assemblies even though it is private? Consider clarifying that. #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
-Method/property declarations in an extension declaration in source are represented as members of the extension type in metadata.  
-The signatures of the original methods are maintained (including attributes), but their bodies are replaced with `throw null`.  
-Those should not be referenced in IL.  
+The content-based name of the extension marker type is based on the following:
+- The fully qualified C# name of the type. 
+    - This will include items like nullable annotations, tuple names, etc ... 
+    - The constraints will have the same ordering as CLR types for the grouping type name
+    - The fully qualified name will not include the containing assembly
+- The name of the extension parameter
+- The modifiers of the extension parameter (`ref`, `ref readonly`, `scoped`, ...)
```
jjonescz: Those should probably be sorted so the name doesn't change after reordering them. #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- The fully qualified name and attribute arguments for any attributes applied to the extension parameter.
```
jjonescz: Attributes and their named arguments should probably be sorted so the content-based name doesn't change after reordering them. #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+  If the extension parameter doesn't specify a name, then the parameter name is empty.  
+- It is marked with the `specialname` flag.  
+
+#### Extension members
 
-Note: we may choose to only emit one extension type in metadata when duplicate extension declarations are found in source.  
+Method/property declarations in an extension block in source are represented as members of the extension grouping type in metadata.  
+- The signatures of the original methods are maintained (including attributes), but their bodies are replaced with `throw NotImplementedException()`.  
+- Those should not be referenced in IL.  
+- They are marked with `[ExtensionMarkerName("...")]` referring to the name of the extension marker type corresponding to the extension block for that member.
```
jjonescz: Is the ExtensionMarkerName attribute going to be synthesized by the compiler? #Resolved

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
 
 ### Metadata for declarations
 
-Each extension declaration is emitted as an extension type with a marker method and extension members.  
-Each extension member is accompanied by a top-level static implementation method with a modified signature.    
-The containing static class for an extension declaration is marked with an `[Extension]` attribute.  
+Extension blocks are grouped by their CLR-level signature. Each CLR equivalency group is emitted as an **extension grouping type** with a content-based name.
+Extension blocks within a CLR equivalency group are then sub-grouped by C# equivalency. Each C# equivalency group is emitted as an **extension marker type** with a content-based name, nested in its corresponding extension grouping type.
+An extension marker type contains a single **extension marker method** which encodes an extension parameter.
+Each extension member is emitted with a stub implementation in the right extension grouping type, refers back to an extension marker type by its name via an attribute, and is accompanied by a top-level static **implementation method** with a modified signature.
```
AlekseyTs: >Each extension member is emitted with a stub implementation [](http://example.com/codeflow?start=0&length=59)

It is not clear to me what does "emitted with a stub implementation" mean. I don't think we established a term  "stub implementation". Are you referring to a declaration representing declaration of an extension member? #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+        ... extension members for ExtensionGroupingType1, each points to its corresponding extension marker type ...
+    }
+
+    ... ExtensionGroupingType2, etc ...
+
+    ... implementation methods ...
+}
+```
+
+The enclosing static class is emitted with an `[Extension]` attribute.
```
AlekseyTs, line 451: >The enclosing static class is emitted with an `[Extension]` attribute. [](http://example.com/codeflow?start=0&length=70)

I think we decided that grouping types get this attribute as well. #Closed

AlekseyTs: > I think we decided that grouping types get this attribute as well.

I see that mentioned below.

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+Note: other constraints are preserved, such as `new()`, `struct`, `class`, `allows ref struct`, `unmanaged`, and type constraints.
 
-Each extension declaration in source is emitted as an extension declaration in metadata (sometimes referred to as skeleton type).  
-- Its name is unspeakable and determined based on the lexical order in the program.  
-  The name is not guaranteed to remain stable across re-compilation. 
-  Below we use `<>E__` followed by an index. For example: `<>E__2`.  
-- Its type parameters are those declared in source (including attributes).  
+#### Extension grouping types
+
+Each extension block with a distinct CLR-level signature in source is emitted as an extension grouping type in metadata (extension grouping types are an evolution of the previous "skeleton type" design).
```
AlekseyTs: >Each extension block with a distinct CLR-level signature in source is emitted as an extension grouping type [](http://example.com/codeflow?start=0&length=107)

I think this is confusing because the statement is worded in such a way that it avoids mentioning merging of several extension blocks into the same grouping type. Basically, I do not see any benefit in saying just that. Consider saying something like: "An extension grouping type is emitted to metadata for each set of extension blocks in source with a the same CLR-level signature." #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+#### Extension grouping types
+
+Each extension block with a distinct CLR-level signature in source is emitted as an extension grouping type in metadata (extension grouping types are an evolution of the previous "skeleton type" design).  
+- Its name is unspeakable and determined based on the contents of the CLR-level signature. More details below.  
+- Its type parameters have normalized names (`T0`, `T1`, ...) and have no attributes.  
 - Its accessibility is public.
+- It is marked with the `specialname` flag and an `[Extension]` attribute.  
+
+The content-based name of the extension grouping type is based on the CLR-level signature and includes the following:
+- The fully qualified CLR name of the type. This will include namespaces, type constraints, named constraints (like `new`).
```
AlekseyTs: >The fully qualified CLR name of the type. [](http://example.com/codeflow?start=2&length=41)

What type? If this is about the extended type, then we should say that explicitly. #Closed

AlekseyTs: >This will include namespaces, type constraints, named constraints (like `new`). [](http://example.com/codeflow?start=44&length=79)

I am not sure what this means. Constraints for type parameters have their own bullet point below. Why would we want to include constraints declared by extended type (this is different from constraints declared by an extension block)? #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    - Type parameter names will be normalized to `T0`, `T1`, etc ... based on the order they appear in the type declaration. This means an extension block with a type parameter `Dictionary<TKey, TValue>` will result in the fully qualified name being `System.Collections.Generic.Dictionary<T0, T1>`.
```
AlekseyTs: >Type parameter names will be normalized to `T0`, `T1` [](http://example.com/codeflow?start=6&length=53)

If we are using IL syntax for type names, the names of type parameters aren't important, they are referenced by ordinal. Why talk about names instead of simply saying that type parameters are referenced by their ordinal. #Closed

proposals/extensions.md
AlekseyTs: >Constraints will included [](http://example.com/codeflow?start=2&length=25)

"Constraints **of type parameters** will **be** included"? #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    - The fully qualified name will not include the containing assembly. It is common for types to be moved between assemblies and that should not break the public API.
+- Constraints will included and sorted such that reordering them in source code does not change the name. Specifically:
+    - Type parameter constraints will be listed in declaration order. The constraints for the Nth type parameter will occur before the Nth+1 type parameter.
+    - Base type and interface constraints will be sorted by comparing the full names ordinally
```
AlekseyTs: >Base type and interface constraints [](http://example.com/codeflow?start=6&length=35)

All these are called type constraints, consider simply using this term. #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    - Non-type constraints will be sorted by comparing the C# text ordinally
+- Since this does not include attributes, it intentionally ignores C#-isms like tuple names, nullability, etc ...
+
+Note: The name is guaranteed to remain stable across re-compilations, re-orderings and changes of C#-isms (ie. that don't affect the CLR-level signature).   
+
+#### Extension marker types
+
+The marker type re-declares the type parameters of its containing type (an extension grouping type) to get full fidelity of the C# view of extension blocks.
+
+Each extension block with a distinct C#-level signature in source results in an extension marker type in metadata.
```
AlekseyTs: Similar feedback as for grouping type. Consider saying something like: "An extension marker type is emitted to metadata for each set of extension blocks in source with a the same C#-level signature." #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+
+Note: The name is guaranteed to remain stable across re-compilations, re-orderings and changes of C#-isms (ie. that don't affect the CLR-level signature).   
+
+#### Extension marker types
+
+The marker type re-declares the type parameters of its containing type (an extension grouping type) to get full fidelity of the C# view of extension blocks.
+
+Each extension block with a distinct C#-level signature in source results in an extension marker type in metadata.  
+- Its name is unspeakable and determined based on the contents of the C#-level signature of the extension block. More details below.  
+- It redeclares the type parameters for its containing type to be those declared in source (including name and attributes).
```
AlekseyTs: >its containing type [](http://example.com/codeflow?start=40&length=19)

"containing grouping type"? #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- It is private and static.
 - It is marked with the `specialname` flag.  
 
-Method/property declarations in an extension declaration in source are represented as members of the extension type in metadata.  
-The signatures of the original methods are maintained (including attributes), but their bodies are replaced with `throw null`.  
-Those should not be referenced in IL.  
+The content-based name of the extension marker type is based on the following:
+- The fully qualified C# name of the type.
```
AlekseyTs: >The fully qualified C# name of the type. [](http://example.com/codeflow?start=2&length=40)

What type? #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- It is private and static.
 - It is marked with the `specialname` flag.  
 
-Method/property declarations in an extension declaration in source are represented as members of the extension type in metadata.  
-The signatures of the original methods are maintained (including attributes), but their bodies are replaced with `throw null`.  
-Those should not be referenced in IL.  
+The content-based name of the extension marker type is based on the following:
+- The fully qualified C# name of the type. 
+    - This will include items like nullable annotations, tuple names, etc ... 
+    - The constraints will have the same ordering as CLR types for the grouping type name
```
AlekseyTs: >- The constraints will have the same ordering as CLR types for the grouping type name [](http://example.com/codeflow?start=4&length=85)

It looks like constraints do not belong in this list #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+Each extension block with a distinct C#-level signature in source results in an extension marker type in metadata.  
+- Its name is unspeakable and determined based on the contents of the C#-level signature of the extension block. More details below.  
+- It redeclares the type parameters for its containing type to be those declared in source (including name and attributes).
+- It is private and static.
 - It is marked with the `specialname` flag.  
 
-Method/property declarations in an extension declaration in source are represented as members of the extension type in metadata.  
-The signatures of the original methods are maintained (including attributes), but their bodies are replaced with `throw null`.  
-Those should not be referenced in IL.  
+The content-based name of the extension marker type is based on the following:
```
AlekseyTs, line 497: >the following: [](http://example.com/codeflow?start=64&length=14)

It looks like the list is missing information about type parameters. Names, constraints, etc.  #Closed

proposals/extensions.md
AlekseyTs: >They are marked with `[ExtensionMarkerName("...")]` [](http://example.com/codeflow?start=2&length=51)

Do property accessors get this attribute as well? #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    }
+}
+```
+is emitted as
+```
+class E
+{
+    public class <>E__ContentName_For_IEnumerable_T<T0>
+    {
+        private static class <>E__ContentName1 // note: re-declares type parameters
```
AlekseyTs: >// note: re-declares type parameters [](http://example.com/codeflow?start=47&length=36)

Not sure how to interpret the comment, I do not see the redeclaration. #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,194 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    }
+
+    extension<T>(IEnumerable<U> source)
+        where T : IEquatable<U>
+    {
+        ... member in extension<T>(IEnumerable<U> source) where T : IEquatable<U>
+    }
+}
+```
+is emitted as
```
AlekseyTs, line 588: It looks like extension attributes are missing #Closed

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
 [Extension]
 static class IEnumerableExtensions
 {
-    public class <>E__1<T>
+    public class <>E__ContentName_For_IEnumerable_T<T0>
```
AlekseyTs: >public class <>E__ContentName_For_IEnumerable_T<T0> [](http://example.com/codeflow?start=4&length=51)

Extension attribute? #Closed

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
...
+
+        [ExtensionMarkerName("<>E__ContentName_For_IEnumerable_T_Source")]
         internal static int Property { get => throw null; set => throw null; }
+
+        [ExtensionMarkerName("<>E__ContentName_For_IEnumerable_T_Source")]
         public int Property2 { get => throw null; set => throw null; }
     }
 
-    public class <>E__2
+    public class <>E__ContentName_For_IAsyncEnumerable_Int
```
AlekseyTs: >public class <>E__ContentName_For_IAsyncEnumerable_Int [](http://example.com/codeflow?start=4&length=54)

Extension attribute? #Closed

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
...
     {
-        private static <Extension>$(IEnumerable<T> source) => throw null;
+        // Extension marker type is emitted as a nested type and re-declares its type parameters to include C#-isms
+        // In this example, the type parameter `T0` is re-declared as `T` with a `notnull` constraint:
+        // .class <>E__IEnumerableOfT<T>.<>E__ContentName_For_IEnumerable_T_Source
+        // .typeparam T
+        //     .custom instance void NullableAttribute::.ctor(uint8) = (...)
+        public static class <>E__ContentName_For_IEnumerable_T_Source
+        {
+            public static <Extension>$(IEnumerable<T> source) => throw null;
```
AlekseyTs: >public [](http://example.com/codeflow?start=12&length=6)

private? #Closed

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
...
         public int Property2 { get => throw null; set => throw null; }
     }
 
-    public class <>E__2
+    public class <>E__ContentName_For_IAsyncEnumerable_Int
     {
-        private static <Extension>$(IAsyncEnumerable<int> values) => throw null;
+        public static class <>E__ContentName_For_IAsyncEnumerable_Int_Values
+        {
+            public static <Extension>$(IAsyncEnumerable<int> values) => throw null;
```
AlekseyTs: >public [](http://example.com/codeflow?start=12&length=6)

private #Closed

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
...
-    public class <>E__1<T>
+    public class <>E__ContentName_For_IEnumerable_T<T0>
     {
-        private static <Extension>$(IEnumerable<T> source) => throw null;
+        // Extension marker type is emitted as a nested type and re-declares its type parameters to include C#-isms
+        // In this example, the type parameter `T0` is re-declared as `T` with a `notnull` constraint:
+        // .class <>E__IEnumerableOfT<T>.<>E__ContentName_For_IEnumerable_T_Source
+        // .typeparam T
+        //     .custom instance void NullableAttribute::.ctor(uint8) = (...)
+        public static class <>E__ContentName_For_IEnumerable_T_Source
```
AlekseyTs: >public [](http://example.com/codeflow?start=8&length=6)

private? #Closed

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
...
+
+        [ExtensionMarkerName("<>E__ContentName_For_IEnumerable_T_Source")]
         public int Property2 { get => throw null; set => throw null; }
     }
 
-    public class <>E__2
+    public class <>E__ContentName_For_IAsyncEnumerable_Int
     {
-        private static <Extension>$(IAsyncEnumerable<int> values) => throw null;
+        public static class <>E__ContentName_For_IAsyncEnumerable_Int_Values
```
AlekseyTs: >public [](http://example.com/codeflow?start=8&length=6)

private? #Closed

proposals/extensions.md
AlekseyTs: Consider also indicating SpecialName

proposals/extensions.md
```diff
@@ -466,17 +606,36 @@ is emitted as
 [Extension]
 static class IEnumerableExtensions
```
AlekseyTs, line 658: Consider indicating SpecialName where relevant #Closed

AlekseyTs:

proposals/extensions.md
jcouv: The extension member in metadata has a `throw ...` body, as opposed to a real body corresponding to source. "Each extension member is emitted as a skeleton member ..." may be better?

proposals/extensions.md
jcouv: Will add a note to be explicit, but that's already how reference assemblies work. FWIW, we verified that in `Implementation_InstanceMethod_01` but I'm planning to add more reference assembly tests.

proposals/extensions.md
jcouv: Numbers can't be used as identifiers in IL: ` error : syntax error at token '0' in:     .method private hidebysig instance void M<0> ( !!0 t ) cil managed `

proposals/extensions.md
jcouv: It would be a break to xml doc references. No source or binary break in traditional sense.

proposals/extensions.md
jcouv: I didn't cover that level of detail yet, but for the grouping type name the idea so far is to use an IL-looking format which can be interpreted unambiguously (ie. no possibility of two meaning for same string).

proposals/extensions.md
jcouv: I don't want to show IL syntax, so I'm using C# syntax and indicating that this nested type re-declared type parameters. In this example, the marker method refers to `T` which results from that re-declaration.
I have a comment above: "since C# cannot represent type parameter re-declaration, the code representing metadata is not valid C# code."

proposals/extensions.md
```diff
@@ -399,54 +399,237 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- Its name is unspeakable and determined based on the contents of the C#-level signature of the extension block. More details below.  
+- It redeclares the type parameters for its containing grouping type to be those declared in source (including name and attributes).
+- It is private and static.
 - It is marked with the `specialname` flag.  
 
-Method/property declarations in an extension declaration in source are represented as members of the extension type in metadata.  
-The signatures of the original methods are maintained (including attributes), but their bodies are replaced with `throw null`.  
-Those should not be referenced in IL.  
+The content-based name of the extension marker type is based on the following:
+- Constraints of type parameters will be included and sorted such that reordering them in source code does not change the name.
```
jcouv, line 500: @jaredpar I'm thinking to relax this requirement. If a change to tuple names or nullability annotations affect the marker name, why not also let a re-ordering of constraints affect the marker name?
It would mean that extension blocks different in constraint order would result in different markers, and we'd allow you to specify doc comments on each (as opposed to merging them with a warning).

I think I'm leaning towards keeping this sorting after all... #Resolved

proposals/extensions.md
jjonescz: Okay, so grouping types need to be stable to avoid breaking xml/EnC/API but also unique enough to avoid ambiguous crefs. But now I'm unsure why we need the C# sub-groups. Is that to allow reconstructing the original blocks from metadata (instead of them being merged by CLR signature)? In any case, I'd like to see the reasoning explained in this doc.

proposals/extensions.md
```diff
@@ -399,54 +399,237 @@ These requirements need more refinement as implementation progresses, and may ne
...
+`Extension.<>E__ExtensionGroupingTypeNameForObject.M(System.Int32)` but rather as `Extension.extension(object).M(int)`. In other words, extension members should be seen as having two sets of
+type parameters and two sets of parameters.
+
+#### Overview
+
+Extension blocks are grouped by their CLR-level signature. Each CLR equivalency group is emitted as an **extension grouping type** with a content-based name.
+Extension blocks within a CLR equivalency group are then sub-grouped by C# equivalency. Each C# equivalency group is emitted as an **extension marker type** with a content-based name, nested in its corresponding extension grouping type.
+An extension marker type contains a single **extension marker method** which encodes an extension parameter.
+Each extension member is emitted with a skeleton body in the right extension grouping type, refers back to an extension marker type by its name via an attribute, and is accompanied by a top-level static **implementation method** with a modified signature.
```
AlekseyTs: >skeleton body [](http://example.com/codeflow?start=40&length=13)

I think we agreed to not use the "skeleton" term. In addition, I do not see any need or benefit in including information about the body in this particular sentence. What body is emitted and where it is emitted is clearly explained later. Therefore, I am suggesting to use the following wording instead: "Declaration of each extension member is emitted in the right extension grouping type, ..."   #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,237 @@ These requirements need more refinement as implementation progresses, and may ne
...
+    }
+}
+```
+is emitted as
+```
+[Extension]
+class E
+{
+    [Extension]
+    public sealed class <>E__ContentName_For_IEnumerable_T<T0> // specialname
```
AlekseyTs: >specialname [](http://example.com/codeflow?start=66&length=11)

I think this could be indicated with a SpecialName attribute #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,237 @@ These requirements need more refinement as implementation progresses, and may ne
...
+- erasing parameter modifiers (like `ref`, `in`, `scoped`, ...)
+- erasing tuple names
+- erasing nullability annotations
+- erasing `notnull` constraints
+
+Note: other constraints are preserved, such as `new()`, `struct`, `class`, `allows ref struct`, `unmanaged`, and type constraints.
+
+#### Extension grouping types
+
+An extension grouping type is emitted to metadata for each set of extension blocks in source with a the same CLR-level signature (extension grouping types are an evolution of the previous "skeleton type" design).
```
AlekseyTs: >(extension grouping types are an evolution of the previous "skeleton type" design) [](http://example.com/codeflow?start=129&length=82)

Consider dropping this note. I do not find it particularly helpful to a reader of the spec. #Closed

proposals/extensions.md
```diff
@@ -399,54 +399,237 @@ These requirements need more refinement as implementation progresses, and may ne
...
+
+An extension grouping type is emitted to metadata for each set of extension blocks in source with a the same CLR-level signature (extension grouping types are an evolution of the previous "skeleton type" design).  
+- Its name is unspeakable and determined based on the contents of the CLR-level signature. More details below.  
+- Its type parameters have normalized names (`T0`, `T1`, ...) and have no attributes.  
+- It is public and sealed.
+- It is marked with the `specialname` flag and an `[Extension]` attribute.  
+
+The content-based name of the extension grouping type is based on the CLR-level signature and includes the following:
+- The fully qualified CLR name of the type of the extension parameter.
+    - Type parameter names will be normalized to `T0`, `T1`, etc ... based on the order they appear in the type declaration. This means an extension block with a type parameter `Dictionary<TKey, TValue>` will result in the fully qualified name being `System.Collections.Generic.Dictionary<T0, T1>`.
```
AlekseyTs: >This means an extension block with a type parameter `Dictionary<TKey, TValue>` will result in the fully qualified name being `System.Collections.Generic.Dictionary<T0, T1>`. [](http://example.com/codeflow?start=125&length=173)

I am not sure what this statement means. `Dictionary<TKey, TValue>` doesn't look like a type parameter. If we are trying to refer to an extension block declaration like 
```
extension<TKey, TValue>(Dictionary<TKey, TValue>)
```

even then I am not sure why do we need to say : "will result in the fully qualified name being `System.Collections.Generic.Dictionary<T0, T1>`" 

1. Given the current implementation, we simply are not going to produce such a name.
2. I do not think we need to have this level of implementation detail in this spec. Therefore, it is not really worth describing what exactly implementation is going to produce.

I would like to suggest dropping this sentence completely. Saying "Type parameter names will be normalized to `T0`, `T1`, etc ... based on the order they appear in the type declaration." is sufficiently clear, in my opinion.  #Closed

AlekseyTs: >Type parameter names [](http://example.com/codeflow?start=6&length=20)

"**Referenced** type parameter names"? #Closed

proposals/extensions.md
AlekseyTs: It still looks like type parameter names aren't mentioned. Aren't they going to affect the name?

proposals/extensions.md
AlekseyTs: For the benefit of the reader, consider making the comment more detailed. Something like: "re-declares type parameter T0 as T." So that one could get an idea where the "T" referenced in the marker method comes from. Similar suggestion for other similar comments in samples.

AlekseyTs:

AlekseyTs:
LGTM (commit 3)

proposals/extensions.md
jcouv: Yes, the purpose of the marker type and method is to encode the full fidelity information of extension block. Things that are lost in the extension type: type parameter and parameter names, attributes and certain C#-isms in parameter and constraints.

PR #9532, opened 2025-07-17
Extensions: collapse the open issues section and add a disclaimer

jcouv:


jcouv:
@BillWagner This is an attempt to mask the work-in-progress and speculative section of the speclet. Relates to https://github.com/dotnet/csharplang/issues/9530#issuecomment-3084371114 
Do you know if this collapsing feature will work when the document is ingested into learn website?

I would rather keep the open issues and alternatives in the same document (splitting our docs feels like a last resort). But I also want to reduce an possible confusion by users or AI reading the speclet.

PR #9560, opened 2025-07-29
Add type parameter restriction to closed hierarchies

MadsTorgersen:
Prevent classes that derive from closed classes from having "free" type parameters, since that would prevent an exhaustive switch.

### DIFF
--- a/proposals/closed-hierarchies.md
+++ b/proposals/closed-hierarchies.md
@@ -14,7 +14,7 @@ public record class Open(float Percent) : GateState;
 public record class Locked : GateState; // ERROR - 'GateState' is a closed class
 ```
 
-A consuming `switch` expression that covers all those derived classes can therefore be concluded to "exhaust" the closed class - it does not need to provide a default case to avoid warnings.
+Since all derived classes are declared in the closed class' assembly, a consuming `switch` expression that covers all of them can be concluded to "exhaust" the closed class - it does not need to provide a default case to avoid warnings.
 
 ``` c#
 // Assembly 3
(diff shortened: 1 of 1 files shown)


proposals/closed-hierarchies.md
```diff
@@ -55,6 +55,19 @@ public class C1 : CC { ... }     // Error, 'CC' is closed and in a different ass
 public class C2 : CO { ... }     // Ok, 'CO' is not closed
 ```
 
+### Type parameter restriction
+
+If a generic class directly derives from a closed class, then all if its type parameters must be used in the base class specification:
```
jnm2:
```suggestion
If a generic class directly derives from a closed class, then all of its type parameters must be used in the base class specification:
```

jnm2: Should we define "used in" precisely? What about:

```
closed class C<T1, T2>;
class D1<T1, T2> : C<int, Tuple<T1, T2>>;
```

mattwar: > Should we define "used in" precisely? What about:
> 
> ```
> closed class C<T1, T2>;
> class D1<T1, T2> : C<int, Tuple<T1, T2>>;
> ```

This still works.  The type parameters in the two types do not have to be the same, the type parameters from the sub-type just need to be used in the reference to the base type.

jnm2: Got it; when switching over a given `C<...>` in that case, `T1` and `T2` are held constant across the switch and exhaustivity doesn't have to worry about a varying type parameter in the derived class.

PR #9576, opened 2025-08-01
Clarify ignored directives speclet

jjonescz:


proposals/ignored-directives.md
```diff
@@ -50,6 +50,8 @@ PP_IgnoredToken
     ;
 ```
 
+These are parsed regardless of language version (the compiler generally dislikes syntax trees to be parsed differently based on LangVersion).
```
jcouv: I think it suffices to say this change is not tied to a specific language version.
The argument about parsing didn't fully make sense to me. It's not that we can't produce an error, it's that we don't want to produce an error :-)
In particular, we don't know the language version until we parse the file with ignored directives, so we don't want parsing of ignored directive to depend on language version.

PR #9607, opened 2025-08-11
Update Design-Process.md

jcouv:
I broke some relative links when moving speclets. Adding a note as reminder.

FYI @BillWagner @333fred 

Relates to https://github.com/dotnet/csharplang/pull/9606

### DIFF
--- a/Design-Process.md
+++ b/Design-Process.md
@@ -190,7 +190,7 @@ times this cannot happen for whatever reason. But the following rules _must_ be 
 Once a feature has been implemented and has or soon will be shipped, take the following steps (these are usually done in bulk when a release nears):
 
 1. If a folder for the C# release does not exist yet, create it.
-2. Move the specification for the feature into that folder.
+2. Move the specification for the feature into that folder and review relative links it contains.
 3. Update the champion issue as follows:
    1. Update the specification link to point at the new location.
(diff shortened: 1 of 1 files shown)


Design-Process.md
```diff
@@ -190,7 +190,7 @@ times this cannot happen for whatever reason. But the following rules _must_ be
 Once a feature has been implemented and has or soon will be shipped, take the following steps (these are usually done in bulk when a release nears):
 
 1. If a folder for the C# release does not exist yet, create it.
-2. Move the specification for the feature into that folder.
+2. Move the specification for the feature into that folder and review relative links it contains.
 3. Update the champion issue as follows:
    1. Update the specification link to point at the new location.
```
333fred, line 195: This line was intended to cover this. Any check-in relative links should be fully-specified to a commit, and shouldn't need to be updated.

BillWagner: It didn't. If a proposal refers to a proposal in an earlier release, that link is broken (e.g. `./csharp-12.0/some-feature.md` now must be `../csharp-12.0/some-feature.md`. And, the speclet disclaimer include changes from `./` to `../`.

333fred: I see. That makes more sense, thanks.

PR #9612, opened 2025-08-12
Reference inference proposals in nesting discussion

MadsTorgersen:


meetings/working-groups/discriminated-unions/to-nest-or-not-to-nest.md
```diff
@@ -173,19 +173,21 @@ Option<int> option = new .Some(5);   // 'Option<int>' inferred
 if (option is .Some(var value)) ...; // 'Option<int>' inferred
 ```
 
-Could target typing help eliminate type arguments in the non-nested case as well? It would have to look something like this:
+For the non-nested case, there are now proposals to:
+
+- Allow target types to be taken into account in generic type inference: [Target-typed generic type inference](https://github.com/dotnet/csharplang/blob/main/proposals/target-typed-generic-type-inference).
+- Allow generic type inference in `new` expressions: [Inference for constructor calls](https://github.com/dotnet/csharplang/blob/main/proposals/inference-for-constructor-calls).
+- Allow generic type inference in type patterns, treating the incoming type (`Option<int>` in this case) as the "target type": [Inference for type patterns](https://github.com/dotnet/csharplang/blob/main/proposals/inference-for-type-patterns).
```
jnm2, line 180: These links 404. Fix: https://github.com/dotnet/csharplang/pull/9615

Issue #9633, opened 2025-08-23
[Proposal]: Unsigned `sizeof`


jnm2:
# Unsigned `sizeof`

- Specification: https://github.com/dotnet/csharplang/pull/9634/files [TEMPORARY URL]
- Discussion: <https://github.com/dotnet/csharplang/discussions/5399>

## Summary

This proposal enables `sizeof` expressions without a constant value to implicitly convert to unsigned integer types the same size as `uint` or larger:

```cs
uint size = sizeof(SomeStruct);
```

## Design meetings

- https://github.com/dotnet/csharplang/blob/main/meetings/2026/LDM-2026-01-12.md#unsigned-sizeof

PR #9634, opened 2025-08-23
Add proposal for unsigned sizeof

jnm2:
Issue: https://github.com/dotnet/csharplang/issues/9633

### DIFF
--- a/proposals/unsigned-sizeof.md
+++ b/proposals/unsigned-sizeof.md
@@ -1,0 +1,85 @@
+# Unsigned `sizeof`
+
+Champion issue: <https://github.com/dotnet/csharplang/issues/9633>
+
+## Summary
+
+This proposal enables `sizeof` expressions without a constant value to implicitly convert to unsigned integer types the same size as `uint` or larger:
+
+```cs
+uint size = sizeof(SomeStruct);
+```
+
+## Motivation
+
+This is a minute language change which alleviates a paper cut within native interop domains.
+
+A common pattern for native Windows APIs involves assigning the size of a struct to one of its own fields, or passing the struct's size as an additional argument to a function along with a pointer to the struct. The language provides the required value easily using `sizeof(NativeStruct)`. This is only permitted in unsafe code, and it is safe so long as the struct is authored to be blittable—to have a memory layout identical to what the native API expects.
+
+These same native APIs make heavy use of unsigned integer types. The C# struct or method representing the native API will often use `uint` or other unsigned types in order to faithfully preserve the native API's distinction between signed and unsigned values. For example, `Microsoft.Windows.CsWin32` is a source generator which autogenerates such structs and methods based on Windows header files.
+
+A size is never negative. Thus, invariably, the native API is defined to take an unsigned integer, and the automated C# projection of that API requires the language user to produce a `uint` value for the size. Since the C# `sizeof` operator produces a non-constant `int` value for a user-defined struct, this results in the user having to insert explicit `uint` casts most of the time that `sizeof` is used with a struct.
+
+For example:
+
+```cs
+var buffer = default(PROCESS_BASIC_INFORMATION);
+
+PInvoke.NtQueryInformationProcess(
+    processHandle,
+    PROCESSINFOCLASS.ProcessBasicInformation,
+    &buffer,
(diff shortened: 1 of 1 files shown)


proposals/unsigned-sizeof.md
```diff
@@ -0,0 +1,85 @@
...
+    public int biWidth;
+    // ...
+}
+```
+
+The frequent insertion of explicit uint casts is ironic: the compiler is emitting the `sizeof` CIL instruction which itself natively produces an `unsigned int32` as defined by ECMA-335, 6th edition, §III.2.45.
+
+## Detailed design
+
+A new implicit conversion is defined, an _unsigned sizeof conversion_, from a `sizeof` expression to `uint`.
```
jcouv: ~~Instead of introducing a new conversion, would it be possible to change the type of `sizeof` to be a `uint` in the first place?~~

On second thought, that would be a breaking change (`int x = sizeof(...)` would now break, since `uint` can't implicitly be converted to `int`)

proposals/unsigned-sizeof.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Drawbacks
+
+No drawbacks are anticipated.
+
+## Answered questions
+
+## Open questions
+
+1. Does the new implicit conversion need to be defined as lower priority than all existing conversions in order to prevent this change from affecting overload resolution? For example:
```
jjonescz: Do conversions really have a priority or does this actually mean "a betterness rule"?

proposals/unsigned-sizeof.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Specification
+
+The [sizeof operator](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#12819-the-sizeof-operator) section is adjusted as follows (additions in **bold**):
+
+> ### 12.8.19 The sizeof operator
+>
+> The `sizeof` operator returns the number of 8-bit bytes occupied by a variable of a given type **as an `int` value**.
+
+Then, a new conversion is added to the [Implicit conversions](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/conversions.md#102-implicit-conversions) section:
```
jjonescz, line 67: Should it be a **standard** implicit conversion?

jnm2: I'm not sure. Where would the difference be noticed?
The default literal conversion is not a standard implicit conversion, for instance.

proposals/unsigned-sizeof.md
jnm2: It sounds better to use the existing mechanism for addressing overload resolution ambiguities. :+1:

proposals/unsigned-sizeof.md
jjonescz: Standard implicit conversions can be applied together with user defined conversion. That doesn't seem relevant here. So I guess it doesn't make a difference unless there is something else I'm missing.

jnm2: Would a standard conversion be needed in order to convert to uint and from there to ulong?

jjonescz: I don't think standard conversion would help there. To be able to convert to `ulong`, I think the new sizeof conversion would need to specify that explicitly. (Maybe we want to allow converting to `long` as well.)

jcouv: > Where would the difference be noticed?

A standard conversion would come into play if you do `M(sizeof(...))` and an `M` overload accepts some type with a user-defined conversion from `uint`.

jnm2: My reading agrees, that standard conversions only come into play with user-defined conversions, but what I am proposing wouldn't be a user-defined conversion. (https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/conversions.md#1041-general: "The standard conversions are those pre-defined conversions that can occur as part of a user-defined conversion.")

jnm2: Perhaps though, making it a standard conversion would enable `sizeof` to be implicitly combined with a user-defined conversion from `uint` to a user type.

jnm2: I have made it a standard implicit conversion and also explicitly added `nuint` and `ulong` as additional allowed types to convert to.

Issue #8606, opened 2024-11-18
[Proposal]: `using` with a discard


333fred:
# `using` with a discard

* Discussion: https://github.com/dotnet/csharplang/discussions/8605

## Summary
[summary]: #summary

Permit `using` variable declarations with a discard variable:

```cs
void M()
{
    using _ = GetDisposable();
}
```

## Design meetings
- https://github.com/dotnet/csharplang/blob/master/meetings/2018/LDM-2018-12-03.md#discussion
- https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-12-17.md#using-local-declaration-improvements

PR #9640, opened 2025-08-23
Create proposal for using discard statements

jnm2:
Champion issue: https://github.com/dotnet/csharplang/issues/8606

/cc @333fred as champion

### DIFF
--- a/proposals/using-with-a-discard.md
+++ b/proposals/using-with-a-discard.md
@@ -56,10 +56,7 @@ using _ = expr1, expr2;
 
 ## Specification
 
-> [!IMPORTANT]
-> This section is based on a draft of the C# 8 Standard specification that does not yet include disposable ref structs. Once the [ongoing specification work](https://github.com/dotnet/csharpstandard/pull/672/files#diff-20796c21eeccfd2c773f2969d23440cbc04e7439f4777729aad5d602477c232fR2057) is complete, this section will be rebased.
-
-[§13.14 The using statement](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/statements.md#1314-the-using-statement) is updated as follows.
+[§13.14 The using statement](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/statements.md#1314-the-using-statement) is updated as follows.
 
 ```diff
  using_statement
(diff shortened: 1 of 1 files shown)


proposals/using-with-a-discard.md
```diff
@@ -0,0 +1,178 @@
...
+using _ = expr1, expr2;
+```
+
+*using_declaration* syntax is not expanded or unified. *using_declaration* is based on *implicitly_typed_local_variable_declaration* and *explicitly_typed_local_variable_declaration* which do not include *simple discards*. A *simple discard* is a special case of a *declaration_expression* ([§12.17](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#1217-declaration-expressions)). (Specification work for *using_declaration* is [ongoing](https://github.com/dotnet/csharpstandard/pull/672/files#diff-20796c21eeccfd2c773f2969d23440cbc04e7439f4777729aad5d602477c232fR2057) at the time of writing.)
+
+## Specification
+
+[§13.14 The using statement](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/statements.md#1314-the-using-statement) is updated as follows.
+
+```diff
```
jcouv, line 61: It looks like the grammar section appears twice with some small differences (the first one has `await?`). Can we dedupe? #Closed

jnm2: Thanks, wdyt now?

proposals/using-with-a-discard.md
```diff
@@ -0,0 +1,162 @@
...
+ using_statement
+     : 'await'? 'using' '(' resource_acquisition ')' embedded_statement
```
333fred, line 63: Where's current using local declarations in this?

proposals/using-with-a-discard.md
```diff
@@ -0,0 +1,162 @@
...
+     ;
+
++using_discard_statement
++   : 'await'? 'using' '_' '=' expression ';'
++   ;
+```
+
+Removals in ~~strikeout~~, additions in **bold**:
+
+> If the form of *resource_acquisition* is *local_variable_declaration* then the type of the *local_variable_declaration* shall be ~~either `dynamic` or~~ a type that can be implicitly converted to `System.IDisposable`. If the form of *resource_acquisition* is *expression* then this expression shall be implicitly convertible to `System.IDisposable`. **If the form of *using_statement* is *using_discard_statement* then its contained *expression* shall be implicitly convertible to `System.IDisposable`.**
```
333fred: This is missing stuff from `ref struct` support.

333fred (commented):
There's a lot of missing stuff in the modifications around current using local declarations and ref struct support. Don't think this can be a pure modification over spec language that doesn't include those things.

proposals/using-with-a-discard.md
jnm2: It is _just_ on the verge of merging at <https://github.com/dotnet/csharpstandard/pull/672>. I'll be sure to come back and update this document, but we could choose to keep `using _ =` conceptually _not_ a using declaration. Things get murky if you define `using _ =` in terms of using declarations instead of in terms of equivalence with using statements.

proposals/using-with-a-discard.md
jnm2: I will update as soon as https://github.com/dotnet/csharpstandard/pull/672 merges. The wording has been in progress.

jnm2: In the meantime I'd like to be able get this on the schedule. What if I remove this for now, or put a WIP marker on it, enabling us to discuss in LDM without having the final spec, as we do for so many proposals?

proposals/using-with-a-discard.md
333fred: I don't know what "murky" means; can you clarify?

proposals/using-with-a-discard.md
333fred: The language isn't required to be final, but it should definitely be mentioned.

proposals/using-with-a-discard.md
jnm2: For sure! I touch on this on line 55:

> *using_declaration* syntax is not expanded or unified. *using_declaration* is based on *implicitly_typed_local_variable_declaration* and *explicitly_typed_local_variable_declaration* which do not include *simple discards*. A *simple discard* is a special case of a *declaration_expression* ([§12.17](https://github.com/dotnet/csharpstandard/blob/draft-v9/standard/expressions.md#1217-declaration-expressions)). (Specification work for *using_declaration* is [ongoing](https://github.com/dotnet/csharpstandard/pull/672/files#diff-20796c21eeccfd2c773f2969d23440cbc04e7439f4777729aad5d602477c232fR2057) at the time of writing.)

In short, using declarations involve actual local declarations, and `_ = expr` is not a local declaration. (It is a special case of a declaration expression, which is something else.) Trying to cast using discards as special case of using declarations felt inappropriate; the split between using discards and using declarations at the grammar level is so deep I felt they may as well not be related. Nor is a using discard in the spirit of a using declaration, whose purpose is presumably to declare a usable variable.

proposals/using-with-a-discard.md
jnm2: Updated.

jnm2: The v8 spec draft is probably-final now in this area, so I have rebased the spec section on it.

jnm2:
@jcouv @333fred Would you please merge?

PR #9682, opened 2025-09-12
Propose change for metadata names of grouping types

jcouv:


meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+- extension block: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.<M>$65CB762EDFDF72BBC048551FDEA778ED\`1"
+- extension member: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.M"
+
+
+**If we don't include an arity suffix**, then:
+- docIDs produced from VB symbols on extension metadata will diverge from those produce from C# symbols. VB doesn't have the concept of extension, so will have regular handling for types (which include an arity suffix)
+- if someone makes metadata for an extension type using some other tool, and they do include the arity suffix in the metadata names of grouping and marker types, then docIDs won't match metadata names again.
+
+The docIDs from C# source or metadata for the example would be:
+- extension block: "E.<G>$8048A6C8BE30A622530249B904B537EB.<M>$65CB762EDFDF72BBC048551FDEA778ED"
```
AlekseyTs: This doesn't look accurate. In this case, "they do include the arity suffix in the metadata names", the suffixes will be preserved by C# because they are in metadata names and C# doesn't remove them from metadata names.  #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+- extension member: "E.<G>$8048A6C8BE30A622530249B904B537EB.M"
```
AlekseyTs: Here as well. #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+    public void M() => throw; // extension member without implementation
+  }
+
+  public static void M(this int i) { } // implementation method
+}
+```
+
+**If we do include an arity suffix** when producing docIDs for extensions, then the docIDs don't match the metadata names.
+
+The docIDs for the example would differ from the names in metadata:
```
AlekseyTs: I think there is another doc id that includes type parameters, which drops the arity suffix, and that also causes slightly different issues. Consider including examples of those too. #Pending

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+Then the docIDs would not match the metadata names:
+- extension block: "E.GroupingName.MarkerName"
+- extension member: "E.GroupingName.M"
+
+# Proposal 
+
+We're proposing to update the metadata design to include arity suffix in the metadata names for grouping and marker types.  
+The metadata names for grouping and marker types would be mangled from the `ExtensionGroupingName` and `ExtensionMarkerName`.
```
AlekseyTs: I think there is no issue with `ExtensionMarkerName`. A conventional name for it wouldn't include any arity suffix, because from language perspective marker type inherits all type parameters from the enclosing grouping type, therefore itself not generic from C# or VB point of view.

``` C#
class C1<T, U>
{
    class C2
    {}
    class C3<S>
    {}
}
```

Checkout metadata name for C2:
``` IL
.class private auto ansi beforefieldinit C1`2<T, U>
    extends [mscorlib]System.Object
{
    // Nested Types
    .class nested private auto ansi beforefieldinit C2<T, U>
        extends [mscorlib]System.Object
    {
    } // end of class C2

    .class nested private auto ansi beforefieldinit C3`1<T, U, S>
        extends [mscorlib]System.Object
    {
    } // end of class C3`1
} // end of class C1`2
```
 #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+We're proposing to update the metadata design to include arity suffix in the metadata names for grouping and marker types.  
+The metadata names for grouping and marker types would be mangled from the `ExtensionGroupingName` and `ExtensionMarkerName`.  
+That would be more conventional.  
+Then we'd produce the docIDs as described above, by taking `ExtensionGroupingName` and `ExtensionMarkerName` and appending an arity suffix as well.
+
+For the above example, the compiler would produce:
+```
+.class E
+{
+  .class '<G>$8048A6C8BE30A622530249B904B537EB`1' // grouping type with arity suffix
```
AlekseyTs: Is there a type parameter? #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+  {  
+    .class '<M>$65CB762EDFDF72BBC048551FDEA778ED`1' // marker type with arity suffix
```
AlekseyTs: Is there a type parameter? And, as I mentioned above, there should be no arity suffix here  #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+  {
+    public void M() { }
+  }
+}
+```
+with corresponding metadata:
+```
+.class E
+{
+  .class '<G>$8048A6C8BE30A622530249B904B537EB' // grouping type
```
AlekseyTs: Is there a type parameter? #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+  {  
+    .class '<M>$65CB762EDFDF72BBC048551FDEA778ED' // marker type
```
AlekseyTs: Is there a type parameter? #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+      .. marker method ..
+    }
+    public void M() => throw; // extension member without implementation
+  }
+
+  public static void M(this int i) { } // implementation method
+}
+```
+
+Then the docIDs would not match the metadata names:
```
AlekseyTs, line 64: Why wouldn't they match? The types aren't generic, therefore suffixes wouldn't be added or removed by compilers. Perhaps there is a mistake in the IL and types should be generic? Suffix doesn't make types generic. #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+    {
+      .. marker method ..
+    }
+    public void M() => throw; // extension member without implementation
+  }
+
+  public static void M(this int i) { } // implementation method
+}
+```
+The `ExtensionGroupingName` would remain "<G>$8048A6C8BE30A622530249B904B537EB" (both for source and metadata symbols).
```
AlekseyTs, line 110: Effectively we are proposing that `ExtensionGroupingName` should reflect name of the emitted grouping type from language perspective, not its emitted name. Emitted name should follow naming convention used by C# for types declared in source. I think this should be clearly stated.  #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,183 @@
...
+The `ExtensionMarkerName` would remain "<M>$65CB762EDFDF72BBC048551FDEA778ED" (both for source and metadata symbols).
```
AlekseyTs, line 111: Taking my previous comments into account. `ExtensionMarkerName` reflects name of the emitted marker type from language perspective and its emitted name at the same time. Both names are equivalent. No need to change here. #Closed

AlekseyTs:

meetings/working-groups/extensions/metadata-names.md
jcouv: This paragraph is about docIDs from C# source or metadata. It relates to the first bullet. I'll make that more explicit.
The illustration for metadata from some other tool (ie. second bullet) is below.

meetings/working-groups/extensions/metadata-names.md
AlekseyTs: IL got adjusted and the thread was resolved, but I am still confused by the names below. What tool/component would produce the names below? Why the arity suffix wouldn't be there? It will be part of `ExtensionGroupingName` for C# and VB would include the suffix as well. Probably in this scenario there could be a problem with a doc id that includes type parameters instead?

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,185 @@
...
+}
+```
+
+Then the docIDs would not match the metadata names:
+- extension block: "E.GroupingName.MarkerName"
+- extension member: "E.GroupingName.M"
+
+# Proposal 
+
+We're proposing to update the metadata design to include arity suffix in the metadata names for grouping and marker types.
```
AlekseyTs: >and marker types [](http://example.com/codeflow?start=105&length=16)

Consider dropping marker types from this sentence since no change is proposed for them. #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,185 @@
...
+- extension member: "E.GroupingName.M"
+
+# Proposal 
+
+We're proposing to update the metadata design to include arity suffix in the metadata names for grouping and marker types.  
+The metadata name for grouping type would be mangled from the `ExtensionGroupingName`.  
+`ExtensionGroupingName` should reflect names of the emitted grouping typs from language perspective, not its emitted names. 
+That would be more conventional.  
+No change to `ExtensionMarkerName` (name and metadata names match).  
+Then we'd produce the docIDs as described above, by appending an arity suffix to `ExtensionGroupingName` as needed.
```
AlekseyTs: >as needed [](http://example.com/codeflow?start=105&length=9)

The "as needed" part is confusing because, in my opinion, in some scenarios compiler adds suffix when it is not needed. However, I think we are not proposing changing that part. Perhaps "in a way consistent to the current handling of regular generic types" would reflect what is proposed more accurately. #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,185 @@
...
+- extension member: "E.GroupingType\`1.M"
+
+Those docIDs are not ideal (they differ from metadata names), but that's not a new problem.
+
+# Practical considerations
+
+If we're going to make this change, we have about 1 week to get this into the RC2 compiler.  
+But the RC2 SDK/BCL will be compiled using the RC1 compiler, so the change would only be visible in the GA SDK/BCL assemblies.  
+The new compiler will still be able to consume extensions produced by the RC1 compiler: when loading metadata symbols, the unmangling is optional (if there is no arity suffix in the metadata name, we just use the whole metadata name as the name).  
+Given that only the compiler and docIDs make use of grouping and marker types, there would be no binary breaking change. Only implementation methods are referenced in IL, and those are unaffected by the change.
```
AlekseyTs: >and marker [](http://example.com/codeflow?start=61&length=10)

Consider dropping marker types, they are not affected #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,185 @@
...
+# Practical considerations
+
+If we're going to make this change, we have about 1 week to get this into the RC2 compiler.  
+But the RC2 SDK/BCL will be compiled using the RC1 compiler, so the change would only be visible in the GA SDK/BCL assemblies.  
+The new compiler will still be able to consume extensions produced by the RC1 compiler: when loading metadata symbols, the unmangling is optional (if there is no arity suffix in the metadata name, we just use the whole metadata name as the name).  
+Given that only the compiler and docIDs make use of grouping and marker types, there would be no binary breaking change. Only implementation methods are referenced in IL, and those are unaffected by the change.  
+
+# Alternative proposals
+
+We also brainstormed a design where docIDs for extensions would be produced by taking `ExtensionGroupingName` and `ExtensionMarkerName`, would strip any arity suffix (to deal with metadata from another tool), then would proceed as usual (add an arity suffix.
```
AlekseyTs: >and `ExtensionMarkerName` [](http://example.com/codeflow?start=110&length=25)

Consider dropping marker types, they are not affected #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,185 @@
...
+Splits identifier into name and arity, then searches symbols with matching Name and Arity (exact).
+
+2. CREF binding
+`BindNameMemberCref`, `ComputeSortedCrefMembers`  
+Takes a name followed by type argument list (which provides arity), looks up by name and arity (exact, for types), then constructs with type arguments
+
+
+## Metadata names and metadata loading
+The C# and VB compilers generally appends the arity suffix for generic types, to make the metadata name for a named type. There's some exceptions (notably EE named type symbols).
+For extension grouping and marker types, we use the grouping and marker names as-is, without appending a suffix. The grouping and marker names are the metadata names.
```
AlekseyTs: >and marker types [](http://example.com/codeflow?start=23&length=16)

Consider dropping marker types, they are not affected #Closed

AlekseyTs: >The grouping and marker names are the metadata names. [](http://example.com/codeflow?start=113&length=53)

Consider dropping this sentence entirely. #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,185 @@
...
+
+When loading types from metadata, the C# and VB compilers remove the arity suffix when it matches the arity found in metadata (see `MetadataHelpers.UnmangleMetadataNameForArity`).
+
+This means that we can load a type whose name doesn't include a suffix.
+
```
AlekseyTs, line 189: Current compiler's behavior is a proof of that because right now it has to load grouping types without suffix. #ByDesign

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,186 @@
...
+- extension grouping type: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1"
+- extension marker type: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.<M>$65CB762EDFDF72BBC048551FDEA778ED"
+- extension member: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.M"
+
+
+# Proposal 
+
+We're proposing to update the metadata design to include arity suffix in the metadata names for grouping types.  
+The metadata name for grouping type would be mangled from the `ExtensionGroupingName`.  
+`ExtensionGroupingName` should reflect names of the emitted grouping typs from language perspective, not its emitted names.
```
jjonescz:
```suggestion
`ExtensionGroupingName` should reflect names of the emitted grouping types from language perspective, not its emitted names. 
``` #Pending

File: meetings/working-groups/extensions/metadata-names.md [hunk 19]
```diff
@@ -0,0 +1,186 @@
...
+# Practical considerations
+
+If we're going to make this change, we have about 1 week to get this into the RC2 compiler.  
+But the RC2 SDK/BCL will be compiled using the RC1 compiler, so the change would only be visible in the GA SDK/BCL assemblies.  
+The new compiler will still be able to consume extensions produced by the RC1 compiler: when loading metadata symbols, the unmangling is optional (if there is no arity suffix in the metadata name, we just use the whole metadata name as the name).  
+Given that only the compiler and docIDs make use of grouping types, there would be no binary breaking change. Only implementation methods are referenced in IL, and those are unaffected by the change.  
+
+# Alternative proposals
+
+We also brainstormed a design where docIDs for extensions would be produced by taking `ExtensionGroupingName`, would strip any arity suffix (to deal with metadata from another tool), then would proceed as usual (add an arity suffix.
```
jjonescz:
```suggestion
We also brainstormed a design where docIDs for extensions would be produced by taking `ExtensionGroupingName`, would strip any arity suffix (to deal with metadata from another tool), then would proceed as usual (add an arity suffix).
``` #Pending

File: meetings/working-groups/extensions/metadata-names.md [hunk 20]
```diff
@@ -0,0 +1,186 @@
...
+But the docIDs from VB metadata would differ from those from C#:
+- extension grouping type: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1"
+- extension marker type: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.<M>$65CB762EDFDF72BBC048551FDEA778ED"
+- extension member: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.M"
+
+
+# Proposal 
+
+We're proposing to update the metadata design to include arity suffix in the metadata names for grouping types.  
+The metadata name for grouping type would be mangled from the `ExtensionGroupingName`.
```
jjonescz: What does it mean that metadata name would be mangled from `ExtensionGroupingName`? As I understand from the examples, the metadata name would be just `ExtensionGroupingName` plus the arity suffix appended. #Pending

jcouv: Yes, mangling means adding an arity suffix

jjonescz (approved):

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,186 @@
...
+}
+```
+
+Then the docIDs would not match the metadata names:
+- extension block: "E.GroupingName\`1\`1.MarkerName"
+- extension member: "E.GroupingName\`1\`1.M"
+
+
+**If we don't include an arity suffix**, then:
+- docIDs produced from VB symbols on extension metadata will diverge from those produce from C# symbols. VB doesn't have the concept of extension, so will have regular handling for types (which include an arity suffix)
```
AlekseyTs: >produce [](http://example.com/codeflow?start=80&length=7)

Typo? "produced" #Closed

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,186 @@
...
+
+But the docIDs from VB metadata would differ from those from C#:
+- extension grouping type: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1"
+- extension marker type: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.<M>$65CB762EDFDF72BBC048551FDEA778ED"
+- extension member: "E.<G>$8048A6C8BE30A622530249B904B537EB\`1.M"
+
+
+# Proposal 
+
+We're proposing to update the metadata design to include arity suffix to the `ExtensionGroupingName`
```
AlekseyTs: >to include arity suffix to the `ExtensionGroupingName` [](http://example.com/codeflow?start=46&length=54)

It feels like this contradicts the next sentence. This now reads as `ExtensionGroupingName` going to match metadata name. Then there is this sentence below: "Then we'd produce the docIDs as described above, by appending an arity suffix to `ExtensionGroupingName` ..."
Why would we need to add suffix if we already added it?

 I think the previous wording of this sentence was fine, we just needed to not mention marker type. For example: "We're proposing to update the metadata design to include arity suffix in the metadata names for grouping types." #Closed

AlekseyTs:
Done with review pass (commit 9)

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,186 @@
...
+  .class 'GroupingType`1'<T0> // grouping type
+  {  
+    .class 'MarkerType'<T> // marker type
+    {
+      .. marker method ..
+    }
+    public void M() => throw; // extension member without implementation
+  }
+
+  public static void M(this int i) { } // implementation method
```
333fred: Type parameter? #Pending

meetings/working-groups/extensions/metadata-names.md
333fred:
```suggestion
- docIDs produced from VB symbols on extension metadata will diverge from those produced from C# symbols. VB doesn't have the concept of extension, so will have regular handling for types (which include an arity suffix)
```

AlekseyTs (approved):
LGTM (commit 10)

meetings/working-groups/extensions/metadata-names.md
```diff
@@ -0,0 +1,189 @@
...
+  .class '<G>$8048A6C8BE30A622530249B904B537EB'<T0> // grouping type
+  {  
+    .class '<M>$65CB762EDFDF72BBC048551FDEA778ED'<T> // marker type
+    {
+      .. marker method ..
+    }
+    public void M() => throw; // extension member without implementation
+  }
+
+  public static void M(this int i) { } // implementation method
```
333fred: Consider doing a find/replace for `this int` and in this doc and fixing up all cases.

jcouv: Done. Thanks

PR #9699, opened 2025-09-23
Fix a variety of links

BillWagner:
These are triggering warnings on the Learn site.

### DIFF
--- a/proposals/csharp-14.0/field-keyword.md
+++ b/proposals/csharp-14.0/field-keyword.md
@@ -291,9 +291,9 @@ The nullable annotation of the backing field is determined as follows:
 
 #### Constructor analysis
 
-Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](../csharp-9.0/nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
+Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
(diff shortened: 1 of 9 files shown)


proposals/csharp-14.0/field-keyword.md
```diff
@@ -291,9 +291,9 @@ The nullable annotation of the backing field is determined as follows:
 
 #### Constructor analysis
 
-Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](../csharp-9.0/nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
+Currently, an auto property is treated very similarly to an ordinary field in [nullable constructor analysis](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-constructor-analysis.md). We extend this treatment to *field-backed properties*, by treating every *field-backed property* as a proxy to its backing field.
 
-We update the following spec language from the previous [proposed approach](../csharp-9.0/nullable-constructor-analysis.md#proposed-approach) to accomplish this (new language in **bold**):
+We update the following spec language from the previous [proposed approach](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-constructor-analysis.md#proposed-approach) to accomplish this (new language in **bold**):
```
RikkiGibson, line 296: Relative urls are discouraged? This is too bad as when working with the specs in the editor I can ctrl+click and have the editor open it instead of browser.

BillWagner: I agree. I wish it worked in the learn publishing system, but it doesn't in a dependent repo.

<Edit> Let me add more context here:  Mostly it can work. Where it doesn't is where the target isn't pulled into the learn site and published. So, mostly it's where we link to the rejected proposals.  This one is unique because it's an implementation spec that doesn't impact the standard semantics. It's not published on learn either.

PR #9719, opened 2025-09-26
Add top-level members proposal

jjonescz:
[Markdown](https://github.com/jjonescz/csharplang/blob/tlm/proposals/top-level-members.md)

Just an outline for now if anyone has any comments before I start filling in details.

cc @jaredpar as the LDM champion for this feature

### DIFF
--- a/proposals/top-level-members.md
+++ b/proposals/top-level-members.md
@@ -1,0 +1,117 @@
+# Top-Level Members
+
+Champion issue: (TODO)
+
+## Summary
+
+Allow some members (methods, operators, extension blocks, and fields) to be declared in namespaces
+and make them available when the corresponding namespace is imported.
+
+```cs
+// util.cs
+namespace MyApp;
+
+void Print(string s) => Console.WriteLine(s);
+
+string Capitalize(this string input) =>
+    input.Length == 0 ? input : char.ToUpper(input[0]) + input[1..];
+```
+
+```cs
+// app.cs
+#!/usr/bin/env dotnet
+
+using MyApp;
+
+Print($"Hello, {args[0].Capitalize()}!");
+```
+
+```cs
+// Fields are useful:
+namespace MyUtils;
+
+static string? cache;
+
+string GetValue() => cache ??= Compute();
+```
+
+```cs
+// Simplifies extensions:
+namespace System.Linq;
+
+extension<T>(IEnumerable<T> e)
+{
+    public IEnumerable<T> AsEnumerable() => e;
+}
+```
+
+## Motivation
+
+TODO: Why are we doing this? What use cases does it support? What is the expected outcome?
+
+- Avoid boilerplate utility static classes.
+- Evolve top-level statements from C# 9.
+
+## Detailed design
+
+TODO: This is the bulk of the proposal. Explain the design in enough detail for somebody familiar with the language to understand, and for somebody familiar with the compiler to implement,  and include examples of how the feature is used. This section can start out light before the prototyping phase but should get into specifics and corner-cases as the feature is iteratively designed and implemented.
+
+- Some members can be declared directly in a namespace (file-scoped or block-scoped).
+  - Allowed kinds currently are: methods, operators, extension blocks, and fields.
+  - Existing declarations like classes still work the same, there shouldn't be any ambiguity.
+  - There is no ambiguity with top-level statements because those are not allowed inside namespaces.
+- It is as if the members were in an "implicit" `static` class
+  whose accessibility is either `internal` (by default) or `public` (if any member is also `public`).
+  For top-level members, this means:
+    - The `static` modifier is disallowed (the members are implicitly static).
+    - The default accessibility is `internal`.
+      `public` and `private` is also allowed.
+      `protected` and `file` is disallowed.
+    - Overloading is supported.
+    - `extern` and `partial` are supported.
+    - XML doc comments work.
+- Metadata:
+  - A type synthesized per namespace and file. That means `private` members are only visible in the file.
+  - Cannot be addressed from C#, but has speakable name `TopLevel` so it is callable from other languages.
+    This means that custom types named `TopLevel` become disallowed in a namespace where top-level members are used.
+- Usage:
+  - `using NS;` implies `using static NS.TopLevel;`.
+  - Lookup for `NS.Method()` can find `NS.TopLevel.Method()`.
+  - Nothing really changes for extensions.
+- Entry points:
+  - Top-level `Main` methods can be entry points.
+  - Top-level statements are generated into `Program.Main` (speakable function).
+    This is a breaking change (previously the main method was unspeakable).
+  - Simplify the logic: TLS entry-points are normal candidates.
+    This is a breaking change (previously they were not considered to be candidates and for example `-main` could not be used to point to them).
+
+## Drawbacks
+
+TODO: Why should we *not* do this?
+
+- Polluting namespaces with loosely organized helpers.
+- Requires tooling updates to properly surface and organize top-level methods in IntelliSense, refactorings, etc.
+- Entry point resolution breaking changes.
+
+## Alternatives
+
+TODO: What other designs have been considered? What is the impact of not doing this?
+
+- Support `args` keyword in top-level members (just like it can be accessed in top-level statements). But we have `System.Environment.GetCommandLineArgs()`.
+- Allow capturing variables from top-level statements inside non-`static` top-level members.
+  Could be used to refactor a single-file program into multi-file program just by extracting functions to separate files.
+  But it would mean that a method's implementation (top-level statements) can influence what other methods see (which variables are available in top-level members).
+- Allow declaring top-level members outside namespaces as well.
+  - Would introduce ambiguities with top-level statements.
+  - Could be brought to scope via `extern alias`.
+    - To avoid needing to specify those in project files (e.g., so file-based apps also work),
+      there could be a syntax for that like `extern alias Util = Util.dll`.
+
+## Open questions
+
+TODO: What parts of the design are still undecided?
+
+- Which member kinds? Methods, fields, properties, indexers, events, constructors, operators.
+- Allow `file` or `private` or both? What should `private` really mean? Visible to file, namespace, or something else?
+- Name for the speakable static class (currently `TopLevel`)? Should it be speakable at all?
+- Should we simplify the TLS entry point logic? Should it be a breaking change?


proposals/top-level-members.md
```diff
@@ -0,0 +1,117 @@
...
+using MyApp;
+
+Print($"Hello, {args[0].Capitalize()}!");
+```
+
+```cs
+// Fields are useful:
+namespace MyUtils;
+
+static string? cache;
```
HaloFour: If all members are `static` is the modifier necessary here?  Without it would that pose an issue with the compiler correctly identifying that `cache` should be a field vs. a local variable in an implicit entry point?

jjonescz: No, good point, that should not be necessary, will update.

jjonescz: The proposal actually already disallows `static` modifier, I just got it wrong in the examples.

proposals/top-level-members.md
```diff
@@ -0,0 +1,118 @@
...
+- Avoid boilerplate utility static classes.
+- Evolve top-level statements from C# 9.
+
+## Detailed design
+
+TODO: This is the bulk of the proposal. Explain the design in enough detail for somebody familiar with the language to understand, and for somebody familiar with the compiler to implement,  and include examples of how the feature is used. This section can start out light before the prototyping phase but should get into specifics and corner-cases as the feature is iteratively designed and implemented.
+
+- Some members can be declared directly in a namespace (file-scoped or block-scoped).
+  - Allowed kinds currently are: methods, operators, extension blocks, and fields.
+  - Existing declarations like classes still work the same, there shouldn't be any ambiguity.
```
jcouv, line 57: Top-level statements can co-exist with type declarations, but must come first. Is there a similar rule for top-level members?

Would such types be in the namespace or nested in the `TopLevel` type? From the spec it should be in the namespace (since types are not allowed as top-level members), but that's surprising. So it may be better to just disallow them...

proposals/top-level-members.md
```diff
@@ -0,0 +1,118 @@
...
+  - There is no ambiguity with top-level statements because those are not allowed inside namespaces.
```
jcouv, line 58: It's a bit unfortunate that top-level members only be allowed inside namespaces. It seems irregular. But it does help make things unambiguous... Probably worth discussing more

proposals/top-level-members.md
```diff
@@ -0,0 +1,118 @@
...
+      `protected` and `file` is disallowed.
+    - Overloading is supported.
+    - `extern` and `partial` are supported.
+    - XML doc comments work.
+- Metadata:
+  - A type synthesized per namespace and file. That means `private` members are only visible in the file.
+  - Cannot be addressed from C#, but has speakable name `TopLevel` so it is callable from other languages.
+    This means that custom types named `TopLevel` become disallowed in a namespace where top-level members are used.
+- Usage:
+  - `using NS;` implies `using static NS.TopLevel;`.
```
jcouv, line 81: If I already have a namespace "N" with a type named "TopLevel" in metadata, will `using N;` gain the new behavior?

proposals/top-level-members.md
jjonescz: That's a good question, I will add it to the list. But I don't find the behavior surprising, consider that you have an existing code like

```cs
namespace N;
class C;
```

and you decide to add a top-level member, for example

```cs
namespace N;
int M() => 42;
class C;
```

that could be disallowed but if it's allowed it seems natural that the class remains directly in the namespace as it was before.

proposals/top-level-members.md
jjonescz: Hm, good question. Perhaps there should be some attribute akin to `[Extension]`.

proposals/top-level-members.md
jcouv: > if it's allowed it seems natural that the class remains directly in the namespace as it was before.

That's arguable, but not obvious. What if I sandwich the type?
```
namespace N;
int M() => 42;
class C;
int M2() => 42;
```

It doesn't seem obvious why `M()` and `M2()` are in `N.TopLevel`, but `C` is directly in `N`

proposals/top-level-members.md
```diff
@@ -0,0 +1,142 @@
...
+  But it would mean that a method's implementation (top-level statements) can influence what other methods see (which variables are available in top-level members).
+- Allow declaring top-level members outside namespaces as well.
+  - Would introduce ambiguities with top-level statements.
+  - Could be brought to scope via `extern alias`.
+    - To avoid needing to specify those in project files (e.g., so file-based apps also work),
+      there could be a syntax for that like `extern alias Util = Util.dll`.
+
+## Open questions
+
+TODO: What parts of the design are still undecided?
```
RikkiGibson: The rules for name conflicts should probably be spelled out.

For example, it seem as if the following declarations would be in conflict, due to how they are represented in the language

```cs
namespace NS;
int Foo;
class Foo { }
```

proposals/top-level-members.md
colejohnson66: My $0.02: There's been many a request to "reduce the indentation" with file-scoped classes (like file-scoped namespaces), and the team has rejected them for a variety of reasons — I would too. Combining top-level members with file-scoped types sounds like a recipe for disaster.

proposals/top-level-members.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Some members can be declared directly in a namespace (file-scoped or block-scoped).
+  - Allowed kinds currently are: methods, operators, extension blocks, and fields.
+  - Existing declarations like classes still work the same, there shouldn't be any ambiguity.
+  - There is no ambiguity with top-level statements because those are not allowed inside namespaces.
+- It is as if the members were in an "implicit" `static` class
+  whose accessibility is either `internal` (by default) or `public` (if any member is also `public`).
+  For top-level members, this means:
+    - The `static` modifier is disallowed (the members are implicitly static).
+    - The default accessibility is `internal`.
```
CyrusNajmabadi, line 68: I definitely don't like this or expect this. I would expect members without modifiers to have the narrowest accessibility (so only visible within the file. 

To be visible outside, you'd need to be explicit about internal/public.

Richiban: I think it *is* expected because it's the default visibility for everything that language currently allows to be top-level (classes/structs/delegates/enums).

For example, I think this is pretty unsatisfactory:

```cs
// util.cs
namespace MyApp;

// Implicitly `internal`
class MyThing(string _) {}

// Implicitly `private` or `file`
MyThing CreateThing(string s) => new MyThing(s);
```

Come to think of it, this raises another question: what happens to types declared inside a file that also contains top-level members? Using the example above, we could take it to mean either (a):

```cs
// util.cs
namespace MyApp;

static class <>__Generated
{
  class MyThing(string _) {}
  
  // Implicitly `private` or `file`
  MyThing CreateThing(string s) => new MyThing(s);
}
```

or (b)

```cs
// util.cs
namespace MyApp;

// Implicitly `internal`
class MyThing(string _) {}

static class <>__Generated
{
  // Implicitly `private` or `file`
  MyThing CreateThing(string s) => new MyThing(s);
}
```

Of the two I think (b) makes more sense†, in which case my initial reaction to visibility stands.

† I think (b) is preferrable to (a) because, otherwise, adding a top-level member to a file would suddenly nest all the classes declared in that file, potentially hiding them.

jjonescz: > I think (b) is preferrable to (a) because, otherwise, adding a top-level member to a file would suddenly nest all the classes declared in that file, potentially hiding them.

Agreed, I said that as well in another discussion above (https://github.com/dotnet/csharplang/pull/9719#discussion_r2390630839).

Although it might just be disallowed to mix existing declarations with the new top-level ones to avoid these kinds of issues.

proposals/top-level-members.md
Richiban: @jcouv 

> It doesn't seem obvious why M() and M2() are in N.TopLevel, but C is directly in N

Aren't you taking on a little too much implementation detail here? As far as the *user* is concerned `M` and `M2` are just functions in the namespace `N`. The fact that a containing class is generated for them would be--and should be--entirely hidden from the user.

You could even _nominally_ think of each top-level member as being given its own, generated, containing class. As long as the members can still see each other it makes no difference from a usability standpoint.

Richiban: To put it another way, my mental model of how this should work goes like this:

Rather than thinking "If this file contains top-level members then lift the contents of the file into a generated class" I think "For each declaration in this file, if the declaration is a type then leave it as-is, if it's a member then lift it into a generated class".

So

```cs
namespace N;

void M1() {}

class C1 {}

void M2() {}

class C2 {}

void M3() {}
```

becomes

```cs
namespace N;

partial static class TopLevel
{
    void M1() {}
}

class C1 {}

partial static class TopLevel
{
    void M2() {}
}

class C2 {}

partial static class TopLevel
{
    void M3() {}
}
```

The-Futurist:
I don't get this. The most frequent argument I see here against many suggestions is why do it, when the language already provides a way to do it. We have static classes, we have using static, why do we need such an esoteric thing as namespace acting like static classes when we have static classes - perhaps I missed something...

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+  ```
+
+## Open questions
+
+- Which member kinds? Methods, fields, properties, indexers, events, constructors, operators.
+- Accessibility: what should be the default and which modifiers should be allowed?
+- Clustering: currently each namespace per assembly gets its `TopLevel` class.
+- Shape of the synthesized static class (currently `[TopLevel] TopLevel`).
+- Should we simplify the TLS entry point logic? Should it be a breaking change?
+- Should we require the `static` modifier (and keep our doors open if we want to introduce some non-`static` top-level members in the future)?
```
CyrusNajmabadi, line 185: fwiw, i think we should require 'static'.  It seems very odd to me that these have 'instance' syntax but aren't referenced/invoked as instancs off of soemthing.  As these are associated with namespaces, and you can't have 'instances' of namespaces, making these look like instance seems weird.  This is esp true as you would be able to write things like `N.SomeNSMethod();`.   That strongly screams "i am static".   

Note: this also follows how we require 'static' on members, even when in a 'static class'.  Just because they must be 'static' doesn't mean we think that 'static' is then not required to say.  

TO me, this makes the declaration/semantics extremely clear, and logically fits in well with other parts of the language (including those that this is directly sitting on top of).

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+  module Utilities =
+    let M() = ()
+
+  open Utilities
+  M()
+  ```
+
+## Open questions
+
+- Which member kinds? Methods, fields, properties, indexers, events, constructors, operators.
```
CyrusNajmabadi, line 180: operators seems weird, as an operator must say which types it operates on, and those must include one of the containing type.

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+- Accessibility: what should be the default and which modifiers should be allowed?
```
CyrusNajmabadi, line 181: minimal accessibility should be hte default.  so absent accessibility, it should be 'private'.

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+
+extension<T>(IEnumerable<T> e)
+{
+    public IEnumerable<T> AsEnumerable() => e;
+}
+```
+
+## Motivation
+
+- Avoid boilerplate utility static classes.
```
CyrusNajmabadi, line 50: tbh, i'm not sure this is waht we need to care that much about.  The declaration side seems fine if you have to say:

```c#
[TopLevel]
public static class TopLevel {

}
```

What matters more (imo) is the consumption side.  So that if we see such a 'TopLevel', we bring its members into scope for code that is pulling in that namespace.

This matches how we pull in extension members into scope when you pull in a namespace.

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+  but their members don't need to be qualified with the module name if they are brought to scope via an import:
+  ```vb
+  Imports N
+
+  Namespace N
+      Module M
+          Sub F()
+          End Sub
+      End Module
+  End Namespace
```
CyrusNajmabadi, line 160: note: i strongly prefer this, as i think it makes things much simpler, and avoids problems listed below (like the scoping confusino issues).  IMO, a good syntax for this would simply be:

```c#
public static implicit class Whatever
{
}
```

Where all that changes from normal C# is the addition of the 'implicit' keyword, which then gives hte meaning that the accessible members of  the static class are in scope when the namepsace is pulled in.

Note that this is really just giving *normal* static class members the same behavior as *class extension methods*.  

The semantics are also really simple.  When you do `using N;`  that is then equivalent:

```c#
using N;
using static N.StaticClass1;
using static N.StaticClass2;
// etc.
```

Being able to have trivial keyword like this, which then maps to extremely trivial concepts, is highly appealing to me for simplicity of design and understanding.  It also makes it easy to update existing code to get this new behavior.

On the declaration side, things barely change.  And you get all the benefits on the consumption side.

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+
+- Some members can be declared directly in a namespace (file-scoped or block-scoped).
+  - Allowed kinds currently are: methods, operators, extension blocks, and fields.
+  - Existing declarations like classes still work the same, there shouldn't be any ambiguity.
+  - There is no ambiguity with top-level statements because those are not allowed inside namespaces.
+
+- Top-level members in a namespace are semantically members of an "implicit" class which:
+  - is `static` and `partial`,
+  - has accessibility either `internal` (by default) or `public` (if any member is also `public`),
+  - is named `TopLevel` (can be addressed even from C# which is useful for extension member disambiguation or source-generated `partial` implementations),
```
jaredpar, line 63: Is the only reason for this having a speakable name so that older C# compilers can access the members? For C# compilers aware of this feature I assume they can disambiguate by using the qualified name of `MyLib.MyNamespace.Print`. 

If this is the case I think we should consider a design where this is an unspeakable name. Having it be speakable creates some edges, like you can't name types `TopLevel`.

proposals/top-level-members.md
```diff
@@ -0,0 +1,216 @@
...
+  - is synthesized per each compilation unit (so having top-level members in the same namespace across assemblies can lead to [ambiguities](#drawbacks)).
```
jaredpar, line 64: There is one `TopLevel` per assembly or one per namespace that declares a top level member?

proposals/top-level-members.md
jjonescz: > and those must include one of the containing type.

That seems no longer true with extension operators.

proposals/top-level-members.md
jjonescz: Not just older compilers but also other languages like VB and F#.

proposals/top-level-members.md
jjonescz: One per namespace (its full name is `Namespace.TopLevel`), will clarify, thanks.

proposals/top-level-members.md
jjonescz: FWIW, members in VB modules are also implicitly static, no need to declare them `Shared`.

But if we go with "explicit" `implicit static class` declarations instead, I agree requiring `static` is a no-brainer.

Issue #9754, opened 2025-10-13
[Proposal]: Immediately Enumerated Collection Expressions


RikkiGibson:
* Specification: https://github.com/dotnet/csharplang/blob/main/proposals/immediately-enumerated-collection-expressions.md
* Discussion: #9755

## Summary
[summary]: #summary

Permit collection expressions in "immediately enumerated" contexts, without requiring a target type.

```cs
foreach (bool b in [true, false])
{
    doMyLogicWith(b);
}

string[] items1 = ["a", "b", .. ["c"]];
string[] items2 = ["a", "b", .. includeRest ? ["c", "d"] : []];
```

See also #9739.

## Design meetings

* https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-10-13.md#immediately-enumerated-collection-expressions
* https://github.com/dotnet/csharplang/blob/main/meetings/2026/LDM-2026-01-12.md#immediately-enumerated-collection-expressions

PR #9759, opened 2025-10-13
Revisions to immediately-enumerated-collection-expressions.md

RikkiGibson:
#9754

Have addressed notes from LDM on Monday 13th October.
Adjusted the definition of *best common element type* to be more clear.
Adjusted the spec so that spread cases like `string?[] arr = [.. [null]];` can work, using the element type from the conversion of the containing collection expression.

### DIFF
--- a/proposals/immediately-enumerated-collection-expressions.md
+++ b/proposals/immediately-enumerated-collection-expressions.md
@@ -104,7 +104,7 @@ This proposal doesn't get us 100% of the way there to "conditional element inclu
 string[] items2 = ["a", "b", .. includeRest ? ["c", "d"] : []];
 ```
 
-We are interested in pursuing type inference improvements which we expect to improve things across the board--for calls, foreach, and spreads--all in a similar way.
+We are interested in pursuing type inference improvements which we expect to improve things across the board—for calls, foreach, and spreads—all in a similar way.
 ```cs
 // Make all of the following work using a future type inference improvement:
 M1(cond ? [1] : [2]);


proposals/immediately-enumerated-collection-expressions.md
```diff
@@ -26,155 +25,103 @@ Choosing a single "natural type" for collection expressions in the general case,
...
-
 ### foreach
 
 Given a foreach statement of the form:
 ```cs
 foreach (iteration_type iteration_variable in collection)
     embedded_stmt
 ```
-- When `collection` lacks a natural type, we determine if a *collection expression iteration conversion* exists, from `collection` to type `TElem[]`, and apply the conversion if it exists.
+- When `collection` lacks a natural type, we determine if a *collection expression iteration conversion* exists, from `collection` to type `IEnumerable<TElem>`, and apply the conversion if it exists.
```
jcouv, line 38: The language is a bit confusing. Why not simply say that if `collection` is a *collection expression* in that position, then it has type `IEnumerable<TElem>` (where `TElem` is determined by rules below)?

We have precedent for expressions that have a type but also can be converted to a target-type: switch expressions.

Here's the [language](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md#switch-expression) for those:

`The type of the switch_expression is the best common type (§12.6.3.15) of the expressions appearing to the right of the => tokens of the switch_expression_arms if such a type exists and the expression in every arm of the switch expression can be implicitly converted to that type. In addition, we add a new switch expression conversion, which is a predefined implicit conversion from a switch expression to every type T for which there exists an implicit conversion from each arm's expression to T.`

jcouv: In addition, maybe we can define the determination of the type by relying on method type inference, like [best common type](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions) does.

> Note: Intuitively this inference is equivalent to calling a method `void M<X>(X x₁ ... X xᵥ)` with the `Eᵢ` as arguments and inferring `X`. end note

We'd say that intuitively this inference is equivalent to calling `void<T>(IEnumerable<T> x)` with `[X1, ...Xn]`.

RikkiGibson: > The language is a bit confusing. Why not simply say that if `collection` is a _collection expression_ in that position, then it has type `IEnumerable<TElem>` (where `TElem` is determined by rules below)?

This language was written with the assumption that the type inference improvement allowing element type info to flow through subexpressions would come at the same time. So, `foreach (var x in cond ? [1] : [2])` was meant to be valid.

Today, we could definitely spec this as a special case for when the collection or spread value is a collection-expression.

I think your suggestion is, that we could just spec the simple case, which is all we need to implement this feature by itself. Then when we do the type inference improvement, we could come back and expand this language, to *also* be able to convert more expressions to `IEnumerable<T>` under certain conditions? (which should cause subexpressions in `cond ? [1] : [2]` to be converted to `IEnumerable<T>`.)

jcouv: > Then when we do the type inference improvement, we could come back and expand this language, to also be able to convert more expressions to IEnumerable<T> under certain conditions?

No, I'm fine with a forward looking spec, so that improving type inference automatically makes things fall out here.

The part I'm reacting to is the first line: "When `collection` lacks a natural type, we determine if a *collection expression iteration conversion* exists, from `collection` to type `IEnumerable<TElem>`, and apply the conversion if it exists."
I think we should just say: "When `collection` is a collection expression, it's type is `IEnumerable<TElem>`". 
I'm not sure that a conversion is involved (in the language at least).

proposals/immediately-enumerated-collection-expressions.md
```diff
@@ -26,155 +25,103 @@ Choosing a single "natural type" for collection expressions in the general case,
 
 However, we do see a significant amount of demand for the ability to use collection expressions in contexts where the collection is immediately enumerated, and the collection value itself is not directly observable in user code. In this case, we should be able to define a solution which is convenient, optimal, and relatively low risk.
 
-Today, we believe that users are using `new[] { ... }` or `(T[])[...]` instead of `[...]`, when possible for "immediately enumerated"/"conditional element inclusion" cases. Today this results in sub-optimal allocation behavior, or verbosity resulting from the limitations of type inference.
```
jnm2, line 29: It seems like this could be applicable for immediately enumerated collections even without conditions. E.g.:

```cs
foreach (var item in new[] { a, b, c })
```

proposals/immediately-enumerated-collection-expressions.md
jnm2: The current specification specifies how the construction works as a conversion: <https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md#conversions>

I'm not sure I understand how the specification works if there is no conversion; how do we specify the construction for the collection expression then?

Rather than introducing a separate new conversion, would it be useful to expand the existing _collection expression conversion_ by appending to this list?

> The following additional implicit conversions exist from a *collection expression*:
> 
> * To a *nullable value type* `T?` where there is a *collection expression conversion* from the collection expression to a value type `T`. The conversion is a *collection expression conversion* to `T` followed by an *implicit nullable conversion* from `T` to `T?`.
> 
> * To a reference type `T` where there is a *[create method](#create-methods)* associated with `T` that returns a type `U` and an *implicit reference conversion* from `U` to `T`. The conversion is a *collection expression conversion* to `U` followed by an *implicit reference conversion* from `U` to `T`.
> 
> * To an interface type `I` where there is a *[create method](#create-methods)* associated with `I` that returns a type `V` and an *implicit boxing conversion* from `V` to `I`. The conversion is a *collection expression conversion* to `V` followed by an *implicit boxing conversion* from `V` to `I`.

proposals/immediately-enumerated-collection-expressions.md
```diff
@@ -26,155 +25,103 @@ Choosing a single "natural type" for collection expressions in the general case,
...
+### Future considerations
+
+This proposal doesn't get us 100% of the way there to "conditional element inclusion" scenarios like the following:
+
+```cs
+// The spread value is erroneous even after this proposal
+string[] items2 = ["a", "b", .. includeRest ? ["c", "d"] : []];
+```
+
+We are interested in pursuing type inference improvements which we expect to improve things across the board--for calls, foreach, and spreads--all in a similar way.
```
jnm2: em dashes (just because)
```suggestion
We are interested in pursuing type inference improvements which we expect to improve things across the board—for calls, foreach, and spreads—all in a similar way.
```

jnm2 (approved):

proposals/immediately-enumerated-collection-expressions.md
RikkiGibson: > I think we should just say: "When `collection` is a collection expression, it's type is `IEnumerable<TElem>`".
> I'm not sure that a conversion is involved (in the language at least).

So, I associate the suggestion to scope the rule to *When `collection` is a collection expression*, with a suggestion to handle only the "base cases", i.e. `foreach (var x in [1,2,3])` or `[.. [1,2,3]]`. The reason is, I think it prevents us from handling cases where `collection` itself is not a collection expression, but it does *contain* collection expressions, i.e. the `foreach (var x in flag ? [1] : [2])` case.

The idea of phrasing it as a conversion was so that it could be pushed down into the subexpressions of `?:`/`switch` or similar, where the collection expressions actually are.

I am unsure what amount of difference there is, for the purposes of the spec, between saying "the expression's type is T under these conditions" versus "a conversion from the expression to T exists and is applied under these conditions". Perhaps there is none, and that is the point?

Another part of why I phrased this as "when it lacks a natural type, then we see if this conversion exists and apply it", was, I thought it would be confusing to try to phrase it as "when it lacks a type, its type is actually `IEnumerable<TElem>`. So, I think there would be a need to find something besides "`collection` lacks a natural type", or "`collection` is a collection expression", as conditions for when to do this.

proposals/immediately-enumerated-collection-expressions.md
RikkiGibson: Yes, I can restore subset of this point. I mainly didn't feel like editing out the conditional element inclusion bit by itself.

proposals/immediately-enumerated-collection-expressions.md
jnm2: Here's where I'm stuck. How do we even have a type in hand at this point, where we can say "the type is `IEnumerable<X>`" here? ~~`X` is only determined inside the existing collection expression conversion.~~ It's determined inside method type inference.

RikkiGibson: > It's determined inside method type inference.

Right, it's determined using *best common element type*, which is a special form of type inference, which is functionally equivalent to the type inference for `M(collection)` in `void M<T>(IEnumerable<T> param) { }`.

jcouv: Sorry, it took me a while to grok something. I was thinking we're giving the collection expression a natural type, but that doesn't work in the future case of `foreach (... in (b ? [e1] : [e2])`. 
So I think the proposed spec is good: it uses a conversion to a target-type with an inferred element type (when `var`) or a specific element type. That mechanism will extend when we improve type inference.

PR #9774, opened 2025-10-21
Extensions: disallow readonly members

jcouv:


proposals/csharp-14.0/extensions.md
```diff
@@ -169,6 +169,7 @@ public static class Enumerable
 
 It is an error to specify the following modifiers on a member of an extension declaration:
 `abstract`, `virtual`, `override`, `new`, `sealed`, `partial`, and `protected` (and related accessibility modifiers).  
+It is an error to specify the `readonly` modifier on a member of an extension declaration.
```
tannergooding, line 172: Just wondering, why disallow this? Are there plans to support it in the future?

Logically given a `struct S { public void M() { ... } public readonly void N() { ... } }`, then `M` has a hidden first parameter `ref S this` while `N` has a hidden first parameter `in S this`.

So the example given below "makes sense" in that it would cause the `getter` to be fine to use with some `readonly ulong bits` but the setter requires mutable.

Would we expect the author to instead define two groups one with `ref ulong bits` and a mutable setter and another with `in ulong bits` and the readonly getter? Does the compiler correctly "merge" these?

jcouv: > why disallow this? 

That functionality didn't ship. We'll have to come back to it

PR #9814, opened 2025-11-10
Add metadata rules for unsafe

333fred:
Expands rules for how metadata will be emitted for the new unsafe feature


proposals/unsafe-evolution.md
```diff
@@ -184,6 +184,43 @@ partial class C2
 
 For properties, `get` and `set/init` members can be independently declared as `unsafe`; marking the entire property as `unsafe` means that both the `get` and `set/init` members are unsafe.
 
+#### Metadata
+
+When an assembly is compiled with the new memory safety rules, it gets marked with `MemorySafetyRulesAttribute` (detailed below), filled in with `15` as the language version. This is a signal to
+any downstream consumers that any members defined in the assembly will be properly attributed with `RequiresUnsafeAttribute` (detailed below) if an `unsafe` context is required to call them.
+Any member in such an assembly that is not marked with `RequiresUnsafeAttribute` does not require an `unsafe` context to be called, regardless of the types in the signature of the member.
```
AaronRobinsonMSFT, line 191: It might be worth stating explicitly that not requiring a context is because that is the "legacy" behavior. Someone reading this in 5 years hopefully should think - "why is this not needed?!" #Closed

proposals/unsafe-evolution.md
```diff
@@ -184,6 +184,43 @@ partial class C2
...
+When a member is marked as `unsafe`, the compiler will synthesize a `RequiresUnsafeAttribute` application on the member in metadata. When a user-facing member marked as `unsafe` generates
+hidden members, such as an auto-property's backing field or get/set methods, both the user-facing member and any hidden members generated by that user-facing member are all marked as `unsafe`,
+and `RequiresUnsafeAttribute` is applied to all of them.
+
+```cs
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>Indicates the language version of the memory safety rules used when the module was compiled.</summary>
+    [AttributeUsage(AttributeTargets.Module, Inherited = false)]
+    public sealed class MemorySafetyRulesAttribute : Attribute
```
AaronRobinsonMSFT, line 202: Will this play well with other languages? Suppose that C# 15 adds new safety rules. Now also suppose that F# adds safety rules in version 8. Does Roslyn understand which langauge version the number is related to? Are attributes under `System.Runtime.CompilerServices` assumed to be limited to C#?

proposals/unsafe-evolution.md
333fred: I honestly don't understand the suggestion.

proposals/unsafe-evolution.md
333fred: It's based on the `RefSafetyRulesAttribute` scheme, which also used language version. Perhaps @jaredpar or @RikkiGibson could comment on that.

proposals/unsafe-evolution.md
AaronRobinsonMSFT: Boo. I think I misinterpreted this statement when I commented. I reread it and it reads correctly.

proposals/unsafe-evolution.md
```diff
@@ -184,6 +184,43 @@ partial class C2
...
+    {
+        /// <summary>Initializes a new instance of the <see cref="MemorySafetyRulesAttribute"/> class.</summary>
+        /// <param name="version">The language version of the memory safety rules used when the module was compiled.</param>
+        public MemorySafetyRulesAttribute(int version) => Version = version;
+ 
+        /// <summary>Gets the language version of the memory safety rules used when the module was compiled.</summary>
+        public int Version { get; }
+    }
+
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Delegate | AttributeTargets.Constructor, AllowMultiple = false, Inherited = true)]
```
jkotas, line 212:
```suggestion
    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor, AllowMultiple = false, Inherited = true)]
```
As discussed offline, we may want to mark delegate `Invoke` method as unsafe so that delegate types are not special. ` AttributeTargets.Delegate` is not needed in that case.

proposals/unsafe-evolution.md
jjonescz: It's documented here: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#refsafetyrulesattribute. The version is actually ignored, but it could be considered in the future if there was another "ref safety rules" version.

AaronRobinsonMSFT: @jjonescz Thanks. I don't see anything specific about the attribute and other languages though. It seems like it is implicitly a C# only attribute. This gets back to my question of, what if F# has their own ref-safety rules? or another .NET compiler decides to express their own memory safety rules?

I'm not saying we should throw it all away or even add a new API or change the existing one. I think I would like to see some clear documentation on the API though say "is specifically for Roslyn" or "is specifically for C#". I'm not going to quibble about potentially other non-Roslyn C# compilers.

RikkiGibson: The attribute is essentially for C#. The version number in the attribute is always a C# major language version. Languages that want to interop with C#'s unsafe code, will need to recognize and understand it, and possibly even emit it themselves if they want their "unsafe methods", etc. to be understood and treated similarly as C#'s.

> This gets back to my question of, what if F# has their own ref-safety rules? or another .NET compiler decides to express their own memory safety rules?

Other languages/compilers can simply be oblivious to this attribute. That is the default state of affairs. If another language decided to implement a divergent concept of `unsafe`, then, we would likely remain in that situation of: we are oblivious to their unsafe mechanics, and they are oblivious to ours. That seems like something we should avoid, and also something which is unlikely to arise inadvertently.

In the past, F# did some work to check ref safety for ref structs, etc. It doesn't look like they recognize RefSafetyRulesAttribute, though, they might wish to do so in the future. That's the nature of the impact I would anticipate on other major .NET languages for this feature as well.

AaronRobinsonMSFT: > The attribute is essentially for C#. 

Then that should be documented cleared on the API.

proposals/unsafe-evolution.md
```diff
@@ -184,6 +184,43 @@ partial class C2
...
+    public sealed class RequiresUnsafeAttribute : Attribute
+    {
+    }
+}
+```
+
+#### Compat mode
```
jcouv, line 219: I think this compat behavior is fine to reduce virality of the feature, but we should consider offering a facility to tighten our enforcement.
The scenario is: I'm enabling the new rules for my project but want to ensure my references all using the new rules too. 
Maybe that's an analyzer or maybe that's a different value to the project-level flag that turns on the new rules.

proposals/unsafe-evolution.md
jjonescz, line 150: Unrelated to this PR, but:

> `unsafe` on a member will mean that that member is `unsafe`, and the body of that member is considered an `unsafe` context.

Was it considered to change this so that `unsafe` on a member has no impact on the body? That's what rust did and their motivation seems to apply to us too (reducing unsafe code): https://github.com/rust-lang/rfcs/blob/master/text/2585-unsafe-block-in-unsafe-fn.md#motivation

333fred: Yes, we discussed this heavily in the most recent LDM. It's not currently in the plan, but we haven't ruled it out.

proposals/unsafe-evolution.md
333fred: That feels like the job of either an analyzer or even an SDK build task, imo.

jcouv: Agreed, I'm not saying it would be implemented in the compiler.

In LDM last week, we discussed the need to review the full picture with all the pieces: compiler+analyzers+guidelines+SDK to see how it all fits. 
If this spec is only meant to cover the language/compiler portion, where would be the overarching design at?

proposals/unsafe-evolution.md
jcouv: nit: I assume you're tracking those offline, but consider adding an open issues section to this doc.

proposals/unsafe-evolution.md
333fred: @agocke's doc [here](https://github.com/dotnet/designs/blob/main/proposed/caller-unsafe.md), also linked in this language proposal.

proposals/unsafe-evolution.md
333fred: It will be added when I get the notes from yesterday typed up :slightly_smiling_face:.

proposals/unsafe-evolution.md
jcouv: @agocke Either you or JanK at LDM last week mentioned that the language rules would not be sufficient by themselves to deliver on the memory safety claim we're trying to offer as part of the feature, and we may need analyzers and coding guidelines to shore up some gaps.

Will you be on point to update the overall design doc with things that fall outside of language/compiler?
- coding guidelines for proper usage of the feature
- analyzers or additional SDK tooling

The reason I'm asking is I'd like to capture a open question: if we design the language/compiler for more compatibility (older compilers and not-opted-into-new-rules compiler are oblivious to new "unsafecaller" modifier) then should some other part of the system allow the user to crank the enforcement to "strict"?
The scenario is: I'm enabling the new rules for my project but want to ensure my references all using the new rules too (no "oblivious" code here).

Before discussing how this would be implemented, my question is whether that's a scenario we're trying to support.

proposals/unsafe-evolution.md
jcouv: @333fred What was the resolution here? `AttributeTargets.Delegate` is still listed. Is this tracked as an open issue for later?

PR #9826, opened 2025-11-17
Add proposal for async main codegen update

333fred:


AlekseyTs (approved):
LGTM (commit 2)

proposals/async-main-update.md
```diff
@@ -0,0 +1,63 @@
...
+follows:
+
+```diff
+- Otherwise, the synthesized entry point waits for the returned task to complete, calling `GetAwaiter().GetResult()` on the task, using either the parameterless instance method or the extension method described by [§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271). If the task fails, `GetResult()` will throw an exception, and this exception is propagated by the synthesized method.
++ Otherwise, the synthesized entry point waits for the returned task to complete, either passing the task to `System.Runtime.CompilerServices.AsyncHelpers.HandleAsyncEntryPoint` (if it exists) or calling `GetAwaiter().GetResult()` on the task, using either the parameterless instance method or the extension method described by [§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271). If the task fails, the calling method form will throw an exception, and this exception is propagated by the synthesized method.
+```
+
+The compiler will look for the following APIs from the core libraries. We do not look at implementations defined outside the core library:
```
jnm2, line 22: How is the determination in the second sentence accomplished?
Also, why add this restriction?

proposals/async-main-update.md
```diff
@@ -0,0 +1,63 @@
...
+public class C
+{
+    public static async Task Main()
+    {
+        await System.Threading.Tasks.Task.Yield();
+        Console.WriteLine("Hello world");
+    }
+}
+
+// Lowered psuedocode, not including async state machine if present
```
jnm2:
```suggestion
// Lowered pseudocode, not including async state machine if present
```

proposals/async-main-update.md
333fred: The determination is done in the same way all core libraries checks are made.

The restriction is there because this is a key runtime scenario. Looking for user-defined handling isn't relevant.

jnm2: I'm not able to find an example of the C# specification stating that any of the standard library APIs must not be found in third-party references. Do you know of an example of such an API where the compiler already has that requirement?

333fred: All of the special types, anything coming from runtime async, all of the first-class span feature work, all of `string.Concat`, anything to do with `decimal`... anything in [this](https://github.com/dotnet/roslyn/blob/670884f61a26150270e41dd281ad2df70e4c1302/src/Compilers/Core/Portable/SpecialMember.cs) file must come from corelib.

333fred: > any of the standard library APIs must not be found in third-party references

Nothing about the specification or this proposal states that other references cannot define methods. It just states that we won't use it.

jnm2: > Nothing about the specification or this proposal states that other references cannot define methods. It just states that we won't use it.

What I was trying to say here is that I can't find where the C# standard states that. Do you know where that is?

333fred: I believe this would be the difference between C.2 and C.3.

jnm2: Where does it define rules about locating the C.2 or C.3 APIs in certain assemblies in certain ways that would _require_ the compiler to implement a distinction between a core library or libraries and other references? I see in these sections "if you're making a standard library for C#, it must have at least this." I'm not seeing something that says the compiler should care which assembly it finds an API in.

333fred: The top of the C annex states:

> A conforming C# implementation shall provide a minimum set of types having specific semantics. These types and their members are listed here, in alphabetical order by namespace and type. For a formal definition of the types and their members identified in ([§C.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)), refer to ISO/IEC 23271:2012 Common Language Infrastructure (CLI), Partition IV; Base Class Library (BCL), Extended Numerics Library, and Extended Array Library, which are included by reference in this specification. For a list of types and their members required beyond those identified in [§C.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/standard-library.md#c2-standard-library-types-defined-in-isoiec-23271), see [§C.3](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271).

jnm2: The current C# implementation depends on a reference being given from the SDK as a way to provide that API set, and thus it errors if these APIs are not present, right? In what way does this text require that the APIs came from a particular _one_ of the references passed to the compiler from the SDK? If the compiler can "provide" one of these APIs by getting it from an SDK-provided reference, why can't it "provide" one of these APIs by getting it from a third-party-provided reference?

333fred: This text requires that these types come from the BCL. I don't see any other way of reading it.

jnm2: I see it saying that the compiler shall provide the APIs in both C.2 and C.3, where only C.2 is related to something called Base Class Library published in ISO/IEC 23271:2012. It seems presented explicitly as happenstance that some of these are in C.2 and some are in C.3, with an explicit reference to this distinction shifting if a potential newer version of ISO/IEC 23271:2012 is produced. I don't see the process defined anywhere for a compiler to consider some references to be BCL references and other references not to be.

333fred: The compiler does not provide any of these APIs. The APIs defined in C.2 must come from the BCL because they are defined in the BCL, as per "For a formal definition of the types and their members identified in ([§C.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)), refer to ISO/IEC 23271:2012 Common Language Infrastructure (CLI), Partition IV; Base Class Library (BCL), Extended Numerics Library, and Extended Array Library, which are included by reference in this specification."

proposals/async-main-update.md
```diff
@@ -0,0 +1,63 @@
...
+
+The compiler will look for the following APIs from the core libraries. We do not look at implementations defined outside the core library:
+
+```cs
+namespace System.Runtime.CompilerServices;
+
+public class AsyncHelpers
+{
+    public static void HandleAsyncEntryPoint(System.Threading.Tasks.Task task);
+    public static void HandleAsyncEntryPoint(System.Threading.Tasks.Task<int> task);
```
jjonescz:
```suggestion
    public static int HandleAsyncEntryPoint(System.Threading.Tasks.Task<int> task);
```

proposals/async-main-update.md
jnm2: I'm seeing a "this is how it already is" but no requirements or instructions for an implementation. I'm not sure how to bring this thread home, so I'll let it go. I don't think this will affect language users, only potentially those who are incorporating the proposals into the standard.

PR #9831, opened 2025-11-19
More `unsafe` spec updates

333fred:
Add a few more open questions and adjust the wording with decisions from the [12th](https://github.com/dotnet/csharplang/blob/5d9c7bec7677a6ce335652712b9fe483680ef487/meetings/2025/LDM-2025-11-12.md).

### DIFF
--- a/proposals/unsafe-evolution.md
+++ b/proposals/unsafe-evolution.md
@@ -135,14 +135,14 @@ is updated to include whether the _anonymous function_ has the `unsafe` keyword,
 it would be if any parameter were a by-`ref`, optional, or `params`.
 
 It is a memory safety error convert a delegate type that is marked as `unsafe` to `System.Delegate`/`System.Linq.Expressions.Expression`/`System.Linq.Expressions.Expression<T>`, or any interface those
-types implement or base type of those types.
+types implement or base type of those types. They also cannot be used as type parameters.
 
 A delegate type that is marked `unsafe` can only be invoked in an `unsafe` context. If a delegate type is `unsafe`, then its `Invoke`, `BeginInvoke`, and `EndInvoke` methods are marked as `unsafe`.
 
 > [!NOTE]
 > We don't actually attribute the delegate type itself, just the `Invoke`, `BeginInvoke`, and `EndInvoke` methods. Determining whether a delegate type is `unsafe` is done by examining those 3 methods.
 > If all are marked as `unsafe`, the delegate type is considered `unsafe`. If only some are marked as `unsafe`, then it is presumed that calling the others is safe and only calling the member that is
-> marked as `unsafe` will cause a memory safety error. It will be a memory safety error to convert an `unsafe` lambda or method group to a delegate type that is does not have all of `Invoke`, `BeginInvoke`,
+> marked as `unsafe` will cause a memory safety error. It will be a memory safety error to convert an `unsafe` lambda or method group to a delegate type that does not have all of `Invoke`, `BeginInvoke`,
 > and `EndInvoke` marked as `unsafe`.
 
 ### `extern`
@@ -231,12 +231,13 @@ for various parts of the ecosystem, particularly any enumerables that are passed
 ### Delegate type `unsafe`ty
 
 We could remove the ability to make delegate types as `unsafe` entirely, and simply require that all conversions of `unsafe` lambdas or method groups to a delegate type occur inside an `unsafe` context.
-This could simplify the model around `unsafe` in C#, but at the risk of forcing `unsafe` annotations in the wrong spot and having an area where the real area of `unsafe`ty isn't properly called out.
+This could simplify the model around `unsafe` in C#, but at the risk of forcing `unsafe` annotations in the wrong spot and having an area where the real area of `unsafe`ty isn't properly called out. There
+are a lot of corner cases here, particularly involving generics and conversions, so it may be better to simply leave the concept for later when we determine it's needed.
 
 ### Lambda/method group natural types
 
-Today, the only real impact on codegen (besides additional metadata) is changing the *function_type* of a lambda or method group when `unsafe` is in the signature. If we were to avoid doing this, then
-there would be no real impact to codegen, which could give adopters more confidence that behavior has not subtly changed under the hood.
+Today, the only real impact on semantics and codegen (besides additional metadata) is changing the *function_type* of a lambda or method group when `unsafe` is in the signature. If we were to avoid doing this, then
+there would be no real impact to either, which could give adopters more confidence that behavior has not subtly changed under the hood.
 
 ### `stackalloc` as initialized
 


proposals/unsafe-evolution.md
```diff
@@ -136,7 +136,13 @@ it would be if any parameter were a by-`ref`, optional, or `params`.
 
 It is a memory safety error convert a delegate type that is marked as `unsafe` to `System.Delegate`/`System.Linq.Expressions.Expression`/`System.Linq.Expressions.Expression<T>`.
```
jkotas: Should this include `object` as well? Otherwise, one can do `unsafe delegate` -> object -> safe delegate.

What can one actually do with the unsafe delegates with all these limitations? #Resolved

proposals/unsafe-evolution.md
```diff
@@ -234,6 +227,16 @@ anonymous and local functions do not keep the unsafe context of their containing
 Is conversion of a lambda or method group marked `unsafe` to a non-unsafe delegate type permitted without warning or error in an `unsafe` context? If we don't do this, then it could be fairly painful
 for various parts of the ecosystem, particularly any enumerables that are passed through LINQ queries.
 
+### Delegate type `unsafe`ty
+
+We could remove the ability to make delegate types as `unsafe` entirely, and simply require that all conversions of `unsafe` lambdas or method groups to a delegate type occur inside an `unsafe` context.
+This could simplify the model around `unsafe` in C#, but at the risk of forcing `unsafe` annotations in the wrong spot and having an area where the real area of `unsafe`ty isn't properly called out.
```
jkotas: If we started without unsafe delegates, can they be added later as an incremental non-breaking change once/if proven necessary? #Resolved

333fred: Yes, I expect so.

proposals/unsafe-evolution.md
333fred: > What can one actually do with the unsafe delegates with all these limitations?

Pass them around and call them in an `unsafe` context. I don't know how much people are generally counting on converting them `object` or similar; I certainly don't use that functionality.

proposals/unsafe-evolution.md
```diff
@@ -134,9 +134,16 @@ It is a memory safety error to convert an `unsafe` member to a delegate type tha
 is updated to include whether the _anonymous function_ has the `unsafe` keyword, or the _method group_ is to a member that is marked `unsafe`. If it is, an anonymous function type is created, just as
 it would be if any parameter were a by-`ref`, optional, or `params`.
 
-It is a memory safety error convert a delegate type that is marked as `unsafe` to `System.Delegate`/`System.Linq.Expressions.Expression`/`System.Linq.Expressions.Expression<T>`.
+It is a memory safety error convert a delegate type that is marked as `unsafe` to `System.Delegate`/`System.Linq.Expressions.Expression`/`System.Linq.Expressions.Expression<T>`, or any interface those
```
jkotas, line 137: Does it mean that the unsafe delegates cannot be used as generic arguments (in safe contexts)?

If they can be used as generic arguments, one can do the offending memory unsafe cast indirectly by doing e..g.:
```
IEnumerable<UnsafeDelegate> e1 = ...!;

IEnumerable<Delegate> e2 = e1.Cast<Delegate>();
```

proposals/unsafe-evolution.md
jkotas: Right, pass them around on your little island. I do not expect libraries are generally going to have APIs that take unsafe delegate arguments.

To me, it feels like that the unsafe delegates are a mine field that will require a lot of special casing and provide very little value. Starting without them may be better.

proposals/unsafe-evolution.md
```diff
@@ -134,9 +134,16 @@ It is a memory safety error to convert an `unsafe` member to a delegate type tha
...
+types implement or base type of those types.
 
-A delegate type that is marked `unsafe` can only be invoked in an `unsafe` context. If a delegate type is `unsafe`, then its `Invoke`, `BeginInvoke`, and `EndInvoke` methods are also marked as `unsafe`.
+A delegate type that is marked `unsafe` can only be invoked in an `unsafe` context. If a delegate type is `unsafe`, then its `Invoke`, `BeginInvoke`, and `EndInvoke` methods are marked as `unsafe`.
+
+> [!NOTE]
+> We don't actually attribute the delegate type itself, just the `Invoke`, `BeginInvoke`, and `EndInvoke` methods. Determining whether a delegate type is `unsafe` is done by examining those 3 methods.
+> If all are marked as `unsafe`, the delegate type is considered `unsafe`. If only some are marked as `unsafe`, then it is presumed that calling the others is safe and only calling the member that is
+> marked as `unsafe` will cause a memory safety error. It will be a memory safety error to convert an `unsafe` lambda or method group to a delegate type that is does not have all of `Invoke`, `BeginInvoke`,
```
jjonescz:
```suggestion
> marked as `unsafe` will cause a memory safety error. It will be a memory safety error to convert an `unsafe` lambda or method group to a delegate type that does not have all of `Invoke`, `BeginInvoke`,
```

proposals/unsafe-evolution.md
```diff
@@ -146,39 +153,26 @@ value cannot be safely called by C#, as the calling convention used for the meth
...
 
 Today, as covered by the [unsafe context specification][unsafe-context-spec], `unsafe` behaves in a lexical manner, marking the entire textual body contained by the `unsafe` block as an `unsafe` context
-(except for iterator bodies). We propose changing this definition from textual to sematic. `unsafe` on a type will now mean that all members declared by that type are considered `unsafe`, and all of the
-member bodies of that type are considered an `unsafe` context. `unsafe` on a member will mean that that member is `unsafe`, and the body of that member is considered an `unsafe` context. For existing code
-moving to the new definition of `unsafe`, this may produce a number of false positives for methods that don't need to be considered `unsafe`; we believe this better than false positives around not doing
-this, or making it an error to put `unsafe` on a type which would easily be the largest breaking change that we've ever introduced in C#.
+(except for iterator bodies). We propose changing this definition from textual to sematic. `unsafe` on a member will mean that that member is `unsafe`, and the body of that member is considered an `unsafe`
```
jjonescz, line 156: Regarding the syntactical meaning of unsafe on bodies, there is no change, right? So for example, iterators still escape out of the unsafe context.

proposals/unsafe-evolution.md
```diff
@@ -234,6 +228,16 @@ anonymous and local functions do not keep the unsafe context of their containing
...
 for various parts of the ecosystem, particularly any enumerables that are passed through LINQ queries.
 
+### Delegate type `unsafe`ty
+
+We could remove the ability to make delegate types as `unsafe` entirely, and simply require that all conversions of `unsafe` lambdas or method groups to a delegate type occur inside an `unsafe` context.
+This could simplify the model around `unsafe` in C#, but at the risk of forcing `unsafe` annotations in the wrong spot and having an area where the real area of `unsafe`ty isn't properly called out.
+
+### Lambda/method group natural types
+
+Today, the only real impact on codegen (besides additional metadata) is changing the *function_type* of a lambda or method group when `unsafe` is in the signature. If we were to avoid doing this, then
```
jjonescz: This is not impacting just codegen, it's impacting semantics too since different delegate types cannot be converted to each other, so for example this snippet will stop working:

```cs
unsafe void M() { }
var x = M;
System.Action y = x;
```

proposals/unsafe-evolution.md
333fred: There's an open question on that.

333fred: There's an open question on that.

proposals/unsafe-evolution.md
jcouv: @333fred Let's capture this in open issues

jkotas: Would it be better to postpone the whole unsafe delegate idea until we have evidence that it is actually worth it to deal with all the complex corner cases that it comes with?

PR #9868, opened 2025-12-11
Update section on unsafe modifier

jjonescz:


proposals/unsafe-evolution.md
```diff
@@ -180,7 +188,9 @@ partial class C1
 }
 ```
 
-For properties, `get` and `set/init` members can be independently declared as `unsafe`; marking the entire property as `unsafe` means that both the `get` and `set/init` members are unsafe.
+For properties, `get` and `set/init` accessors can be independently declared as `unsafe`; marking the entire property as `unsafe` means that both the `get` and `set/init` accessors are unsafe.
+It is currently not possible to place any modifiers on event accessors, and this proposal doesn't change that, i.e., `add` and `remove` event accessors cannot be independently declared as `unsafe`.
+Only if the entire event is marked as `unsafe`, it means that the accessors are unsafe; otherwise they are safe.
```
jcouv, line 193: nit: Did you also want to capture question about allowing `unsafe` modifier on lambdas?

jcouv (approved):

PR #9869, opened 2025-12-11
Extension indexers: flesh out the spec

jcouv:


proposals/extension-indexers.md
```diff
@@ -4,18 +4,210 @@
...
 
-Extension indexers declarations generally follow the rules for 
-non-extension [indexers in the Standard](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/classes.md#159-indexers).
+All rules from the C# standard that apply to ordinary indexers apply to extension indexers,
+but extension members do not have an implicit or explicit `this`.
+
+### `IndexerName` attribute
+
+`IndexerNameAttribute` may be applied to an extension indexer. The attribute is
+carried over to the extension member in metadata and its value determines
```
AlekseyTs: This feels somewhat misleading. It looks like the attribute isn't actually emitted, but "carried over" can be interpreted as it is. #Closed

jcouv:
@AlekseyTs for another look. Thanks

PR #9883, opened 2025-12-15
Clarify compat modes in unsafe evolution proposal

jjonescz:


proposals/unsafe-evolution.md
```diff
@@ -154,6 +154,14 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
 Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`. Even methods that only take `unmanaged` parameters by
 value cannot be safely called by C#, as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
 
+`extern` methods are considered `unsafe` when called from code using the updated rules regardless of the memory safety rules version of the callee.
+
+Only `extern` methods which are recognizable from metadata are handled, i.e., methods with the `pinvokeimpl` flag in IL.
+Therefore, `extern` methods without `DllImport` attribute are not recognized as implicitly `unsafe` when used from a different assembly.
+Such `extern` methods exposed as API should not be a common scenario anyway and there is already a warning for `extern` methods without any attribute
+(albeit absence of that warning doesn't guarantee the method will be recognized as `unsafe` in metadata).
+One can mark such methods as `unsafe` explicitly if needed.
```
333fred: Why would this be a caveat? The goal is that any `extern` method is `unsafe`, period.

jjonescz: extern without DllImport are not useful anyway I think and we would need to synthesize the RequiresUnsafe attribute for them. Which also means it would be different if such API was compiled by old or new compiler (whereas as currently proposed, even extern apis compiled by old compilers would be recognized in the same way).

333fred: Why would we need to synthesize anything for them? They're extern, they're just always unsafe.

333fred: Why would we need to synthesize anything for them? They're extern, they're just always unsafe.

jjonescz: Because we can't tell whether they are extern from metadata, that's what I'm trying to say by this PR.

jjonescz: Extern methods without any attribute look just like plain ordinary empty methods in IL.

jkotas: I think you should be able to tell by checking whether the method has IL implementation (method is implemented by IL).

If the method has no implementation in IL and it is not abstract (there may be a few more cases like that), it is extern -> unsafe.

jkotas: > extern without DllImport are not useful anyway 

extern without DllImport are used by https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute

jjonescz: You're right, it seems we can recognize extern methods by them having RVA 0, thanks.

jjonescz: Not sure what to do with this: `object.GetType` (and probably other methods) on .net framework is marked as `extern`, so it would be an error to use it outside `unsafe` under the new rules.

https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/mscorlib/system/object.cs#L105

333fred (dismissed):

proposals/unsafe-evolution.md
```diff
@@ -151,8 +151,15 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
...
+Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`
+if compiled under the updated memory safety rules (when it gets the `RequiresUnsafeAttribute`).
+Even methods that only take `unmanaged` parameters by value cannot be safely called by C#,
+as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+
+`extern` methods from assemblies using the legacy memory safety rules are not considered implicitly `unsafe` because that would be a large breaking change
+(for example, `object.GetType` is `extern` in .NET Framework [reference assemblies](https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/mscorlib/system/object.cs#L105)).
+Note that this is different from the [compat mode](#compat-mode) which applies to legacy-rules assemblies too
+because methods with pointers in signature would always need an unsafe context at the call site.
```
333fred, line 164: It will definitely be good to have tests covering `extern`s from a legacy assembly with pointers in their signature.

proposals/unsafe-evolution.md
```diff
@@ -151,8 +151,15 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
...
-value cannot be safely called by C#, as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`
+if compiled under the updated memory safety rules (when it gets the `RequiresUnsafeAttribute`).
+Even methods that only take `unmanaged` parameters by value cannot be safely called by C#,
+as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+
+`extern` methods from assemblies using the legacy memory safety rules are not considered implicitly `unsafe` because that would be a large breaking change
+(for example, `object.GetType` is `extern` in .NET Framework [reference assemblies](https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/mscorlib/system/object.cs#L105)).
```
jkotas: This link does not point to a reference assembly. It points to an implementation assembly sources.

jjonescz: Ah, good catch. Is there a way to get an URL to reference assembly though? The API reference doesn't seem to have this level of detail: https://learn.microsoft.com/en-us/dotnet/api/system.object.gettype?view=netframework-4.8.1#system-object-gettype

proposals/unsafe-evolution.md
```diff
@@ -151,8 +151,15 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
 
 ### `extern`
 
-Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`. Even methods that only take `unmanaged` parameters by
-value cannot be safely called by C#, as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`
+if compiled under the updated memory safety rules (when it gets the `RequiresUnsafeAttribute`).
+Even methods that only take `unmanaged` parameters by value cannot be safely called by C#,
+as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
```
jkotas, line 157: What do you think about the @agocke's proposal in the internal chat:

> My proposal is: the compiler doesn't look for extern for metadata symbols, it only looks for the [RequiresUnsafe] attribute, and when compiling with the new rules extern methods get annotated with [RequiresUnsafe].

I think it may be a cleaner design.

Alternatively, we can make `extern` methods safe by default and have analyzers that will warn if an extern method is not marked as `unsafe`. It is similar to how we are dealing with pointers in signatures. We do not make methods with a pointer in signature implicitly unsafe since there is a small chance that it is actually safe, and I expect that we are going to have analyzers that will warn you if the method with pointers in the signature is not marked as unsafe.

jjonescz: > My proposal is: the compiler doesn't look for extern for metadata symbols, it only looks for the [RequiresUnsafe] attribute, and when compiling with the new rules extern methods get annotated with [RequiresUnsafe].

I agree. My current proposal should be equivalent to that.

proposals/unsafe-evolution.md
jjonescz: I guess I can leave the link and update the description.

jkotas: The sources for .NET Framework reference assemblies are not on github as far as I know. If they were on github, Type.GetType would not be an extern in them. `extern` is treated as an implementation detail. It is not part of the public surface and thus it is not preserved in reference assemblies. Changing method back and forth between extern and non-extern is not changing public surface.

(`extern` happens to be preserved in reference assemblies produced by Roslyn, but that does not make it part of public surface.)

proposals/unsafe-evolution.md
```diff
@@ -151,8 +151,15 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
 
 ### `extern`
 
-Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`. Even methods that only take `unmanaged` parameters by
-value cannot be safely called by C#, as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`
+if compiled under the updated memory safety rules (when it gets the `RequiresUnsafeAttribute`).
```
jkotas: CoreLib exposes many extern methods (FCalls) as safe today. Treating extern methods as implicitly unsafe will require wrapping the implicitly unsafe extern methods with a safe wrapper. We may run into situations where adding the extra wrapper is difficult due to runtime implementation details. We can give it a try and revisit if necessary.

proposals/unsafe-evolution.md
jjonescz: > (`extern` happens to be preserved in reference assemblies produced by Roslyn, but that does not make it part of public surface.)

Okay, makes sense to me, then I guess the real reason we cannot do this is that `extern` is not actually guaranteed to be preserved in metadata of reference assemblies.

proposals/unsafe-evolution.md
```diff
@@ -151,8 +151,18 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
...
+Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`
+if compiled under the updated memory safety rules (i.e., it gets the `RequiresUnsafeAttribute`).
+Even methods that only take `unmanaged` parameters by value cannot be safely called by C#,
+as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+
+`extern` methods from assemblies using the legacy memory safety rules are not considered implicitly `unsafe` because
+- it is not guaranteed that `extern` will be always preserved in reference assemblies (i.e., compiler cannot reliably recognize `extern` methods from metadata), and
+- it would be a large breaking change (e.g., `object.GetType` currently appears as `extern` in .NET Framework reference assemblies
+  because `extern` is preserved in reference assemblies by current tooling from the [implementation assembly](https://github.com/microsoft/referencesource/blob/ec9fa9ae770d522a5b5f0607898044b7478574a3/mscorlib/system/object.cs#L105)).
```
jkotas: .NET Framework tooling for reference assemblies creates *all* methods as externs. If you run ildasm on "C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.8\mscorlib.dll", you should see that all methods have IL bodies stripped that maps to extern in C#.

jkotas:
```suggestion
`extern` methods from assemblies using the legacy memory safety rules are not considered implicitly `unsafe` because it is not guaranteed that `extern` will be always preserved in reference assemblies (i.e., compiler cannot reliably recognize `extern` methods from metadata). `extern` is considered implementation detail that is not part of public surface, and thus `extern` is not always preserved in reference assemblies.
```

proposals/unsafe-evolution.md
```diff
@@ -151,8 +151,18 @@ A delegate type that is marked `unsafe` can only be invoked in an `unsafe` conte
 
 ### `extern`
 
-Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`. Even methods that only take `unmanaged` parameters by
-value cannot be safely called by C#, as the calling convention used for the method could be incorrectly specified by the user and must be manually verified by review.
+Because `extern` methods are to native locations that cannot be guaranteed by the runtime, any `extern` method is automatically considered `unsafe`
+if compiled under the updated memory safety rules (i.e., it gets the `RequiresUnsafeAttribute`).
```
jkotas, line 155: Is it worth pointing out that this is the only place (so far) where the compiler produces `RequiresUnsafeAttribute` annotation without the user actually writing `unsafe` keyword? It feels like an outlier. Are we happy with that?

proposals/unsafe-evolution.md
jjonescz: That's not what I see - other methods have an empty body, but they still have a body (i.e., non-zero RVA):

image

jjonescz: FWIW, that is from ILSpy on `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.7.2\mscorlib.dll` (I don't seem to have v4.8 on this machine)

jkotas: > they still have a body (i.e., non-zero RVA)

Yes, you are right. It is a weird body. I think if one does ildasm/ilasm, on it, it won't roundtrip and it will turn into zero RVA. Still, I would simplify the discussion in this doc to "extern is not part of public surface" and end with that.

jjonescz: > Still, I would simplify the discussion in this doc to "extern is not part of public surface" and end with that.

Yep, I did just that. Thanks!

PR #9941, opened 2026-01-16
Create unsafe-alternative-syntax.md

MadsTorgersen:
This proposal amends and references [Unsafe evolution](https://github.com/dotnet/csharplang/blob/main/proposals/unsafe-evolution.md). 

It proposes directly using the `[RequiresUnsafe]` attribute to mark caller-unsafe members instead of repurposing the `unsafe` modifier.

### DIFF
--- a/proposals/unsafe-alternative-syntax.md
+++ b/proposals/unsafe-alternative-syntax.md
@@ -103,6 +103,12 @@ This means that the member's author doesn't get more fine-grained control over w
 
 **With this proposal** the `[RequiresUnsafe]` attribute does not imply that the member is an unsafe context. The author is free to mark the boundaries that make the most sense.
 
+### Meaningless uses of unsafe
+
+The original proposal introduces a warning for a few places where the `unsafe` modifier is allowed only for compat purposes, but no longer has an effect.
+
+**With this proposal** there is no need for such a warning, since the meaning of `unsafe` doesn't change.
+
 ## Detailed design
 
 - Keep the `unsafe` keyword with the same meaning in the same places as today: In every location it simply denotes a region of code as an unsafe context.


jcouv (approved):

proposals/unsafe-alternative-syntax.md
```diff
@@ -0,0 +1,147 @@
...
+
+This allows some freedom in the order that compilations are opted-in, while preserving a measure of safety when consuming assemblies that are not.
+
+The original proposal uses the presence of `[MemorySafetyRules]` in an assembly to determine whether it was annotated, which has some validity because if it didn't have the attribute it *couldn't* have been annotated. Essentially there's an assumption that when an assembly opts in it takes responsibility for appropriately annotating its members.
+
+In this alternative proposal, is it still reasonable to use `[MemorySafetyRules]` as the only indication that members have been properly annotated and compat mode should not be used? After all they could have been annotated with `[RequiresUnsafe]` even *without* the assembly having been opted in. Perhaps any assembly with at least one `[RequiresUnsafe]` should also be considered annotated?
```
jjonescz, line 152: > Perhaps any assembly with at least one `[RequiresUnsafe]` should also be considered annotated?

This doesn't sound like something that could be implemented in a performant fashion in the compiler (we would need to look at all members up front).

proposals/unsafe-alternative-syntax.md
```diff
@@ -0,0 +1,147 @@
...
+
+- Remove certain operations from the set of language-defined unsafe operations, exactly as in the original proposal. This takes effect unconditionally for the C# version where it is introduced.
+
+- Designate caller-unsafe members directly with the `[RequiresUnsafe]` attribute instead of repurposing the `unsafe` keyword. The attribute does *not* automatically make member bodies unsafe contexts; an explicit `unsafe` region is needed for that.
+
+- Introduce a compiler opt-in mechanism that controls whether caller-unsafe methods are prevented from being called outside of unsafe contexts. When opted-in, the assembly gets marked with `[MemorySafetyRules]` just as in the original proposal.
+
+- Consider augmenting `[RequiresUnsafe]` with arguments to distinguish new vs existing caller-unsafe members and regulate the impact of the opt-in mechanism on the members' allowed use.
+
+- Other aspects remain unchanged from the original proposal, e.g.:
```
jjonescz, line 124: We can probably also get rid of most of the "meaningless `unsafe`" warnings (from https://github.com/dotnet/csharplang/blob/79da45137622071d3c80ebe0a1296387af67afb9/proposals/unsafe-evolution.md#unsafe-modifiers-and-contexts).

Even with the original proposal, the warnings are a bit weird. Consider:

```cs
class B { public unsafe B() { } }
class C1 : B; // error: cannot use the unsafe base constructor
unsafe class C2 : B // ok: we are in unsafe context
```

So the `unsafe` on a `class` isn't "meaningless". It just doesn't mean what you might think (the class isn't caller-unsafe when it has the `unsafe` modifier).

Similar for other member kinds like fields, destructors, etc.
We just may need to preserve the warning for a few constructs where `unsafe` really is meaningless, like `using` alias.

But with this alternative proposal, that will get much nicer - `unsafe` always means an unsafe context, so no need for a warning, and we can disallow `[RequiresUnsafe]` on classes,destructors,etc. where it doesn't make sense.

proposals/unsafe-alternative-syntax.md
```diff
@@ -0,0 +1,147 @@
...
+    - Placement of `[RequiresUnsafe]` needs to obey the same restrictions as the original proposal lays out for `unsafe` on members.
+    - Delegates and lambdas must obey equivalent rules to the original proposal.
+    - Members marked `extern` are implicitly caller-unsafe.
+
+There is an open question around "compat mode" below.
+
+## Drawbacks
+
+- **No dedicated syntax:** Unlike the original proposal, this version falls back to attributes to express the intended semantics. It wouldn't be the only place in the language where member annotations are by attribute, but maybe there are arguments why this situation is more common or important and deserves a keyword.
```
jjonescz, line 133: Perhaps the lack of some sort of "uniformity" is another drawback - with the original proposal, if I have an `unsafe` member, I can use it in an `unsafe` context. Sounds simple. (Similar to if I have an `Obsolete` member, I can use it in another `Obsolete` member, i.e., `Obsolete` context).

proposals/unsafe-alternative-syntax.md
MadsTorgersen: True. This is a very speculative section - if we wanted to do something like that, I feel like we could work around the performance issue. For instance, add `[RequiresUnsafe]` to the assembly itself when compiling, if we encounter it on any member.

proposals/unsafe-alternative-syntax.md
MadsTorgersen: That's a good point. I'll add that.

proposals/unsafe-alternative-syntax.md
MadsTorgersen: Good point. Adding it to the "Motivation" section.

proposals/unsafe-alternative-syntax.md
```diff
@@ -103,6 +103,12 @@ This means that the member's author doesn't get more fine-grained control over w
 
 **With this proposal** the `[RequiresUnsafe]` attribute does not imply that the member is an unsafe context. The author is free to mark the boundaries that make the most sense.
 
+### Meaningless uses of unsafe
+
+The original proposal introduces a warning for a few places where the `unsafe` modifier is allowed only for compat purposes, but no longer has an effect.
+
+**With this proposal** there is no need for such a warning, since the meaning of `unsafe` doesn't change.
```
jjonescz, line 110: That's not entirely true. For example, `unsafe using X = List<int*[]>` should probably still result in a "meaningless `unsafe`" warning since the pointer type is now safe. But most other cases indeed do not need the warning anymore.

proposals/unsafe-alternative-syntax.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Language breaking changes
+
+Both proposals support users introducing *API breaking changes* (mitigated by an opt-in mechanism) by marking existing APIs caller-unsafe.
+
+The original proposal additionally introduces two *language breaking changes*:
+
+1. **Unsafe members**: Existing uses of `unsafe` on a member are generally intended only to make the member body an unsafe context. They may be used as a more convenient shorthand for wrapping the whole member body in an `unsafe` statement. However, with the new semantics of the original proposal, an `unsafe` member is now considered caller-unsafe, and existing calls to the member will break.
+2. **Unsafe types**: Existing uses of `unsafe` on a type are intended to make the type body an unsafe context. They may be used as a more convenient shorthand for marking each member unsafe. However, with the new semantics of the original proposal, an `unsafe` type is no longer an unsafe context, and unsafe operations within the type will break.
```
jjonescz, line 56: ~I think in the original proposal, `unsafe` on a `class` still introduces an unsafe context. Yes, there is a "meaningless" warning but I guess that's only to make you aware `unsafe class` is not "caller-unsafe".~ Actually that's not true, it really shouldn't introduce the unsafe context, I have misunderstood.

PR #9959, opened 2026-01-27
Unsafe evolution speclet: use attribute to denote caller-unsafe members

jjonescz:


proposals/unsafe-evolution.md
```diff
@@ -35,16 +35,18 @@ potentially occur, making it easier for reviewers and auditors to understand the
...
+the member has memory safety concerns and any usages must be manually validated by the programmer using the member (the error will go away if the member is used inside an `unsafe` context).
 
-This is a potentially large breaking change for particular segments of the C# user base. Our hope is that, for many of our users, this is effectively transparent, and updating to the new rules
+Nevertheless, this is still a breaking change for particular segments of the C# user base. Our hope is that, for many of our users, this is effectively transparent, and updating to the new rules
 will be seamless. However, given that some large API surfaces like large parts of reflection may need to be marked `unsafe`, we do think it likely that there will need to be a decent on-ramp to
 the new rules to avoid entirely bifurcating the ecosystem.
 
 ## Detailed Design
 
+Terminology: members which [have the `RequiresUnsafe` attribute](#metadata) or [are `extern`](#extern) or [contain pointers in signature](#compat-mode) are called *caller-unsafe*.
```
jcouv: nit: the extern and pointer cases are only for compat mode, and the RequiresUnsafe case is only for the new mode #Closed

jcouv: It would be good for the spec to include the definition for RequiresUnsafe attribute as reference #Closed

proposals/unsafe-evolution.md
```diff
@@ -129,29 +131,30 @@ For other types, `sizeof` used to require unsafe context ([§24.6.9][sizeof-unsa
...
-is updated to include whether the _anonymous function_ has the `unsafe` keyword, or the _method group_ is to a member that is marked `unsafe`. If it is, an anonymous function type is created, just as
+It is a memory safety error to convert a caller-unsafe member to a delegate type that is not caller-unsafe. The [_function type_](csharp-10.0/lambda-improvements.md#natural-function-type) definition
+is updated to include whether the _anonymous function_ has the `RequiresUnsafe` attribute, or the _method group_ is to a member that is caller-unsafe. If it is, a caller-unsafe anonymous function type is created, just as
 it would be if any parameter were a by-`ref`, optional, or `params`.
 
-It is a memory safety error convert a delegate type that is marked as `unsafe` to `System.Delegate`/`System.Linq.Expressions.Expression`/`System.Linq.Expressions.Expression<T>`, or any interface those
+It is a memory safety error convert a delegate type that is caller-unsafe to `System.Delegate`/`System.Linq.Expressions.Expression`/`System.Linq.Expressions.Expression<T>`, or any interface those
```
jcouv: >convert [](http://example.com/codeflow?start=28&length=7)

typo: convert => to convert #Closed

proposals/unsafe-evolution.md
jcouv: >*caller-unsafe* [](http://example.com/codeflow?start=163&length=15)

nit: regarding "caller-unsafe" terminology, we leaned towards RequiresUnsafe instead of CallerUnsafe for the attribute because of fields. We should consider aligning terminology #Closed

jjonescz: extern is for new mode only actually; I will clarify though, thanks

jjonescz: There is one in the metadata section (this sentence even links to it)

proposals/unsafe-evolution.md
```diff
@@ -166,49 +171,28 @@ because methods with pointers in signature would always need an unsafe context a
...
+(except for iterator bodies). Since pointer types are now safe, `unsafe` modifier on declarations without bodies is not necessary anymore. Hence `unsafe` on the following declarations will produce a warning:
 - `using static`,
 - `using` alias.
 
-Note that `unsafe` on the following declarations will not have any effect on the callers but it will have an effect on the body/initializer of the member:
-- field (there is [an open question](#unsafe-fields) for this),
-- destructor.
-
-`unsafe` on a member is _not_ applied to any nested anonymous or local functions inside the member. To mark a anonymous or local function as `unsafe`, it must manually be marked as `unsafe`. The same goes for
+`RequiresUnsafe` on a member is _not_ applied to any nested anonymous or local functions inside the member. To mark a anonymous or local function as requires-unsafe, it must manually be marked as `RequiresUnsafe`. The same goes for
```
jcouv: >To mark a anonymous [](http://example.com/codeflow?start=108&length=19)

typo: To mark a anonymous => To mark an anonymous

proposals/unsafe-evolution.md
```diff
@@ -166,49 +171,28 @@ because methods with pointers in signature would always need an unsafe context a
 ### Unsafe modifiers and contexts
 
 Today, as covered by the [unsafe context specification][unsafe-context-spec], `unsafe` behaves in a lexical manner, marking the entire textual body contained by the `unsafe` block as an `unsafe` context
-(except for iterator bodies). We propose changing this definition from textual to sematic. `unsafe` on a member will mean that that member is `unsafe`, and the body of that member is considered an `unsafe`
-context.
-
-`unsafe` on the following declarations will be permitted for source compatibility purposes only; it will have no meaning, and the compiler will produce a warning informing the user that it does not have any effect:
-- type (except delegate type),
+(except for iterator bodies). Since pointer types are now safe, `unsafe` modifier on declarations without bodies is not necessary anymore. Hence `unsafe` on the following declarations will produce a warning:
```
jcouv: >`unsafe` modifier [](http://example.com/codeflow?start=64&length=17)

nit: "the" or "an" `unsafe` modifier

proposals/unsafe-evolution.md
```diff
@@ -35,16 +35,20 @@ potentially occur, making it easier for reviewers and auditors to understand the
...
 types cannot be inherently unsafe. It is only the action of using a type that could be unsafe, not the existence of that type.
 
-In order for this information to flow through the system, we therefore need to have a way to mark methods themselves as `unsafe`. Today, `unsafe` as a method modifier has no external impact,
-it only allows pointers to be used in the signature and body of the member. Going forward, `unsafe` as a modifier will actually publicly change the meaning of the member; it will indicate that
-the member has memory safety concerns and any usages must be manually validated by the programmer using the member.
+In order for this information to flow through the system, we therefore need to have a way to mark methods themselves as unsafe. Today, `unsafe` as a method modifier has no external impact,
+it only allows pointers to be used in the signature and body of the member. To avoid a breaking change, we are not going to change that, instead an attribute (`RequiresUnsafe`) will be used to indicate that
```
333fred: We _are_ still changing that; you won't need to have `unsafe` in the signature to have a pointer type.

proposals/unsafe-evolution.md
```diff
@@ -69,8 +73,8 @@ The following expression require an `unsafe` context when used:
 * `stackalloc` under the conditions defined [below](#stack-allocation)
 
 In addition to these expressions, expressions and statements can also conditionally require an `unsafe` context if they depend on any symbol that is marked as `unsafe`. For example, calling a method
-that is marked as `unsafe` will cause the _invocation_expression_ to require an `unsafe` context. Statements with invocations embedded (such as `using`s, `foreach`, and similar) can also require an
-`unsafe` context when they use a member that is marked as `unsafe`.
+that is requires-unsafe will cause the _invocation_expression_ to require an `unsafe` context. Statements with invocations embedded (such as `using`s, `foreach`, and similar) can also require an
```
333fred:
```suggestion
that is *requires-unsafe* will cause the _invocation_expression_ to require an `unsafe` context. Statements with invocations embedded (such as `using`s, `foreach`, and similar) can also require an
```

And other locations, to match the definition.

proposals/unsafe-evolution.md
```diff
@@ -35,16 +35,20 @@ potentially occur, making it easier for reviewers and auditors to understand the
...
+the member has memory safety concerns and any usages must be manually validated by the programmer using the member (the error will go away if the member is used inside an `unsafe` context).
 
-This is a potentially large breaking change for particular segments of the C# user base. Our hope is that, for many of our users, this is effectively transparent, and updating to the new rules
+Nevertheless, this is still a breaking change for particular segments of the C# user base. Our hope is that, for many of our users, this is effectively transparent, and updating to the new rules
 will be seamless. However, given that some large API surfaces like large parts of reflection may need to be marked `unsafe`, we do think it likely that there will need to be a decent on-ramp to
 the new rules to avoid entirely bifurcating the ecosystem.
 
 ## Detailed Design
 
+Terminology: we call members *requires-unsafe* (a.k.a. *caller-unsafe*) if
```
333fred:
```suggestion
Terminology: we call members *requires-unsafe* if
```

Don't think that aka should be included. If it's mentioned somewhere else in the speclet, it should be removed, and we should just use `*requires-unsafe*`

proposals/unsafe-evolution.md
```diff
@@ -69,8 +73,8 @@ The following expression require an `unsafe` context when used:
 * `stackalloc` under the conditions defined [below](#stack-allocation)
 
 In addition to these expressions, expressions and statements can also conditionally require an `unsafe` context if they depend on any symbol that is marked as `unsafe`. For example, calling a method
-that is marked as `unsafe` will cause the _invocation_expression_ to require an `unsafe` context. Statements with invocations embedded (such as `using`s, `foreach`, and similar) can also require an
-`unsafe` context when they use a member that is marked as `unsafe`.
+that is requires-unsafe will cause the _invocation_expression_ to require an `unsafe` context. Statements with invocations embedded (such as `using`s, `foreach`, and similar) can also require an
+`unsafe` context when they use a requires-unsafe member.
 
 When we say "requires an unsafe context" or similar in this document, it means emitting an error that the construct requires an `unsafe` context to be used.
```
333fred, line 80: Does this need to be updated?

proposals/unsafe-evolution.md
```diff
@@ -166,49 +171,28 @@ because methods with pointers in signature would always need an unsafe context a
...
-    public partial unsafe void M2();
-}
-
-partial class C1
-{
-    public unsafe partial void M1() => Console.WriteLine("hello world");
-    public partial void M2() => Console.WriteLine("hello world"); // Error: both parts must be unsafe, or neither can be
-}
-```
+When a member is `partial`, both parts must agree on the `unsafe` modifier, but only one can specify the `RequiresUnsafe` attribute, unchanged from C# rules today.
```
333fred, line 197: This (only able to applied to one part) is actually somewhat of a downside, imo. Maybe something to raise as a question?

proposals/unsafe-evolution.md
jjonescz: We use "caller unsafe" throughout the implementation, in other speclets, etc, so I think it would be good to mention it just once.

proposals/unsafe-evolution.md
jjonescz: I usually just make the definition italic, the other usages don't need to be italic IMO. (At least it seems like overkill for speclets.)

proposals/unsafe-evolution.md
jjonescz: If you are talking about this sentence "When we say "requires an unsafe context" or similar in this document, it means emitting an error that the construct requires an `unsafe` context to be used.", that doesn't look like it needs any updating to me.

proposals/unsafe-evolution.md
333fred: If it's a specific concept, it should be italicized throughout, as that's how the spec will be written as well.

proposals/unsafe-evolution.md
333fred: We should remove mentions of caller unsafe from the implementation at least.

proposals/unsafe-evolution.md
```diff
@@ -170,8 +171,23 @@ because methods with pointers in signature would always need an unsafe context a
...
+    [RequiresUnsafe] public A() { }
+}
+class C
+{
+    [A] void M1() { } // error: cannot use `A..ctor` in safe context
+    [A] unsafe void M1() { } // ok: the `unsafe` context applies to the `A..ctor` usage
+}
+```
+
+Since pointer types are now safe, an `unsafe` modifier on declarations without bodies does not have a meaning anymore. Hence `unsafe` on the following declarations will produce a warning:
```
333fred, line 190: Add fields without initializers?

proposals/unsafe-evolution.md
jjonescz: Added to test plan.

proposals/unsafe-evolution.md
jjonescz: There is an open question for that:

> ### More meaningless `unsafe` warnings
>
> Should more declarations produce the meaningless `unsafe` warning?
For example, fields without initializers (assuming we don't support [requires-unsafe fields](#requires-unsafe-fields)), methods with empty bodies (or `extern`), etc.
We already have an IDE analyzer for unnecessary `unsafe` though.

jjonescz: But we should probably warn about meaningless `RequiresUnsafe` applications too. Added to the speclet.

proposals/unsafe-evolution.md
jjonescz: Agreed that's how the spec will be written, but it seems overkill for the speclet. Especially for such specific terms as "requires-unsafe" where there is no ambiguity.

But alright, I will do find+replace.

PR #9964, opened 2026-01-29
Add closed classes open questions

RikkiGibson:


proposals/closed-hierarchies.md
```diff
@@ -137,4 +137,60 @@ Closed classes are generated with a `Closed` attribute, to allow them to be reco
...
+```cs
+closed abstract class C { } // equivalent to:
+closed class C { }
+
+// compare with:
+abstract interface I { } // error
+internal class C { } // ok, explicitly specifying the default accessibility
+```
+
+### TODO: Should subtypes be marked in metadata?
```
RikkiGibson, line 194: @AlekseyTs I would be interested in any thoughts you may have on this question

proposals/closed-hierarchies.md
```diff
@@ -137,4 +137,60 @@ Closed classes are generated with a `Closed` attribute, to allow them to be reco
...
+### Permit explicit use of abstract modifier
+
+The `closed` modifier also makes the class abstract. Should we permit both `closed` and `abstract` to be used on a declaration? Permitting it may give an impression that the `abstract` modifier is making a difference.
+
+```cs
+closed abstract class C { } // equivalent to:
+closed class C { }
+
+// compare with:
+abstract interface I { } // error
```
AlekseyTs: At the same time abstract is allowed on interface members, even when it is implied

proposals/closed-hierarchies.md
AlekseyTs: > I would be interested in any thoughts you may have on this question

To optimize the process of finding direct derived classes? I think we should try instead of searching through symbols, search for `TypeDef`  entries referring to the closed class in its `Extends` column. That probably will be much faster. And then we can locate `NamedTypeSymbol`s corresponding to those entries. I think we have readily available hash tables to do that. We are resolving token references from metadata signatures to type symbols a lot.

RikkiGibson: > To optimize the process of finding direct derived classes?

Yes exactly. I will give what you are suggesting a try in the next implementation PR.

proposals/closed-hierarchies.md
RikkiGibson: Updating doc to show this.

```suggestion
abstract interface I // error
{
    abstract void M(); // ok
}
```

PR #9990, opened 2026-02-13
Incorporate recent design decisions in unions.md

MadsTorgersen:


proposals/unions.md
```diff
@@ -7,16 +7,16 @@ Champion issue: https://github.com/dotnet/csharplang/issues/9662
 
 *Unions* is a set of interlinked features, that combine to provide C# support for union types:
 
-- *Union types*: Classes and structs that implement the `IUnion` interface are recognized as *union types*, and support the *union behaviors*.
-- *Case types*: Union types have a set of *case types*, which is given by the set of public instance constructors with exactly one parameter. The type of each such constructor parameter is a case type for the union type.
-- *Union behaviors*: All union types support the following *union behaviors*:
+- *Union types*: Structs and classes that have a `[Union]` attribute are recognized as *union types*, and support the *union behaviors*.
```
AlekseyTs, line 10: It would be good to include a proposed definition for the attribute. For example, I assume that the name of the type is actually "UnionAttribute". Another important aspect is the containing namespace. A check for presence of the attribute cannot be implemented without this information. I guess we can start with "System.Runtime.CompilerServices" namespace, which can be adjusted later, but we have to pick something specific and record the choice in this document to unblock implementation. #Closed

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
+The proposed unions in C# are unions of *types* and not "discriminated" or "tagged". "Discriminated unions" can be expressed in terms of "type unions" by using fresh type declarations as case types. Alternatively they can be implemented as a [closed hierarchy](https://github.com/dotnet/csharplang/blob/main/proposals/closed-hierarchies.md), which is another, related, upcoming C# feature focused on exhaustiveness.
 
 ## Detailed design
 [design]: #detailed-design
 
-### Union interfaces
+### Union types
 
-The following interfaces are recognized by the language in its implementation of union features.
+Any class or struct type with a `[Union]` attribute is considered a *union type*. It is an error to apply the `[Union]` attribute to a type declaration that isn't a class or a struct.
```
AlekseyTs: It is not clear whether the compiler is expected to block the attribute application regardless of the attribute usage settings for the attribute type, or the attribute type should be declared in such a way that allows its application only on classes and structs. #Closed

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
-    throw ...;
-}
-```
+The `HasValue` and `TryGetValue` members are collectively referred to as the *non-boxing union access pattern*.
 
-### Union types
+The different union members are described in the following.
 
-Any non-abstract class or struct type that implements the `IUnion` interface is considered a *union type*.
-For each public constructor with exactly one by-value or `in` parameter, the type of that parameter is considered
```
AlekseyTs, line 91: It looks like the requirement for ref-ness of the parameter got deleted. Please restore. #Closed

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
-```
+Union constructors and union factory methods are referred to collectively as *union creation members*.
+
+A union type must have at least one union creation member, and therefore at least one case type.
+
+#### Value property
+
+The `Value` property allows access to the value contained in a union, regardless of its case type.
+
+Every union-defining type must declare a `Value` property of type `object?` or `object`. The property must have a `get` accessor and may optionally have an `init` or `set` accessor, which can be of any accessibility and is not used by the compiler.
```
AlekseyTs, line 101: Are we sticking with `object` and not going to allow a more specific type mentioned at https://github.com/dotnet/csharplang/blob/main/meetings/working-groups/discriminated-unions/union-patterns-update.md#more-specific-shared-base-types ?

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
+
+A union type can choose to additionally implement the *non-boxing union access pattern*, which allows strongly typed conditional access to each case type, as well as a way to check for null.
+
+This allows the compiler to implement pattern matching more efficiently when case types are value types and stored as such within the union.
+
+The non-boxing access members are:
+
+- A `HasValue` property of type `bool` with a public `get` accessor. It may optionally have an `init` or `set` accessor, which can be of any accessibility and is not used by the compiler. 
+- A `TryGetValue` method for each case type. The method returns `bool` and takes an out-parameter of a type that corresponds to the given case type in the following way:
+    - If the case type is a nullable value type, the type of the parameter should be identity-convertible to the underlying type
```
AlekseyTs, line 118: It looks like this nullable unwrapping rule is new. I think adding it makes sense, however, since, according to my recollection, the addition was not discussed, consider also adding an entry to the "Open Questions" section for the working group to discuss/confirm. #Closed

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
+    - Otherwise, the type should be identity-convertible to the case type.
+    
+`HasValue` should return true if and only if the union's `Value` is not null.
 
-`MyUnion` has the two case types `Case1` and `Case2`.
+`TryGetValue` should return true if and only if the union's `Value` is of the given case type, and if so, deliver that value in the method's out parameter.
 
-*Note:* By this definition, the `IUnion` is not a union type and therefore does not have union behaviors.
+It is an error to have some but not all of the non-boxing access members.
```
AlekseyTs: I don't think I agree with this statement. Obviously, to get the most benefits from the APIs, it is recommended to provide complete set of them. But, I do not think that it is appropriate for a compiler to report any errors on declarations that do not do that. And I think it is even more inappropriate for compiler to report any errors for a pattern matching against such types.  #Closed

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
+```csharp
+public record class Result<T> : Result<T>.IUnionMembers
+{
+    object? _value;
 
-* *Soundness*: The `Value` property always evaluates to null or to a value of a case type.
-* *Stability*: If a union value is constructed from a case type, the `Value` property will match that case type or null. If a union value is constructed from a `null` value, the `Value` property will be `null`.
-* *Constructor equivalence*: If a value is implicitly convertible to two different case types then the constructor for either of those case types has the same observable behavior when called with that value.
-* *Access pattern consistency*: The behavior of `HasValue` and `TryGetValue` members of the non-boxing access pattern, if present, is observably equivalent to that of checking against the `Value` property directly.
-* *Default value*: If a union type is a value type, it's default value has `null` as its `Value`.
```
AlekseyTs, line 174: It looks like this rule and the following rule got deleted. I think this is the right thing to do, but I do not remember us discussing this explicitly. Also, I think there is an open question that refers to the deleted rules. After the deletion, it probably becomes obsolete. Consider adjusting accordingly. Perhaps adding a note that the rules were deleted.   #Closed

proposals/unions.md
```diff
@@ -25,154 +25,154 @@ Unions are a long-requested C# feature, which allows expressing values from a cl
...
-        Kind.Case1 => value1,
-        Kind.Case2 => value2,
-        _ => null;
-    };
 }
 ```
 
 *Note:* This is just an example of how the non-boxing access pattern might be implemented. The user code can store the content any way it likes. In particular, it does not prevent the implementation from boxing! The `non-boxing` in its name refers to allowing the compiler's pattern matching implementation to access each case type in a strongly typed way, as opposed to the `object?`-typed `Value` property.
 
 *Note:* This is an example of an implementation strategy that may give rise to "large" (bigger than a pointer) structs. This means they may be vulnerable to "tearing", where copying of values isn't atomic, and concurrent threads may observe corrupted values that are "torn" between before and after parts. This is an orthogonal concern that applies to all large structs, but worth keeping in mind for union scenarios. The unions generated by [Union declarations](#union-declarations) are immune to this issue, as they always generate pointer-length structs.
```
AlekseyTs: This section isn't changed. At the same time, as was discussed in the past, it makes an incorrect claim. Given the current design, the union declarations are not immune to the issue. Since you are making broad changes to the spec, consider removing the claim and/or adding open design question to adjust the design for addressing the "tearing" problem #Closed

proposals/unions.md
```diff
@@ -318,6 +328,59 @@ var value = pet switch
...
+```csharp
+if (value is IUnion { Value: null }) { ... }
+```
+
+#### Union creation interface
+
+The `IUnion<TUnion>` interface provides a means to create union instances at runtime.
+
+```csharp
+public interface IUnion<TUnion> : IUnion where TUnion : IUnion<TUnion>
```
AlekseyTs, line 387: I think it was discussed on several occasions that the design of this interface is questionable. Both the inheritance from ```IUnion``` and the ```IUnion<TUnion>``` constraint. Please add an open design question so that we finally get this discussed and resoleved. #Closed

proposals/unions.md
MadsTorgersen: I've added a declaration and included a namespace requirement.

proposals/unions.md
MadsTorgersen: I've removed the requirement. It follows from the `[AttributeUsage]` of the attribute declaration.

proposals/unions.md
AlekseyTs: It looks like the lowering should be adjusted. The attribute is missing. Also, I think it would be good to explicitly state what is going to happen when the attribute and the interfaces aren't found. Is compiler expected to synthesize them, or report an error?  #Closed

proposals/unions.md
MadsTorgersen: Done.

proposals/unions.md
MadsTorgersen: Yep, I hadn't thought about it before, but it seems obviously right. I've added an open question to confirm.

proposals/unions.md
MadsTorgersen: You're right. I've removed the requirement.

proposals/unions.md
MadsTorgersen: I added a note at the beginning of that open question.

proposals/unions.md
MadsTorgersen: I've removed the paragraph about tearing.

proposals/unions.md
MadsTorgersen: Open question added.

proposals/unions.md
MadsTorgersen: I've added the attribute. I've added an open question about what happens when types are missing.

proposals/unions.md
```diff
@@ -68,10 +76,22 @@ Union creation members are used to create new union values from a case type valu
...
 If the union-defining type is a union member provider, each static `Create` method with a single parameter and a return type that is identity-convertible to the union type itself is a *union factory method*. The case types of the union are identified as the set of parameter types from these factory methods.
 
+```csharp
+// Union factory method making `Cat` a case type
+public static Pet Create(Cat value) { ... }
+```
+
 Union constructors and union factory methods are referred to collectively as *union creation members*.
 
+The single parameter of a union creation member must be an by-value or `in` parameter.
```
AlekseyTs: "an by-value" vs. "a by-value"

proposals/unions.md
AlekseyTs: No answer?

proposals/unions.md
```diff
@@ -318,6 +355,59 @@ var value = pet switch
 }
 ```
 
+### Union interfaces
+
+The following interfaces are recognized by the language in its implementation of union features.
```
mattwar: Instead of 'recognized' maybe 'used'?  Recognized implies that the interfaces are significant to the compiler understanding the union.  But this is not the case. It is merely also implementing them.

proposals/unions.md
MadsTorgersen: Sorry, somehow missed this one! 

We didn't agree on allowing more specific types, so I didn't incorporate it. I think it would be an easy and useful addition. Maybe I'll try it out in a separate PR to see if I'm right about it being easy! ;-)

proposals/unions.md
MadsTorgersen: Oops, thanks!

proposals/unions.md
MadsTorgersen: Good point, will change!

proposals/unions.md
AlekseyTs: > We didn't agree on allowing more specific types, so I didn't incorporate it.

I see. Makes sense.

PR #10088, opened 2026-04-08
Add ref safety rules for IEnumerable ref struct collection expressions

### LINKED ISSUE #80750: Ref safety spec gap for collection expression with ref struct element type
The following program reports an error on `return spans`, but nothing in [the spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md#ref-safety) indicates that should be the case. Strictly, the code seems valid to me. I think there is a spec gap due to the fact that ref structs were not anticipated to be valid element types under any conditions.

```cs
[Fact, WorkItem("https://github.com/dotnet/roslyn/issues/77827")]
public void RefStructElementType_SafeContext_06()
{
    var source = """
        #nullable enable
        using System;

        partial class Program
        {
            public static SpanCollection<Span<int>> Test(Span<int> span1, Span<int> span2)
            {
                SpanCollection<Span<int>> spans = [span1, span2];
                return spans;
            }
        }
        """;

    var comp = CreateCompilation([source, s_collectionOfRefStructsSource], targetFramework: TargetFramework.Net90);
    comp.VerifyDiagnostics(
        // (9,16): error CS8352: Cannot use variable 'spans' in this context because it may expose referenced variables outside of their declaration scope
        //         return spans;
        Diagnostic(ErrorCode.ERR_EscapeVariable, "spans").WithArguments("spans").WithLocation(9, 16));
}

private static readonly string s_collectionOfRefStructsSource = """
        #nullable enable
        using System;
        using System.Collections;

        ref struct SpanCollection<T>() : IEnumerable<T> where T : allows ref struct
        {
            int _count = 0;
            T? _item0;
            T? _item1;

            public SpanEnumerator GetEnumerator()
                => new SpanEnumerator(this);

            public void Add(T element)
            {
                switch (_count)
                {
                    case 0: _item0 = element; break;
                    case 1: _item1 = element; break;
                    default: throw new InvalidOperationException();
                }
                _count++;
            }

            IEnumerator<T> IEnumerable<T>.GetEnumerator()
                => throw new NotImplementedException();

            IEnumerator IEnumerable.GetEnumerator()
                => throw new NotImplementedException();

            public ref struct SpanEnumerator(SpanCollection<T> spans)
            {
                private SpanCollection<T> spans = spans;
                private int _index = -1;
                public T Current
                {
                    get
                    {
                        switch (_index)
                        {
                            case 0: return spans._item0!;
                            case 1: return spans._item1!;
                            default: throw new InvalidOperationException();
                        }
                    }
                }
                public bool MoveNext() => ++_index < spans._count;
            }
        }
        """;
```

jjonescz:
Resolves https://github.com/dotnet/roslyn/issues/80750. Implementation: https://github.com/dotnet/roslyn/pull/82401

### DIFF
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -353,6 +353,11 @@ The *safe-context* of a collection expression is:
 
 * If the target type is a *ref struct type* with a [*create method*](#create-methods), the safe-context of the collection expression is the [*safe-context of an invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) of the create method where the collection expression is the span argument to the method.
 
+* If the target type is a *ref struct type* that implements `System.Collections.IEnumerable` and does not have a *[create method](#create-methods)*, the collection expression is treated equivalently to an *object creation expression* with a *collection initializer* for ref safety purposes. The safe-context of the collection expression is the narrowest of:
+  * the [*safe-context of an invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) of the applicable constructor, and
+  * for each expression element, the safe-context that the element expression can contribute to the collection through the applicable `Add` method invocation, as determined by the [*method and property invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) rules applied to the `Add` call where the collection is the receiver, and
+  * for each spread element, the safe-context determined similarly for the `Add` method invocation where the argument is the iteration variable (with safe-context of *declaration-block*).
+
 * Otherwise the safe-context of the collection expression is the *caller-context*.
 
 A collection expression with a safe-context of *declaration-block* cannot escape the enclosing scope, and the compiler *may* store the collection on the stack rather than the heap.
--- a/proposals/csharp-12.0/collection-expressions.md
+++ b/proposals/csharp-12.0/collection-expressions.md
@@ -353,10 +353,12 @@ The *safe-context* of a collection expression is:
 
 * If the target type is a *ref struct type* with a [*create method*](#create-methods), the safe-context of the collection expression is the [*safe-context of an invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) of the create method where the collection expression is the span argument to the method.
 
-* If the target type is a *ref struct type* that implements `System.Collections.IEnumerable` and does not have a *[create method](#create-methods)*, the collection expression is treated equivalently to an *object creation expression* with a *collection initializer* for ref safety purposes. The safe-context of the collection expression is the narrowest of:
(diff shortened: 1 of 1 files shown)


proposals/csharp-12.0/collection-expressions.md
```diff
@@ -353,6 +353,11 @@ The *safe-context* of a collection expression is:
...
 
+* If the target type is a *ref struct type* that implements `System.Collections.IEnumerable` and does not have a *[create method](#create-methods)*, the collection expression is treated equivalently to an *object creation expression* with a *collection initializer* for ref safety purposes. The safe-context of the collection expression is the narrowest of:
+  * the [*safe-context of an invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) of the applicable constructor, and
+  * for each expression element, the safe-context that the element expression can contribute to the collection through the applicable `Add` method invocation, as determined by the [*method and property invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) rules applied to the `Add` call where the collection is the receiver, and
+  * for each spread element, the safe-context determined similarly for the `Add` method invocation where the argument is the iteration variable (with safe-context of *declaration-block*).
```
RikkiGibson: It feels surprising for `declaration-block` to be hard-coded here in a sense. I think this won't always be correct, and, could lead to ref safety bugs.

(At the same time, I think it's OK for us to take this PR in order to move to what seems a "more correct" state than before.)

Have we defined the safe-context of an iteration variable? For example, if I have this code:

```cs
RefCollection rc = Create(...);
foreach (var item in rc)
{
    // Is this allowed or not?
    // Does it depend on safe-context of `rc`? Safe-context of `rc.GetEnumerator().Current`? Something else?
    return M(item);
}
```

Basically, it feels reasonable to define the safe-context of an iteration variable in general, and then use that concept here.

proposals/csharp-12.0/collection-expressions.md
```diff
@@ -353,6 +353,11 @@ The *safe-context* of a collection expression is:
 
 * If the target type is a *ref struct type* with a [*create method*](#create-methods), the safe-context of the collection expression is the [*safe-context of an invocation*](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/structs.md#164126-method-and-property-invocation) of the create method where the collection expression is the span argument to the method.
 
+* If the target type is a *ref struct type* that implements `System.Collections.IEnumerable` and does not have a *[create method](#create-methods)*, the collection expression is treated equivalently to an *object creation expression* with a *collection initializer* for ref safety purposes. The safe-context of the collection expression is the narrowest of:
```
RikkiGibson: > treated equivalently to an *object creation expression* with a *collection initializer* for ref safety purposes.

I think this could be more precise. Is this referring to the set of arg mixing checks that are performed involving the collection expression? Or more than that?

jjonescz: Yes, arg mixing, will specify that explicitly, thanks.

proposals/csharp-12.0/collection-expressions.md
jjonescz: Good catch, thanks. We have a safe context of iteration variable specified, I will reference that. And fix up the implementation too.

RikkiGibson (approved):

PR #10091, opened 2026-04-09
Unsafe evolution speclet: back to unsafe keyword

jjonescz:
Incorporating decisions from [LDM 2026-04-06](https://github.com/dotnet/csharplang/blob/main/meetings/2026/LDM-2026-04-06.md): `unsafe` keyword denotes *requires-unsafe* members, and pointers are safe regardless of memory safety rules version. Didn't incorporate changes around `extern` as it's not yet decided how `safe`/`unsafe` is going to be required there.


proposals/unsafe-evolution.md
```diff
@@ -150,19 +154,19 @@ contract of `Span<T>` and `ReadOnlySpan<T>`, and so must be subject to extra scr
...
-definition and not see any addition of `RequiresUnsafe` by a derived implementation.
+It is a memory safety error to add `unsafe` at the member level in any override or implementation of a member that is not *requires-unsafe* originally, because callers may be using the base
+definition and not see any addition of `unsafe` by a derived implementation.
 
 ### Delegates and lambdas
 
 It is a memory safety error to convert a *requires-unsafe* member to a delegate type outside the `unsafe` context.
 Delegate types and [_function types_](csharp-10.0/lambda-improvements.md#natural-function-type) cannot be *requires-unsafe*.
-It is a compile-time error to apply `RequiresUnsafe` on a lambda symbol.
+It is a parse error to apply `unsafe` on a lambda symbol.
```
333fred: I wouldn't specify parse error. The existing wording is fine and as specific as we should be.  #Closed

jjonescz:
```suggestion
It is a compile-time error to apply `unsafe` on a lambda symbol.
```

333fred (approved):

proposals/unsafe-evolution.md
```diff
@@ -196,18 +200,36 @@ class C
 }
 ```
 
+We propose extending this definition from textual to semantic.
```
jcouv: There's a remaining mention of the attribute above (line 194) #Closed

proposals/unsafe-evolution.md
```diff
@@ -80,6 +79,11 @@ unsafe in itself. [Pointer types][pointer-types-spec], [Fixed and moveable varia
 are all no longer considered `unsafe`, and exist in normal C# with no requirement to be used in an `unsafe` context. Similarly, declaring a [fixed size buffer][fixed-size-buffer-declarations] or
 an initialized [`stackalloc`][stack-allocation-spec] are also perfectly legal in safe C#. For all of these cases, it is only _accessing_ the memory that is unsafe.
 
+Importantly, these pointer relaxations apply **regardless of whether an assembly opts in** to the [updated memory safety rules](#metadata).
```
jcouv, line 82: >these pointer relaxations apply [](http://example.com/codeflow?start=13&length=31)

Would we want these relaxations to apply regardless of language version as well?

proposals/unsafe-evolution.md
jcouv: I didn't understand what this means.

By the way, although it's hard to imagine someone doing that, the example above is another scenario that needs a way of declaring an unsafe-context on a declaration while not necessarily passing the obligation to caller. Some other scenarios in that category are declarations with `: base(...)` or `: this(...)` where that constructor is caller-unsafe. #Closed

proposals/unsafe-evolution.md
```diff
@@ -217,18 +239,13 @@ It is an error to apply the `MemorySafetyRulesAttribute` to any symbol explicitl
 
 The compiler ignores `RequiresUnsafeAttribute`-marked members from assemblies that are using the legacy memory safety rules (instead, the [compat mode](#compat-mode) is used there).
 
-The compiler will emit a warning if `RequiresUnsafe` is used under the legacy memory safety rules and an error if it is applied to unsupported symbol kinds
-(note that this excludes symbol kinds that should be banned already by `AttributeUsageAttribute` on the attribute's definition):
-- destructors,
-- static constructors,
-- lambdas.
 
+When a member is marked as `unsafe`, the compiler will synthesize a `RequiresUnsafeAttribute` application on the member in metadata.
```
jcouv:
```suggestion
When a non-type member is marked as `unsafe`, the compiler will synthesize a `RequiresUnsafeAttribute` application on the member in metadata.
``` #Closed

jcouv: Should we check usages of `[RequiresUnsafeAttribute]` in source, or just allow them? Consider calling it out either way #Closed

jcouv (commented):

proposals/unsafe-evolution.md
jjonescz: That doesn't seem to be decided by LDM, I can add an open question for that.

But I think we would want to gate by LangVersion. It seems that's exactly what LangVersion was invented for.

proposals/unsafe-evolution.md
jjonescz: > I didn't understand what this means.

It means `unsafe` doesn't just introduce a syntactic unsafe block, but also semantic meaning for the symbol (requires-unsafe). It's actually the wording that the proposal had before we switched to attribute (since we are now switching back, I was trying to revert parts of the proposal). But I guess the sentence can be removed, it doesn't add much value.

jjonescz: > By the way, although it's hard to imagine someone doing that, the example above is another scenario that needs a way of declaring an unsafe-context on a declaration while not necessarily passing the obligation to caller. Some other scenarios in that category are declarations with `: base(...)` or `: this(...)` where that constructor is caller-unsafe.

Yes, I know. See also "`unsafe` on types" section which mentions a similar example. I can mention this in the alternative too.

jcouv (approved):
LGTM Thanks (commit 4)

PR #10098, opened 2026-04-13
unions.md: Expand usage of `TryGetValue(Nullable<T>)` methods for a pattern that implies checking for a specific type `T`

AlekseyTs:
Otherwise the following scenario doesn't use `TryGetValue`:
```
[System.Runtime.CompilerServices.Union]
struct S1
{
    private readonly object _value;
    public S1(int? x) { _value = x; }
    public S1(string x) { _value = x; }
    public object Value => _value;
    public bool TryGetValue(out int? x) { if (_value is int v) { x = v; return true; } x = null; return false; }

    static bool Test1(S1 u)
    {
        return u is int;
    }   
}
```


MadsTorgersen (approved):
You're right, good catch!

PR #10103, opened 2026-04-14
Unsafe evolution speclet: record all LDM decisions

jjonescz:


333fred (approved):

proposals/unsafe-evolution.md
```diff
@@ -300,7 +300,10 @@ Advantages of an attribute (or another keyword):
 - allows suppressing all *requires-unsafe* errors without needing to mark the member itself as *requires-unsafe*
   ([examples](#allow-suppressing-requires-unsafe-errors-in-edge-case-scenarios)).
 
-## Open questions
+Discussions:
```
jcouv, line 303: nit: for my own features, I've tried to record a one-line summary of decision (how was the open issue resolved?). I've found that useful

jcouv (approved):

proposals/unsafe-evolution.md
jjonescz: I'm doing that here too - after each LDM link, there is a few words about the decision

jjonescz: I guess you mean one summary per open question (not one per LDM), that's a good point, I can add that in a follow up.

jcouv: FWIW, here's what I used in recent work: "Answer: yes, disallow, LDM 2025-06-11" or "Decision (LDM 2026-02-02): extensions should contribute everywhere, including countable properties and implicit indexer fallback."

PR #10104, opened 2026-04-15
Unsafe evolution speclet: unsafe accessors question

jjonescz:


jcouv (approved):

proposals/unsafe-evolution.md
```diff
@@ -398,6 +398,30 @@ class A : Attribute
 - [LDM 2025-11-12](https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-11-12.md#unsafe-evolution): `unsafe` on a type will have no meaning
 - TODO: warning or error? also could revisit in light of more recent decisions
 
+### `unsafe` on accessors
```
333fred, line 401: It feels like this section is missing an actual question/recommmendation?

jjonescz: It's implied - "are we okay with all of these points?", I can clarify

Issue #9704, opened 2025-09-23
[Proposal]: Unsafe evolution


333fred:
# Unsafe evolution

* Specification: https://github.com/dotnet/csharplang/blob/main/proposals/unsafe-evolution.md
* .NET Design: https://github.com/dotnet/designs/blob/main/proposed/caller-unsafe.md 
* Discussion: https://github.com/dotnet/csharplang/discussions/9705
* Implementation: https://github.com/dotnet/roslyn/issues/81207

## Summary
[summary]: #summary

Evolving unsafe in C# to cover general memory unsafety.

## Design meetings

* https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-09-17.md#unsafe-evolution
* https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-10-29.md#unsafe-evolution
* https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-11-05.md#unsafe-evolution
* https://github.com/dotnet/csharplang/blob/main/meetings/2025/LDM-2025-11-12.md#unsafe-evolution

PR #10109, opened 2026-04-16
Unsafe evolution speclet: updates for pointers to managed types

jjonescz:
Relates to championed proposal: https://github.com/dotnet/csharplang/issues/9704

### DIFF
--- a/proposals/unsafe-evolution.md
+++ b/proposals/unsafe-evolution.md
@@ -129,8 +129,7 @@ The [fixed statement][fixed-statement] moves to [§13](https://github.com/dotnet
 Function pointers are not yet incorporated into the main C# specification, but they are similarly affected; everything but function pointer invocation is moved into the standard specification.
 A function pointer invocation expression must always occur in an `unsafe` context.
 
-Although the C# specification restricts the underlying type (_value_type_) of a [pointer types][pointer-types-spec] to be an unmanaged type,
-the Roslyn compiler already only warns in case of a pointer to a managed type and we should codify this behavior in the new specification.
+C# 11 [allows pointers to managed types with a warning][csharp-11-unsafe-changes] which we relax by allowing that without a warning.
 
 #### Fixed-size buffers
 
@@ -154,6 +153,8 @@ contract of `Span<T>` and `ReadOnlySpan<T>`, and so must be subject to extra scr
 
 > [!NOTE]
 > This means that assigning a `stackalloc` to a pointer is _always_ safe, regardless of context.
+
+Note that a `stackalloc` of a managed type remains an error.
 
 #### `sizeof`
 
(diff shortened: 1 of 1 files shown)


proposals/unsafe-evolution.md
```diff
@@ -129,6 +129,9 @@ The [fixed statement][fixed-statement] moves to [§13](https://github.com/dotnet
 Function pointers are not yet incorporated into the main C# specification, but they are similarly affected; everything but function pointer invocation is moved into the standard specification.
 A function pointer invocation expression must always occur in an `unsafe` context.
 
+Although the C# specification restricts the underlying type (_value_type_) of a [pointer types][pointer-types-spec] to be an unmanaged type,
```
333fred: This was part of C# 11, which the spec hasn't covered yet: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#changes-in-unsafe-context. Don't think there's any questions to answer or changes to make.

proposals/unsafe-evolution.md
```diff
@@ -129,6 +129,9 @@ The [fixed statement][fixed-statement] moves to [§13](https://github.com/dotnet
...
+the Roslyn compiler already only warns in case of a pointer to a managed type and we should codify this behavior in the new specification.
```
jcouv: nit: I would have put this comment under Open Issues section which tracks follow-ups

proposals/unsafe-evolution.md
jcouv: From quick chat with Fred, we think the reason there's an error for stackalloc scenario is that the underlying IL instruction `localloc` allocates memory that is not tracked by the GC. The instruction takes a size, but no type. So stackalloc of managed type should likely remain an error (`ERR_ManagedAddr`).
It would be good to check with @RikkiGibson 

For non-stackalloc scenarios, take a pointer to a managed type would become "safe" under the new memory safety rules (ie. no longer produce `WRN_ManagedAddr`. The GC is tracking types and references just fine. The problem is only when the user dereferences the pointer, which falls under normal unsafe evolution rules.

jcouv (commented):

proposals/unsafe-evolution.md
RikkiGibson: > So stackalloc of managed type should likely remain an error (`ERR_ManagedAddr`).

Agree with this.

jjonescz: > For non-stackalloc scenarios, take a pointer to a managed type would become "safe" under the new memory safety rules (ie. no longer produce `WRN_ManagedAddr`. The GC is tracking types and references just fine. The problem is only when the user dereferences the pointer, which falls under normal unsafe evolution rules.

But today the warning is there even if you are under `unsafe` context. Isn't that because it's more unsafe and that wouldn't change? Similarly, this same warning is reported for `sizeof` of a managed type - should we remove that warning too and just make that completely safe?

I think I will keep the warning for now and turn this into an open question.

333fred (approved):

jcouv (approved):

PR #10114, opened 2026-04-18
Flesh out null-conditional-await proposal

CyrusNajmabadi:
Expanding the existing draft spec for `await?` with real rules that can be taken to LDM. Adds grammar, applicability (operand-nullability), awaitable-pattern resolution, classification, and run-time evaluation against the existing spec sections (ECMA §11.8.8 / §11.7.7 / §11.7.9), with spec deltas shown as bold additions in the style of `anonymous-using-declarations.md`. Also adds a worked-examples table and an interactions/edge-cases section.

Happy to iterate on anything here.


CyrusNajmabadi:
@jnm2 ptal :)

PR #10117, opened 2026-04-19
Add chained-comparison proposal

CyrusNajmabadi:
Specifies `a < b < c`, `min <= x <= max`, `a < b > c` and longer chains of `<`, `<=`, `>`, `>=` as a single-node recursive extension to relational binding: the classical left-associative meaning is preserved when it compiles, and chain semantics fire only at nodes where classical binding would have been a binding-time error. Middle operands are evaluated exactly once, left-to-right, with short-circuit on the first false link.


CyrusNajmabadi:
@jnm2 ptal :)

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,321 @@
...
+operand twice:
+
+```csharp
+if (min <= x && x <= max) { … }
+if (0 <= i && i < array.Length) { … }
+```
+
+This is awkward in three distinct ways:
+
+**It does not match mathematical notation.** Every student learns to write
```
CyrusNajmabadi:
```suggestion
**It does not match mathematical notation.** Students commonly learn to write
```

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,321 @@
...
+`min ≤ x ≤ max`. C# is the only place the same inequality needs to be written
+as two separate comparisons joined by `&&`, with the pivot repeated.
+
+**It is a double-evaluation correctness trap.** The natural rewrite of
+`Min < ComputeVal() < Max` as `Min < ComputeVal() && ComputeVal() < Max`
+evaluates `ComputeVal()` *twice*. That is wrong when the middle expression has
+side-effects, is non-deterministic, or is merely expensive. The typical
+workaround — binding the middle expression into a local — destroys the
```
CyrusNajmabadi:
```suggestion
workaround (binding the middle expression into a local), destroys the
```

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,321 @@
...
+```csharp
+if (min <= x && x <= max) { … }
+if (0 <= i && i < array.Length) { … }
+```
+
+This is awkward in three distinct ways:
+
+**It does not match mathematical notation.** Students commonly learn to write
+`min ≤ x ≤ max`. C# is the only place the same inequality needs to be written
+as two separate comparisons joined by `&&`, with the pivot repeated.
```
CyrusNajmabadi:
```suggestion
`min ≤ x ≤ max`.
```

CyrusNajmabadi:
Note: please do not merge yet. There's a simplification I want to do later today to remove some repetition from this spec.

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,320 @@
...
+
+// What the user has to write today to avoid double-evaluation:
+var tmp = ComputeVal();
+if (Min < tmp && tmp < Max) { … }
+```
+
+**It is a copy-paste hazard.** The two-comparison form `a < b && c < d` reads
+naturally as a chained comparison at a glance, but silently fails to test
+a chain if `c` was meant to be `b`. There is no shorthand that makes the
+correct spelling the *only* spelling.
```
jnm2: I get the point here, but I don't see what it has to do with copy-paste.

Also, in general, this type of substitution error is just one example of a far wider class of substitution errors which are not typically solved by introducing new syntax. Introducing new syntax for just this one case falls a bit short of being motivation for the feature personally. I see other benefits, but "moves the needle 1/10000th on substitution errors" isn't it (for me).

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,320 @@
...
+
+There is even a third-party NuGet workaround (referenced in the linked
+discussion thread for this proposal) that uses operator-overload trickery to
+approximate the feature, which is evidence of real demand in the ecosystem.
+
+## Detailed design
+
+The following updates are presented as a diff against the corresponding
+sections of the C# 6 standard
+([expressions.md](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md)).
```
jnm2, line 97: Same comment as in other PR: would prefer draft-v8, or at least standard-v7

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,320 @@
...
+
+### §11.11.13 Chained relational comparison (new subsection)
+
+Add the following new subsection at the end of
+[§11.11](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#1111-relational-and-type-testing-operators),
+after [§11.11.12 The as operator](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#111112-the-as-operator):
+
+**This subclause defines the binding and run-time evaluation of
+*relational_expression*s whose top-level operator is one of `<`, `<=`, `>`, or
+`>=` and whose left operand is itself such a *relational_expression*. Because
```
jnm2: Is this worded strictly enough to not be applying to `((a | b) < c) < d`?
- The top level operator is `<` (from `< d`)
- the left operand is `(a | b) < c`, which is itself a relational expression due to the `<` in `< c`

proposals/chained-relational-comparison.md
```diff
@@ -0,0 +1,320 @@
...
+the existing specification, and **bold** indicates text being added. Unchanged
+prose is quoted verbatim for context.
+
+### Grammar
+
+No grammar change. The existing production for *relational_expression* in
+[§11.11.1](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11111-general)
+already parses `a < b < c < …` left-associatively as
+`((a < b) < c) < …`. This proposal changes only how such a tree is bound and
+evaluated.
```
jnm2: Can you clarify somewhere that `(a < b) < c` (with the explicit parenthesized_expression) is not proposed to behave as `a < b && b < c`?

jnm2 (approved):
Very neat. The binding of the operators matches the binding used in the extended form with `&&`.

proposals/chained-relational-comparison.md
CyrusNajmabadi: Fair. This is just a restatement that a duplication can lead to bugs. I'll trim.

proposals/chained-relational-comparison.md
CyrusNajmabadi: My upcoming edit will make this clear. This is not ready for review :-)

proposals/chained-relational-comparison.md
CyrusNajmabadi: Absolutely.  It does fall out. But I had the same thought in the bus that we should be explicit here. As sometimes there use a view that parens don't change anything. And here it should.

CyrusNajmabadi:
@jnm2  feel free to do another pass if you want.  otherwise, merge if you're happy.

proposals/chained-relational-comparison.md
```diff
@@ -6,10 +6,11 @@
 ## Summary
 
 Allow *relational_expression*s such as `a < b < c`, `min <= x <= max`, and
-`0 <= i < array.Length` to have the intuitive chained-comparison meaning — roughly
-`a < b && b < c`, `min <= x && x <= max`, `0 <= i && i < array.Length` —
-without requiring the middle operand to be spelled twice. This extends naturally
-to `<`, `<=`, `>`, `>=` combined in any order and at any length.
+`0 <= i < array.Length` to have the intuitive chained-comparison meaning,
+without requiring the middle operand to be spelled twice. The semantics are
+roughly those of `a < b && b < c`, `min <= x && x <= max`, and
```
jnm2, line 11: Instead of "roughly" which leaves things open and vague, maybe it could be "the semantics are those of" with the explicit exception that the middle operand is evaluated only once? (Or is there more difference than that?)

jnm2 (approved):

PR #10126, opened 2026-04-21
Partial extension members: initial proposal

jcouv:


proposals/partial-extension-members.md
```diff
@@ -0,0 +1,149 @@
...
+// generated:
+public static partial class E
+{
+    extension(int i)
+    {
+        public partial void M() { } // error: A partial member must be declared within a partial type
+    }
+}
+```
+
```
michael-hawker, line 60: @Sergio0694 would it be helpful here to call our specific use-case for the Toolkit Attached Property generator as motivation? i.e. this isn't a hypothetical to allow SG, we have a practical use for this we want to allow for consumption to simplify a complex API surface? (It's also interesting as it's on the app dev side vs. library dev side, we're making a SG to aid the end-dev with an extension pattern here.)

Sergio0694: We already mentioned that use case scenario in the linked issue, and I've chatted about it with Julien as well. I believe this doc is just meant to be purely a language specification thing, but we'll definitely use the Toolkit as a motivation scenario when discussing this in LDM. That said of course, @jcouv if you think we should also include any of those motivating scenarios in this doc as well, then sure 😄

jcouv: Yup, I have a link below to the issue with your motivating scenario and some other customers' as well. Thanks!

proposals/partial-extension-members.md
```diff
@@ -0,0 +1,149 @@
...
+E.M(t: 1, i1: 2); // ok
+1.M(i1: 2); // ok
+
+E.M(t2: 1, 2); // error
+E.M(1, i2: 2); // error
+1.M(i2: 2); // error
+
+partial class E
+{
+    extension<T>(T t) { public partial T M(int i1);
```
jjonescz:
```suggestion
    extension<T>(T t) { public partial T M(int i1); }
```

proposals/partial-extension-members.md
```diff
@@ -0,0 +1,149 @@
...
+
+Attributes on regular partial methods or their parameters get merged across all parts.
+But extension block members are associated with a specific marker type name
+which includes any attributes on the extension parameter.
+
+What should we do if the extension parameter has different attributes in the definition and implementation parts?
+```cs
+extension([A] string s) { public partial void M(); }
+extension([B] string s) { public partial void M() { } }
+ ```
```
jjonescz:
````suggestion
```
````

jjonescz (approved):

Issue #9896, opened 2025-12-19
[Proposal]: Compound assignment in object initializer and `with` expression


jnm2:
# Compound assignment in object initializer and `with` expression

- Specification: https://github.com/dotnet/csharplang/blob/main/proposals/compound-assignment-in-initializer-and-with.md
- Discussion: https://github.com/dotnet/csharplang/discussions/8650

## Summary

Allow compound assignments in an object initializer:

```cs
var timer = new DispatcherTimer
{
    Interval = TimeSpan.FromSeconds(1d),
    Tick += (_, _) => { /*actual work*/ },
};
```

Or a `with` expression:

```cs
var newCounter = counter with { Value -= 1 };
```

## Design meetings

https://github.com/dotnet/csharplang/blob/main/meetings/2021/LDM-2021-09-20.md#object-initializer-event-hookup

PR #10129, opened 2026-04-21
Flesh out detailed design for compound assignment in object initializer and `with`

CyrusNajmabadi:
Expands the Detailed design section of `proposals/compound-assignment-in-initializer-and-with.md` (champion #9896) so the proposal carries real normative rules rather than a sketch.

## What changes

Against the existing proposal, the Detailed design is rewritten as a standard-style diff against §11.7.15.3 and §11.18 of the C# 6 standard, plus the with-expression subsection of the C# 9 records proposal. In summary:

* **Grammar.** Adds a new `compound_assignment_operator` production and extends `member_initializer` (both object initializer and `with`) with a second alternative so that the compound form admits only *expression*, not a nested *object_or_collection_initializer*. This makes `P += { 1, 2 }` ill-formed at the grammar level rather than by prose.
* **Targets.** Allows events as valid `initializer_target`s (and as `with` targets), with the constraint that only `+=` / `-=` are permitted on an event target, per §11.18.4 / §11.18.1.
* **Semantics via lowering.** Specifies a `member_initializer` `target op value` as being equivalent to the statement_expression `x.target op value;` on the temporary holding the new instance (or the converted clone, for `with`). This makes simple assignment, compound assignment, and event assignment all fall out of §11.18 uniformly, and in particular satisfies §11.18.4's statement_expression-context requirement for event assignment by construction.
* **Same-target uniqueness.** Relaxes the existing "at most one member initializer per field or property" rule to permit any number of compound member initializers for the same target, with at most one `=` for that target and the `=` (if present) appearing first in lexical order. This is what makes `Click += h1, Click += h2` and `Value = 10, Value += 5` legal.
* **Accessor requirements.** Adds an explicit rule: compound member initializers require a readable target (via a `get` accessor, or a ref-returning getter which classifies the access as a variable). On the write side, object initializers and `with` differ: object initializers require `set`, whereas `with` allows `set` or `init` because the whole `{ ... }` is an init-allowed context for the clone.
* **Worked examples.** Adds the `__tmp`-style expansions used elsewhere in §11.7.15.3, for both object initializers and `with`.
* **Interactions.** Adds short notes for indexer targets, dynamic, collection initializers (unaffected), and the read-from-clone semantics for `with`.

## Open for LDM

The proposal currently picks a specific answer for each of the design questions (events in `with` yes, same-target compound yes, `init` allowed on `with`-compound but not on object-initializer-compound). Any of those can be reversed without cascade; they are all called out explicitly in the Design decisions section.


CyrusNajmabadi:
@jnm2 ptal.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -22,70 +22,223 @@ var newCounter = counter with { Value -= 1 };
...
+
+ initializer_value
+     : expression
+     | object_or_collection_initializer
      ;
+
++compound_assignment_operator
++    : '+=' | '-=' | '*=' | '/=' | '%=' | '&=' | '|=' | '^=' | '<<='
++    | right_shift_assignment
++    ;
```
jnm2, line 69: Can `compound_assignment_operator` now be used in the place that these productions copied from?

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -22,70 +22,223 @@ var newCounter = counter with { Value -= 1 };
...
 
-```diff
-member_initializer
--    : identifier '=' expression
-+    : identifier assignment_operator expression
-    ;
+~~Each *initializer_target* is followed by an equals sign and either an expression, an object initializer or a collection initializer.~~ **When the *member_initializer* uses the `=` operator, the *initializer_target* is followed by either an expression, an object initializer, or a collection initializer. When the *member_initializer* uses a *compound_assignment_operator*, the *initializer_target* is followed by an expression.** ~~It is not possible~~ **When the `=` operator is used, it is not possible** for expressions within the object initializer to refer to the newly created object it is initializing. **When a *compound_assignment_operator* is used, both the read and the write of the target occur on the newly created object. Member initializers are processed in lexical order, so the read side of a compound member initializer observes the value established by any preceding member initializer for the same target.**
```
jnm2, line 80: There is no guarantee that the read side observes the value established by a preceding member because there is no guarantee by the language that a property will return the same value that was just assigned.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -22,70 +22,223 @@ var newCounter = counter with { Value -= 1 };
...
 ```
 
-By requiring that all collection elements are `non_assignment_expression`, `a += b` is already disallowed as that is an assignment_expression.
+where *compound_assignment_operator* is the production defined in [§11.7.15.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#117153-object-initializers).
 
-## Open questions
+The prose of the [`with` expression](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/records.md#with-expression) subsection is updated as follows.
 
-Is this feature needed? For example, users could support some of these scenarios doing something like the following:
+On the right hand side of the `with` expression is a `member_initializer_list` with a sequence of assignments to *identifier*, which must be an accessible instance field**,** ~~or~~ property**, or event** of the receiver's type. **An *identifier* that names an event is a valid target only in combination with the `+=` or `-=` operator; every other *assignment_operator* on an event target is a compile-time error.**
```
jnm2, line 147: The way the `**` is added is not rendering as markdown

Image

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -22,70 +22,223 @@ var newCounter = counter with { Value -= 1 };
...
+
+First, receiver's "clone" method (specified above) is invoked and its result is converted to the receiver's type. ~~Then, each `member_initializer` is processed the same way as an assignment to a field or property access of the result of the conversion. Assignments are processed in lexical order.~~ **Then, for each `member_initializer` `target op value` in lexical order, the *statement_expression* `x.target op value;` is executed, where `x` is the otherwise invisible and inaccessible temporary variable holding the converted clone. The meaning of that *statement_expression* is given by [§11.18](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#1118-assignment-operators), as for an object initializer. When `op` is a *compound_assignment_operator*, both the read and the write of `target` are performed on the clone `x`; the original receiver of the `with` expression is not read again after its clone method has returned.**
```
jnm2, line 151: If this was worth saying for object initializers, it seems worth saying the equivalent for `with` expressions.

> The equivalence is evaluated in the init-allowed context of the enclosing object initializer, so an init-only accessor remains callable as the write side regardless of whether op is = or a compound_assignment_operator.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -22,70 +22,223 @@ var newCounter = counter with { Value -= 1 };
...
...ard-v6/standard/expressions.md#11182-simple-assignment)) when `op` is `=`, compound assignment ([§11.18.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11183-compound-assignment)) when `op` is a *compound_assignment_operator* and the target is not an event, and event assignment ([§11.18.4](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11184-event-assignment)) when `op` is `+=` or `-=` and the target is an event. In particular, the *statement_expression* context required by [§11.18.4](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11184-event-assignment) is satisfied by this lowering, and the get-and-set requirement on property and indexer targets imposed by [§11.18.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11183-compound-assignment) applies to the target of a *compound_assignment_operator* member initializer. The equivalence is evaluated in the init-allowed context of the enclosing object initializer, so an `init`-only accessor remains callable as the write side regardless of whether `op` is `=` or a *compound_assignment_operator*.**
```
jnm2, line 82: Here's an example where it's too bad you're using v6 and not v7, where ref-returning properties exist, or draft-v8 where ref-returning properties exist and are being better specced. The get-and-set requirement here misses ref-returning properties. The "interactions with other features" section mentions ref-returning properties being an alternative to needing a setter.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -22,70 +22,223 @@ var newCounter = counter with { Value -= 1 };
...
+
+  This works for non-`init` members, but not for `init`-only properties, and loses the first-class expression shape.
+
+- **Restrict to events only.** A narrower feature that would permit `Event += handler` in an object initializer but not `Property += 1`. This addresses the headline UI scenario but excludes the accumulation-into-property and accumulation-into-cloned-value scenarios that motivate the `with` case.
+
+## Design decisions
+
+### Why allow multiple member initializers for the same target?
+
+The single most compelling motivation for this feature is events, and events naturally chain: subscribing two handlers to the same event in one initializer is useful and unsurprising. Forbidding this would make the common case (`Click += h1, Click += h2`) illegal in exactly the contexts where the feature is most valuable. The same reasoning extends to compound operators on properties (`Value = 10, Value += 5`), where each step has an observable effect.
```
jnm2, line 222: Events are trivially solved as `Click += h1 + h2` though.

jnm2 (approved):

proposals/compound-assignment-in-initializer-and-with.md
CyrusNajmabadi: I'm not saying you get the same value. But that you can observe since you have done a read.

proposals/compound-assignment-in-initializer-and-with.md
CyrusNajmabadi: It doesn't render in PR view. It renders fine when finally merged and you're just in the mark down view.

proposals/compound-assignment-in-initializer-and-with.md
CyrusNajmabadi: You still need this because you might do: `e -= oldTarget, e += newTarget`

hez2010:
> Forbidding this would make the common case (Click += h1, Click += h2) illegal in exactly the contexts where the feature is most valuable. 

I don't think so, because you will be able to do `Click += (SomeEventHandler)h1 + h2`. I think a better way is to allow operators on method group directly without requiring it to be converted to a delegate first. Then `Click += h1 + h2` will be enabled automatically.

proposals/compound-assignment-in-initializer-and-with.md
jnm2: I think the wording is inaccurate because the value returned from a property might not be a value established by a preceding call to the setter.

```cs
int Prop { get => condition ? 42 : field; set; }
```

If `condition` is met, the value observed is specifically not the value established by any preceding member initializer.

proposals/compound-assignment-in-initializer-and-with.md
jnm2: It's not PR view, it's GitHub file view that I was using. This will be the same view that this proposal is primarily viewed on.

proposals/compound-assignment-in-initializer-and-with.md
jnm2: That would be a stronger case to state!

PR #10132, opened 2026-04-22
Re-anchor compound assignment proposal to standard-v7

CyrusNajmabadi:
Follow-up to the merged #10129. Updates `proposals/compound-assignment-in-initializer-and-with.md` to track the current standard and clean up a few points raised in review.

## Standard update

All cross-references and URL anchors are re-anchored from `standard-v6` to `standard-v7`. Section numbers shift:

* §11.7.15.3 → §12.8.16.3 (Object initializers)
* §11.7.15.4 → §12.8.16.4 (Collection initializers)
* §11.18 → §12.21 (Assignment operators)
* §11.18.1 → §12.21.1 (General)
* §11.18.2 → §12.21.2 (Simple assignment)
* §11.18.3 → §12.21.4 (Compound assignment; v7 inserted §12.21.3 Ref assignment)
* §11.18.4 → §12.21.5 (Event assignment)
* §11.19 → §12.22 (Expression)
* §11.7.6 → §12.8.7 (Member access)

## Reviewer feedback addressed

* **GitHub rendering of the "field or property or event" diff.** The previous form used `**,** ~~or~~ property**, or event**`, which GitHub renders awkwardly because of the bare bolded comma. Switched to append-only `or property **or event**` in both the object initializer prose and the `with` prose.
* **"observes the value established by a preceding member initializer"** was reworded so it is about sequencing rather than value equivalence. A get accessor is not obliged to return the value that was last assigned, so the spec should not imply that. The new wording says the compound form's read takes place after every preceding member initializer has executed, and reflects whatever state the target's get accessor reports at that point. An equivalent sentence is now in the `with` prose for parity.
* **"Evaluated in the init-allowed context"** phrasing was dropped from the main normative paragraph and from the Interactions bullet. The practical rule (init is accepted wherever set would be, on the same terms as for a direct `=` member initializer) is stated plainly.
* **Can `compound_assignment_operator` be used elsewhere in the standard?** Added a Design decisions subsection surveying the question. One candidate site: §12.21.1's definitional sentence for "compound assignment operators" could name the production. Other occurrences in §15.7.3 and §15.7.5 are prose references to the concept and already link to §12.21.4, so they do not need grammar updates.


CyrusNajmabadi:
@jnm2 ptal.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -135,20 +158,20 @@ The following diff is applied to the grammar of the [`with` expression](https://
...
  member_initializer
-     : identifier '=' expression
-+    | identifier compound_assignment_operator expression
+-    : identifier '=' expression
++    : identifier assignment_operator expression
      ;
 ```
 
-where *compound_assignment_operator* is the production defined in [§11.7.15.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#117153-object-initializers).
+Because both the previous `=` form and every *compound_assignment_operator* form share the same right-hand side (an *expression*), the two alternatives collapse into a single *assignment_operator* production from [§12.21.1](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12211-general). The `= ref` form of *assignment_operator* is not applicable to a `member_initializer` target (a field, property, or event access is not a reference variable, so [§12.21.3](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12213-ref-assignment)'s left-operand requirement rejects it at binding time).
```
jnm2, line 166: This is not correct, C# 11 adds ref fields (https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#provide-ref-fields-and-scoped)

so `new C { RefField = ref x }`  would be allowed, contrary to this paragraph.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -62,26 +88,23 @@ The following diff is applied to the grammar in [§11.7.15.3](https://github.com
...
...nt_expression* is given by [§12.21](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#1221-assignment-operators): simple assignment ([§12.21.2](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12212-simple-assignment)) when `op` is `=`, compound assignment ([§12.21.4](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12214-compound-assignment)) when `op` is a *compound_assignment_operator* and the target is not an event, and event assignment ([§12.21.5](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12215-event-assignment)) when `op` is `+=` or `-=` and the target is an event. In particular, the *statement_expression* context required by [§12.21.5](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12215-event-assignment) is satisfied by this lowering, and the get-and-set requirement on property and indexer targets imposed by [§12.21.4](https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/expressions.md#12214-compound-assignment) applies to the target of a *compound_assignment_operator* member initializer.**
```
jnm2, line 105: This mentions get-and-set requirement, that's not a great term. Ref-returning properties do not need a set accessor. This is mentioned in the feature interaction summary, but it is not recognized here.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -219,21 +242,27 @@ As with any language feature, the additional specification complexity must be we
...
 
-### Why require `get` on a compound target?
+## Open LDM questions
 
-Compound assignment is defined by [§11.18.3](https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/expressions.md#11183-compound-assignment) as a read-modify-write: the target must be both readable and writable. A compound *member_initializer* therefore requires a readable target, whereas a plain-`=` member initializer requires only a writable one. A ref-returning `get` produces a variable-classified access and is sufficient for both the read and the write, as it is in any other compound assignment.
+### Does a *compound_assignment_operator* member initializer satisfy a `required` member?
 
-### Why specify the semantics as a lowering to *statement_expression*?
+A `required` field or property must be assigned by every object creation that does not forward the requirement (`SetsRequiredMembersAttribute`). The existing rule is expressed in terms of `=` member initializers; this proposal introduces a second form of member initializer that targets the same member, so the question is whether the compound form also discharges the `required` obligation.
```
jnm2, line 257: The question for me is not whether it discharges the obligation (it clearly assigns), but rather whether the `required` feature is intended to prevent default values from being preserved.

For instance, you could do `new C { RequiredProp += 0 }` to explicitly preserve a default rather than bringing your own value.

jnm2 (approved):

PR #10137, opened 2026-04-22
Include `??=` in compound_assignment_operator

CyrusNajmabadi:
`??=` was added to the language in C# 8, after the ECMA-334 v7 standard this proposal diffs against. Include it in the `compound_assignment_operator` production so the rest of the proposal covers it uniformly — no separate interaction rules needed.


CyrusNajmabadi:
@jnm2 plz merge.

PR #10139, opened 2026-04-23
Unsafe evolution speclet: update attribute synthesis to match implementation

jjonescz:
The implementation (https://github.com/dotnet/roslyn/pull/83295) does not synthesize the RequiresUnsafe attribute (I don't see a reason to do that and I didn't want to complicate the implementation).


proposals/unsafe-evolution.md
```diff
@@ -247,7 +247,8 @@ When a member under the new memory safety rules is `extern`, the compiler will i
 When a user-facing *requires-unsafe* member generates hidden members, such as an auto-property's get/set methods,
 both the user-facing member and any hidden members generated by that user-facing member are all *requires-unsafe*, and `RequiresUnsafeAttribute` is applied to all of them.
 
-The `MemorySafetyRulesAttribute` and `RequiresUnsafeAttribute` definition is synthesized by the compiler if necessary per standard well-known member rules.
+The `MemorySafetyRulesAttribute` definition is synthesized by the compiler if necessary per standard well-known member rules.
+The `RequiresUnsafeAttribute` definition is *not* synthesized, and the compiler reports an error if the attribute is needed but its definition is missing.
```
333fred, line 251: Now that we're back to `unsafe` as a keyword, we probably need to go back to synthesizing.

jjonescz: Why though? There are many attributes we don't synthesize (e.g., RequiredMemberAttribute), I'm not actually sure how we decide that.

But I don't have a particularly strong reason to not synthesize, so I can change the implementation if you prefer that.

333fred: Yes, we are not super consistent about it. In more recent years, we've tried to come with a general rule of "If a user expected to apply the type by hand, don't synthesize it". But, of course, we do have features like `required` that break this.

PR #10142, opened 2026-04-25
Add proposal: extension members on typeless receivers

CyrusNajmabadi:
Add a new proposal that lifts the requirement that the receiver of an extension method or extension member call have a type.

The motivating scenario is `[a, b, c].ToImmutableArray()` on a collection expression. Lambdas without a natural type, overloaded method groups (`Console.WriteLine`, `Math.Max`, `string.Format`), and conditional or switch expressions without a common arm type fall out under the same uniform rule.

The diff is a single bullet against §12.8.9.3 of the C# 7 standard and against the *Consumption* section of the C# 14 extension members speclet, adding a typeless branch to the existing receiver-eligibility test. Existing typed-receiver behavior is unchanged; this is a pure extension.

LDM 2023-07-12 named this as a separate orthogonal track from a collection-expressions-only ergonomic; this proposal is that track. Champion issue and discussion will be filed against the merged spec.


CyrusNajmabadi:
@jnm2 ptal.  Feedback welcome.  If you're happy, merge as is.  Minor feedback can be incorporated later. A followup pr will be made to link to the issue/discussion for this proposal doc, and can include tweaks at that time as well.

CyrusNajmabadi:
@jcouv ptal.

jnm2 (approved):

PR #10144, opened 2026-04-27
Simplify null-conditional-await proposal: defer to §12.8.8 / §12.8.10

CyrusNajmabadi:
## Summary

The detailed-design section of the `null-conditional-await` proposal previously enumerated five classification cases, four operand-applicability cases, and four runtime-evaluation steps in parallel to the existing rules for *null-conditional member access* (§12.8.8) and *null-conditional invocation expression* (§12.8.10). Every one of those cases was a mechanical cross-product of `?.` and `await`.

Recast around a single intuition: **`await? t` is the *null_conditional_member_access* `(t)?.GetAwaiter().GetResult()` per §12.8.8, with `await` semantics applied on the non-null branch.** The classification (lifting to `Nullable<R>`, the type-parameter result restriction, the pointer-type result case, the already-`Nullable<V>` case, NRT annotation), the *void* → §12.8.10 routing, the null-test shape, the `Nullable<V>` operand handling (§12.8.8's `P.Value.A` rule), and the single-evaluation guarantee all fall out of §12.8.8 / §12.8.10 without restatement. Future evolution of those sections (e.g. recent LDM clarifications around pointer results) is automatically inherited.

The only carve-out kept normatively is the operand-side compile-time error on a known non-nullable value-type operand (concrete struct, `where T : struct`, `where T : unmanaged`): `await?` rejects these even though §12.8.8 would admit them with a trivially-false null-test.

## What changed

- **New** *"`await? t` builds on `?.` and `await`"* intuition section. States the equivalence both as `(t)?.GetAwaiter().GetResult()` and as the operational form `((object)t == null) ? default(X) : await t` / `(!t.HasValue) ? default(X) : await t.Value`. A reader who stops at the end of this section has a complete operational mental model.
- **§12.9.8.5 Applicability** collapses to: (a) `S` admissible as `P` in `P?.A` per §12.8.8, (b) value-type-operand carve-out, (c) awaitable pattern on `(t)?.GetAwaiter()`.
- **§12.9.8.2 / §12.9.8.3 / §12.9.8.4** each become a single short paragraph cross-referencing §12.8.8 / §12.8.10 and naming what `await?` plugs in.
- **"Underlying type `U`" auxiliary concept dropped** throughout. §12.8.8's existing nullable-value-type-receiver rule supplies the same behavior.
- **Worked-examples Tables A and B** retained verbatim as non-normative; their headers updated to point at the inheriting subsections.

## Net effect

- No observable behavior change.
- The detailed-design body is shorter and easier to skim.
- Future spec changes to §12.8.8 / §12.8.10 (the LDM has been clarifying these recently around pointer results, type-parameter receivers, etc.) propagate automatically into `await?`.

### DIFF
--- a/proposals/null-conditional-await.md
+++ b/proposals/null-conditional-await.md
@@ -29,14 +29,31 @@ which does nothing if `GetX()` was null, and otherwise awaits the task. This com
 ## Detailed design
 [design]: #detailed-design
 
-### Two independent rules
+### `await? t` builds on `?.` and `await`
 
-The semantics of `await? e` are described by two orthogonal rules that are applied independently. Every case in the worked-examples table at the end of this section is a mechanical cross-product of these two rules; the behaviors of `Task`, `Task<T>`, `ValueTask`, `ValueTask<T>`, and arbitrary user-defined awaitables are all consequences of them, not normative cases of their own.
+The semantics of `await? t` are, both intuitively and literally, the composition of two existing language features:
 
-- **Operand-nullability rule**: what does the static type `S` of `e` tell us about whether `e` can be null at runtime? This drives the compile-time error cases and the shape of the runtime null-test. It is entirely independent of what the awaitable produces.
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
@jnm2

PR #10148, opened 2026-04-27
Apply LDM resolutions on `required` and `target = { ... }` exclusivity in compound-assignment-in-initializer

CyrusNajmabadi:
Two LDM resolutions on the [Compound assignment in object initializer and \`with\` expression](https://github.com/dotnet/csharplang/blob/main/proposals/compound-assignment-in-initializer-and-with.md) proposal. Both are narrow rules added to the object-initializer prose with short motivation notes.

### \`required\` discharge

In an object initializer, a \`required\` member is satisfied only by a \`=\` member initializer; a *compound_assignment_operator* member initializer does not on its own discharge the requirement. A \`required\` target may additionally appear under compound member initializers in the same object initializer, subject to the existing "\`=\` before any compound" ordering. In a \`with\` expression no such restriction applies — the receiver has already been constructed and its \`required\` members satisfied by that earlier construction.

*Motivation:* \`required\` ensures each marked slot is in a valid state before any read of it. A compound member initializer reads the slot before writing it; on a freshly constructed object that read sees the default value, which is exactly the state \`required\` is meant to prevent observation in. A \`with\` clone has already been initialized once, so the read is safe.

This resolves the open LDM question previously in the spec ("Does a *compound_assignment_operator* member initializer satisfy a \`required\` member?"), which is removed.

### \`target = { ... }\` exclusivity

A *member_initializer* whose *initializer_value* is an *object_or_collection_initializer* (the \`target = { ... }\` form) is now *exclusive*: at most one such *member_initializer* per target, and no other *member_initializer* (whether \`=\`, another \`= { ... }\`, or a compound) may target the same field, property, event, or indexer-with-equivalent-arguments.

*Motivation:* the \`= { ... }\` form configures the nested instance the target already references; combining it with a slot-overwriting \`=\` or a read-modify-write compound would either discard the nested configuration or read it mid-configuration.

(The \`with\` grammar already excludes nested initializers — \`identifier assignment_operator expression\` admits only \`expression\` on the right — so this rule has no effect on \`with\` clauses.)


CyrusNajmabadi:
@jcouv ptal.

jnm2 (approved):

PR #10151, opened 2026-04-28
Drop redundant event-operator-restriction text in compound-assignment-in-initializer spec

CyrusNajmabadi:
Follow-up cleanup of the recently-merged spec for `compound-assignment-in-initializer-and-with`. Two redundant sentences (one in the object-initializer paragraph, one in the `with`-expression paragraph) call out that an event target is "only valid in combination with `+=` or `-=`" — but that already falls out from:

1. **The lowering paragraph** specifies `target op value` as semantically equivalent to the *statement_expression* `x.target op value;` and defers to §12.21.
2. **§12.21.1** restricts compound assignment to "variable, property, or indexer element" left operands — an event access doesn't qualify.
3. **§12.21.5** covers the only two valid event forms (`+=` and `-=`).

So `Event *= h`, `Event ??= h`, etc. are ill-formed by exclusion through the existing C# Standard machinery. The "Why specify the semantics as a lowering to *statement_expression*?" design-decisions section already explicitly argues this:

> "[§12.21.5] already requires event assignment to appear in a *statement_expression* context; the lowering satisfies that requirement by construction, **without special-casing events at the initializer level**."

The two sentences this PR removes were exactly that special-casing — they contradicted our own design narrative.

## Also folds events into the uniqueness-rule scope

The other small change: events are now subject to the same "at most one `=`, any number of compound, `=` before any compound" rule as fields and properties, *vacuously*. `=` on an event isn't valid in the first place, so the `=`-related parts of the rule are trivially satisfied; "any number of compound permitted" is exactly what enables `Click += h1, Click += h2` handler stacking. No need for an event-specific carve-out.

Indexers stay carved out — the same indexer with different argument lists names different keys, so the rule's "same target" notion is ambiguous (the existing "same indexer arguments multiple times" allowance is preserved by the *Note* a couple paragraphs down).

## Diff

The two paragraphs become:

```diff
- An *identifier* shall name an accessible field or property **or event** of the object being initialized, … . **An *identifier* that names an event is a valid target only in combination with the `+=` or `-=` operator (§12.21.5); every other *assignment_operator* on an event target is a compile-time error, because no other assignment operator is valid with an event access as the left operand (§12.21.1).** ~~It is an error for an object initializer to include more than one member initializer for the same field or property.~~ **For any given field or property target, … . No such restriction applies to event or indexer targets.**
+ An *identifier* shall name an accessible field or property **or event** of the object being initialized, … . ~~It is an error for an object initializer to include more than one member initializer for the same field or property.~~ **For any given field, property, or event target, … . No such restriction applies to indexer targets.**
```

```diff
- On the right hand side of the `with` expression … must be an accessible instance field or property **or event** of the receiver's type. **An *identifier* that names an event is a valid target only in combination with the `+=` or `-=` operator; every other *assignment_operator* on an event target is a compile-time error.**
- **For any given field or property target, … . No such restriction applies to event targets.**
+ On the right hand side of the `with` expression … must be an accessible instance field or property **or event** of the receiver's type.
+ **For any given field, property, or event target, … .**
```

Net change: -3 lines / +3 lines, spec is shorter, self-consistent with its own design notes, and tracks the C# Standard's existing event-access rules without restating them.

## Related

- Champion issue: dotnet/csharplang#9896
- Prior LDM resolutions PR: #10148


CyrusNajmabadi:
@jnm2 ptal.

proposals/compound-assignment-in-initializer-and-with.md
```diff
@@ -97,7 +97,7 @@ where *compound_assignment_operator* is the production introduced in [§12.21.1]
...
+An object initializer consists of a sequence of member initializers, enclosed by `{` and `}` tokens and separated by commas. Each *member_initializer* shall designate a target for the initialization. An *identifier* shall name an accessible field or property **or event** of the object being initialized, whereas an *argument_list* enclosed in square brackets shall specify arguments for an accessible indexer on the object being initialized. ~~It is an error for an object initializer to include more than one member initializer for the same field or property.~~ **For any given field, property, or event target, at most one member initializer may use the `=` operator. Any number of member initializers using a *compound_assignment_operator* are permitted for the same target. If both are present for the same target, the `=` member initializer shall appear in lexical order before any *compound_assignment_operator* member initializer for that target. No such restriction applies to indexer targets.**
```
jnm2: Instead of "if both are present," wdyt of saying "If present, an `=` member initializer shall appear in lexical order before any other member initializer for that target."

This wording points simplifies slightly and also blocks having two `=` initializers of the same target in one swoop. This feels like it points more at the fundamental constraint IMO.

jnm2 (approved):

proposals/compound-assignment-in-initializer-and-with.md
CyrusNajmabadi: Done.  merge plz @jnm2

PR #10158, opened 2026-04-30
Update labeled-break-continue.md with code example

jcouv:
FYI @CyrusNajmabadi @jjonescz. Adding a code sample and some references for discussion

### DIFF
--- a/proposals/labeled-break-continue.md
+++ b/proposals/labeled-break-continue.md
@@ -322,6 +322,17 @@ void M()
 }
 ```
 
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
It should fail because a continue/break with a label references the innermost applicable loop/switch with that label directly on it. 

That said. It makes no difference ultimately. The same set of programs are accepted and rejected. So there is no actual distinction here. The rules are isomorphic.

PR #10173, opened 2026-05-12
Unsafe evolution speclet: bring back overload resolution break

jjonescz:
That break actually doesn't go away even with extending compat mode to non-opted-in callers. See the implementation PR: https://github.com/dotnet/roslyn/pull/83660


jcouv (approved):
LGTM Thanks (commit 1)

333fred (approved):

PR #10180, opened 2026-05-14
Unsafe evolution speclet: incorporate LDM decisions

jjonescz:


333fred (approved):

proposals/unsafe-evolution.md
```diff
@@ -209,12 +212,17 @@ class C
...
 and does *not* introduce an `unsafe` context (instead, only explicit `unsafe` regions in the body establish `unsafe` contexts).
 
-`unsafe` on the following declarations produces a warning because it does not have a meaning anymore:
+`unsafe` on the following declarations produces an error because it does not have a meaning anymore:
 - `delegate`,
 - static constructor,
 - destructor,
 - type declaration (`class`, `struct`, etc.).
 
+Only an `unsafe` constructor's initializer may call a *requires-unsafe* `base` or `this` constructor.
```
jcouv: Does that mean that `unsafe` on constructor introduces an unsafe-context just for the constructor initializer?
I think we should. 
So far, we only have one mechanism to suppress the diagnostic from requires-unsafe calls/usages, it's unsafe-contexts.

(Comment not blocking merging)

jjonescz: Yes, that's what it should mean, I will clarify, thanks.

proposals/unsafe-evolution.md
```diff
@@ -62,15 +62,21 @@ The following breaking changes can be observed when updating to a compiler imple
...
 ## Detailed Design
 
 Terminology: we call members *requires-unsafe* (previously known as *caller-unsafe*) if
-- under [the updated memory safety rules](#metadata) they [are marked `unsafe`](#metadata) or [are `extern`](#extern),
+- under [the updated memory safety rules](#metadata) they [are marked `unsafe`](#metadata),
 - under [the legacy memory safety rules](#metadata) they [contain pointers in signature](#compat-mode).
 
+### Syntax
+
+We introduce new `safe` contextual keyword. It can be applied to [`extern` members](#extern) and [fields in explicit layout](#fields).
```
jcouv, line 75: Can you use the `safe` keyword in a compilation where `AllowUnsafeBlocks==false`?

jcouv: Nevermind, I see the question is already captured (line 397)

PR #10184, opened 2026-05-18
Mixed object and collection initializers: initial proposal

CyrusNajmabadi:
Initial proposal to relax the existing restriction that an object initializer following `new T(...)` must contain **only** member initializers or **only** element initializers. A single initializer would be allowed to mix the two kinds in any order; each element keeps its existing per-kind meaning (member initializers assign to the named field/property/event/indexer, element initializers invoke `Add`).

* Champion issue: #10185
* Discussion: https://github.com/dotnet/csharplang/discussions/10186
* Earlier community discussion: https://github.com/dotnet/csharplang/discussions/4879

Open LDM question (called out in the proposal):

- Should members and elements be required to appear in groups, or may they be interleaved? Recommendation in the draft is to allow arbitrary interleaving.

### DIFF
--- a/proposals/mixed-object-and-collection-initializers.md
+++ b/proposals/mixed-object-and-collection-initializers.md
@@ -1,0 +1,184 @@
+# Mixed object and collection initializers
+
+Champion issue: <https://github.com/dotnet/csharplang/issues/10185>
+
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
@jnm2 ptal.

jnm2 (approved):

PR #10205, opened 2026-05-27
Resolve closed classes open questions

RikkiGibson:


proposals/closed-hierarchies.md
```diff
@@ -274,116 +360,4 @@ class C1
...
-
-    int M2(C c)
-        => c switch
-        {
-            C => 1, // ok
-        };
-}
-```
-
-### Allow matching the base type after matching all subtypes
```
RikkiGibson, line 302: This item was deleted rather than moved because the resolution was already captured in the text of **Exhaustiveness in switches**:

> On the other hand this also means that it can be an error for the closed base class to occur as a case after all its direct descendants:
> 
> ``` c#
> _ = cc switch
> {
>     CO co => ...,
>     CC cc => ..., // Error, case cannot be reached
> };
> ```

PR #10221, opened 2026-06-08
Add agent instructions

jcouv:
FYI @CyrusNajmabadi


CyrusNajmabadi:
Love it. 

Perhaps something about keeping the terminology and cadence of the existing spec?

HaloFour:
Why are you using both files?  Github Copilot is going to add both of their contents to the prompt anyway.  And for any other agent you're going to be relying on the LLM to decide to read `copilot-instructions.md` based on having `AGENTS.md` injected into the prompt, which even if that happens every time would just be a waste of a turn and tokens.

.github/copilot-instructions.md
333fred, line 1: Let's just use `AGENTS.md`, as we just moved to in roslyn.

333fred (approved):

PR #10235, opened 2026-06-18
Add proposal for await in unsafe contexts

333fred:
## Summary
- allow await expressions in unsafe contexts in the unsafe evolution speclet
- keep await disallowed inside fixed statements
- update unsafe expression examples now that await can appear in unsafe blocks

### DIFF
--- a/proposals/unsafe-evolution.md
+++ b/proposals/unsafe-evolution.md
@@ -169,6 +169,37 @@ Unlike other changes to `unsafe` rules which are relaxations, this is a tighteni
 
 Note that a `stackalloc` of a managed type remains an error.
 
+#### `await` in `unsafe` contexts
+
+`await` expressions are allowed in `unsafe` contexts. This permits an async method to directly await a *requires-unsafe* operation without first spilling the awaitable to a temporary variable:
+
+```cs
+async Task M()
+{
+    unsafe
+    {
+        await DoUnsafeAsync();
+    }
+}
+
+unsafe Task DoUnsafeAsync() => ...;
+```
+
+The `unsafe` context still represents the programmer's manual safety audit boundary. If a pointer, function pointer, or other unsafe value is kept live across an `await`, the language does not attempt to prove that the use is safe after resumption; that responsibility remains with the `unsafe` code author. This is consistent with the rest of unsafe code, which permits operations such as storing pointers in fields or using pointers after their referent's lifetime has ended.
+
+However, `await` remains disallowed inside a `fixed` statement. A `fixed` statement pins a moveable variable for the duration of the statement, and allowing suspension while the statement is active would require the pinning lifetime to span an async suspension point.
+
+The corresponding csharpstandard update is in [§15.15.1 Async functions > General][async-funcs-general]. Throughout this section, ~~strikethrough~~ indicates text being removed from the existing specification, and **bold** indicates text being added.
+
+> It is a compile-time error for the formal parameter list of an async function to specify
+> any `in`, `out`, or `ref` parameters, or any parameter of a `ref struct` type.
+>
+> It is a compile-time error for an unsafe context ([§24.2][unsafe-context-spec]) to contain
+> ~~an `await` expression ([§12.9.8][await-expressions]) or~~ a `yield return` statement ([§13.15][yield-statement]).
+>
+> **It is a compile-time error for a `fixed` statement ([§24.7][fixed-statement]) to contain
+> an `await` expression ([§12.9.8][await-expressions]).**
+
 #### `sizeof`
 
 For certain predefined types, `sizeof` has always been constant and safe ([§12.8.19][sizeof-const]) and that remains unchanged.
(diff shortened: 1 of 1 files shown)


jcouv (approved):

proposals/unsafe-evolution.md
```diff
@@ -568,12 +599,16 @@ Console.WriteLine(result);
...
+Should we relax restrictions around `await` expressions in `unsafe` contexts?
+Especially allowing `await UnsafeMethod()` is useful because otherwise users have to rewrite that to `Task t; unsafe { t = UnsafeMethod(); } await t;`.
 See [ref/unsafe in iterators/async](./csharp-13.0/ref-unsafe-in-iterators-async.md#alternatives) for more details.
 
+Answer: yes. See [`await` in `unsafe` contexts](#await-in-unsafe-contexts).
+
+### More `unsafe` contexts and relaxations
+
 Should we also allow `&UnsafeMethod` in safe context? Today as the proposal stands, this requires `unsafe` context if the method is marked as `unsafe`.
 But since we are just getting its address, which will need `unsafe` context when dereferenced/called, we could allow the address-of itself in a safe context.
```
jjonescz, line 614: Can you merge main in? There is a link to LDM that will probably end up in a wrong section now:

https://github.com/dotnet/csharplang/blob/37400caa75a4d3a558b603de4754b1fe81ac09cb/proposals/unsafe-evolution.md?plain=1#L580

proposals/unsafe-evolution.md
```diff
@@ -568,12 +599,16 @@ Console.WriteLine(result);
 
 Answer: yes. See the [detailed design section](#unsafe-expressions).
 
-### More `unsafe` contexts and relaxations
+### (answered) `await` in `unsafe` contexts
```
jjonescz, line 602: Marking this as answered might have us forget to bring it back to LDM. Consider at least adding it to schedule.

proposals/unsafe-evolution.md
```diff
@@ -568,12 +599,16 @@ Console.WriteLine(result);
...
 
-Should we relax more restrictions around `unsafe` and pointer parameters in iterators and async methods?
-Especially allowing `await UnsafeMethod()` would be useful because now users have to rewrite that to `Task t; unsafe { t = UnsafeMethod(); } await t;`.
+Should we relax restrictions around `await` expressions in `unsafe` contexts?
+Especially allowing `await UnsafeMethod()` is useful because otherwise users have to rewrite that to `Task t; unsafe { t = UnsafeMethod(); } await t;`.
 See [ref/unsafe in iterators/async](./csharp-13.0/ref-unsafe-in-iterators-async.md#alternatives) for more details.
```
jjonescz, line 606: There were more open questions in the ref/unsafe in iterators/async speclet. I think we could resolve them as part of this or we should extract them out as separate open questions.
- Whether iterator/async methods should be allowed to contain pointer parameters (currently that's disallowed).
- Whether `yield` should be allowed in `unsafe` context.

Also, is there a problem with `stackalloc` as noted in the ref/unsafe in iterators/async speclet section which talks about allowing `await` in `unsafe` context? Specifically, does the spec need to explicitly specify that `stackalloc` doesn't survive `await` or is that already implied?

proposals/unsafe-evolution.md
333fred: https://github.com/dotnet/csharplang/blob/main/Design-Process.md#proposed-specification-updates

I'm not merging this until LDM :slightly_smiling_face:.

proposals/unsafe-evolution.md
333fred: > Specifically, does the spec need to explicitly specify that stackalloc doesn't survive await or is that already implied?

I don't believe the spec needs to specifically say this wrt stackalloc; You can already stackalloc in `async` methods today.

proposals/unsafe-evolution.md
333fred: It'll be some time before we merge.

proposals/unsafe-evolution.md
jjonescz: Works for me. But there is no rule to bring PRs to LDM, right? It's possible to just merge and bring a merged open question. (Might be easier in case there are multiple spec updates, eg, see comment above, this PR is already outdated...)

proposals/unsafe-evolution.md
jjonescz: > You can already stackalloc in `async` methods today.

But only to pointers which are disallowed in async methods in unsafe v1 or to a Span which is a ref struct and hence also disallowed across `await`.

proposals/unsafe-evolution.md
333fred: We don't have to bring PRs, no, but I like the PR as a diff of "here's the section we're adding with the wording".

jjonescz:
This was already discussed in LDM, I'm going to merge and update in a follow up.

jjonescz (approved):

Issue #9007, opened 2025-01-06
[Proposal]: Type inference using method group natural type


333fred:
# Type inference using method group natural type

* Specification: In the discussion for now
* Discussion: https://github.com/dotnet/csharplang/discussions/9006

## Summary
[summary]: #summary

It allows the natural type of a method group to contribute to method type inference

## Design meetings

- https://github.com/dotnet/csharplang/blob/main/meetings/2024/LDM-2024-09-06.md#type-inference-using-method-group-natural-type

PR #10239, opened 2026-06-19
Add proposal for type inference from method group

jcouv:
Relates to championed issue: https://github.com/dotnet/csharplang/issues/9007

### DIFF
--- a/proposals/type-inference-from-method-group.md
+++ b/proposals/type-inference-from-method-group.md
@@ -1,0 +1,32 @@
+# Type inference using method group natural type
+
+Champion issue: https://github.com/dotnet/csharplang/issues/9007
+
+## Summary
+[summary]: #summary
+
+It allows the natural type of a method group to contribute to method type inference
+
+The current [type inference rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1263-type-inference) allow 
+- method groups to contribute to output type inference
+- infering bounds for the `Ui` type parameters in `C<U1 ... Uk>` when given a `C<V1...Vk>` and `C` is a class, struct, interface or delegate type.  
+But they don't allow method groups to contribute to bounds.
+
+```
+Test(IsEven); // Error CS0411	The type arguments for method 'Program.Test<T>(Func<T, bool>)' cannot be inferred from the usage. Try specifying the type arguments explicitly.
+
+partial class Program
+{
+    public static bool IsEven(int x) => x % 2 == 0;
+    public static void Test<T>(Func<T, bool> predicate) { }
+}
+```
+
+## Design
+
+We modify the [explicit parameter type inference rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#12638-explicit-parameter-type-inferences) to not just apply to explicitly-typed lambdas, but also to method groups:
+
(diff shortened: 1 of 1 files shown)


proposals/type-inference-from-method-group.md
```diff
@@ -0,0 +1,32 @@
...
+Champion issue: https://github.com/dotnet/csharplang/issues/9007
+
+## Summary
+[summary]: #summary
+
+It allows the natural type of a method group to contribute to method type inference
+
+The current [type inference rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1263-type-inference) allow 
+- method groups to contribute to output type inference
+- infering bounds for the `Ui` type parameters in `C<U1 ... Uk>` when given a `C<V1...Vk>` and `C` is a class, struct, interface or delegate type.
```
jnm2, line 12:
```suggestion
- inferring bounds for the `Ui` type parameters in `C<U1 ... Uk>` when given a `C<V1...Vk>` and `C` is a class, struct, interface or delegate type.  
```

proposals/type-inference-from-method-group.md
```diff
@@ -0,0 +1,32 @@
+# Type inference using method group natural type
+
+Champion issue: https://github.com/dotnet/csharplang/issues/9007
+
+## Summary
+[summary]: #summary
+
+It allows the natural type of a method group to contribute to method type inference
```
jnm2, line 8:
```suggestion
It allows a method group with a unique signature to contribute to method type inference.
```
Because as written I think the spec would still work even if method groups had no natural type.

proposals/type-inference-from-method-group.md
```diff
@@ -0,0 +1,32 @@
...
+partial class Program
+{
+    public static bool IsEven(int x) => x % 2 == 0;
+    public static void Test<T>(Func<T, bool> predicate) { }
+}
+```
+
+## Design
+
+We modify the [explicit parameter type inference rules](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#12638-explicit-parameter-type-inferences) to not just apply to explicitly-typed lambdas, but also to method groups:
```
333fred, line 27: Perhaps expand on potential breaking changes that will come from this? IIRC that was a major reason we didn't do this in the first place.

jnm2: E.g. the exact inference causes:
```cs
static bool P(object o) => true;
static void M<T>(Func<T, bool> f, T value) => Console.WriteLine(typeof(T));

M(P, "hello"); // Today, prints System.String; after proposal, prints System.Object
M(bool (object o) => true, "hello"); // Prints System.Object
```

jnm2: Similarly, this would stop compiling:
```cs
static bool P(object o) => true;
static void M<T>(Func<T, bool> f, Func<T, bool> g) { }

M((string s) => true, P); // Compiles now, would behave like the next line after the proposal
M((string s) => true, bool (object o) => true); // Fails to compile
```

333fred: Those are good to include, but neither would have worked before the original lambda changes. The breaks I'm referring to are when an inner scope method suddenly becomes applicable where it wasn't before, such as an instance generic now working instead of going to an extension method, or a derived overload now working where previously it went to a base type.

PR #10241, opened 2026-06-20
Extension constants: add proposal

jcouv:
Relates to championed proposal: https://github.com/dotnet/csharplang/issues/10242

### DIFF
--- a/proposals/extension-constants.md
+++ b/proposals/extension-constants.md
@@ -1,0 +1,159 @@
+# Extension constants
+
+[!INCLUDE[Specletdisclaimer](../speclet-disclaimer.md)]
+
+Champion issue: https://github.com/dotnet/csharplang/issues/10242
+
+## Summary
+
+Allow `const` declarations inside extension blocks:
+
+```cs
+public static class E
+{
+    extension(object)
+    {
+        public const int Member = 42;
+    }
+}
+```
+
+The constant is then consumable through extension member lookup on the receiver type:
+
+```cs
+_ = object.Member; // 42
+```
+
+## Motivation
+
+Extension blocks currently support methods, properties, operators, and indexers, but not constants. This leaves a gap for API shapes where a compile-time constant is naturally part of an extension surface.
+
+Today, authors must choose between:
+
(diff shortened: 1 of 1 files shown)


proposals/extension-constants.md
```diff
@@ -0,0 +1,159 @@
...
+
+Only `constant_declaration` is permitted. Non-const fields remain disallowed in extension blocks.
+
+```cs
+public static class E
+{
+    extension(object)
+    {
+        public const int I = 1; // OK
+        public int F;           // Error: not allowed in an extension block
```
jjonescz: Could we allow `static` fields too? Feels like that could be part of the same feature without much additional effort.

jjonescz (approved):

proposals/extension-constants.md
jcouv: `static` fields raise a few problems:

First, they clash with support disambiguation syntax. Disambiguation syntax requires that we emit "implementation" members directly in the enclosing type. With fields, that means we now have two fields.

Second, they raise some questions with generics:
```cs
static class E
{
  extension<T>(C<T>)
  {
     public static int field = ...;
  }
}
```
Do we expect `C<int>.field` and `C<string>.field` to be the same or different fields?

Finally, since the static field would be emitted in the marker type, any C#-level change to the extension block would be a breaking change. For example, changing from `extension(List<(int, int)>) { public static int field = ... }` to `extension(List<(int a, int b)>) { public static int field = ... }` would be a binary break, even if we'd think of this as a backwards compatible change from the C# perspective.

jjonescz: > `static` fields raise a few problems

Aren't `decimal` constants also backed by fields, so aren't those same problems shared? I guess those are only `static readonly` fields so they are easier (e.g., we might not mind they are duplicated).

jjonescz (approved):

proposals/extension-constants.md
jcouv: Yes, I have tests for this `decimal` scenario. It's supported and because constants are inlined at call-site, there's not issue with sharing.

